PR #29, opened 2021-02-17
Fix build to publish artifacts and run SDL checks on packages.

jander-msft:
Overview:
- Publish artifacts for internal builds.
- Set _BuildConfig variable.
- Execute SDL checks on PackageArtifacts artifact.

Sample internal build: https://dev.azure.com/dnceng/internal/_build/results?buildId=999305

### DIFF
--- a/eng/build.yml
+++ b/eng/build.yml
@@ -24,7 +24,9 @@ jobs:
     artifacts:
       publish:
         logs: true
-        manifests: true
+        ${{ if ne(variables['System.TeamProject'], 'public') }}:
+          artifacts: true
+          manifests: true
 
     pool:
       # Public Linux Build Pool
@@ -64,6 +66,7 @@ jobs:
 
     variables: 
     - ${{ insert }}: ${{ parameters.variables }}
+    - _BuildConfig: ${{ parameters.configuration }}
     - _HelixType: build/product
     - _HelixBuildConfig: ${{ parameters.configuration }}
     - _SignType: test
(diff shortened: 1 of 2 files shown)


eng/build.yml
```diff
@@ -24,7 +24,9 @@ jobs:
     artifacts:
       publish:
         logs: true
-        manifests: true
+        ${{ if ne(variables['System.TeamProject'], 'public') }}:
+          artifacts: true
```
hoyosjs, line 28: artifacts too? Do you need anything outside the packages?

eng/build.yml
```diff
@@ -64,6 +66,7 @@ jobs:
 
     variables: 
     - ${{ insert }}: ${{ parameters.variables }}
+    - _BuildConfig: ${{ parameters.configuration }}
```
hoyosjs, line 69: I opened https://github.com/dotnet/arcade/issues/6972 because this shouldn't be necessary.

hoyosjs (approved):

eng/build.yml
jander-msft: When setting publish.artifacts = true, it will pull out the files from artifacts/bin and artifacts/packages and publish them. It'll be nice to have the binaries published so that when we add tests, we could add a testing stage (which would allow reruns for the tests in case there is anything flaky, instead of having to rerun the entire pipeline or the build stage) to pull down the binaries and packages and run them.

PR #38, opened 2021-03-02
OpenAPI document generation and unit tests

### LINKED ISSUE #33: Create OpenAPI document for HTTP surface

jander-msft:
These changes use Swashbuckle.AspNetCore to generate the OpenAPI document from the dotnet-monitor binaries. The dotnet-monitor tool itself is not aware of the OpenAPI document generation, as that code is in the Microsoft.Diagnostics.Monitoring.OpenApiGen executable. This generation executable is used in the unit tests to make sure what is generated from the binaries is the same as the committed documentation. The baseline OpenAPI document can be found at /documentation/openapi.json.

Some other minor changes such as using just controllers instead of all of MVC, installing ASP.NET Core as part of tools setup, running tests, and moving all version information to eng/Version.props is also included.

Overview:
- Refactor RestServer to be better suited for OpenAPI documentation.
- Use Swashbuckle.AspNetCore for OpenAPI generation.
- Update SDK and runtime versions and install ASP.NET Core runtimes.
- Unit tests to validate the OpenAPI document.

closes #33


src/Microsoft.Diagnostics.Monitoring.RestServer/Models/EventPipeProvider.cs
```diff
@@ -9,8 +9,9 @@
 
 namespace Microsoft.Diagnostics.Monitoring.RestServer.Models
 {
-    [DataContract(Name = "EventPipeProvider")]
-    public class EventPipeProviderModel
+    [DataContract]
+    public class EventPipeProvider
```
wiktork, line 13: Any chance we're going to have an api naming clash with other types here?

src/Tests/Microsoft.Diagnostics.Monitoring.OpenApiGen.UnitTests/OpenApiGeneratorTests.cs
```diff
@@ -0,0 +1,130 @@
...
+namespace Microsoft.Diagnostics.Monitoring.OpenApiGen.UnitTests
+{
+    public class OpenApiGeneratorTests
+    {
+        private const int GenerationTimemoutMs = 30_000;
+
+        private const string OpenApiBaselineName = "openapi.json";
+        private const string OpenApiGenName = "Microsoft.Diagnostics.Monitoring.OpenApiGen";
+
+        private static string CurrentExecutingAssemblyPath =>
```
wiktork, line 27: nit: readonly

src/Tools/dotnet-monitor/Startup.cs
```diff
@@ -33,24 +34,32 @@ public Startup(IConfiguration configuration)
...
                     var details = new ValidationProblemDetails(context.ModelState);
                     var result = new BadRequestObjectResult(details);
-                    result.ContentTypes.Add("application/problem+json");
+                    result.ContentTypes.Add(ContentTypes.ApplicationProblemJson);
                     return result;
                 };
+
+                // Suppress the automatic mapping of 4XX status codes to the ProblemDetails type.
+                // The tool uses ValidationProblemDetails and explicitly handles exceptions with this type.
+                options.SuppressMapClientErrors = true;
```
wiktork, line 62: Have we looked into what this looks like if we do automatic mapping?

src/Microsoft.Diagnostics.Monitoring.RestServer/Controllers/DiagController.cs
```diff
@@ -175,59 +218,89 @@ public Task<ActionResult<IEnumerable<ProcessIdentifierModel>>> GetProcesses()
...
+        /// </summary>
+        /// <param name="processKey">Value used to identify the target process, either the process ID or the runtime instance cookie.</param>
+        /// <param name="configuration">The trace configuration describing which events to capture.</param>
+        /// <param name="durationSeconds">The duration of the trace session (in seconds).</param>
+        /// <param name="egressProvider">The egress provider to which the trace is saved.</param>
+        [HttpPost("trace/{processKey?}", Name = nameof(CaptureTraceCustom))]
+        [Produces(ContentTypes.ApplicationOctectStream, ContentTypes.ApplicationProblemJson)]
+        // FileResult is the closest representation of the output so that the OpenAPI document correctly
+        // describes the result as a binary file.
+        [ProducesResponseType(typeof(FileResult), StatusCodes.Status200OK)]
```
wiktork, line 287: IIRC part of the Api review was to produce 202's if we egress. That doesn't have to be done right now, but I'm assuming that's something we can easily add later.

src/Tools/dotnet-monitor/Startup.cs
```diff
@@ -163,7 +174,11 @@ public void ConfigureServices(IServiceCollection services)
             }
 
             app.UseResponseCompression();
-            app.UseMvc();
+
+            app.UseEndpoints(builder =>
```
wiktork, line 178: Can you make sure our route restrictions that prevent sensitive endpoints on the metrics port still work as expected?

src/Tools/dotnet-monitor/Startup.cs
```diff
@@ -33,24 +34,32 @@ public Startup(IConfiguration configuration)
...
-            services.AddMvc(options =>
-                {
-                    options.Filters.Add(new ProducesAttribute("application/json"));
-
-                    options.EnableEndpointRouting = false;
-                })
-                .SetCompatibilityVersion(CompatibilityVersion.Latest)
-                .AddApplicationPart(typeof(DiagController).Assembly);
+            // AddControllers is sufficient because the tool does not use Razor nor Views.
+            services.AddControllers(options =>
```
wiktork, line 38: We disable endpoint routing here but enable it later on. Is this intended?

src/Microsoft.Diagnostics.Monitoring.RestServer/Controllers/DiagController.cs
jander-msft: Correct, that should be "easily" done, but I do not want to do that at this point. I want to wait until we get agreement on the HTTP surface area before making any significant change.

src/Microsoft.Diagnostics.Monitoring.RestServer/Controllers/DiagController.cs
```diff
@@ -28,14 +28,16 @@ namespace Microsoft.Diagnostics.Monitoring.RestServer.Controllers
     [ApiController]
     [HostRestriction]
     [Authorize(Policy = AuthConstants.PolicyName)]
+    [ProducesResponseType(typeof(ValidationProblemDetails), StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
```
wiktork, line 32: Just for my own understanding, the OpenApi generator uses these attributes to figure out where to apply things like the UnauthorizedResponseFilter correct?

wiktork (approved):

src/Microsoft.Diagnostics.Monitoring.RestServer/Controllers/DiagController.cs
jander-msft: These attributes tell the API Explorer what types of status codes (other than 200) the routes will return. Then the OpenAPI generator will query the API Explorer via Swashbuckle to get this information, which will write out these as responses for each route. However, the content type of the responses are dictated by the ProducesAttribute (and if that does not exist, the default content types).

Using the logs route as an example, it would say that 401 produces application/x-ndjson, text/event-stream, and application/problem_json. These are not correct and there is no way to describe using attribution that certain responses only have certain content types. So I use IOperationFilter, such as UnauthorizedResponseFilter to fix up the response (specifically, clear it out and set a reference to a global response description).

src/Tools/dotnet-monitor/Startup.cs
jander-msft: Without suppressing this, the 401 responses have an additional content types of application/json and application/problem+json with a structure of ProblemDetails, which is also added to the schema. This occurs because we state that 401 does not have a response body (using typeof(void)) and API behavior middleware interprets that it should override that with the ProblemDetails type.

I can work around this by specifying the ProducesErrorResponseType attribute in stead of suppressing the mapping of client errors.

src/Microsoft.Diagnostics.Monitoring.RestServer/Models/EventPipeProvider.cs
jander-msft: Yes, they do clash with the ones under Microsoft.Diagnostics.NETCore.Client only if the Microsoft.Diagnostics.Monitoring.RestServer.Models is used. So I've just used these model classes with the Model. namespace qualifier since these are largely just used in the controllers.

src/Tools/dotnet-monitor/Startup.cs
jander-msft: It was there before, but doesn't seem to be necessary.

src/Tools/dotnet-monitor/Startup.cs
jander-msft: Seems to be working correctly: I can get to /processes and /metrics on 52323 but only to /metrics on 52325.

PR #45, opened 2021-03-03
Upgrade ASP.NET Core 2.1 packages to secure versions

jander-msft:
Upgrades the following packages to the minimal version that are secure.
- Microsoft.AspNetCore.Http
- Microsoft.AspNetCore.Server.Kestrel.Core
- System.IO.Pipelines

Removed other packages that are not referenced:
- Microsoft.AspNetCore
- Microsoft.AspNetCore.HttpsPolicy
- Microsoft.AspNetCore.ResponseCompression

Lastly, change ASP.NET Core references to version ranges that end at 2.2.0 exclusively.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -11,17 +11,17 @@
     <UsingToolNetFrameworkReferenceAssemblies>false</UsingToolNetFrameworkReferenceAssemblies>
     <!-- Other libs -->
     <AzureStorageBlobsVersion>12.6.0</AzureStorageBlobsVersion>
-    <MicrosoftAspNetCoreVersion>2.1.7</MicrosoftAspNetCoreVersion>
     <MicrosoftAspNetCoreAuthenticationNegotiateVersion>3.1.10</MicrosoftAspNetCoreAuthenticationNegotiateVersion>
-    <MicrosoftAspNetCoreHttpsPolicyVersion>2.1.1</MicrosoftAspNetCoreHttpsPolicyVersion>
+    <MicrosoftAspNetCoreHttpVersion>2.1.22</MicrosoftAspNetCoreHttpVersion>
     <MicrosoftAspNetCoreMvcVersion>2.1.3</MicrosoftAspNetCoreMvcVersion>
-    <MicrosoftAspNetCoreResponseCompressionVersion>2.1.1</MicrosoftAspNetCoreResponseCompressionVersion>
+    <MicrosoftAspNetCoreServerKestrelCoreVersion>2.1.7</MicrosoftAspNetCoreServerKestrelCoreVersion>
     <MicrosoftBclHashCodeVersion>1.1.0</MicrosoftBclHashCodeVersion>
     <MicrosoftExtensionsConfigurationKeyPerFileVersion>5.0.2</MicrosoftExtensionsConfigurationKeyPerFileVersion>
     <MicrosoftExtensionsLoggingConsoleVersion>5.0.0</MicrosoftExtensionsLoggingConsoleVersion>
     <!-- We use a newer version of LoggingEventSource due to a bug in an older version-->
     <MicrosoftExtensionsLoggingEventSourceVersion>3.1.4</MicrosoftExtensionsLoggingEventSourceVersion>
     <SystemCommandLineVersion>2.0.0-beta1.20468.1</SystemCommandLineVersion>
+    <SystemIOPipelinesVersion>4.5.1</SystemIOPipelinesVersion>
     <!-- dotnet-monitor references -->
     <MicrosoftDiagnosticsMonitoringVersion>5.0.0-preview.21151.1</MicrosoftDiagnosticsMonitoringVersion>
     <MicrosoftDiagnosticsMonitoringEventPipeVersion>5.0.0-preview.21151.1</MicrosoftDiagnosticsMonitoringEventPipeVersion>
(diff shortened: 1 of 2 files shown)


eng/Versions.props
```diff
@@ -11,17 +11,17 @@
...
     <AzureStorageBlobsVersion>12.6.0</AzureStorageBlobsVersion>
-    <MicrosoftAspNetCoreVersion>2.1.7</MicrosoftAspNetCoreVersion>
     <MicrosoftAspNetCoreAuthenticationNegotiateVersion>3.1.10</MicrosoftAspNetCoreAuthenticationNegotiateVersion>
-    <MicrosoftAspNetCoreHttpsPolicyVersion>2.1.1</MicrosoftAspNetCoreHttpsPolicyVersion>
+    <MicrosoftAspNetCoreHttpVersion>2.1.22</MicrosoftAspNetCoreHttpVersion>
     <MicrosoftAspNetCoreMvcVersion>2.1.3</MicrosoftAspNetCoreMvcVersion>
-    <MicrosoftAspNetCoreResponseCompressionVersion>2.1.1</MicrosoftAspNetCoreResponseCompressionVersion>
+    <MicrosoftAspNetCoreServerKestrelCoreVersion>2.1.7</MicrosoftAspNetCoreServerKestrelCoreVersion>
     <MicrosoftBclHashCodeVersion>1.1.0</MicrosoftBclHashCodeVersion>
     <MicrosoftExtensionsConfigurationKeyPerFileVersion>5.0.2</MicrosoftExtensionsConfigurationKeyPerFileVersion>
```
wiktork, line 19: Perhaps we should do ranges for all our packages? For example, 5.0.3 was recently released, and I'm sure 3.1.* is getting periodic security updates as well.

shirhatti:
Why are we package refs here and not a `<FrameworkReference Include="Microsoft.AspNetCore.App" />`?

wiktork:
> Why are we package refs here and not a `<FrameworkReference Include="Microsoft.AspNetCore.App" />`?

We build in both core3.1 and netstandard. The package references are conditional:
 ` <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">`

eng/Versions.props
jander-msft: From a packaging perspective, if you just specify an exact version, then the package dependency will be ">= {version}". For example, the Microsoft.Bcl.HasCode package dependency will be ">= 1.1.0". So all of the package references in the RestServer assembly had ">= {version}". This change is only constraining the ASP.NET Core 2.1 versions.

Other packages that are used only by the dotnet-monitor tool do not need version ranges because the tool itself is not a library so no one can reference it.

jander-msft:
> > Why are we package refs here and not a `<FrameworkReference Include="Microsoft.AspNetCore.App" />`?
> 
> We build in both core3.1 and netstandard. The package references are conditional:
> ` <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">`

Just FYI for everyone, netstandard cannot use FrameworkReference.

I can look into converting this project to be netcoreapp2.1 and netcoreapp3.1 and just using the FrameworkReference for 2.1, but I would like to make that separately with a little more investigation.

wiktork:
> I can look into converting this project to be netcoreapp2.1 and netcoreapp3.1 and just using the FrameworkReference for 2.1, but I would like to make that separately with a little more investigation.

That will cause problems with the library scenario. For example, if you have a net 5 application and you want to reference our rest server that is netcore 2.1, it won't work so well, esp if you are already referencing asp.net 5. Basically you cannot reference a specific metapackage and at the same time claim you are netstandard compatible.

shirhatti:
> claim you are netstandard compatible.

Unless you plan to run on net462, why be netstandard2.0 compatible 😛

jander-msft:
@shirhatti @wiktork , I won't be making any target framework changes at this time. We can investigate later what is the best approach.

eng/Versions.props
```diff
@@ -11,17 +11,17 @@
     <UsingToolNetFrameworkReferenceAssemblies>false</UsingToolNetFrameworkReferenceAssemblies>
     <!-- Other libs -->
     <AzureStorageBlobsVersion>12.6.0</AzureStorageBlobsVersion>
-    <MicrosoftAspNetCoreVersion>2.1.7</MicrosoftAspNetCoreVersion>
     <MicrosoftAspNetCoreAuthenticationNegotiateVersion>3.1.10</MicrosoftAspNetCoreAuthenticationNegotiateVersion>
-    <MicrosoftAspNetCoreHttpsPolicyVersion>2.1.1</MicrosoftAspNetCoreHttpsPolicyVersion>
+    <MicrosoftAspNetCoreHttpVersion>2.1.22</MicrosoftAspNetCoreHttpVersion>
     <MicrosoftAspNetCoreMvcVersion>2.1.3</MicrosoftAspNetCoreMvcVersion>
```
shirhatti, line 16: Perhaps we should upgrade all these version?

shirhatti:
> packages to the minimal version that are secure.

Why the minimal version. Shouldn't we pick latest patch?

jander-msft:
> > packages to the minimal version that are secure.
> 
> Why the minimal version. Shouldn't we pick latest patch?

So that it's possible to integrate this with any ASP.NET Core 2.1 application that is referencing secure assemblies. If we go to the latest, we force anyone who wants to use this library on ASP.NET Core 2.1 to also use the latest version of those packages.

shirhatti (approved):

PR #47, opened 2021-03-05
Log bound default vs metrics addresses.

jander-msft:
This change allows dotnet-monitor to report which addresses are default and which are metrics. This will be used for testing when telling dotnet-monitor to bind to any port for each of these types of addresses, e.g.

```
dotnet-monitor collect --urls https://127.0.0.1:0 --metricUrls http://127.0.0.1:0
```

Output (with Debug log level set for Microsoft.Diagnostics logging category):
```
17:33:31 warn: Microsoft.Diagnostics.Tools.Monitor.ExperimentalToolLogger[0]
      WARNING: dotnet-monitor is experimental and is not intended for production environments yet.
17:33:32 warn: Microsoft.AspNetCore.Server.Kestrel[0]
      Overriding address(es) 'https://127.0.0.1:0'. Binding to endpoints defined in UseKestrel() instead.
17:33:32 info: Microsoft.Hosting.Lifetime[0]
      Now listening on: https://127.0.0.1:53960
17:33:32 info: Microsoft.Hosting.Lifetime[0]
      Now listening on: http://127.0.0.1:53961
17:33:32 dbug: Microsoft.Diagnostics.Tools.Monitor.Startup[16]
      Bound default address: https://127.0.0.1:53960
17:33:32 dbug: Microsoft.Diagnostics.Tools.Monitor.Startup[17]
      Bound metrics address: http://127.0.0.1:53961
17:33:32 info: Microsoft.Hosting.Lifetime[0]
17:33:32 info: Microsoft.Hosting.Lifetime[0]
17:33:32 info: Microsoft.Hosting.Lifetime[0]
      Content root path: D:\src\dotnet\dotnet-monitor\artifacts\bin\dotnet-monitor\Debug\netcoreapp3.1\
```

### DIFF
--- a/src/Tools/dotnet-monitor/Startup.cs
+++ b/src/Tools/dotnet-monitor/Startup.cs
@@ -4,7 +4,9 @@
 
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Hosting.Server.Features;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.ResponseCompression;
 using Microsoft.AspNetCore.Server.Kestrel.Core;
@@ -17,6 +19,7 @@ using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO.Compression;
 using System.Text.Json.Serialization;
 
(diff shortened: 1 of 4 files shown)


src/Tools/dotnet-monitor/Startup.cs
```diff
@@ -173,5 +179,27 @@ public void ConfigureServices(IServiceCollection services)
                 builder.MapControllers();
             });
         }
+
+        private static void LogBoundAddresses(IFeatureCollection features, AddressListenResults results, ILogger logger)
+        {
+            IServerAddressesFeature serverAddresses = features.Get<IServerAddressesFeature>();
```
wiktork, line 185: Is there a reason we fall back to the IServerAddressFeature instead of simply passing the urls along to the AddressListenResults if they succeed in binding, and having that structure contain two separate collections, one for default and one for metric urls?

jander-msft: We aren't actually binding the addresses, we are setting up the options for each address when calling KestrelServerOptions.Listen*. Binding happens latter when the Kestrel server starts up. So the urls that we have while setting up the listening options still have 0 for their ports if 0 was specified on the command line.

wiktork (approved):

PR #50, opened 2021-03-09
Use App filters

wiktork:
UseAppFilters for logs in .net 5+. See https://github.com/dotnet/runtime/pull/40165

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -7,6 +7,12 @@
     <DotNetUseShippingVersions>true</DotNetUseShippingVersions>
   </PropertyGroup>
   <PropertyGroup>
+    <!-- These versions are not used directly. For Dev workflows, nuget requires these to properly follow
+    project references for command line builds. They should match the values in the diagnostics repo. -->
+    <MicrosoftBclAsyncInterfacesVersion>1.1.0</MicrosoftBclAsyncInterfacesVersion>
+    <MicrosoftDiagnosticsTracingTraceEventVersion>2.0.64</MicrosoftDiagnosticsTracingTraceEventVersion>
+    <MicrosoftExtensionsLoggingVersion>2.1.1</MicrosoftExtensionsLoggingVersion>
+    <SystemTextJsonVersion>4.7.1</SystemTextJsonVersion>
     <!-- Runtime versions to test -->
     <MicrosoftNETCoreApp31Version>3.1.12</MicrosoftNETCoreApp31Version>
     <MicrosoftAspNetCoreApp31Version>$(MicrosoftNETCoreApp31Version)</MicrosoftAspNetCoreApp31Version>
--- a/src/Microsoft.Diagnostics.Monitoring.RestServer/Controllers/DiagController.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.RestServer/Controllers/DiagController.cs
@@ -341,7 +341,7 @@ namespace Microsoft.Diagnostics.Monitoring.RestServer.Controllers
             [FromQuery][Range(-1, int.MaxValue)]
             int durationSeconds = 30,
             [FromQuery]
-            LogLevel level = LogLevel.Debug,
+            LogLevel? level = null,
             [FromQuery]
             string egressProvider = null)
         {
@@ -364,11 +364,21 @@ namespace Microsoft.Diagnostics.Monitoring.RestServer.Controllers
 
                     loggerFactory.AddProvider(new StreamingLoggerProvider(outputStream, format, level));
 
-                    var settings = new EventLogsPipelineSettings
+                    var settings = new EventLogsPipelineSettings()
                     {
-                        Duration = duration,
-                        LogLevel = level,
+                        Duration = duration
                     };
+
+                    //If the user does not explicitly specify a log level, we will use the apps defaults for .net 5.
+                    //Otherwise, we use the log level specified by the query.
+                    if ((!level.HasValue) && (processInfo.EndpointInfo.RuntimeInstanceCookie != Guid.Empty))
+                    {
+                        settings.UseAppFilters = true;
+                    }
+                    else
+                    {
+                        settings.LogLevel = level.GetValueOrDefault(LogLevel.Warning);
+                    }
 
                     var client = new DiagnosticsClient(processInfo.EndpointInfo.Endpoint);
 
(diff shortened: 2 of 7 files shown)


dotnet-monitor.dev.sln
```diff
@@ -0,0 +1,352 @@
...
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tests", "Tests", "{C7568468-1C79-4944-8136-18812A7F9EA7}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Diagnostics.Monitoring.TestCommon", "src\Tests\Microsoft.Diagnostics.Monitoring.TestCommon\Microsoft.Diagnostics.Monitoring.TestCommon.csproj", "{863AA12D-0918-49CE-9E76-45A22680268B}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Diagnostics.Monitoring.OpenApiGen", "src\Tests\Microsoft.Diagnostics.Monitoring.OpenApiGen\Microsoft.Diagnostics.Monitoring.OpenApiGen.csproj", "{5AB5A4B4-3A13-4D42-B42C-B9193E60F426}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Diagnostics.Monitoring.OpenApiGen.UnitTests", "src\Tests\Microsoft.Diagnostics.Monitoring.OpenApiGen.UnitTests\Microsoft.Diagnostics.Monitoring.OpenApiGen.UnitTests.csproj", "{FCEC25E0-9EE1-416E-A7C4-2898E5640157}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Diagnostics.Monitoring", "%DIAGNOSTICS_REPO_ROOT%\src\Microsoft.Diagnostics.Monitoring\Microsoft.Diagnostics.Monitoring.csproj", "{6DBF4829-AF59-48F4-8BC8-3A7E873B41A8}"
```
wiktork, line 27: Note the environment variables. Long term, I would like to generate the sln instead of adding it to the repo.

src/Microsoft.Diagnostics.Monitoring.RestServer/Controllers/DiagController.cs
```diff
@@ -341,7 +341,7 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
             [FromQuery][Range(-1, int.MaxValue)]
             int durationSeconds = 30,
             [FromQuery]
-            LogLevel level = LogLevel.Debug,
+            LogLevel level = LogLevel.Warning,
```
WeihanLi, line 344: Maybe we could make `level` nullable, so that we could use `settings.LogLevel = level.GetValueOrDefault(LogLevel.Warning)`

dotnet-monitor.dev.sln
jander-msft: I think you can remove this solution for now. It should work okay to just open the diagnostics.sln and dotnet-monitor.sln in two separate VS instances until we are able to (1) automatically generate this one, say, from a script, and (2) exclude it via .gitignore file. The repository build is failing because it's trying to restore the new solution and I think it scans for all *.sln at the repository root.

wiktork: If I don't have a sln, project references won't work and i'll have to fall back to the previous workflow, which I found too cumbersome. I'll add the sln generation.

src/Microsoft.Diagnostics.Monitoring.RestServer/Controllers/DiagController.cs
```diff
@@ -364,12 +364,22 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
...
+                    var settings = new EventLogsPipelineSettings()
                     {
-                        Duration = duration,
-                        LogLevel = level,
+                        Duration = duration
                     };
 
+                    //If the user does not explicitly specify a log level, we will use the apps defaults for .net 5.
+                    //Otherwise, we use the log level specified by the query.
+                    if ((!level.HasValue) && (processInfo.EndpointInfo.RuntimeInstanceCookie != Guid.Empty))
```
jander-msft, line 374: Might have to update this heuristic when we generate artificial cookie values for .NET Core 3.1, but we can probably also add another field to specify whether the cookie is real or generated in order to differentiate their usage.

wiktork: I was thinking we would another field to indicated generated runtime id's.

src/Tools/dotnet-monitor/dotnet-monitor.csproj
```diff
@@ -26,10 +26,17 @@
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="$(MicrosoftExtensionsLoggingConsoleVersion)" />
     <PackageReference Include="System.CommandLine" Version="$(SystemCommandLineVersion)" />
     <PackageReference Include="Microsoft.AspNetCore.Authentication.Negotiate" Version="$(MicrosoftAspNetCoreAuthenticationNegotiateVersion)" />
-    <PackageReference Include="Microsoft.Diagnostics.Monitoring" Version="$(MicrosoftDiagnosticsMonitoringVersion)" />
   </ItemGroup>
 
-  <ItemGroup>
+    <ItemGroup Condition="'$(DIAGNOSTICS_REPO_ROOT)' == ''">
```
jander-msft, line 31: nit: can you fix the ident on these ItemGroup tags?

eng/Versions.props
```diff
@@ -7,7 +7,13 @@
...
   </PropertyGroup>
   <PropertyGroup>
-    <!-- Runtime versions to test -->
+    <!-- These versions are not used directly. For Dev workflows, nuget requires these to properly follow
+    project references for command line builds. They should match the values in the diagnostics repo. -->
+    <MicrosoftBclAsyncInterfacesVersion>1.1.0</MicrosoftBclAsyncInterfacesVersion>
+    <MicrosoftDiagnosticsTracingTraceEventVersion>2.0.64</MicrosoftDiagnosticsTracingTraceEventVersion>
+    <MicrosoftExtensionsLoggingVersion>2.1.1</MicrosoftExtensionsLoggingVersion>
+    <SystemTextJsonVersion>4.7.1</SystemTextJsonVersion>
+      <!-- Runtime versions to test -->
```
jander-msft, line 16: nit: Fix indent

jander-msft (approved):

sywhang (approved):

PR #51, opened 2021-03-09
Have HostRestrictionAttribute use bound addresses instead of configured addresses.

jander-msft:
This change makes sure that the HostRestrictionAttribute uses the bound addresses rather than the configured addresses. This fixes an issue where if dynamic port binding is used, then the HostRestrictionAttribute will test the dynamic port values rather than port 0. Example usage of dynamic port binding: `dotnet-monitor collect --urls http://127.0.0.1:0 --metricUrls http://127.0.0.1:0`

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.RestServer/IUrlPortsProvider.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.RestServer/IUrlPortsProvider.cs
@@ -1,0 +1,19 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Diagnostics.Monitoring.RestServer
+{
+    /// <summary>
+    /// Provides port information about the URLs the were successfully bound.
+    /// </summary>
+    internal interface IUrlPortsProvider
+    {
+        /// <summary>
+        /// Get the ports of the metrics URLs.
+        /// </summary>
+        IEnumerable<int> MetricsPorts { get; }
+    }
+}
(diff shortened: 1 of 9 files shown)


src/Microsoft.Diagnostics.Monitoring.RestServer/IUrlPortsProvider.cs
```diff
@@ -0,0 +1,19 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Diagnostics.Monitoring.RestServer
+{
+    /// <summary>
+    /// Provides port information about the URLs the were successfully bound.
+    /// </summary>
+    internal interface IUrlPortsProvider
```
jander-msft, line 12: Not happy with the name; open to suggestions (and for anything else). Maybe rename to IBoundPortsProvider?

wiktork: Given the functionality, you may want to call it something like IMetricPortsProvider.

src/Tools/dotnet-monitor/Startup.cs
```diff
@@ -183,22 +184,18 @@ public void ConfigureServices(IServiceCollection services)
...
             IServerAddressesFeature serverAddresses = features.Get<IServerAddressesFeature>();
-            Debug.Assert(serverAddresses.Addresses.Count == results.AddressesCount + results.MetricAddressesCount);
 
             // This logging allows the tool to differentiate which addresses
             // are default address and which are metrics addresses.
-            int urlCount = results.AddressesCount;
-            foreach (string serverAddress in serverAddresses.Addresses)
+
+            foreach (string defaultAddress in results.GetDefaultAddresses(serverAddresses))
             {
```
wiktork, line 192: Can you make sure to test this with metrics turned off as well?

wiktork (approved):

src/Tools/dotnet-monitor/Startup.cs
jander-msft: Seems to be working correctly. No addresses are reported via GetMetricsAddresses.

PR #86, opened 2021-03-17
Add all configuration capabilities, metrics disablement tests, and authentication tests.

jander-msft:
- Add configuration via environment variables, settings, and key-per-file configuration.
- Make all options properties nullable to allow reuse in test configuration.
- Test disablement of metrics via configuration.
- Test API Key authentication and key rotation.
- Test Negotiate authentication.


src/Microsoft.Diagnostics.Monitoring.RestServer/CorsConfiguration.cs
```diff
@@ -2,12 +2,12 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Threading.Tasks;
 
+#if UNITTEST
+namespace Microsoft.Diagnostics.Monitoring.UnitTests.Options
```
wiktork, line 7: Why a custom namespace for unit test versions of these?

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/Microsoft.Diagnostics.Monitoring.UnitTests.csproj
```diff
@@ -10,7 +10,11 @@
   </ItemGroup>
 
   <ItemGroup>
+    <Compile Include="..\..\Microsoft.Diagnostics.Monitoring.RestServer\CorsConfiguration.cs" Link="Options\CorsConfiguration.cs" />
```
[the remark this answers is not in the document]
jander-msft: This (the CorsConfiguration) is linked in from the rest assembly.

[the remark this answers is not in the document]
jander-msft: Are you asking why not use the type from the REST assembly? The unit test project compiles for both ASP.NET Core 3.1 and 5.0; I don't think I can reference the REST assembly on the 5.0 compilation due to FrameworkReference differences.

src/Microsoft.Diagnostics.Monitoring.RestServer/CorsConfiguration.cs
jander-msft: Because these are compiled into the unit test assembly, I like to keep the root namespace the same with the assembly. It's not important to do it though.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
```diff
@@ -0,0 +1,232 @@
...
+
+            await using DotNetMonitorRunner toolRunner = new DotNetMonitorRunner(_outputHelper);
+
+            _outputHelper.WriteLine("Generating API key.");
+
+            // Generate initial API key
+            byte[] apiKey = GenerateApiKey();
+
+            // Set API key via key-per-file
+            RootOptions options = new();
```
wiktork, line 125: Now to start the debate between:
var options = new RootOptions() vs. RootOptions options = new()?

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
```diff
@@ -0,0 +1,232 @@
...
+                try
+                {
+                    await client.GetProcessesAsync(DefaultApiTimeout);
+                }
+                catch (ApiStatusCodeException ex) when (ex.StatusCode == HttpStatusCode.Unauthorized)
+                {
+                    break;
+                }
+
+                await Task.Delay(TimeSpan.FromSeconds(3));
```
wiktork, line 169: nit: Consider delay followed by attempt. I'm guessing it never succeeds on the first try.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
```diff
@@ -0,0 +1,232 @@
...
+            using HttpClientHandler handler = new HttpClientHandler();
+            handler.Credentials = CredentialCache.DefaultCredentials;
+            using HttpClient httpClient = new HttpClient(handler);
+
+            using ApiClient client = new ApiClient(_outputHelper, await toolRunner.GetDefaultAddressAsync(DefaultAddressTimeout), httpClient);
+
+            // TODO: Split test into elevated vs non-elevated tests and skip
+            // when not running in the corresponding context. Possibly unelevate
+            // dotnet-monitor when running tests elevated.
+            if (EnvironmentInformation.IsElevated)
```
wiktork, line 208: Do we even have a way of running this elevated during PR checks?

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/Microsoft.Diagnostics.Monitoring.UnitTests.csproj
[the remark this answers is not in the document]
wiktork: That makes sense, but makes me wonder about the dotnet-monitor reference? Is that one purely for build ordering?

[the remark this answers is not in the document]
wiktork: Thinking about it some more, do we want to make this test 3.1 only? I realize that the test could really be anything, but if we want to reuse the types in restserver/dotnet-monitor, making the test 3.1 would make it easier.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/Options/EgressOptions.cs
```diff
@@ -0,0 +1,18 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Diagnostics.Monitoring.UnitTests.Options
+{
+    internal class EgressOptions
+    {
+        public Dictionary<string, Dictionary<string, string>> Providers { get; }
+            = new Dictionary<string, Dictionary<string, string>>(StringComparer.OrdinalIgnoreCase);
```
wiktork, line 13: nit: new(StringComparer.OrdinalIgnoreCase)?

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/Options/OptionsExtensions.cs
```diff
@@ -0,0 +1,147 @@
...
+        /// <summary>
+        /// Generates a key-per-file map of the options.
+        /// </summary>
+        /// <remarks>
+        /// Each key is the file name; each value is the file content.
+        /// </remarks>
+        public static IDictionary<string, string> ToKeyPerFileConfiguration(this RootOptions options)
+        {
+            Dictionary<string, string> variables = new Dictionary<string, string>();
+            MapObject(options, string.Empty, variables);
```
wiktork, line 40: Will this work properly for array based options? E.g. Something__0__Name, Something__1__Name

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/Runners/DotNetMonitorRunner.cs
```diff
@@ -150,24 +157,38 @@ public async Task StartAsync(CancellationToken token)
             _runner.SetEnvironmentVariable("Logging__Console__FormatterName", "json");
             // Enable Debug on Startup class to get lifetime and address events
             _runner.SetEnvironmentVariable("Logging__LogLevel__Microsoft.Diagnostics.Tools.Monitor.Startup", "Debug");
+            
+            // Override the shared config directory
+            _runner.SetEnvironmentVariable("DotnetMonitor_SharedConfigDirectoryOverride", SharedConfigDirectoryPath);
```
wiktork, line 162: Can we make sure the metasettings don't look like normal settings? These match our config prefix.

wiktork (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
jander-msft: Fun thing I learned today: the CI build runs them elevated. So what I really need to do if figure out a way to de-elevate. I think I found some Win32 APIs for creating an access token with normal privileges and then can use CreateProcessAsUser to launch the executables, so I can rewrite the DotNetRunner to de-elevate (maybe by default).

jander-msft: I don't plan on making the de-elevation change in this PR.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
jander-msft: It typically does not succeed on the first try. However, I plan to replace this with some log listening code once #74 is completed. I will switch the order as you requested.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
jander-msft: What do you think? I can switch all of the test code to be either way.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/Runners/DotNetMonitorRunner.cs
jander-msft: Do you have a proposed naming convention? `DotnetMonitor_Testing__`?

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/Microsoft.Diagnostics.Monitoring.UnitTests.csproj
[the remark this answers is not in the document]
jander-msft: If I changed it to be only 3.1, I'd have to figure out a way to run each test for dotnet-monitor running as 3.1 and 5.0. I want the tests to flex the runtime roll-forward capabilities so we aren't blindsided with an incompatibility. Also, the test app that I will want to write should run as both 3.1 and 5.0 so we can test different polyfill capabilities (cookie generation, process information gathering, `/logs` behavior, etc). As it is right now, the tests build for both 3.1 and 5.0 and they run dotnet-monitor (and eventually the test app) with the same runtime version. Then it'll get even more complicated when I go to add 6.0 previews.

So I could change it to just be 3.1, but I think I'd have to convert the Fact tests to be Theory tests that pull in the runtime versions so that each test is run for each runtime version.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/Options/OptionsExtensions.cs
jander-msft: I believe that `MapList` is doing this. We can add other type handling in the future if we use other list types in the configuration options.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/Microsoft.Diagnostics.Monitoring.UnitTests.csproj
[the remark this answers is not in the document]
jander-msft: Alternative: put all of the shared types in a new netstandard2.0 library (one for REST assembly, one for dotnet-monitor)

Alternative: put all of the shared types into shared projects (does not produce assembly, but effectively links in types into each dependent assembly).

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
wiktork: I have no strong feelings either way.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/Runners/DotNetMonitorRunner.cs
wiktork: I would avoid the DotnetMonitor_ prefix, since you don't want this to merge into configuration right? Maybe DotnetMonitorTestSettings_?

jander-msft: All environment variables are merged into configuration regardless of the prefix. Configuring with the prefix `DotnetMonitor_` just means "also add these without this prefix". For example, the variables `DotnetMonitor_Metrics__Enabled` and `Metrics__Enabled` are merged into the same key `Metrics:Enabled`.

jander-msft: I'll use your suggested prefix `DotnetMonitorTestSettings__` for the variables and not configure this prefix.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
jander-msft: I've changed all of the test code to remove the type from the ctor call if the type is explicitly used on the member of local.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/Microsoft.Diagnostics.Monitoring.UnitTests.csproj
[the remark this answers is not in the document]
jander-msft: I'm going to keep the changes as-is for now and then revisit this after thinking about it and discussing it more.

PR #97, opened 2021-03-18
Skip negotiate test on non-Windows platforms.

### LINKED ISSUE #92: Support skippable tests
Need support for skippable tests, especially for running tests on certain platforms e.g. the Negotiate authentication test should only be run on Windows.

jander-msft:
Everything under Xunit.Extensions was copied from dotnet/diagnostics and I updated the THIRD-PARTY-NOTICES.txt to point to the correct location in dotnet/dotnet-monitor.

closes #92


wiktork:
Is there any way we could share this logic instead of duplicating it?

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
```diff
@@ -185,9 +187,11 @@ public async Task ApiKeyAuthenticationSchemeTest()
         /// <summary>
         /// Tests that Negotiate authentication can be used for authentication.
         /// </summary>
-        [Fact]
+        [SkippableFact]
```
hoyosjs, line 190: The other option you can have here is to reference `Xunit.SkippableFact@1.4.13` if you don't want to maintain the fork of the code. There's also a way to get the diagnostics version of this code, but I don't know if it adds more value than the other package.

THIRD-PARTY-NOTICES.TXT
```diff
@@ -278,7 +278,7 @@ SOFTWARE.
 
 -------------------------------------------------
 
-Source in src/Microsoft.Diagnostics.TestHelpers/Xunit.Extensions/* is largely derived from source
+Source in src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Xunit.Extensions/* is largely derived from source
```
jander-msft, line 281: Remove this since we are not using this code in this repo.

wiktork (approved):

PR #99, opened 2021-03-22
Increase test timeouts to 1 minute.

jander-msft:
Increase test timeout to 1 minute to allow for tests to pass on resource constrained machines. The purpose of the tests is not to measure performance; the timeouts are just a means of giving up on the test so that the builds do not hang for the maximum timeout.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
@@ -19,9 +19,7 @@ namespace Microsoft.Diagnostics.Monitoring.UnitTests
 {
     public class AuthenticationTests
     {
-        private static readonly TimeSpan DefaultAddressTimeout = TimeSpan.FromSeconds(10);
-        private static readonly TimeSpan DefaultApiTimeout = TimeSpan.FromSeconds(10);
-        private static readonly TimeSpan DefaultStartTimeout = TimeSpan.FromSeconds(30);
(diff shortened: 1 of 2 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
```diff
@@ -19,9 +19,7 @@ namespace Microsoft.Diagnostics.Monitoring.UnitTests
 {
     public class AuthenticationTests
     {
-        private static readonly TimeSpan DefaultAddressTimeout = TimeSpan.FromSeconds(10);
-        private static readonly TimeSpan DefaultApiTimeout = TimeSpan.FromSeconds(10);
-        private static readonly TimeSpan DefaultStartTimeout = TimeSpan.FromSeconds(30);
```
wiktork, line 24: How long does the full test pass take?

wiktork (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
jander-msft: The PR build shows that the ApiKeyAuthenticationSchemeTest is the slowest test at about 5 seconds. The total time for all of these tests is about 10 seconds (x2 because they run for netcoreapp3.1 and net5.0).

I've typically seen the vast majority of the tests run within 2 seconds per test; it's just that some times attempting to access an HTTP API route will timeout on the public build machines.

PR #100, opened 2021-03-22
Have blob group include release subfolder based on build quality.

### LINKED ISSUE #80: Only publish aka.ms latest versions from release branches
Alternative: Set the blob group prefix for the version and hash files to include the preview version to simulate channels. For example, if release branch is building 5.0.0 Preview 4 and main branch is building 5.0.0 Preview 5, then the blob group prefixes can be monitor5.0p4 and monitor5.0p5 respectively. This allows the release branch to stabilize the prior preview version and allows the main branch to continue development. The dotnet-docker repo can use 5.0p4 in nightly until the preview is published in its main and switch the nightly update to 5.0p5

jander-msft:
The current aka.ms link generation for the latest version produces a link that represents the absolute latest version within a given release. For example, the link 'aka.ms/dotnet/diagnostics/monitor5.0/dotnet-monitor.nupkg.version' only allows for pointing to a single version within the 5.0.X release.

In order to allow dotnet-docker to consume a specific preview version, the link generation needs to be updated to include the prerelease information. This change adds the prerelease information as a subfolder to the existing blob group name calculation. It also introduces a build quality notion that is similar to aka.ms channel build quality so that the main branch does not have to declare its release name.

For example, we could fork main into a release branch (which would build preview 4 bits) and update main to preview 5. The dotnet-docker repo would then be able to monitoring for the stabilizing versions under preview 4 without being impacted by the main branch producing preview 5 bits. The release branch would have a build quality set to 'prerelease'. Thus, the release branch would update the 'aka.ms/dotnet/diagnostics/monitor5.0/preview.4/dotnet-monitor.nupkg.version' link with its latest build whereas the main branch would update 'aka.ms/dotnet/diagnostics/monitor5.0/daily/dotnet-monitor.nupkg.version'. When it comes time to produce the final release, the release branch build quality would be set to 'release' and would update the 'aka.ms/dotnet/diagnostics/monitor5.0/release/dotnet-monitor.nupkg.version' link.

closes #80

### DIFF
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -16,9 +16,52 @@
           AfterTargets="Pack"
           Condition="'$(IsPackable)' == 'true' and '$(BlobGroupPrefix)' != ''">
     <PropertyGroup>
-      <_BlobGroupVersionMajor>$(PackageVersion.Split('.')[0])</_BlobGroupVersionMajor>
-      <_BlobGroupVersionMinor>$(PackageVersion.Split('.')[1])</_BlobGroupVersionMinor>
-      <_BlobGroupName>$(BlobGroupPrefix)$(_BlobGroupVersionMajor).$(_BlobGroupVersionMinor)</_BlobGroupName>
+      <!--
+        These properties take a package version and transform it into a blob group name so that
+        all builds from the same product and release version are grouped together. This code has
+        to consider when the version is a release version (e.g. 5.0.0) or has a prerelease label
+        (e.g. 5.0.0-preview.1). The former is transformed into '$(BlobGroupPrefix)5.0/release' whereas
+        the latter is transformed into '$(BlobGroupPrefix)5.0/preview.1'. It also accounts for the
+        BlobGroupBuildQuality defined in Version.props, which determines if the prerelease information
+        should be used in the final blob group name.
+        -->
+      <_PreReleaseSeperatorIndex>$(PackageVersion.IndexOf('-'))</_PreReleaseSeperatorIndex>
+      
+      <!-- Prerelease: '5.0.0-preview.1' -> '5.0.0' and 'preview.1' -->
+      <_BlobGroupVersion Condition="'$(_PreReleaseSeperatorIndex)' != '-1'">$(PackageVersion.Substring(0, $(_PreReleaseSeperatorIndex)))</_BlobGroupVersion>
+      <_BlobGroupPreRelease Condition="'$(_PreReleaseSeperatorIndex)' != '-1'">$(PackageVersion.Substring($([MSBuild]::Add($(_PreReleaseSeperatorIndex), 1))))</_BlobGroupPreRelease>
+      
+      <!-- Release: take the package version as-is. -->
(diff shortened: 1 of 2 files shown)


wiktork (approved):

jander-msft:
@hoyosjs, let me know what you think of this change. If you have questions, or an alternative, please let me know! I'd like to get this in this week and work out any issues next week before we start branching.

src/Directory.Build.targets
```diff
@@ -16,9 +16,52 @@
...
+
+      <!--
+        Calculate release name based on version parts and blob group build quality.
+        Daily: use 'daily'
+        Prerelease: 'preview' and '1' -> 'preview.1'
+        Release: use 'release'
+        -->
+      <_BlobGroupReleaseName  Condition="'$(BlobGroupBuildQuality)' == 'daily'">daily</_BlobGroupReleaseName>
+      <_BlobGroupReleaseName  Condition="'$(BlobGroupBuildQuality)' == 'prerelease'">$(_BlobGroupPreReleaseLabel).$(_BlobGroupPreReleaseIteration)</_BlobGroupReleaseName>
+      <_BlobGroupReleaseName  Condition="'$(BlobGroupBuildQuality)' == 'release'">release</_BlobGroupReleaseName>
```
hoyosjs, line 52: Maybe error out if `_BlobGroupReleaseName` is empty after this point. 
```suggestion
      <_BlobGroupReleaseName  Condition="'$(BlobGroupBuildQuality)' == 'release'">release</_BlobGroupReleaseName>
      <Error
            Text="The BlobGroupBuildQuality property must be set to one of: daily, release, prerelease."
            Condition="'$(_BlobGroupReleaseName)' == ''" />
```

hoyosjs (approved):

src/Directory.Build.targets
```diff
@@ -16,9 +16,52 @@
...
+      
+      <!-- Release: take the package version as-is. -->
+      <_BlobGroupVersion Condition="'$(_PreReleaseSeperatorIndex)' == '-1'">$(PackageVersion)</_BlobGroupVersion>
+
+      <!--
+        Prerelease may contain additional fields; only want the label and first number
+        (e.g. 'preview.1.12345' -> 'preview' and '1').
+        -->
+      <_BlobGroupPreReleaseLabel Condition="'$(_BlobGroupPreRelease)' != ''">$(_BlobGroupPreRelease.Split('.')[0])</_BlobGroupPreReleaseLabel>
+      <_BlobGroupPreReleaseIteration Condition="'$(_BlobGroupPreRelease)' != ''">$(_BlobGroupPreRelease.Split('.')[1])</_BlobGroupPreReleaseIteration>
```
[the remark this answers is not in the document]
jander-msft: It actually was X.Y.Z-preview.##### and now is X.Y.Z-preview.N.#####, so either way it will always have the first "." in the label portion that we'll be looking for. So it will produce monitorX.Y/preview.##### in the former case or monitorX.Y/preview.N in the latter case (which is the current versioning schema).

src/Directory.Build.targets
jander-msft: I'll make this change, but it'll have to be outside of the PropertyGroup.

jander-msft: I created an item group of the valid names and validated against that as well as validating the blob group parts are not empty.

Issue #91, opened 2021-03-18
Run tests on non-Windows platforms in build pipeline.


jander-msft:
Possibly use Helix to run tests.

PR #101, opened 2021-03-22
Build and test on non-Windows platforms.

jander-msft:
Builds and runs tests on the core OS platforms that dotnet-monitor targets. It runs both on CentOS and Alpine Linux distros to make sure it does not have any failures based on glibc vs musl libc. It also builds the Debug configuration on all OS platforms to catch and failed debug assertions during testing. Only the binaries from the Windows Release build are published. Finally, the solution file has been cleaned up to remove all unused configurations and platforms.

closes: #91


jander-msft:
The test failures in the PR build should be resolved by #97 and #99

dotnet-monitor.yml
```diff
@@ -23,23 +23,71 @@ variables:
...
         configuration: Release
-        platform: x64
     # This registers the build with BAR.
     - ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
       - template: /eng/common/templates/job/publish-build-assets.yml
         parameters:
           configuration: Release
           dependsOn:
-          - Windows
+          - WindowsDebug
```
wiktork, line 83: We only build all of these in non-PR builds correct?

wiktork (approved):

dotnet-monitor.yml
jander-msft: The publish-build-assets.yml is only run on the internal project. All of the (Windows|CentOs|Alpine|MacOS)(Debug|Release) builds are run for both PR and CI, however they only publish logs when running in PR.

eng/build.yml
```diff
@@ -55,7 +57,7 @@ jobs:
     ${{ if eq(parameters.osGroup, 'Linux') }}:
       container: mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7-359e48e-20200313130914
 
-    ${{ if eq(parameters.osGroup, 'Linux-musl') }}:
+    ${{ if eq(parameters.osGroup, 'Linux_Musl') }}:
```
hoyosjs, line 60: On this change, can you update the vmImage to ubuntu-18.04 or ubuntu-latest? 16.04 is EoL pretty soon (not the buildpool part).

hoyosjs (approved):
Did you run these in internal just to test them out?

dotnet-monitor.yml
```diff
@@ -23,23 +23,71 @@ variables:
...
+          - WindowsRelease
+          - CentOSDebug
```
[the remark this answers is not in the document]
jander-msft: Yes, the debug legs are to catch debug assertions during test.

jander-msft:
> Did you run these in internal just to test them out?

Yes, I did. I'll update with the new Ubuntu image, pull in the latest from main, and rerun internally.

hoyosjs:
Also, if you are seeing timeouts being tricky (you already solved it nonetheless) I'd say switch to dotnet-test, which has the advantage of being able to take dumps on test timeout.

PR #102, opened 2021-03-22
[NO REVIEW, NO MERGE] Investigating test timeouts

jander-msft:


WeihanLi:
Maybe we could mark this as draft so that it will not be merged by mistake

jander-msft:
My current theory is that this is an HttpClient socket exhaustion problem. Thinking of trying a Xunit collection fixture to set up an IHttpClientFactory that is shared among all of the tests.

WeihanLi:
> My current theory is that this is an HttpClient socket exhaustion problem. Thinking of trying a Xunit collection fixture to set up an IHttpClientFactory that is shared among all of the tests.

We wanna make the `ApiClient` singleton, right?
Maybe we can try this xunit dependency injection extension, that will make the test code much more clean than collection fixture I think, maybe you wanna have a look at the project <https://github.com/pengweiqhca/Xunit.DependencyInjection>

Issue #12, opened 2020-10-16
container-to-container dump retrieval


josalem:
dotnet-monitor always requests that memory dumps be written to `$TMPDIR`.  In the .net5.0 container-to-container configuration, `$TMPDIR` is not guaranteed to be the shared file mount.  In that case, dotnet-monitor needs a solution for specifying where the dump should be created (from the target application's perspective).  The location of the dump is sent via the Diagnostic Port to the target process.  The requested dump location is resolved by the target application.  If the shared file mount is not at the same path in both the target container and the dotnet-monitor container, then dotnet-monitor won't be able to access the dump.  dotnet-monitor needs to be made aware of the relative path difference of the shared file mount between the two containers.  This becomes complex if there are more than 2 containers in a Pod and each container has mounted the shared file mount to a different path.

CC @wiktork @noahfalk @jander-msft @shirhatti

jander-msft:
At this time, allowing the definition of where the shared dump location exists between target application containers vs the .NET Monitor container will not be added. The current expectation, at is has been since the beginning of the tool offering, is that the volume is mounted at the same location in all involved containers.

PR #104, opened 2021-03-23
Configure temporary storage options for dump operation

wiktork:
Also add warning if we upgrade metrics endpoint to https

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.RestServer/StorageOptions.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.RestServer/StorageOptions.cs
@@ -1,9 +1,17 @@
-﻿using System;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 
+#if UNITTEST
+namespace Microsoft.Diagnostics.Monitoring.UnitTests.Options
+#else
 namespace Microsoft.Diagnostics.Monitoring.RestServer
+#endif
 {
     internal class StorageOptions
     {
--- a/src/Microsoft.Diagnostics.Monitoring.RestServer/StorageOptions.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.RestServer/StorageOptions.cs
@@ -1,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Monitoring.RestServer
+{
+    internal class StorageOptions
+    {
+        public string DumpTempFolder {get; set; }
+    }
+}
--- a/src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
+++ b/src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
@@ -217,7 +217,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor
                         //non-sensitive data such as metrics. We may be missing a certificate for https binding. We want to continue with the
                         //http binding in that scenario.
                         metricUrls = metricsOptions.Enabled.GetValueOrDefault(MetricsOptionsDefaults.Enabled) ?
-                            ProcessMetricUrls(metricUrls, metricsOptions) :
+                            ProcessMetricUrls(metricUrls, metricsOptions, listenResults) :
                             Array.Empty<string>();
 
                         listenResults.Listen(options, urls, metricUrls);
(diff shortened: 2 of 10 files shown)


src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
```diff
@@ -252,13 +255,24 @@ private static string[] ProcessMetricUrls(string[] metricUrls, MetricsOptions me
                             ":",
                             metricUrl.Port.ToString(System.Globalization.CultureInfo.InvariantCulture),
                             metricUrl.PathBase);
+
+                        results.Warnings.Add($"Upgraded {metricUrl} to {metricUrls[i]} due to custom metrics. To host custom metrics over http " +
```
jander-msft, line 259: This message should be moved to `LoggingExtensions` with a new `EventId`. If you take the structured information change in `AddressListenResults`, this becomes just adding a new instance of `UrlModification`.

src/Tools/dotnet-monitor/Startup.cs
```diff
@@ -98,6 +98,11 @@ public void ConfigureServices(IServiceCollection services)
         {
             exprLogger.LogExperimentMessage();
 
+            foreach (string warning in listenResults.Warnings)
+            {
+                logger.LogWarning(warning);
```
jander-msft, line 103: This should be moved to `LoggingExtensions` so that we can give it an `EventId`; maybe call it `UpgradeMetricsUrl`.

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
jander-msft: Optionally, the second part of the message ("To host custom metrics...") can be split into its own logging message so that if there is more than one upgraded URL we don't repeat the same message.

src/Microsoft.Diagnostics.Monitoring.RestServer/StorageOptions.cs
```diff
@@ -0,0 +1,20 @@
...
+
+#if UNITTEST
+namespace Microsoft.Diagnostics.Monitoring.UnitTests.Options
+#else
+namespace Microsoft.Diagnostics.Monitoring.RestServer
+#endif
+{
+    internal class StorageOptions
+    {
+        public string DumpTempFolder {get; set; }
```
jander-msft, line 18: This will be really useful in partially alleviating #12 such that containers don't need to share their /tmp when using listening mode in dotnet-monitor, but they'll still need to mount an intermediate volume; at least now it can just share the dumps instead of everything in /tmp on every container.

jander-msft (changes requested):

src/Microsoft.Diagnostics.Monitoring.RestServer/StorageOptions.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+
+#if UNITTEST
+namespace Microsoft.Diagnostics.Monitoring.UnitTests.Options
+#else
+namespace Microsoft.Diagnostics.Monitoring.RestServer
+#endif
+{
+    internal class StorageOptions
```
jander-msft, line 16: Do we want to add another option to allows specifying a dump location in the "remote" container that is different in the dotnet-monitor container so that way we don't have to force a symmetry of the dump location in both containers? Maybe `public string RemoteDumpTempFolder { get; set; }`?

wiktork: Given the common usage for containers, I don't think we would need something like this.

src/Tools/dotnet-monitor/LoggingExtensions.cs
```diff
@@ -111,6 +112,18 @@ internal static class LoggingExtensions
                 logLevel: LogLevel.Debug,
                 formatString: "Bound metrics address: {address}");
 
+        private static readonly Action<ILogger, Exception> _metricUrlsUpdated =
+            LoggerMessage.Define(
+                eventId: new EventId(17, "MetricUrlsUpdated"),
```
jander-msft, line 117: Needs to be 18

src/Tools/dotnet-monitor/LoggingExtensions.cs
```diff
@@ -111,6 +112,18 @@ internal static class LoggingExtensions
...
+                logLevel: LogLevel.Warning,
+                formatString: $"Metric bindings changed. To host custom metrics over http set {ConfigurationHelper.MakeKey(ConfigurationKeys.Metrics, nameof(MetricsOptions.AllowInsecureChannelForCustomMetrics))} to {true}");
+
+        private static readonly Action<ILogger, string, string, Exception> _metricUrlUpdated =
+            LoggerMessage.Define<string, string>(
+                eventId: new EventId(18, "MetricUrlUpdated"),
```
jander-msft, line 123: Needs to be 19

jander-msft (approved):

PR #107, opened 2021-03-24
Add unit test application and add tests for /processes routes.

jander-msft:
- Refactor common runner logic into base class.
- Initial test app with additional logging.
- Add GET /processes/{processKey} testing.
- Use ConfigureAwait(false) to help avoid async deadlocks.


src/Microsoft.Diagnostics.Monitoring.RestServer/Models/ProcessInfo.cs
```diff
@@ -16,28 +20,15 @@ public class ProcessInfo
         public Guid Uid { get; set; }
 
         [JsonPropertyName("name")]
-        public string Name { get; private set; }
+        public string Name { get; internal set; }
 
         [JsonPropertyName("commandLine")]
-        public string CommandLine { get; private set; }
+        public string CommandLine { get; internal set; }
```
wiktork, line 26: Since this is really just a plain old object, you could set these to public.

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/TestAppScenarios.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public static class TestAppScenarios
+    {
+        public static class Commands
+        {
+            public const string EndScenario = nameof(EndScenario);
+            public const string StartScenario = nameof(StartScenario);
+        }
+
+        public static class SpinWait
```
wiktork, line 15: nit: Rename since the name clashes with a bcl type.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/LoggingExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Extensions.Logging;
+using System;
+
+namespace Microsoft.Diagnostics.Monitoring.UnitTestApp
+{
+    internal static class LoggingExtensions
+    {
+        private static readonly Action<ILogger, Exception> _scenarioReady =
```
wiktork, line 12: While I understand this is objectively better, is it necessary for our tests? It's a lot simpler to just LogInformation.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Program.cs
```diff
@@ -0,0 +1,78 @@
...
+            return new CommandLineBuilder()
+                .AddCommand(SpinWaitScenario.Command())
+                .UseDefaults()
+                .Build()
+                .InvokeAsync(args);
+        }
+
+        public static async Task<int> RunScenarioAsync(Func<ILogger, Task<int>> func, CancellationToken token)
+        {
+            // Create JSON console logger so that app can communicate with test host
```
wiktork, line 30: This is much better than my protocol of sending '0' and '1'

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Program.cs
```diff
@@ -0,0 +1,78 @@
...
+            // with strucutured responses.
```
wiktork, line 31: nit: typo

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Program.cs
```diff
@@ -0,0 +1,78 @@
...
+            using ServiceProvider hostServices = new ServiceCollection()
+                .AddLogging(builder =>
+                {
+                    builder.AddFilter(typeof(Program).FullName, LogLevel.Debug)
```
wiktork, line 35: It would be nice to push all the helper functions here to a separate class, instead of Program. I know it's very aspirational but maybe at some point we can unify and redist these test runners as one package across diag/dotnet-monitor repos.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Program.cs
```diff
@@ -0,0 +1,78 @@
...
+                .CreateLogger<Program>();
+
+            logger.ScenarioReady();
+
+            // Wait for test host before executing scenario
+            await WaitForCommandAsync(TestAppScenarios.Commands.StartScenario, logger);
+
+            logger.ScenarioExecuting();
+
+            int result = await func(logger);
```
wiktork, line 53: I found it handy to try/catch in most of the infrastructure code on the app side and spit it out to stderror. Then the host could print it, for easier debugging.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/ProcessTests.cs
```diff
@@ -0,0 +1,244 @@
...
+                appProcessId = appRunner.ProcessId;
+
+                await appRunner.SendStartScenarioAsync(DefaultTimeout);
+
+                await VerifyProcessAsync(client, await client.GetProcessesAsync(DefaultTimeout), appProcessId);
+
+                await EndSpinWaitScenarioAsync(appRunner);
+            }
+
+            await Task.Delay(TimeSpan.FromSeconds(1));
```
wiktork, line 89: Maybe guard this around a method that we can change later + a TODO

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/Runners/AppRunner.cs
```diff
@@ -0,0 +1,178 @@
...
+                }
+
+                environment.Add("DOTNET_DiagnosticPorts", DiagnosticPortPath);
+            }
+
+            return environment;
+        }
+
+        protected override void OnStandardOutputLine(string line)
+        {
```
wiktork, line 89: As mentioned before, we should also be logging stderr

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/TestAppScenarios.cs
wiktork, line 15: Do you we have to use the spin scenario for our test apps? It seems incredibly cpu intensive, but i'm not sure what benefit it provides for tests such as /processes.

jander-msft: I can create a non spinning wait, more of an async wait. We'll probably want a spin wait for testing high cpu usage in metrics and eventually triggers. I'll change this to an async wait scenario.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/ProcessTests.cs
jander-msft: I fixed it to wait for the diagnostic pipes to no longer be reported.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Program.cs
jander-msft: Fixed

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/Runners/AppRunner.cs
jander-msft: Fixed

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/TestAppScenarios.cs
jander-msft: Changed to async wait.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Program.cs
jander-msft: Done

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/LoggingExtensions.cs
jander-msft: I simplified this a bit. We're using the EventId in the test host to be able to identify the different event types, so it's better to have those IDs consolidated into one file.

jander-msft:
The Linux and MacOS tests are hanging on waiting for the diagnostic pipes to be cleaned up. I believe the issue is that the tests are killing the app processes rather than waiting for them to exit, which causes the Unix socket files to be left behind. I'll try to have the tests wait for the processes to exit. If that doesn't work, I'll revert the behavior and use Tast.Delay for now.

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/LoggingRunnerAdapter.cs
```diff
@@ -0,0 +1,124 @@
...
+        private readonly List<string> _standardErrorLines = new();
+        private readonly List<string> _standardOutputLines = new();
+
+        private bool _isDiposed;
+        private Task _standardErrorTask;
+        private Task _standardOutputTask;
+
+        public Dictionary<string, string> Environment { get; } = new();
+
+        public Action<string> StandardErrorCallback { get; set; }
```
wiktork, line 28: Nit: did you want these to be events rather than delegates?

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/LoggingRunnerAdapter.cs
```diff
@@ -0,0 +1,124 @@
...
+                _outputHelper.WriteLine("Exit Code: {0}", _runner.ExitCode);
+            }
+            return _runner.ExitCode;
+        }
+
+        private static async Task ReadLinesAsync(StreamReader reader, List<string> lines, Action<string> callback, CancellationToken token)
+        {
+            try
+            {
+                await foreach (string line in reader.ReadLinesAsync(token))
```
wiktork, line 113: Another option here is to register for cancellation and simply dispose the underlying reader (at least i would expect that to work, i have not tried it)

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/ProcessTests.cs
```diff
@@ -0,0 +1,275 @@
...
+            Assert.NotNull(info);
+            Assert.Equal(identifier.Pid, info.Pid);
+            Assert.Equal(identifier.Uid, info.Uid);
+#endif
+        }
+
+        /// <summary>
+        /// Has the SpinWait scenario end.
+        /// </summary>
+        private static async Task EndSpinWaitScenarioAsync(AppRunner runner)
```
wiktork, line 246: Rename?

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/Runners/AppRunner.cs
```diff
@@ -0,0 +1,196 @@
...
+        /// </summary>
+        private static string CurrentExecutingAssemblyPath =>
+            Assembly.GetExecutingAssembly().Location;
+
+        /// <summary>
+        /// The path to the application.
+        /// </summary>
+        private static string AppPath =>
+            CurrentExecutingAssemblyPath
+                .Replace(Assembly.GetExecutingAssembly().GetName().Name, "Microsoft.Diagnostics.Monitoring.UnitTestApp");
```
wiktork, line 46: Nit: There's a subtle bug here if you have a subfolder with the same name as the assembly. Consider Path.Combine(Path.GetDirectoryName(...),...)

jander-msft: We need to do replace because this binary is at:

`artifacts\bin\Microsoft.Diagnostics.Monitoring.UnitTests\Debug\net5.0\Microsoft.Diagnostics.Monitoring.UnitTests.dll`

which needs to be transformed to:

`artifacts\bin\Microsoft.Diagnostics.Monitoring.UnitTestApp\Debug\net5.0\Microsoft.Diagnostics.Monitoring.UnitTestApp.dll`

So, we are intentionally replacing the folder name as well as the file name. I don't think there will be another folder that coincidentally has the same name, unless we really go out of our way to bin-place into particular folders.

If we did build up the path, we'd have to start at `artifacts\bin`, append the new binary name, append the configuration (which we don't have available in C# without a bunch of #ifdef), append the TFM (which we don't have available in C# without a bunch of #ifdef, and then append the binary name again.

wiktork: Thanks for the detailed explanation. I was under the false assumption that execution would combine these into one place for the test run (i.e. the out folder).

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/LoggingRunnerAdapter.cs
jander-msft: Sure? At the end of the all, they would basically be the same, just that this is not multicast (by default).

wiktork (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/Runners/AppRunner.cs
jander-msft: > Thanks for the detailed explanation. I was under the false assumption that execution would combine these into one place for the test run (i.e. the out folder).

For .NET Framework, that is the case. For .NET Core, they are executed in-place.

PR #110, opened 2021-03-25
Use HttpClientFactory shared among tests to avoid socket exhaustion.

jander-msft:
There is a problem in the tests where HttpClient.SendAsync will hang and only return when cancelled. This happens due to exhaustion of the available sockets for use by HttpClientHandler.

To mitigate this, create an HttpClientFactory (via the AddHttpClient service method in the ServiceProviderFixture) which pools HttpClientHandlers. The ServiceProviderFixture provides for two HttpCliendHandlers: a default one and one that authenticates with the current user credentials. All of the tests use the default one except for NegotiateAuthenticationSchemeTest which uses the one with default credentials.

Additional refactoring to the ApiClient was done to remove the privately created HttpClient since it shouldn't encourage the use of separately instantiated HttpClient instances.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
@@ -3,9 +3,11 @@
 // See the LICENSE file in the project root for more information.
 
 using Microsoft.Diagnostics.Monitoring.TestCommon;
+using Microsoft.Diagnostics.Monitoring.UnitTests.Fixtures;
 using Microsoft.Diagnostics.Monitoring.UnitTests.HttpApi;
 using Microsoft.Diagnostics.Monitoring.UnitTests.Options;
 using Microsoft.Diagnostics.Monitoring.UnitTests.Runners;
+using Microsoft.Extensions.DependencyInjection;
 using System;
 using System.Net;
 using System.Net.Http;
@@ -15,20 +17,22 @@ using System.Security.Cryptography;
 using System.Threading.Tasks;
 using Xunit;
 using Xunit.Abstractions;
-using Xunit.Extensions;
 
 namespace Microsoft.Diagnostics.Monitoring.UnitTests
 {
+    [Collection(DefaultCollectionFixture.Name)]
     public class AuthenticationTests
     {
         private static readonly TimeSpan DefaultTimeout = TimeSpan.FromMinutes(1);
 
         private const string ApiKeyScheme = "MonitorApiKey";
 
+        private readonly IHttpClientFactory _httpClientFactory;
         private readonly ITestOutputHelper _outputHelper;
 
-        public AuthenticationTests(ITestOutputHelper outputHelper)
+        public AuthenticationTests(ITestOutputHelper outputHelper, ServiceProviderFixture serviceProviderFixture)
         {
+            _httpClientFactory = serviceProviderFixture.ServiceProvider.GetService<IHttpClientFactory>();
             _outputHelper = outputHelper;
         }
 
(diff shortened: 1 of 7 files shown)


jander-msft:
I've now run this several times through the PR build and run it locally many times and have not encountered the
```
System.Threading.Tasks.TaskCanceledException : The operation was canceled.\r\n---- System.IO.IOException : Unable to read data from the transport connection: The I/O operation has been aborted because of either a thread exit or an application request..\r\n-------- System.Net.Sockets.SocketException : The I/O operation has been aborted because of either a thread exit or an application request.
```
issue at all so I believe it is fixed with this change.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
```diff
@@ -43,19 +47,21 @@ public async Task DefaultAddressTest()
             
             await toolRunner.StartAsync(DefaultTimeout);
 
-            using ApiClient client = new(_outputHelper, await toolRunner.GetDefaultAddressAsync(DefaultTimeout));
+            using HttpClient httpClient = _httpClientFactory.CreateClient();
```
jander-msft, line 50: I'm thinking I can create an extension method that combines creation to the http client and setting the base address in order to minimize the boilerplate code.

shirhatti:
I haven't been following your investigation too closely, but I'm curious how you ran into socket exhaustion.

`SocketsHttpHandler` does connection pooling by default for this reason. I wonder if fiddling with the `PooledConnectionIdleTimeout` might have solved your problem?

hoyosjs (approved):

wiktork (approved):

jander-msft:
> I haven't been following your investigation too closely, but I'm curious how you ran into socket exhaustion.

Each test would create a new HttpClient and attempt to send with it. Fairly frequently, several of them would hang deep in HttpCient.SendAsync code path, the cancellation token would kick in after one minute, and the following would be thrown.

Error:
```
System.Threading.Tasks.TaskCanceledException : The operation was canceled.\r\n---- System.IO.IOException : Unable to read data from the transport connection: The I/O operation has been aborted because of either a thread exit or an application request..\r\n-------- System.Net.Sockets.SocketException : The I/O operation has been aborted because of either a thread exit or an application request.
```
Stack Trace:
```
   at System.Net.Http.HttpConnection.SendAsyncCore(HttpRequestMessage request, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.SendWithNtConnectionAuthAsync(HttpConnection connection, HttpRequestMessage request, Boolean doRequestAuth, CancellationToken cancellationToken)
   at System.Net.Http.HttpConnectionPool.SendWithRetryAsync(HttpRequestMessage request, Boolean doRequestAuth, CancellationToken cancellationToken)
   at System.Net.Http.RedirectHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at System.Net.Http.HttpClient.FinishSendAsyncBuffered(Task`1 sendTask, HttpRequestMessage request, CancellationTokenSource cts, Boolean disposeCts)
   at Microsoft.Diagnostics.Monitoring.UnitTests.HttpApi.ApiClient.SendAndLogAsync(HttpRequestMessage request, CancellationToken token) in /_/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/HttpApi/ApiClient.cs:line 200
   at Microsoft.Diagnostics.Monitoring.UnitTests.HttpApi.ApiClient.GetProcessesAsync(CancellationToken token) in /_/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/HttpApi/ApiClient.cs:line 59
   at Microsoft.Diagnostics.Monitoring.UnitTests.HttpApi.ApiClient.GetProcessesAsync(TimeSpan timeout) in /_/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/HttpApi/ApiClient.cs:line 84
   at Microsoft.Diagnostics.Monitoring.UnitTests.ProcessTests.SingleProcessCore(DiagnosticPortConnectionMode monitorConnectionMode) in /_/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/ProcessTests.cs:line 92
   at Microsoft.Diagnostics.Monitoring.UnitTests.ProcessTests.SingleProcessCore(DiagnosticPortConnectionMode monitorConnectionMode) in /_/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/ProcessTests.cs:line 95
--- End of stack trace from previous location where exception was thrown ---
----- Inner Stack Trace -----
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken)
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.GetResult(Int16 token)
   at System.Net.Http.HttpConnection.SendAsyncCore(HttpRequestMessage request, CancellationToken cancellationToken)
----- Inner Stack Trace -----
```

Did some quick searching and found that using too many HttpClient clients can cause this issue and that using pooling should alleviate it.

> `SocketsHttpHandler` does connection pooling by default for this reason. I wonder if fiddling with the `PooledConnectionIdleTimeout` might have solved your problem?

When searching, most of the suggestions I was getting was to use IHttpClientFactory since it handles pooling automatically and allows for some other configurability. I didn't run across using SocketsHttpHandler.PooledConnectionIdleTimeout, but good to know. I think I'll keep the code as-is since it does seem to work and it can allow us to provide other services the tests need in the future.

shirhatti:
> When searching, most of the suggestions I was getting was to use IHttpClientFactory since it handles pooling automatically and allows for some other configurability. I didn't run across using SocketsHttpHandler.PooledConnectionIdleTimeout, but good to know. I think I'll keep the code as-is since it does seem to work and it can allow us to provide other services the tests need in the future.

Unfortunately, we're very aware of this. There's a lot of FUD on how to use HttpClient out there. WinHttpHandler and ergo HttpClient on Framework doesn't have pooling. `SocketsHttpHandler` only recently (I think 2.1) introduced connection pooling and it's hard to beat out 15 years of bad advice.

HttpClientFactory is great and by all means use it 😄. I just don't want good old `HttpClient` users to continue to suffer  because we something hard to use. @samsp-msft might be interested in your struggles.

PR #113, opened 2021-03-26
Set package and informational versions to the version prefix and optional prerelease label

jander-msft:
If the current main branch were to be shipped as the final release for 5.0.0, the following versions would be produced:
- Assembly: 5.0.0.0
- AssemblyFile: 5.0.21.17602
- AssemblyInformational: 5.0.217602+0a3cbd909e6c089eb78b0e2f4ee80ef156980288
- Package: 5.0.217602

`dotnet monitor --version` will return `5.0.217602+0a3cbd909e6c089eb78b0e2f4ee80ef156980288`

At least, the package version needs to be fixed such that it represents the actual final version (5.0.0). The assembly informational version should also be fixed similarly so that when a user runs "dotnet monitor --version", they get a version string that corresponds with the publicly expected version.

These fixes are accomplished by resetting the PackageVersion and InformationalVersion properties to be just the original version prefix + the optional prerelease label (e.g. "preview.4"). After these changes, the same build produces:
- Assembly: 5.0.0.0
- AssemblyFile: 5.0.21.17602
- AssemblyInformational: 5.0.0+0a3cbd909e6c089eb78b0e2f4ee80ef156980288
- Package: 5.0.0

`dotnet monitor --version` will return `5.0.0+0a3cbd909e6c089eb78b0e2f4ee80ef156980288`

Prerelease packages are not impacted by this change since Arcade performs the same computation for prerelease versions. For example, `dotnet monitor --version` would return something that looks like `5.0.0-preview.4.21176.2+0a3cbd909e6c089eb78b0e2f4ee80ef156980288` for prerelease versions.

### DIFF
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -2,6 +2,29 @@
 <Project>
   <Import Project="Sdk.targets" Sdk="Microsoft.DotNet.Arcade.Sdk" />
   
+  <!--
+    Set InformationalVersion to only include the original version prefix and any prerelease suffix.
+    It will be updated later with commit hash information. Without this setting to the original
+    version prefix, the version would include a patch number that is computed from the build number,
+    which makes it difficult to trace back to the exact release version. This information is also
+    used when invoking the global version option for tools using System.CommandLine.
+  -->
+  <PropertyGroup>
+    <InformationalVersion>$(_OriginalVersionPrefix)</InformationalVersion>
+    <InformationalVersion Condition="'$(VersionSuffix)' != ''">$(InformationalVersion)-$(VersionSuffix)</InformationalVersion>
+  </PropertyGroup>
+
+  <!--
+    Set PackageVersion to only include the original version prefix and any prerelease suffix.
+    Without this setting to the original version prefix, the version would include a patch number
+    that is computed from the build number, which makes it difficult to trace back to the exact
+    release version. Only set this if PackageVersion was not explicitly set.
+  -->
+  <PropertyGroup Condition="'$(PackageVersion)' == ''">
+    <PackageVersion>$(_OriginalVersionPrefix)</PackageVersion>
+    <PackageVersion Condition="'$(VersionSuffix)' != ''">$(PackageVersion)-$(VersionSuffix)</PackageVersion>
+  </PropertyGroup>
+
   <!-- Work around https://github.com/dotnet/sourcelink/issues/572
   Remove once we build using an SDK that contains https://github.com/dotnet/sdk/pull/10613 -->
   <PropertyGroup>


src/Directory.Build.targets
```diff
@@ -2,6 +2,29 @@
...
+  </PropertyGroup>
+
+  <!--
+    Set PackageVersion to only include the original version prefix and any prerelease suffix.
+    Without this setting to the original version prefix, the version would include a patch number
+    that is computed from the build number, which makes it difficult to trace back to the exact
+    release version. Only set this if PackageVersion was not explicitly set.
+  -->
+  <PropertyGroup Condition="'$(PackageVersion)' == ''">
+    <PackageVersion>$(_OriginalVersionPrefix)</PackageVersion>
```
wiktork, line 24: How often does the originalversionprefix rev? For example, is there going to be a 5.0.1 when we do a small patch? Or would we expect an update for 6.0 only?

jander-msft: `_OriginalVersionPrefix` has the value of `VersionPrefix` in eng/Version.props, so it revs as fast as we want it to. If we do small fixes, it will change because we change `VersionPrefix` (to 5.0.1 in your example). If we do next feature release, it will change because we change `VersionPrefix` (to 5.1.0 or 6.0.0, however we decide to do "the next release").

wiktork (approved):

src/Directory.Build.targets
jander-msft: The `_OriginalVersionPrefix` property is Arcade's way of preserving our stated `VersionPrefix` because at this point in the targets file, Arcade has overwritten `VersionPrefix` with whatever version it decided to calculate. If that is confusing, then I can create a new property only known to our repo, which is set from `VersionPrefix` in the eng/Version.props file, and used here.

hoyosjs, line 24: Can we use a non-private property here? if this doesn't change from what we set originally, we could just generate a copy property for us to use.

Also, `PreReleaseVersionLabel` set to blank implies you want to generate that package version scheme. Usually we don't remove PreReleaseVersionLabel for box releases (set it to rtm or servicing). 

Instead, if you set `IsShipping` for monitor, and set DotNetFinalVersionKind the package would look like `5.0.0`.
So:
- Previews (like the upcoming ones), do nothing and you end up with: 
  - AssemblyFile:  5.0.21.18101
  - PackageVersion: dotnet-monitor.5.0.0-preview.4.21181.1.nupkg
  - AssemblyInformational: 5.0.0-preview.4.21181.1+b223bea10a931df4f993727a593321655d95dd30
  - Assembly: 5.0.0.0
- Releases by setting DotNetFinalVersionKind=release and PreReleaseVersionLabel=non-blank : 
  - AssemblyFile:  5.0.21.18101
  - PackageVersion: dotnet-monitor.5.0.0.nupkg
  - AssemblyInformational: 5.0.0+b223bea10a931df4f993727a593321655d95dd30
  - Assembly: 5.0.0.0

And the `--version` flag just returns whatever the informational version is. Is that enough? That way we don't need any custom logic.

jander-msft: Thanks for that information! Seems to be working as expected, so I'll close this PR. When I was attempting to figure this out previously, it was counter-intuitive to specify `IsShipping=true`, `DotNetFinalVersionKind=true`, and `PreReleaseVersionLabel=<non-empty>`.

hoyosjs: Yeah, prerelease to non-empty is the crazy part... And for release you set `DotNetFinalVersionKind` to `release`. For previews it can remain empty.

PR #129, opened 2021-04-02
Add storage schema

wiktork:
Also fix some bugs

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.RestServer/StorageOptions.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.RestServer/StorageOptions.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Diagnostics.Monitoring.RestServer/StorageOptions.cs
```diff
@@ -15,6 +16,7 @@ namespace Microsoft.Diagnostics.Monitoring.RestServer
 {
     internal class StorageOptions
     {
+        [Display(Description = "The location for temporary artifacts, such as dump files. Defaults to the temp folder.")]
```
jander-msft, line 19: I know I'm being pedantic, but since it is named `DumpTempFolder`, it should be the temp folder ONLY for dump files.

jander-msft (approved):

Issue #127, opened 2021-04-02
Investigate: Does schema.json need a "$schema" property?


jander-msft:
In VS Code, the following warning is shown when using the schema in the settings.json file:
![schema](https://user-images.githubusercontent.com/17344677/113441239-ce4bef00-93a2-11eb-85bd-1cbfa23198dd.png)

This is a benign error and the IntelliSense seems to work.

Should we add a "$schema" property to the schema to avoid the error? Does leaving a "$schema" property in the settings.json file cause issues with the configuration system consuming the settings.json file (if it does, then maybe we don't want to add "$schema" to the schema)?

cc @wiktork, @shirhatti

wiktork:
I believe this is by design. See https://code.visualstudio.com/Docs/languages/json#_mapping-in-the-json

jander-msft:
JSON Schema has $schema as a property of their self-describing schema: https://json-schema.org/draft/2020-12/meta/core

https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.8.1.1 also says that "The "$schema" keyword SHOULD be used in the document root schema object"

jander-msft:
> I believe this is by design. See https://code.visualstudio.com/Docs/languages/json#_mapping-in-the-json

So then would our recommendation be this? https://code.visualstudio.com/Docs/languages/json#_mapping-in-the-user-settings

jander-msft:
I think we should add it to the schema so long as it doesn't mess with the configuration system. I assume that the JSON deserialization is smart enough to know that $schema is not actually part of the objects.

wiktork:
Based on another discussion, we likely want the user to freely merge other configuration into this file anyway, so I am relaxing the schema to allow other properties.

Issue #23, opened 2020-02-26
Create a configuration schema for dotnet-monitor


shirhatti:

shirhatti:
I took an initial stab at it by basing it off the ASP.NET schema. Try:

```json
{
  "$schema": "https://gist.githubusercontent.com/shirhatti/880c71e47dc95d9099df1b240119850f/raw/d72b3d32d6554d9a9ce88ed0dbef06b894476cbe/settings.json"
}
```

shirhatti:
https://aka.ms/dotnet-monitor-schema

shirhatti:
I just realized this schema doesn't include ASP.NET Core schema. No completions for any kestrel server options 😭

jander-msft:
> I just realized this schema doesn't include ASP.NET Core schema. No completions for any kestrel server options 😭

I wonder if we can do something like:
```
{
  "allOf": [
    { "$ref": "<address of asp.net schema so we don't have to do it ourselves>" },
    { <dotnet-monitor specific schema> }
  ]
}
```

for the dotnet-monitor schema.

jander-msft:
I looked at https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/appsettings.json as an example, and it's got a lot more than we'd want (also 3rd party properties) and yet doesn't have enough of the more standard properties, such as console logging and event logging objects. So maybe we do need to maintain our own self-contained schema with the more common ASP.NET Core properties.

wiktork:
I think we're approaching this the wrong way; us maintaining or referencing asp.net schema shouldn't be necessary. We should:
- Rely on tools to support multiple schemas. (And probably make sure we are not restrictive in our schema if users want to combine settings)
- Possibly support more than 1 settings file, to make it easier to configure disparate settings.

wiktork:
Sample multi-schema in vs code settings:
    "json.schemas": [
    {
        "fileMatch": [
            "settings.json"
        ],
        "url": "https://aka.ms/dotnet-monitor-schema",

    },
    {
        "fileMatch": [
            "settings.json"
        ],
        "url": "https://json.schemastore.org/appsettings"
    }
    ],

jander-msft:
I tried searching a bit for how to do something similar in Visual Studio (not Code) and all I've been able to find is specifying $schema in the json file.

kelltrick:
Close this and make a new issue to track the additional kestrel schema work. @wiktork

PR #131, opened 2021-04-02
Change default values for top level properties.

wiktork:
- Also relax schema restrictions
#127 
#23

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
@@ -16,10 +16,23 @@ namespace Microsoft.Diagnostics.Monitoring.ConfigurationSchema
             schema.Id = @"https://www.github.com/dotnet/dotnet-monitor";
             schema.Title = "DotnetMonitorConfiguration";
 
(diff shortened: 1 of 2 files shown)


jander-msft (approved):
If you can, I'd like to see the following have defaults as well:
- Egress.Providers has a default of {}
- Egress.Properties has a default of {}

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
```diff
@@ -16,11 +16,24 @@ public string GenerateSchema()
...
 
+            //Allow other properties in the schema.
+            schema.AdditionalPropertiesSchema = JsonSchema.CreateAnySchema();
+
             //HACK Even though Properties is defined as JsonDataExtension, it still emits a property
             //called 'Properties' into the schema, instead of just specifying additionalProperties.
             //There is likely a more elegant way to fix this.
             schema.Definitions[nameof(EgressProvider)].Properties.Remove(nameof(EgressProvider.Properties));
 
+            //TODO Figure out a better way to add object defaults
```
jander-msft, line 27: Probably could create a ISchemaProcessor that applies this default to any property that is an object and doesn't have a default value. But you don't have to do this now.

PR #154, opened 2021-04-12
Validate that diagnostic port endpoint name is specified if running in listen mode.

jander-msft:
If listen connection mode is specified via configuration without specifying the diagnostic port name, the reversed server will silently fail to bind, thus no applications can be discovered and the user has no indication of the failure.

This change validates that the diagnostic port endpoint name is specified if running in listen mode. If not specified, the tool will fail to startup and the validation failures are sent through logging:

```
21:48:42 crit: Microsoft.Diagnostics.Tools.Monitor.DiagnosticsMonitorCommandHandler[20]
      In 'Listen' mode, the diagnostic port endpoint name must be specified.
```


jander-msft:
@dotnet/dotnet-monitor, if exception stack is too much, I can catch the OptionsValidationException, write out its message, and have the tool exit.

wiktork (approved):

PR #156, opened 2021-04-12
Additional model validation and content type fixes.

jander-msft:
Add content type when custom action result has exception.
Set Bad Request content type only to application/problem+json.

This is a cherry-pick of #148

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.RestServer/ProducesWithProblemDetailsAttribute.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.RestServer/ProducesWithProblemDetailsAttribute.cs
@@ -1,0 +1,46 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Filters;
+using System;
+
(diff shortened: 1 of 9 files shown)


src/Microsoft.Diagnostics.Monitoring.RestServer/ProducesWithProblemDetailsAttribute.cs
```diff
@@ -0,0 +1,46 @@
...
+        public ProducesWithProblemDetailsAttribute(string contentType, params string[] additionalContentTypes)
+            : base(contentType, additionalContentTypes)
+        {
+        }
+
+        public override void OnResultExecuting(ResultExecutingContext context)
+        {
+            if (context.Result is BadRequestObjectResult badRequestResult && badRequestResult.Value is ProblemDetails)
+            {
+                badRequestResult.ContentTypes.Add(RestServer.ContentTypes.ApplicationProblemJson);
```
kelltrick, line 38: Does `[Result].ContentTypes` describe all the MIME types of the current response or all possible responses from this API.

kelltrick (approved):

src/Microsoft.Diagnostics.Monitoring.RestServer/ProducesWithProblemDetailsAttribute.cs
jander-msft: It describes the allowable content types of the returned response instance. There are other attributes that can modify the applicable list of content types (typically its just the ProducesAttribute, but we're overriding the behavior here).

PR #157, opened 2021-04-12
Validate that diagnostic port endpoint name is specified if running in listen mode.

jander-msft:
* Validate that diagnostic port endpoint name is specified if running in listen mode.

* Send options validation failures through logging.

This is a cherry-pick of #154

### DIFF
--- a/src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
+++ b/src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
@@ -11,6 +11,8 @@ using Microsoft.Diagnostics.Monitoring.RestServer;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 using System;
 using System.Collections.Generic;
 using System.CommandLine;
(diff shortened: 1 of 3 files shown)


src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
```diff
@@ -68,13 +70,33 @@ public async Task<int> Start(CancellationToken token, IConsole console, string[]
...
+            catch (OptionsValidationException ex)
+            {
+                host.Services.GetRequiredService<ILoggerFactory>()
+                    .CreateLogger(typeof(DiagnosticsMonitorCommandHandler))
+                    .OptionsValidationFailure(ex);
+                return -1;
+            }
+            finally
+            {
+                if (host is IAsyncDisposable asyncDisposable)
```
kelltrick, line 91: Does C#'s `await using (...)` do this same logic and accept an `IDisposable` or `IAsyncDisposable`, or does it require an `IAsyncDisposable`?

kelltrick (approved):

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
jander-msft: I'll give that a try, but I believe that `await using` requires it to implement IAsyncDisposable.

jander-msft: Results in compilation error CS8417.

PR #160, opened 2021-04-13
Add .NET 6 runtimes and run tests on .NET 6.

### LINKED ISSUE #144: Run tests on .NET 6
Add .NET 6 to the test matrix (running test app in .NET 6, roll-forward dotnet-monitor onto .NET 6 runtime).

jander-msft:
This change allows running tests on .NET 6, which required splitting the `DotNetHost.CurrentRuntimeVersion` property into two (one for the runtime version and one for the ASP.NET version) so that dotnet-monitor can be run on the ASP.NET runtime and the unit test app can be run .NET runtime.

closes #144

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -12,23 +12,23 @@
       <Uri>https://github.com/dotnet/diagnostics</Uri>
       <Sha>c13a550fd8135e031678015743557ea2a543403f</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.DotNet.XUnitExtensions" Version="6.0.0-beta.21209.17">
-      <Uri>https://github.com/dotnet/arcade</Uri>
-      <Sha>28d9452d7e2ae4e98a1df735b90b03d3cac1f4e7</Sha>
-    </Dependency>
-    <Dependency Name="Microsoft.NETCore.App.Runtime.win-x64" Version="6.0.0-preview.4.21211.7">
-      <Uri>https://github.com/dotnet/runtime</Uri>
-      <Sha>355eff52bed00e7ca9d4a6d769ddbe2bbadbea47</Sha>
-    </Dependency>
+  </ProductDependencies>
+  <ToolsetDependencies>
     <Dependency Name="Microsoft.AspNetCore.App.Ref" Version="6.0.0-preview.4.21211.1">
       <Uri>https://github.com/dotnet/aspnetcore</Uri>
       <Sha>1870441efdc00a6c253c2a5c54c20a313fb56ee2</Sha>
     </Dependency>
-  </ProductDependencies>
-  <ToolsetDependencies>
     <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="6.0.0-beta.21209.17">
       <Uri>https://github.com/dotnet/arcade</Uri>
       <Sha>28d9452d7e2ae4e98a1df735b90b03d3cac1f4e7</Sha>
+    </Dependency>
+    <Dependency Name="Microsoft.DotNet.XUnitExtensions" Version="6.0.0-beta.21209.17">
+      <Uri>https://github.com/dotnet/arcade</Uri>
+      <Sha>28d9452d7e2ae4e98a1df735b90b03d3cac1f4e7</Sha>
+    </Dependency>
+    <Dependency Name="VS.Redist.Common.NetCore.SharedFramework.x64.6.0" Version="6.0.0-preview.4.21211.7">
+      <Uri>https://github.com/dotnet/runtime</Uri>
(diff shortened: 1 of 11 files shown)


eng/Versions.props
```diff
@@ -1,5 +1,5 @@
 <Project>
```
jander-msft, line 1: Everything in this file is the same except:

1. I added labels to the PropertyGroups. DARC seems to find the Automated label and add new dependencies to that group.
2. I added the MicrosoftAspNetCoreAppRefVersion and MicrosoftNETCoreAppRuntimewinx64Version property versions via DARC.

Otherwise, everything else has been preserved but rearranged according to the type of dependency it is.

eng/Version.Details.xml
```diff
@@ -16,6 +16,14 @@
       <Uri>https://github.com/dotnet/arcade</Uri>
       <Sha>28d9452d7e2ae4e98a1df735b90b03d3cac1f4e7</Sha>
     </Dependency>
+    <Dependency Name="Microsoft.NETCore.App.Runtime.win-x64" Version="6.0.0-preview.4.21211.7">
```
hoyosjs, line 19: I would move these two and XUnit under toolset - they are not really product impacting dependencies, right?

hoyosjs: I am also trying to remember if these are good packages to depend on as their version becomes stable near the time of releasing a runtime (say, 6.0), but their blob version is unstable. It might be better to depend from something that's always unstable (say, the VS ingestion artifacts).

jander-msft: So you think I should be using something like VS.Redist.Common.NetCore.SharedFramework.x64.6.0 instead of Microsoft.NETCore.App.Runtime.win-x64?

jander-msft: Fixed

hoyosjs: Yes, and an equivalent one for ASP

eng/Version.Details.xml
```diff
@@ -4,23 +4,31 @@
       <Uri>https://github.com/dotnet/command-line-api</Uri>
       <Sha>166610c56ff732093f0145a2911d4f6c40b786da</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.Diagnostics.Monitoring" Version="5.0.0-preview.21212.1">
+    <Dependency Name="Microsoft.Diagnostics.Monitoring" Version="5.0.0-preview.21209.1">
```
jander-msft, line 7: Fix the diagnostics dependencies due to bad merge.

eng/Versions.props
```diff
@@ -1,5 +1,5 @@
...
-  <PropertyGroup>
+  <PropertyGroup Label="Versioning">
     <RepositoryUrl>https://github.com/dotnet/dotnet-monitor</RepositoryUrl>
     <VersionPrefix>5.0.0</VersionPrefix>
```
kelltrick, line 4: should we rev this?

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/Runners/MonitorRunner.cs
```diff
@@ -261,7 +267,11 @@ public async Task WriteUserSettingsAsync(RootOptions options, CancellationToken
 
             JsonSerializerOptions serializerOptions = new()
             {
```
kelltrick, line 269: I couldn't find it but isn't there a particular setting in `JsonSerializerOptions` that is required security?

kelltrick: Wait, nevermind, this is only a serializer, the rules don't apply.

kelltrick (approved):

eng/Version.Details.xml
jander-msft: Fixed (again) :-)

eng/Versions.props
jander-msft: We're still developing the 5.0.0 version; what is getting changed for each release is the PreReleaseVersionIteration property. The release branch has it set to 4 whereas the main branch has it set to 5.

hoyosjs (approved):
LGTM. The only rough edge is it isn't obvious in the test result what variant passes/fails.

jander-msft:
> LGTM. The only rough edge is it isn't obvious in the test result what variant passes/fails.

Yeah, I was thinking about that too. I might update the yaml to specify a custom title for each tested framework (that would require filtering testing per framework version). Alternatively, add the framework version to each test (that seems a bit messy to me and easily forgettable). I can also add the information in the test output, but that requires digging into the test results to find that information.

PR #172, opened 2021-04-14
Add process environment tests and dump tests

jander-msft:
Summary of changes
- Add process environment tests and dump tests
- Don't fail test when detecting segfault on OSX when requesting dump.
- Add extension methods for executing the app runner to remove boilerplate code from tests.
- Refactor all test timeouts into consolidated class.
- Verify test app exit codes.


src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/DumpTests.cs
```diff
@@ -0,0 +1,128 @@
...
+                    // file content.
+                    Assert.True(total > 10_000);
+
+                    await appRunner.SendCommandAsync(TestAppScenarios.AsyncWait.Commands.Continue, TestTimeouts.SendCommand);
+                });
+            }
+            catch (ApiStatusCodeException ex) when (IsSegFaultOnOSX(appRunner, ex.StatusCode))
+            {
+                // Don't fail the test if the test app segfaults due to calling the dump commmand.
+                _outputHelper.WriteLine("WARNING: Test app segfaulted while producing dump type '{0}'.", type);
```
jander-msft, line 111: @josalem, I've been seeing on occasion that asking the .NET 5 runtime to produce to a dump on MacOS will segfault the runtime occasionally. Is this a known issue?

josalem: I'm not aware of one.  @mikem8361 are you aware of anything?

hoyosjs: Not something we've seen.

jander-msft: I'm trying to think if there's a way that I can collect the core dump of the process as part of the test; I would remove this guard to allow the test to fail so we can get that data.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
```diff
@@ -195,14 +192,14 @@ public async Task ApiKeyAuthenticationSchemeTest()
...
-            await toolRunner.StartAsync(DefaultTimeout);
+            await toolRunner.StartAsync(TestTimeouts.StartProcess);
 
             // Create HttpClient and HttpClientHandler that uses the current
             // user's credentials from the test process. Since dotnet-monitor
             // is launched by the test process, the usage of these credentials
             // should authenticate correctly (except when elevated, which the
             // tool will deny authorization).
-            using HttpClient httpClient = await toolRunner.CreateHttpClientDefaultAddressAsync(_httpClientFactory, DefaultTimeout, ServiceProviderFixture.HttpClientName_DefaultCredentials);
+            using HttpClient httpClient = await toolRunner.CreateHttpClientDefaultAddressAsync(_httpClientFactory, TestTimeouts.HttpApi, ServiceProviderFixture.HttpClientName_DefaultCredentials);
```
wiktork, line 202: nit: Should we create overloads in the client that automatically pass the default timeout

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/DumpTests.cs
```diff
@@ -0,0 +1,128 @@
...
+                        {
+                            total += read;
+                        }
+                        return total;
+                    });
+
+                    _outputHelper.WriteLine("Dump Size: {0} bytes", total);
+                    // Dumps should have at least 10k of data (this is an arbitrary threshold).
+                    // CONSIDER: Is there a better lightweight way to check if the dump is viable?
+                    // e.g. install dotnet-dump and do a brief analysis; or check the header on the
```
wiktork, line 101: +1 on the header idea.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/DumpTests.cs
```diff
@@ -0,0 +1,128 @@
...
+            }
+            Assert.Equal(0, appRunner.ExitCode);
+        }
+
+        private static bool IsSegFaultOnOSX(AppRunner runner, HttpStatusCode? statusCode)
+        {
+            // Requesting a dump of a dotnet process on OSX sometimes causes the process to segfault.
+            // It returns an exit code of 139 (128 + 11), which indicates SIGSEGV. This causes the
+            // HTTP request to fail (500) because the diagnostics connection is terminated.
+            return RuntimeInformation.IsOSPlatform(OSPlatform.OSX)
```
wiktork, line 122: TODO or create issue?

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/HttpApi/ApiStreamHolder.cs
```diff
@@ -0,0 +1,36 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.IO;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Monitoring.UnitTests.HttpApi
+{
+    internal class ApiStreamHolder : IDisposable
```
wiktork, line 12: Maybe just ResponseStreamHolder?

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/HttpApi/ApiClient.cs
```diff
@@ -124,6 +131,112 @@ private async Task<Models.ProcessInfo> GetProcessAsync(string processKey, Cancel
...
+                await SendAndLogAsync(
+                    request,
+                    HttpCompletionOption.ResponseHeadersRead,
+                    token).ConfigureAwait(false));
+
+            switch (responseBox.Value.StatusCode)
+            {
+                case HttpStatusCode.OK:
+                    ValidateContentType(responseBox.Value, ContentTypes.ApplicationOctetStream);
+                    return await responseBox.GetAndReleaseAsync(ApiStreamHolder.CreateAsync).ConfigureAwait(false);
```
wiktork, line 226: Don't you already ConfigureAwait inside GetAndReleaseAsync?

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/HttpApi/ApiStreamHolder.cs
```diff
@@ -0,0 +1,36 @@
...
+        public void Dispose()
+        {
+            _response.Dispose();
+            Stream?.Dispose();
+        }
+
+        public static async Task<ApiStreamHolder> CreateAsync(HttpResponseMessage response)
+        {
+            ApiStreamHolder holder = new(response);
+            holder.Stream = await response.Content.ReadAsStreamAsync();
```
wiktork, line 32: Looking at HttpContent, this may not be necessary. Disposing the response also cleans up the underlying content stream.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/DisposableBox.cs
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        public void Dispose()
+        {
+            Release()?.Dispose();
+        }
+
+        public async Task<U> GetAndReleaseAsync<U>(Func<T, Task<U>> func)
+        {
+            U value = await func(Value).ConfigureAwait(false);
```
wiktork, line 22: Can you add some comments to this class? It would help to understand how it transfers lifetime ownership.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/DisposableBox.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            Value = value ?? throw new ArgumentNullException(nameof(value));
+        }
+
+        public void Dispose()
+        {
+            Release()?.Dispose();
+        }
+
+        public async Task<U> GetAndReleaseAsync<U>(Func<T, Task<U>> func)
```
wiktork, line 20: Given the specific usage of this api, do we have to keep this so generic? For example, could DisposableBox be a wrapper from HttpResponse only, with a direct conversion to ApiStreamHolder? I'm just not sure we would ever use this class for lifetime management of other types.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/HttpApi/ApiClient.cs
jander-msft: We should technically call ConfigureAwait(false) for any Task that is awaited... except for Tasks that are directly in the context of an xUnit test (so that the continuations are resumed on the xUnit SynchronizationContext). So everything should be calling ConfigureAwait(false) unless it is directly in the test method.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/DisposableBox.cs
jander-msft: Sure, I'll either fold it into ApiStreamHolder/ResponseStreamHolder or some other class that is specifically for those types. But if I run into another type that I need to conditionally dispose, then I'll likely refactor it back out into its own class again.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/DumpTests.cs
jander-msft: Done

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/HttpApi/ResponseStreamHolder.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public void Dispose()
+        {
+            // The response disposes the stream when disposed.
+            _response.Dispose();
+        }
+
+        public static async Task<ResponseStreamHolder> CreateAsync(DisposableBox<HttpResponseMessage> responseBox)
+        {
+            using DisposableBox<ResponseStreamHolder> holderBox = new(new(responseBox.Release()));
```
wiktork, line 35: I'm not sure why we need holder box here? Doesn't the lifetime ultimately transfer to ResponseStreamHolder?

jander-msft: It does, but if `holderBox.Value._response.Content.ReadAsStreamAsync().ConfigureAwait(false);` throws an exception, we want to dispose the ResponseStreamHolder (since it owns the HttpResponseMessage at that point. If it doesn't throw, then we want to not dispose ResponseStreamHolder and just return it.

wiktork (approved):

PR #173, opened 2021-04-14
More /processes route notes and clarifications.

jander-msft:
Fix authentication.md links.

### DIFF
--- a/documentation/api/pidvsuid.md
+++ b/documentation/api/pidvsuid.md
@@ -1,0 +1,10 @@
+# Process ID `pid` vs Unique ID `uid`
+
+Many of the HTTP routes allow specifying either the process ID `pid` or the unique ID `uid`. Which one to use depends on the target process and the environment in which the process is running.
+
+The `uid` value of a process is guaranteed to be unique, regardless of environment, as long is the application is running on .NET 5+. Applications running on .NET Core 3.1 will have an empty value of `00000000-0000-0000-0000-000000000000` for `uid`.
+
+Recommendations:
+- For applications running on .NET 5+, use the `uid` parameter. This is especially beneficial when running in Docker or Kubernetes, since containerized .NET applications within the same pod will likely report process IDs of 1.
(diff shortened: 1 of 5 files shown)


documentation/api/pidvsuid.md
```diff
@@ -0,0 +1,10 @@
+# Process ID `pid` vs Unique ID `uid`
+
+Many of the HTTP routes allow specifying either the process ID `pid` or the unique ID `uid`. Which one to use depends on the target process and the environment in which the process is running.
+
+The `uid` value of a process is guaranteed to be unique, regardless of environment, as long is the application is running on .NET 5+. Applications running on .NET Core 3.1 will have an empty value of `00000000-0000-0000-0000-000000000000` for `uid`.
+
+Recommendations:
+- For applications running on .NET 5+, use the `uid` parameter. This is especially beneficial when running in Docker or Kubernetes, since containerized .NET applications within the same pod will likely report process IDs of 1.
+- For applications running on .NET Core 3.1, the `pid` parameter is the only option since .NET Core 3.1 processes have a `uid` of `00000000-0000-0000-0000-000000000000`.
+- If `dotnet-monitor` is **NOT** set up to observe processes across process namespaces, then it is allowable to use the `pid` parameter regardless of the runtime version.
```
wiktork, line 10: This statement is very confusing, since normally the share process namespace with respect to the pod creates unique pids. Having only 1 process also creates a scenario where a pid is acceptable and optional.

wiktork (approved):

documentation/api/pidvsuid.md
jander-msft: It is confusing, but I think its correct. If your containers are using a shared process namespace and dotnet-monitor is one of those containers, then it is not observing across different process namespaces. How about:

"If `dotnet-monitor` is set up to observe processes from only one process namespace (e.g. all concerned containers share a process namespace, the tool is not running in a container, etc), then it is allowable to use the `pid` parameter regardless of the runtime version."

jander-msft: I updated the clarification. Let me know what you think.

Issue #125, opened 2021-04-02
Add new command option to enable configuration output


wiktork:
Add a switch to collect command to enable a configuration display only run. I.e. the user will run the tool but it will only calculate and display configuration instead of actually running the host. 
- We should opt in to configuration paths to prevent accidentally leaking sensitive information.
- Provide an option to display all configuration.

PR #179, opened 2021-04-15
Adding new command to display configuration

wiktork:
Addresses issue #125

``` json
Sample output:
{
  "urls": "https://localhost:52323",
  "Kestrel": {
    "Certificates": {
      "Default": {
        "Path": "/etc/aspnet/customcert/customcert.pfx",
        "Password": "REDACTED"
      }
    }
  },
  "CorsConfiguration": "NOT PRESENT",
  "DiagnosticPort": {
    "ConnectionMode": "Connect",
    "EndpointName": null
  },
  "Metrics": {
    "AllowInsecureChannelForCustomMetrics": "False",
    "Enabled": "True",
    "Endpoints": "http://localhost:52325",
    "IncludeDefaultProviders": "True",
    "MetricCount": "3",
    "UpdateIntervalSeconds": "10"
  },
  "Storage": {
    "DumpTempFolder": "/tmp/"
  },
  "ApiAuthentication": {
    "ApiKeyHash": "REDACTED",
    "ApiKeyHashType": "SHA256"
  },
  "Egress": {
    "Properties": {
      "MonitorBlobAccountKey": "REDACTED"
    },
    "Providers": {
      "monitorBlob": {
        "Type": "azureBlobStorage",
        "AccountUri": "https://wikaks.blob.core.windows.net",
        "BlobPrefix": "artifacts",
        "ContainerName": "dotnet-monitor",
        "CopyBufferSize": "NOT PRESENT",
        "SharedAccessSignature": "NOT PRESENT",
        "AccountKey": "NOT PRESENT",
        "SharedAccessSignatureName": "NOT PRESENT",
        "AccountKeyName": "MonitorBlobAccountKey"
      }
    }
  }
}
```

### DIFF
--- a/src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
+++ b/src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
@@ -1,0 +1,213 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Diagnostics.Tools.Monitor.Egress;
+using Microsoft.Diagnostics.Tools.Monitor.Egress.AzureStorage;
+using Microsoft.Diagnostics.Tools.Monitor.Egress.Configuration;
+using Microsoft.Diagnostics.Tools.Monitor.Egress.FileSystem;
+using Microsoft.Extensions.Configuration;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Text.Json;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    /// <summary>
+    /// Writes out configuration as JSON.
+    /// </summary>
+    internal sealed class ConfigurationJsonWriter : IDisposable
+    {
+        private readonly Utf8JsonWriter _writer;
+        public ConfigurationJsonWriter(Stream outputStream)
+        {
+            JsonWriterOptions options = new() { Indented = true };
+            _writer = new Utf8JsonWriter(outputStream, options);
+        }
+
+        public void Write(IConfiguration configuration, bool full)
+        {
+            //Note that we avoid IConfigurationRoot.GetDebugView because it shows everything
+            //CONSIDER Should we show this in json, since it cannot represent complete configuration?
+            //CONSIDER Should we convert number based names to arrays?
+            //CONSIDER There's probably room for making this code more generalized, to separate the output from the traversal
+            //but it seemed like over-engineering.
+
+            _writer.WriteStartObject();
+
+            ProcessSection(configuration, WebHostDefaults.ServerUrlsKey);
+            IConfigurationSection kestrel = ProcessSection(configuration, "Kestrel", includeChildSections: false);
+            if (kestrel != null)
+            {
+                _writer.WriteStartObject();
+                IConfigurationSection certificates = ProcessSection(kestrel, "Certificates", includeChildSections: false);
+                if (certificates != null)
+                {
+                    _writer.WriteStartObject();
+                    IConfigurationSection defaultCert = ProcessSection(certificates, "Default", includeChildSections: false);
+                    if (defaultCert != null)
+                    {
+                        _writer.WriteStartObject();
+                        ProcessSection(defaultCert, "Path", includeChildSections: false);
+                        ProcessSection(defaultCert, "Password", includeChildSections: false, redact: !full);
+                        _writer.WriteEndObject();
+                    }
+                    _writer.WriteEndObject();
+                }
+                _writer.WriteEndObject();
+            }
+
+            //No sensitive information
+            ProcessSection(configuration, ConfigurationKeys.CorsConfiguration, includeChildSections: true);
+            ProcessSection(configuration, ConfigurationKeys.DiagnosticPort, includeChildSections: true);
+            ProcessSection(configuration, ConfigurationKeys.Metrics, includeChildSections: true);
+            ProcessSection(configuration, ConfigurationKeys.Storage, includeChildSections: true);
+
+            if (full)
+            {
+                ProcessSection(configuration, ConfigurationKeys.ApiAuthentication, includeChildSections: true);
+                ProcessSection(configuration, ConfigurationKeys.Egress, includeChildSections: true);
+            }
+            else
+            {
+                //Do not emit ApiKeyHash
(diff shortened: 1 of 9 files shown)


src/Tools/dotnet-monitor/Program.cs
```diff
@@ -76,6 +77,22 @@ class Program
                 Argument = new Argument<bool>(name: "noAuth", getDefaultValue: () => false)
             };
 
+        private static Option ShowRedactedConfig() =>
```
jander-msft, line 80: I was expecting this to be a new command rather than flags added to the existing collect command. Here, you have, for example:

`dotnet monitor collect <other args> --redacted-config`

whereas I was expecting something like:

`dotnet monitor printconfig <other args> --level redacted`

I personally don't see the configuration printing analogous to a dry-run notion because other tools that use dry-run actually execute whatever the command is without effecting external state whereas this is just printing the understood configuration.

@shirhatti, do you have an opinion on using the existing collect command vs adding a new command (or some other option)?

src/Tools/dotnet-monitor/Program.cs
```diff
@@ -76,6 +77,22 @@ class Program
...
+            new Option(
+                alias: "--redacted-config",
+                description: "Show configuration.")
+            {
+                Argument = new Argument<bool>(name: "redactedConfig", getDefaultValue: () => false)
+            };
+
+        private static Option ShowFullConfig() =>
```
jander-msft, line 88: I'm wondering if these two flags could be combined into a single enum rather than as separate flags. If we happen to add another level of configuration printing, it would be easier and cleaner to add and document a new enum value rather than adding a new flag. Do something like `--printConfig [Redacted|Full]`

src/Tools/dotnet-monitor/Program.cs
jander-msft: Had offline discussion, we landed on something like `dotnet monitor config show --level [Redacted|Full]` where the level parameter is optional.

src/Tools/dotnet-monitor/Program.cs
```diff
@@ -35,6 +35,22 @@ class Program
...
+                  description: "Configuration related commands for dotnet-monitor.")
+              {
+                new Command(
+                  name: "show",
+                  description: "Shows configuration, as if dotnet-monitor collect was executed with these parameters.")
+                {
+                        // Handler
+                        CommandHandler.Create(Delegate.CreateDelegate(typeof(Func<CancellationToken, IConsole, string[], string[], bool, string, bool, ConfigDisplayLevel, Task<int>>),
+                            new DiagnosticsMonitorCommandHandler(), nameof(DiagnosticsMonitorCommandHandler.ShowConfig))),
+                        Urls(), MetricUrls(), ProvideMetrics(), DiagnosticPort(), NoAuth(), ConfigLevel()
```
jander-msft, line 50: I have a feeling we're going to forget to add new options here when adding them to the `collect` command. Is there a way you can refactor this so that they are shared among both commands? I know this is using initializer syntax, which might prevent the sharing, but I think a refactor could save us in the future.

Feel free to not do this now. We can refactor stuff later in the next preview.

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -0,0 +1,213 @@
...
+                _writer.WriteStringValue("NOT PRESENT");
+                return null;
+            }
+
+            ProcessSection(section, includeChildSections, redact);
+
+            return section;
+        }
+
+        private void ProcessSection(IConfigurationSection section, bool includeChildSections = true, bool redact = false)
```
jander-msft, line 177: This looks like it could write a property name without writing a value for it if (1) the section doesn't have any children but it was told include the children, or (2) not to include children but does have children.

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -0,0 +1,213 @@
...
+                            }
+                        }
+                    }
+                    _writer.WriteEndObject();
+                }
+                _writer.WriteEndObject();
+            }
+        }
+
+        private IConfigurationSection ProcessSection(IConfiguration parentSection, string key, bool includeChildSections = true, bool redact = false)
```
jander-msft, line 162: `ProcessChildSection`?

jander-msft (approved):

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
wiktork: Will add comments.

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
```diff
@@ -100,10 +100,20 @@ public async Task<int> Start(CancellationToken token, IConsole console, string[]
             return 0;
         }
 
-        public static IHostBuilder CreateHostBuilder(IConsole console, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth)
+        public Task<int> ShowConfig(CancellationToken token, IConsole console, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, ConfigDisplayLevel level)
```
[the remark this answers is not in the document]
wiktork: I figured it would make it easier to add async code here in the future if necessary.

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -0,0 +1,214 @@
...
+                }
+                _writer.WriteEndObject();
+            }
+            else
+            {
+                if (!children.Any())
+                {
+                    if (redact)
+                    {
+                        _writer.WriteStringValue("REDACTED");
```
jander-msft, line 199: nit: Maybe make the "REDACTED" and "NOT PRESENT" value more likely to not be misinterpreted as actual values or collide with possible usage of "REDACTED" and "NOT PRESENT". Maybe something like:
- "REDACTED" -> "\<redacted>" or "!redacted!" or ":redacted:"
- "NOT PRESENT" -> "\<null>" or "!null!" or ":null:"

Something with delimiters that are more likely to make these not actual values that someone might put into configuration, but avoid $ and # since they have special meaning as JSON references.

wiktork: I initially tried <REDACTED> but this gets escaped n json. I will try some other characters.

wiktork: Using :REDACTED: and :NOT PRESENT:. Note null handling is already built into the system and a null value is different than a missing section.

PR #181, opened 2021-04-15
Additional API docs for dumps

jander-msft:


documentation/api/dump.md
```diff
@@ -0,0 +1,92 @@
...
+GET https://localhost:52323/dump/{uid}?type={type}
+```
+
+or
+
+```http
+GET https://localhost:52323/dump?type={type}
+```
+
+> **NOTE:** Process information (IDs, names, environment, etc) may change between invocations of these APIs. Processes may start or stop between API invocations, causing this information to change.
```
shirhatti, line 23: Do we already have the change merged where we only allow one concurrent dump collection API call? Might be worth calling out the limitation

shirhatti: We should also mention the observability overhead of collecting a dump which results in the entire process being suspended

shirhatti (approved):

Issue #5, opened 2021-02-16
Add warning when dotnet-monitor when running as an elevated processes


shirhatti:
There are security implications when running as an elevated process. We should add a warning (with a preferably an fwlink pointing to a doc).

jander-msft:
@shirhatti, can we define what this warning says this week? I can create the fwlink and point it to #27 for now.

jander-msft:
fwlink: https://go.microsoft.com/fwlink/?linkid=2160204

jander-msft:
Two messages we need to get across:
- Negotiate authentication (includes NTLM) is not enabled when dotnet-monitor is elevated. 
- Elevated dotnet-monitor means the ability to collect artifacts from any .NET Core 3.1+ process regardless of user running it. This should be a warning, not to discourage the use of being elevated, but to make plainly aware the consequences of running elevated.

jander-msft:
Logging goes here: https://github.com/dotnet/dotnet-monitor/blob/main/src/Tools/dotnet-monitor/LoggingExtensions.cs
Caller likely goes here: https://github.com/dotnet/dotnet-monitor/blob/69542c44b7d70a83901b93a73d0ea09b90870bcd/src/Tools/dotnet-monitor/Startup.cs#L91

PR #188, opened 2021-04-16
Add logged warnings for elevated launch on windows

kelltrick:
Resolve issue #5: Add warning when running as an elevated processes
- Adds a warning when launched as admin on windows:
```
11:53:37 warn: Microsoft.Diagnostics.Tools.Monitor.Startup[21]
      The process was launched as an administrator level and will have access to all processes on the system. The dotnet-monitor process should be launched as the same user as the target dotnet process.
```
- Adds warning when launched as admin on windows and NTLM auth is enabled:
```
11:53:37 warn: Microsoft.Diagnostics.Tools.Monitor.Startup[22]
      While running as administrator NTLM authentication is not possible, NTLM has been disabled.
```

### DIFF
--- a/src/Tools/dotnet-monitor/LoggingExtensions.cs
+++ b/src/Tools/dotnet-monitor/LoggingExtensions.cs
@@ -131,6 +131,18 @@ namespace Microsoft.Diagnostics.Tools.Monitor
                 logLevel: LogLevel.Critical,
                 formatString: "{failure}");
 
+        private static readonly Action<ILogger, Exception> _runningElevated =
+            LoggerMessage.Define(
+                eventId: new EventId(21, "RunningElevated"),
+                logLevel: LogLevel.Warning,
+                formatString: "The process was launched elevated and will have access to all processes on the system. Do not run elevated unless you need to monitor processes launched by another user (e.g., IIS worker processes)");
+
+        private static readonly Action<ILogger, Exception> _disabledNegotiateWhileElevated =
+            LoggerMessage.Define(
+                eventId: new EventId(22, "DisabledNegotiateWhileElevated"),
+                logLevel: LogLevel.Warning,
+                formatString: "Negotiate, Kerberos, and NTLM authentication are not enabled when running with elevated permissions.");
+
         public static void EgressProviderAdded(this ILogger logger, string providerName)
         {
             _egressProviderAdded(logger, providerName, null);
@@ -245,6 +257,16 @@ namespace Microsoft.Diagnostics.Tools.Monitor
         {
             foreach (string failure in exception.Failures)
                 _optionsValidationFalure(logger, failure, null);
+        }
+
+        public static void RunningElevated(this ILogger logger)
+        {
+            _runningElevated(logger, null);
+        }
+
+        public static void DisabledNegotiateWhileElevated(this ILogger logger)
+        {
+            _disabledNegotiateWhileElevated(logger, null);
         }
 
         private static string Redact(string value)
--- a/src/Tools/dotnet-monitor/Startup.cs
+++ b/src/Tools/dotnet-monitor/Startup.cs
@@ -21,6 +21,8 @@ using System;
 using System.Collections.Generic;
 using System.IO.Compression;
 using System.Linq;
+using System.Runtime.InteropServices;
+using System.Security.Principal;
 using System.Text.Json.Serialization;
 
 namespace Microsoft.Diagnostics.Tools.Monitor
@@ -130,6 +132,8 @@ namespace Microsoft.Diagnostics.Tools.Monitor
 
             lifetime.ApplicationStarted.Register(() => LogBoundAddresses(app.ServerFeatures, listenResults, logger));
 
+            LogElevatedPermissions(options, logger);
+
             if (options.KeyAuthenticationMode == KeyAuthenticationMode.NoAuth)
             {
                 logger.NoAuthentication();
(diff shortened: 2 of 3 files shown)


src/Tools/dotnet-monitor/LoggingExtensions.cs
```diff
@@ -131,6 +131,18 @@ internal static class LoggingExtensions
                 logLevel: LogLevel.Critical,
                 formatString: "{failure}");
 
+        private static readonly Action<ILogger, Exception> _excessivePermissions =
+            LoggerMessage.Define(
+                eventId: new EventId(21, "ExcessivePermissions"),
+                logLevel: LogLevel.Warning,
+                formatString: "The process was launched at an administrative level and will have access to all processes on the system. The dotnet-monitor process should be launched as the same user as the target dotnet process.");
```
wiktork, line 138: I'm not completely sure this is how we want to phrase this guidance. For example, if the user wants to examine the state of w3wp processes, i believe they are forced to run as admin but they just can't use ntlm auth for that scenario.

src/Tools/dotnet-monitor/Startup.cs
```diff
@@ -209,5 +213,25 @@ private static void LogBoundAddresses(IFeatureCollection features, AddressListen
...
+        private static void LogElevatedPermissions(IAuthOptions options, ILogger logger)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                WindowsIdentity currentUser = WindowsIdentity.GetCurrent();
+                WindowsPrincipal principal = new WindowsPrincipal(currentUser);
+                if (principal.IsInRole(WindowsBuiltInRole.Administrator))
+                {
+                    logger.ExcessivePermissions();
+                    // In the future this will need to be modified when ephemeral keys are setup
```
wiktork, line 226: I think your modification covers the future temporary key scenario as well; in temporary key mode, we should still probably turn on ntlm on windows, but we will disable it in admin mode.

src/Tools/dotnet-monitor/LoggingExtensions.cs
kelltrick: Should this just be `The process was launched as an administrator level and will have access to all processes on the system.`

jander-msft: @shirhatti, do you have any opinion on the wording of this warning?

src/Tools/dotnet-monitor/LoggingExtensions.cs
```diff
@@ -131,6 +131,18 @@ internal static class LoggingExtensions
                 logLevel: LogLevel.Critical,
                 formatString: "{failure}");
 
+        private static readonly Action<ILogger, Exception> _excessivePermissions =
+            LoggerMessage.Define(
+                eventId: new EventId(21, "ExcessivePermissions"),
```
jander-msft, line 136: Let's call this "RunningElevated" so as to not imply that running elevated is bad in some way.

src/Tools/dotnet-monitor/LoggingExtensions.cs
```diff
@@ -131,6 +131,18 @@ internal static class LoggingExtensions
...
+                logLevel: LogLevel.Warning,
+                formatString: "The process was launched at an administrative level and will have access to all processes on the system. The dotnet-monitor process should be launched as the same user as the target dotnet process.");
+
+        private static readonly Action<ILogger, Exception> _ntlmDisabled =
+            LoggerMessage.Define(
+                eventId: new EventId(22, "ExcessivePermissions"),
```
jander-msft, line 142: This event name needs to be unique. I would suggest "DisabledNegotiateWhileElevated".

src/Tools/dotnet-monitor/LoggingExtensions.cs
```diff
@@ -131,6 +131,18 @@ internal static class LoggingExtensions
...
+                logLevel: LogLevel.Warning,
+                formatString: "While running as administrator NTLM authentication is not possible, NTLM has been disabled.");
```
jander-msft, line 144: Suggested rewording: "Negotiate, Kerberos, and NTLM authentication are not enabled when running with elevated permissions."

src/Tools/dotnet-monitor/LoggingExtensions.cs
jander-msft: And rename the field and method to correspond.

src/Tools/dotnet-monitor/LoggingExtensions.cs
jander-msft: And rename the field and method to correspond.

jander-msft:
@wiktork @shirhatti, did we want this for Preview 4 (the upcoming preview)? If so, this should target release/5.0 and then be backported to main.

src/Tools/dotnet-monitor/Startup.cs
kelltrick: That's what I thought, if this doesn't work we can modify `options.EnableNegotiate` if necessary.

shirhatti:
I don't think it's necessary to re-target. Preview5 is fine

src/Tools/dotnet-monitor/LoggingExtensions.cs
shirhatti: Actually phrasing aside, the desired outcome is that a user doesn't run in any elevated context on accident. Simply put, only elevate if you need to.

As for my attempt, I'd probably say 

> The process was launched as an administrator and will have access to all processes on the system. Do not run elevated unless you need to monitor processes launched by other user (e.g., IIS worker processes)

jander-msft (approved):

PR #197, opened 2021-04-19
Additional API docs for /logs

jander-msft:
Minor fix ups for other docs.

### DIFF
--- a/documentation/api/logs.md
+++ b/documentation/api/logs.md
@@ -1,0 +1,109 @@
+# Logs - Get
+
+Captures log statements that are logged to the [ILogger<> infrastructure](https://docs.microsoft.com/aspnet/core/fundamentals/logging) within a specified process.
+
+> **NOTE:** The [`LoggingEventSource`](https://docs.microsoft.com/aspnet/core/fundamentals/logging#event-source) provider must be enabled in the process in order to capture logs.
+
+## HTTP Route
+
+```http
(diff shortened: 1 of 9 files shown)


documentation/api/logs.md
```diff
@@ -0,0 +1,109 @@
...
+
+> **NOTE:** Process information (IDs, names, environment, etc) may change between invocations of these APIs. Processes may start or stop between API invocations, causing this information to change.
+
+## URI Parameters
+
+| Name | In | Required | Type | Description |
+|---|---|---|---|---|
+| `pid` | path | false | int | The ID of the process. |
+| `uid` | path | false | guid | A value that uniquely identifies a runtime instance within a process. |
+| `level` | query | false | [LogLevel](definitions.md#LogLevel) | The name of the log level at which log events are collected. For .NET Core 3.1, the default is `Warning`. For .NET 5+, logs are collected at the levels configured by the application for each logging category; this can be overriden by setting `level`, which will fallback to collecting all categories at or above the specified level. |
```
wiktork, line 33: Did you want to reference UseAppFilter explicitly here? I'm not sure it's documented or an implementation detail.

wiktork (approved):

documentation/api/logs.md
jander-msft: I think that's an implementation detail. The effect is described here, but I don't think UseAppFilters is documented other than in the runtime itself and there's no configuration for it at this time (because it is the default).

When we go to add `POST /logs`, then I think it makes sense to mention it there.

PR #204, opened 2021-04-20
Add /metrics API docs.

jander-msft:
Add "Host Address" sections stating default address and availability information.
Update "Authentication" sections with whether authentication is enforced or not.


documentation/api/metrics.md
```diff
@@ -0,0 +1,164 @@
...
+|---|---|
+| Windows | .NET Core 3.1, .NET 5+ |
+| Linux | .NET Core 3.1, .NET 5+ |
+| MacOS | .NET Core 3.1, .NET 5+ |
+
+## Additional Notes
+
+### Custom Metrics
+
+The metrics providers and counter names to return from this route can be specified via configuration. See [TBD]() for more information.
```
shirhatti, line 164: ## Additional Configuration

### Metrics Urls

In addition to the ordinary diagnostics urls that `dotnet-monitor` binds to, it also binds to metric urls that only expose the `/metrics` endpoint. Unlike the other endpoints, the metrics urls do not require authentication. Unless you enable collection of custom providers that may contain sensitive business logic, it is generally consider safe to expose metrics endpoints. 

Metrics urls can configured via the command line:

```cmd
dotnet monitor collect --metricUrls http://*:52325/
```

Or configured via configuration:

```json
{
  "Metrics": {
    "Endpoints": "http://localhost:52325"
  }
}
```

### Customize collection interval and counts

In the default configuration, `dotnet-monitor` requests that the connected runtime provides updated counter values every 10 seconds and will retain 3 data point for every collected metric. When using a collection tool like Prometheus, it is recommended that you set your scrape interval to `MetricCount` * `UpdateIntervalSeconds`. In the default configuration, we recommend you scrape `dotnet-monitor` for metrics every 30 seconds.

You can customize the number of data points stored per metric and the frequency at which the runtime updates each metric via the following configuration:

```json
{
  "Metrics": {
    "MetricCount": 3,
    "UpdateIntervalSeconds": 10,
  }
}
```

### Custom Metrics

Additional metrics providers and counter names to return from this route can be specified via configuration. 

```json
{
  "Metrics": {
    "Providers": [
      {
        "ProviderName": "Microsoft-AspNetCore-Server-Kestrel",
        "CounterNames": [
          "connections-per-second",
          "total-connections"
        ]
      }
    ]
  }
}
```

#### Allow insecure channel for custom metrics

In the default configuration, enabling of custom metrics changes the default binding address from `http` to `https` for the metrics urls. To continue egressing custom metric via an endpoint with `http` scheme, you explicit opt-in to it. You can do so via the following configuration:

```json
{
  "Metrics": {
    "AllowInsecureChannelForCustomMetrics": true
  }
}
```

### Disable default providers

In addition to enabling custom providers, `dotnet-monitor` also allows you to disable to collection of the default providers. You can do so via the following configuration:

```json
{
  "Metrics": {
    "IncludeDefaultProviders": false
  }
}

Comment (jander-msft) [reply documentation/api/metrics.md:164]: I think all of this should go into the configuration doc you wrote and that the HTTP API really should only document the HTTP aspects. I have a TBD link to point to the metrics section when you merge the configuration PR.

File: documentation/api/processes-list.md [hunk 2]
```diff
@@ -10,14 +10,18 @@ GET /processes HTTP/1.1
 
 > **NOTE:** Process information (IDs, names, environment, etc) may change between invocations of these APIs. Processes may start or stop between API invocations, causing this information to change.
 
+## Host Address
+
+The default host address for these routes is `https://localhost:52323`. This route is only available on the addresses configured via the `--urls` command line parameter and the `DOTNETMONITOR_URLS` environment variable.
```
wiktork, line 15: nit: I'm not sure we need to call this out in every doc.

documentation/api/metrics.md
```diff
@@ -0,0 +1,164 @@
...
+Gets a snapshot of metrics in the Prometheus exposition format of a single process.
+
+The metrics are collected from the following providers by default:
+- `System.Runtime`
+- `Microsoft.AspNetCore.Hosting`
+- `Grpc.AspNetCore.Server`
+
+All of the counters for each of these providers are collected by default.
+
+> **NOTE:** This route collects metrics only from a single process. If there are no processes or more than one process, the endpoint will not return information.
```
wiktork, line 12: I would also add a link to reverse server docs here as a means of constraining how many processes are visible to dotnet-monitor. This acts as a workaround to forcing one process.

wiktork (approved):

documentation/api/processes-list.md
jander-msft: Some doc needs to call out which routes are available on which addresses. This seemed like a natural place.

This is alos not a boilerplate statement; this statement is in contrast to the /metrics.md doc, which calls out that it is available on all addresses.

wiktork: That's a good point, that restriction is not very obvious.

documentation/api/metrics.md
shirhatti: Agreed

shirhatti (approved):

documentation/api/metrics.md
jander-msft: I'll add this once the reversed server docs are added.

PR #205, opened 2021-04-20
Add curl --negotiate usage

shirhatti:


documentation/authentication.md
```diff
@@ -134,6 +134,11 @@ If using PowerShell, be sure to use `curl.exe`, as `curl` is an alias for `Invok
 curl.exe -H "Authorization: MonitorApiKey Uf6Yq8ZGkcn+ltq2QLHxuXpKA6/Q5VH5Mb5aSIJBxRc=" https://localhost:52323/processes
 ```
 
+To use Windows authentication with PowerShell, you can specify the `--negotiate` flag
```
wiktork, line 137: nit: Do you want to tie this specifically to powershell? Does it matter if the user invokes from ps or cmd?

wiktork (approved):

documentation/authentication.md
jander-msft: I think he's trying to subtly say "if you are using PowerShell, use `curl.exe` instead of `curl`.

shirhatti: What Justin said and that I've included the `$(whoami)` bit which only works in PowerShell

PR #217, opened 2021-04-21
Mirror scrips from diagnostic for fixes

hoyosjs:


hoyosjs:
Direct port used in release. Merging as it's purely a mirror need.

PR #248, opened 2021-04-27
Add ability to specify process name to process selection algorithm.

### LINKED ISSUE #48: Proposal: Allow process selection by process name
Current support for process selection in the HTTP API allows for selection by process ID and runtime instance identifier. Add support for selecting by process name. This may not be useful for processes hosted by the default host (dotnet) but is useful for application hosts that are created when building the application e.g. "WebApplication1.exe"

### LINKED ISSUE #73: Add process names to /processes

jander-msft:
This change allows process selection based on the process name.

For example, a request such as

```http
GET /processes/pwsh
```

would yield a response such as

```json
{
    "pid": 13564,
    "uid": "199efe46-595b-4640-b957-1fbf7ee78c26",
    "name": "pwsh",
    "commandLine": "\"C:\\Program Files\\PowerShell\\7\\pwsh.exe\"",
    "operatingSystem": "Windows",
    "processArchitecture": "x64"
}
```

The resolution of zero or more than one process remains unchanged.

Additional changes include adding the process name to the `/processes` route:

```json
[
    {
        "pid": 13564,
        "uid": "199efe46-595b-4640-b957-1fbf7ee78c26",
        "name": "pwsh"
    },
    {
        "pid": 20412,
        "uid": "87596ba0-b92f-47c9-b065-1cb271d41645",
        "name": "dotnet"
    },
    {
        "pid": 21308,
        "uid": "00000000-0000-0000-0000-000000000000",
        "name": "WebApplication3"
    }
]
```

and updating the OpenAPI document with the additional `ProcessName` field on the `ProcessKey` object.

closes #48
closes #73

### DIFF
--- a/documentation/api/logs.md
+++ b/documentation/api/logs.md
@@ -16,7 +16,7 @@ or
 GET /logs/{uid}?level={level}&durationSeconds={durationSeconds} HTTP/1.1
 ```
 
-or (Preview 5+)
+or
 
 ```http
 GET /logs/{name}?level={level}&durationSeconds={durationSeconds} HTTP/1.1
@@ -40,7 +40,7 @@ The default host address for these routes is `https://localhost:52323`. This rou
 |---|---|---|---|---|
 | `pid` | path | false | int | The ID of the process. |
 | `uid` | path | false | guid | A value that uniquely identifies a runtime instance within a process. |
-| `name` | path | false | string | (Preview 5+) The name of the process. |
+| `name` | path | false | string | The name of the process. |
 | `level` | query | false | [LogLevel](definitions.md#LogLevel) | The name of the log level at which log events are collected. For .NET Core 3.1, the default is `Warning`. For .NET 5+, logs are collected at the levels configured by the application for each logging category; this can be overriden by setting `level`, which will fallback to collecting all categories at or above the specified level. |
 | `durationSeconds` | query | false | int | The duration of the log collection operation in seconds. Default is `30`. Min is `-1` (indefinite duration). Max is `2147483647`. |
 | `egressProvider` | query | false | string | If specified, uses the named egress provider for egressing the collected logs. When not specified, the logs are written to the HTTP response stream. See [Egress Providers](../egress.md) for more details. |
--- a/documentation/api/logs.md
+++ b/documentation/api/logs.md
@@ -16,6 +16,12 @@ or
 GET /logs/{uid}?level={level}&durationSeconds={durationSeconds} HTTP/1.1
 ```
 
+or (Preview 5+)
+
+```http
+GET /logs/{name}?level={level}&durationSeconds={durationSeconds} HTTP/1.1
+```
+
 or
 
 ```http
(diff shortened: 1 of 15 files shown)


src/Microsoft.Diagnostics.Monitoring.RestServer/DiagnosticServices.cs
```diff
@@ -50,6 +50,12 @@ public async Task<IEnumerable<IProcessInfo>> GetProcessesAsync(CancellationToken
                 IList<Task<ProcessInfo>> processInfoTasks = new List<Task<ProcessInfo>>();
                 foreach (IEndpointInfo endpointInfo in await _endpointInfoSource.GetEndpointInfoAsync(token))
                 {
+                    // CONSIDER: Can this processing be pushed into the IEndpointInfoSource implementation and cached
```
wiktork, line 53: Outside the reverse scenario, I think we could only reliably cache entries for runtime id based queries.

src/Microsoft.Diagnostics.Monitoring.RestServer/DiagnosticServices.cs
```diff
@@ -141,41 +147,47 @@ public async Task<IProcessInfo> GetProcessAsync(ProcessKey? processKey, Cancella
...
                 case 0:
                     throw new ArgumentException("Unable to discover a target process.");
                 case 1:
-                    return await ProcessInfo.FromEndpointInfoAsync(endpointInfoArray[0]);
+                    return processInfosArray[0];
                 default:
 #if DEBUG
-                    IEndpointInfo endpointInfo = endpointInfoArray.FirstOrDefault(info => string.Equals(Process.GetProcessById(info.ProcessId).ProcessName, "iisexpress", StringComparison.OrdinalIgnoreCase));
-                    if (endpointInfo != null)
+                    IProcessInfo processInfo = processInfosArray.FirstOrDefault(info => string.Equals(Process.GetProcessById(info.EndpointInfo.ProcessId).ProcessName, "iisexpress", StringComparison.OrdinalIgnoreCase));
```
wiktork, line 187: Were you thinking that we would further expand ProcessKey to support other scenarios as well? For example, I might want to specify the metrics default by using command line substring, since the most common process name of dotnet.exe won't work very well.

jander-msft: Sure, we can add a "command line contains the following list of values" filter in the future. I would say that is something that would NOT be available in the type conversion for use in the path/query though.

src/Microsoft.Diagnostics.Monitoring.RestServer/DiagnosticServices.cs
jander-msft: I have a proposal for generating instance cookies for .NET Core 3.1 that is stable for the lifetime of each process based on the diagnostic pipe name components. Then that or the runtime instance cookie can be used for identifying cached items. But we don't have that in place yet.

jander-msft:
Need to update API documentation under /documentation.

documentation/api/logs.md
```diff
@@ -34,13 +40,14 @@ The default host address for these routes is `https://localhost:52323`. This rou
 |---|---|---|---|---|
 | `pid` | path | false | int | The ID of the process. |
 | `uid` | path | false | guid | A value that uniquely identifies a runtime instance within a process. |
+| `name` | path | false | string | (Preview 5+) The name of the process. |
```
wiktork, line 43: nit: I understand we want to call out Preview 5+ now in documentation now since the feature hasn't shipped yet, but I feel like we should remove these at some point.

wiktork (approved):

wiktork:
We may want to call out the limitations of process name in shared framework scenarios in the documentation.

documentation/api/logs.md
shirhatti: Omit the preview5+ in here. Let's version the docs with the code.

Perhaps, we have a separate PR to move the current docs to the release/5.0 branch and then modify the docs as we see fit?

jander-msft: If we do that, then we need to have some better differentiation (maybe with different links) where released bits point to the docs in the release branch rather than the main branch.

shirhatti: We could update the fwlinks to point to docs in the most recent release branch?

shirhatti: The other thing I've found that works well with mardown doc is add a yaml header with metadata like what version the docs address

shirhatti: https://github.com/dotnet/docs/blob/main/docs/core/diagnostics/index.md

PR #256, opened 2021-04-29
Add timestamp to logs

wiktork:
Depends on https://github.com/dotnet/diagnostics/pull/2229

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.RestServer/StreamingLogger.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.RestServer/StreamingLogger.cs
@@ -87,8 +87,7 @@ namespace Microsoft.Diagnostics.Monitoring.RestServer
                 // Matches the format of JsonConsoleFormatter
 
                 jsonWriter.WriteStartObject();
-                // TODO: Write timestamp as string
-                //jsonWriter.WriteString("Timestamp", DateTime.UtcNow.ToString());
+                jsonWriter.WriteString("Timestamp", (state is IStateWithTimestamp stateWithTimestamp) ? FormatTimestamp(stateWithTimestamp): string.Empty);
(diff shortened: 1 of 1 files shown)


noahfalk:
@shirhatti and I happened to be talking about this so I figured its better to put my musings here. Don't take any of this to mean I'm unhappy with the PR as-is, its more a few thoughts on what an ideal/long-term log forwarding pipeline might be.

I'm skeptical if ILogger is best internal interface to be using for log handling for a few reasons:
1. The interface was designed for developers creating logs, not for forwarding pre-existing ones. Timestamps is one problematic example but others would be trying to fake up an exception object, or reconstitute and replay scopes to set up the right context. 
2. Using the interface nearly guarantees that every log message needs to be fully deserialized into a .NET object (and additional scopes?), then that new object is re-serialized into some output format. It's probably not a pressing concern, but if someday we want to get into the _really_ fast/low resource usage territory I suspect we'd have to re-implement without using ILogger. 
3. Its not clear to me how much value we get from being able to plug in any arbitrary ILogger sink. I suspect there are only one or two pre-existing ones we care about and we have their source. Rather than spending effort trying to make our input data look like the ILogger interface they expect (and still losing some fidelity in the reconstruction regardless) it might be just as easy to fork and change them. Even if we do want the ability to plug in arbitrary ILoggers and accept some loss of fidelity we could still do that as an optional transformation rather than defining ILogger as the canonical interface that all log messages flow through.

And again not implying you guys should jump into any major change or that this is a priority. Maybe ILogger is more valuable than I've recognized. At this point its just an idea to think upon and you can let me know at some point if you think it would help : )

wiktork:
>it might be just as easy to fork and change them. Even if we do want the ability to plug in arbitrary ILoggers and accept some loss of fidelity we could still do that as an optional transformation rather than defining ILogger as the canonical interface that all log messages flow through.

I think this is probably the right approach going forward.

PR #299, opened 2021-05-12
Add default process configuration

wiktork:
Sample configuration (matches ConsoleApp24)

```json
  "DefaultProcess":
  {
      "ProcessFilters": [
      {
      "Key": "CommandLine",
      "Value": "App24",
      "MatchType": "Loose"
    }]
  }
```


src/Microsoft.Diagnostics.Monitoring.RestServer/DiagnosticServices.cs
```diff
@@ -114,82 +112,46 @@ private static DumpType MapDumpType(Models.DumpType dumpType)
...
-            var processInfos = await GetProcessesAsync(token);
+            IEnumerable<IProcessInfo> processes = await GetProcessesAsync(token);
 
-            if (processKey.HasValue)
+            if (processFilterConfig != null)
             {
-                return GetSingleProcessInfo(
-                    processInfos,
-                    processKey);
+                processes = processes.Where(p => processFilterConfig.ProcessFilters.All(c => ProcessFilterOptionsUtils.MatchFilter(c, p)));
```
jander-msft, line 121: I think this should be folded into `Task<IEnumerable<IProcessInfo>> GetProcessesAsync(CancellationToken token)` as a extra filtering step and then add the ProcessFilterOptions processFilterConfig parameter to the method. I don't think anything is calling that method other than this one and the separation doesn't seem necessary.

src/Microsoft.Diagnostics.Monitoring.RestServer/ProcessFilterOptions.cs
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    [JsonConverter(typeof(JsonStringEnumConverter))]
+    public enum ProcessFilterType
+    {
+        [Display(Description = "Performs a case-insensitive string comparison.")]
+        Exact,
+
+        [Display(Description = "Performs a case-insensitive substring search.")]
+        Loose,
```
jander-msft, line 34: Subset? Contains?

src/Microsoft.Diagnostics.Monitoring.RestServer/ProcessFilterOptions.cs
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public sealed class ProcessFilterOptions
+    {
+        [Display(Description = "Process filters used to determine the process when collecting metrics. All filters must match.")]
+        public List<ProcessFilterDescriptor> ProcessFilters { get; set; } = new List<ProcessFilterDescriptor>(0);
```
jander-msft, line 40: nit: just call it `Filters`. Given that this property is on a class called `ProcessFilterOptions` and the options in the configuration are under `DefaultProcess`, I don't think repeating Process here adds anything.

src/Microsoft.Diagnostics.Monitoring.RestServer/ProcessFilterOptionsUtils.cs
```diff
@@ -0,0 +1,60 @@
...
+            }
+            if (filter.Key == ProcessFilterKey.Name)
+            {
+                return Compare(filter, processInfo.ProcessName);
+            }
+            if ((filter.Key == ProcessFilterKey.RuntimeId) && (processInfo.EndpointInfo.RuntimeInstanceCookie != Guid.Empty))
+            {
+                return ExactCompare(filter, processInfo.EndpointInfo.RuntimeInstanceCookie.ToString("D"));
+            }
+
```
jander-msft, line 33: Can you Debug.Assert/Fail here so that if we add another filter criteria, it's easier to find this if a mistake was made.

src/Microsoft.Diagnostics.Monitoring.RestServer/ProcessFilterOptionsUtils.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            if (filter.MatchType == ProcessFilterType.Exact)
+            {
+                return ExactCompare(filter, value);
+            }
+            if (filter.MatchType == ProcessFilterType.Loose)
+            {
+                return LooseCompare(filter, value);
+            }
+            return false;
```
jander-msft, line 47: Can you Debug.Assert/Fail here so that if we add another filter type, it's easier to find this if a mistake was made.

jander-msft (changes requested):

src/Microsoft.Diagnostics.Monitoring.RestServer/ProcessFilterOptions.cs
```diff
@@ -0,0 +1,54 @@
...
+#else
+namespace Microsoft.Diagnostics.Monitoring.RestServer
+#endif
+{
+    [JsonConverter(typeof(JsonStringEnumConverter))]
+    public enum ProcessFilterKey
+    {
+        Pid,
+        RuntimeId,
+        Name,
```
jander-msft, line 23: nit: ProcessName, to differentiate from other future values that will contain name e.g. EntrypointName

src/Microsoft.Diagnostics.Monitoring.RestServer/ProcessFilterOptions.cs
```diff
@@ -0,0 +1,54 @@
...
+#if UNITTEST
+namespace Microsoft.Diagnostics.Monitoring.UnitTests.Options
+#else
+namespace Microsoft.Diagnostics.Monitoring.RestServer
+#endif
+{
+    [JsonConverter(typeof(JsonStringEnumConverter))]
+    public enum ProcessFilterKey
+    {
+        Pid,
```
jander-msft, line 21: nit: ProcessId to be more explicit that this is the ID of the process.

src/Microsoft.Diagnostics.Monitoring.RestServer/DiagnosticServices.cs
```diff
@@ -22,27 +23,24 @@ internal sealed class DiagnosticServices : IDiagnosticServices
...
         private static readonly TimeSpan ExtendedProcessInfoTimeout = TimeSpan.FromMilliseconds(500);
 
         private readonly IEndpointInfoSourceInternal _endpointInfoSource;
         private readonly CancellationTokenSource _tokenSource = new CancellationTokenSource();
         private readonly StorageOptions _storageOptions;
+        private readonly ProcessFilterOptions _defaultProcessOptions;
 
-        public DiagnosticServices(IEndpointInfoSource endpointInfoSource, IOptions<StorageOptions> storageOptions)
+        public DiagnosticServices(IEndpointInfoSource endpointInfoSource, IOptions<StorageOptions> storageOptions,
+            IOptions<ProcessFilterOptions> defaultProcessOptions)
```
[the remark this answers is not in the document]
jander-msft: Without reloadability, using RuntimeId in the process filter is probably not going to work very well.

src/Microsoft.Diagnostics.Monitoring.RestServer/DiagnosticServices.cs
```diff
@@ -114,82 +112,46 @@ private static DumpType MapDumpType(Models.DumpType dumpType)
...
-                if (!string.IsNullOrEmpty(processName))
-                {
-                    processInfos = processInfos.Where(info => string.Equals(info.ProcessName, processName, StringComparison.OrdinalIgnoreCase));
-                }
+        private async Task<IProcessInfo> GetProcessAsync(ProcessFilterOptions processFilterConfig, CancellationToken token)
+        {
+            //Short circuit when we are missing default process config
+            if (!processFilterConfig.ProcessFilters.Any())
+            {
+                throw new InvalidOperationException("No default process configuration has been set.");
```
jander-msft, line 144: Is this the message in the HTTP response when the default process is not configured?

src/Microsoft.Diagnostics.Monitoring.RestServer/DiagnosticServices.cs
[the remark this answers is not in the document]
wiktork: I will add reload support here, but I really wouldn't expect anyone to use runtime id in configuration.

src/Microsoft.Diagnostics.Monitoring.RestServer/Controllers/DiagController.cs
```diff
@@ -63,7 +63,7 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
             return this.InvokeService(async () =>
             {
                 IList<Models.ProcessIdentifier> processesIdentifiers = new List<Models.ProcessIdentifier>();
-                foreach (IProcessInfo p in await _diagnosticServices.GetProcessesAsync(HttpContext.RequestAborted))
+                foreach (IProcessInfo p in await _diagnosticServices.GetProcessesAsync(processFilterOptions: null, HttpContext.RequestAborted))
```
jander-msft, line 66: I think we might want to add a "new" route such as `/processes/$default` that will give you the process information of the configured default process.

jander-msft: Maybe leave this future change.

src/Microsoft.Diagnostics.Monitoring.RestServer/DiagnosticServices.cs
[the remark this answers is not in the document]
jander-msft: If we don't think anyone is going to use runtime ID in configuration, then let's not offer it, especially since it has some (minor) special casing for it.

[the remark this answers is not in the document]
wiktork: What is the special casing for it?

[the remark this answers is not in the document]
jander-msft: `if ((filter.Key == ProcessFilterKey.RuntimeId) && (processInfo.EndpointInfo.RuntimeInstanceCookie != Guid.Empty))`

If that wasn't there, then the whole method could be a switch rather that if-if-if-...

src/Microsoft.Diagnostics.Monitoring.RestServer/ProcessFilterOptions.cs
wiktork: I initially intended for these to be vague in case we have some other criteria, but I think that's unlikely anyway.

src/Microsoft.Diagnostics.Monitoring.RestServer/Controllers/DiagController.cs
wiktork: Let's discuss in the other issue.

src/Microsoft.Diagnostics.Monitoring.RestServer/DiagnosticServices.cs
wiktork: Yep

documentation/configuration.md
```diff
@@ -224,6 +224,49 @@ Unlike the other diagnostic artifacts (for example, traces), memory dumps aren't
...
+Match the iisexpress process by name
+
+```json
+{
+  "DefaultProcess":
+  {
+      "Filters": [
+      {
+      "Key": "ProcessName",
+      "Value": "iisexpress.exe",
```
jander-msft, line 248: Are you certain that it's "iisexpress.exe" and not "iisexpress"? I believe dotnet-monitor removes the extension on Windows to be consistent with Process.ProcessName and with Linux/OS X.

documentation/schema.json
```diff
@@ -300,6 +311,77 @@
...
+    "ProcessFilterOptions": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "Filters": {
+          "type": [
+            "array",
+            "null"
+          ],
+          "description": "Process filters used to determine the process when collecting metrics. All filters must match.",
```
jander-msft, line 324: Not just for metrics.

src/Microsoft.Diagnostics.Monitoring.RestServer/DiagProcessFilter.cs
```diff
@@ -0,0 +1,153 @@
...
+
+namespace Microsoft.Diagnostics.Monitoring.RestServer
+{
+    /// <summary>
+    /// There are effectively 3 sets of models for this:
+    /// The configuration model (ProcessFilterOptions), which does not contain criteria such as RuntimeId, and is always string based.
+    /// The REST call model (ProcessKey), which does not contain criteria such as Command Line, and is strongly typed.
+    /// This internal model, which contains all possible criteria and is string based. Only this model is used for the actual comparison.
+    /// </summary>
+    internal sealed class DiagProcessFilter
```
jander-msft, line 20: Why not convert ProcessKey to ProcessFilterOptions and then make of this an implementation detail within the DiagnosicServices class?

wiktork: That's not convertible, since ProcessKey supports RuntimeId while ProcessFilterOptions doesn't.

documentation/configuration.md
wiktork: Good catch.

src/Microsoft.Diagnostics.Monitoring.RestServer/DiagProcessFilter.cs
jander-msft: For sake of argument, if that's the only thing blocking this conversion, ProcessFilterOptions _could_ support RuntimeInstanceCookie and be documented, but in reality just won't be used via configuration very often. That would lower the complexity of this code quite a bit.

wiktork: - I don't think the current complexity is really all that high, it's just conversion from one object model to another.
- We no longer have coupling between ProcessKey and ProcessFilterOptions, we can change them independently.

documentation/configuration.md
```diff
@@ -223,6 +223,47 @@ Unlike the other diagnostic artifacts (for example, traces), memory dumps aren't
...
+|---|---|---|
+| Key | string | Specifies which criteria to match on the process. Can be `ProcessId`, `ProcessName`, `CommandLine`. |
+| Value | string | The value to match against the process. |
+| MatchType | string | The type of match to perform. Can be `Exact` or `Contains` for sub-string matching. Both are case-insensitive.|
+
+### Examples
+
+Match the iisexpress process by name
+
+```json
```
jander-msft, line 240: Please format the examples to be consistent with the rest of the existing documentation.

src/Microsoft.Diagnostics.Monitoring.RestServer/DiagProcessFilter.cs
```diff
@@ -0,0 +1,153 @@
...
+
+            throw new ArgumentException($"Invalid {nameof(processKey)}");
+        }
+
+        private static DiagProcessFilterEntry TransformDescriptor(ProcessFilterDescriptor processFilterDescriptor)
+        {
+            switch (processFilterDescriptor.Key)
+            {
+                case ProcessFilterKey.ProcessId:
+                    return new DiagProcessFilterEntry { Criteria = DiagProcessFilterCriteria.ProcessId, MatchType = DiagProcessFilterMatchType.Exact, Value = processFilterDescriptor.Value };
```
jander-msft, line 64: I think if we are going to only allow exact match (which makes sense), we need to validate that the ProcessFilterDescriptor does the same for ProcessId (not fail the check, but maybe write a warning that states it's going to only use Exact).

src/Microsoft.Diagnostics.Monitoring.RestServer/DiagProcessFilter.cs
```diff
@@ -0,0 +1,153 @@
...
+
+            return false;
+        }
+
+        private bool ExactCompare(string value)
+        {
+            return string.Equals(Value, value, StringComparison.OrdinalIgnoreCase);
+        }
+
+        private bool LooseCompare(string value)
```
jander-msft, line 134: nit: ContainsCompare, SubsetCompare

src/Microsoft.Diagnostics.Monitoring.RestServer/DiagnosticServices.cs
```diff
@@ -114,82 +120,37 @@ private static DumpType MapDumpType(Models.DumpType dumpType)
...
-                }
-
-                string processName = processKey.Value.ProcessName;
-                if (!string.IsNullOrEmpty(processName))
-                {
-                    processInfos = processInfos.Where(info => string.Equals(info.ProcessName, processName, StringComparison.OrdinalIgnoreCase));
-                }
+                throw new InvalidOperationException("No default process configuration has been set.");
             }
+            IEnumerable<IProcessInfo> matchingProcesses = await GetProcessesAsync(processFilterConfig, token);
```
jander-msft, line 145: I would suggest a ToArray or ToList on this so that you avoid multiple evaluations of the underlying LINQ expression. Or have GetProcessesAsync does this.

src/Microsoft.Diagnostics.Monitoring.RestServer/IDiagnosticServices.cs
```diff
@@ -16,8 +16,19 @@ namespace Microsoft.Diagnostics.Monitoring.RestServer
...
     {
-        Task<IEnumerable<IProcessInfo>> GetProcessesAsync(CancellationToken token);
-
+        /// <summary>
+        /// Returns running processes, optionally based on filter criteria.
+        /// </summary>
+        Task<IEnumerable<IProcessInfo>> GetProcessesAsync(DiagProcessFilter processFilter, CancellationToken token);
+
+        /// <summary>
+        /// Returns back a process based on a key. If no key is specified, the DefaultProcess configuration is used.
```
jander-msft, line 25: nit: Don't need the word "back".

src/Microsoft.Diagnostics.Monitoring.RestServer/DiagProcessFilter.cs
jander-msft: I don't think I agree however there's nothing wrong with it, so I won't press further at this time.

jander-msft (approved):

src/Microsoft.Diagnostics.Monitoring.RestServer/DiagProcessFilter.cs
wiktork: I will defer this for now. I would rather do a broader set of ValidationAttribute/ValidationContext for all the config.

PR #305, opened 2021-05-13
Add POST /logs route

### LINKED ISSUE #54: Add the ability to specify log level on a per-category basis
```
GET /logs HTTP/1.1
Host: localhost:52323
User-Agent: curl/7.54.0
Accept: application/x-ndjson
{
  "Default": "Information",
  "Microsoft": "Warning",
  "Microsoft.Hosting.Lifetime": "Information"
}
```

jander-msft:
Add POST /logs route. The body is a object with the following properties:
- `logLevel`: Default log level for any category without a log level.
- `filterSpecs`: The key-value pairs of categories and log levels that should be collected.
- `useAppFilters`: Include the application-defined filters.

Example request:
```http
POST /logs/2584?durationSeconds=15 HTTP/1.1

{
    "filterSpecs": {
        "Microsoft.AspNetCore.Hosting": "Information"
    },
    "useAppFilters": false
}
```

Example response:
```http
HTTP/1.1 200 OK
Content-Type: text/event-stream

﻿data: Information Microsoft.AspNetCore.Hosting.Diagnostics[1]
data: 2021-05-13 18:06:41Z
data: Request starting HTTP/1.1 GET http://localhost:5000/  
data: => RequestId:0HM8M726ENU3K:0000002B, RequestPath:/, SpanId:|4791a4a7-433aa59a9e362743., TraceId:4791a4a7-433aa59a9e362743, ParentId:

﻿data: Information Microsoft.AspNetCore.Hosting.Diagnostics[2]
data: 2021-05-13 18:06:41Z
data: Request finished in 6.8026ms 200 text/html; charset=utf-8
data: => RequestId:0HM8M726ENU3K:0000002B, RequestPath:/, SpanId:|4791a4a7-433aa59a9e362743., TraceId:4791a4a7-433aa59a9e362743, ParentId:
```

closes #54

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.RestServer/Models/LogsConfiguration.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.RestServer/Models/LogsConfiguration.cs
@@ -11,14 +11,23 @@ namespace Microsoft.Diagnostics.Monitoring.RestServer.Models
 {
     public class LogsConfiguration
     {
+        /// <summary>
+        /// The default level at which logs are collected.
+        /// </summary>
         [JsonPropertyName("logLevel")]
         [EnumDataType(typeof(LogLevel))]
         [Required]
         public LogLevel LogLevel { get; set; } = LogLevel.Warning;
 
+        /// <summary>
+        /// The logger categories and levels at which logs are collected. Setting the log level to null will have logs collected from the corresponding category at the level set in the LogLevel property.
+        /// </summary>
         [JsonPropertyName("filterSpecs")]
         public Dictionary<string, LogLevel?> FilterSpecs { get; set; }
 
+        /// <summary>
+        /// Set to true to collect logs at the application-defined categories and levels.
+        /// </summary>
         [JsonPropertyName("useAppFilters")]
         public bool UseAppFilters { get; set; } = true;
     }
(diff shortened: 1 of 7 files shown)


src/Microsoft.Diagnostics.Monitoring.RestServer/Models/LogsConfiguration.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Diagnostics.Monitoring.RestServer.Models
+{
+    public class LogsConfiguration
+    {
+        [JsonPropertyName("logLevel")]
+        [EnumDataType(typeof(LogLevel))]
+        [Required]
```
wiktork, line 16: Shouldn't level be optional? It's not really needed if we useAppfilters

wiktork (approved):

src/Microsoft.Diagnostics.Monitoring.RestServer/Models/LogsConfiguration.cs
jander-msft: It's require and has a default value. This effectively prevents someone from stating:
```json
{
    "logLevel": null
}
```
I don't think this is really a valid scenario. Also, it prevents the following:
```json
{
    "filterSpecs": {
        "Microsoft.AspNetCore.Hosting": null
    },
    "logLevel": null
}
```
Which is most definitely not allowed. Because the category has a null level, it would fallback to the default log level (which is also null).

We cannot use null when setting up the EventPipeProvider. Just based on how event source works, the log level (which is converted to event level) is required.

We could allow for nullability and then provide a default value if it is null, but that's not obvious what that value will be (maybe `LogLevel.Trace`/`EventLevel.Verbose`)

jander-msft: I suppose that since it is not a nullable property, then I can remove the Required attribute.

PR #318, opened 2021-05-17
projected volume config docs

shirhatti:


documentation/configuration.md
```diff
@@ -49,22 +49,70 @@ kubectl create secret generic apikey \
...
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: my-configmap
+data:
+  Metrics__MetricCount: "2"
+```
+
+You can then use a Kubernetes volument mount to supply the configuration map to the container at runtime
```
wiktork, line 79: nit: typo

documentation/configuration.md
```diff
@@ -49,22 +49,70 @@ kubectl create secret generic apikey \
...
+
+```yaml 
+spec:
+  volumes:
+  - name: apikey
```
wiktork, line 84: volume name is wrong?

documentation/configuration.md
```diff
@@ -49,22 +49,70 @@ kubectl create secret generic apikey \
...
+
+Alternatively, you can also configuration maps to specify configuration to the container at runtime.
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: my-configmap
+data:
+  Metrics__MetricCount: "2"
```
wiktork, line 76: I think we should have a more intuitive configuration here, such as Metrics__UpdateIntervalSeconds. The other option would be to demonstrate how an egress provider can separate its secret and non-secret data.

wiktork (approved):

documentation/configuration.md
shirhatti: In general the issue with scenario specific docs (like the egress provider) is that they tend to go stale when the config schema changes. As a first step, let's just have good conceptual docs.

I'm happy to reconsider once we've completed all the basics 😄

Issue #216, opened 2021-04-21
Allow specifying multiple API keys with different authorizations


jander-msft:
It appears that Prometheus does support authorization: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config

This means we can allow securing the `/metrics` endpoint with API key authorization. However, we would likely want to configure a separate API key for metrics that is different than the API key used for the other routes.

Configuration could look like:
```json
{
    "ApiAuthentication": {
        "Hashes": [{
            "Name": "MyKey",
            "Value": "<Base 64 Encoded Hash of Key>",
            "Algorithm": "SHA256",
            "Authorizations": [
                "all"
            ]
        },{
            "Name": "PrometheusKey",
            "Value": "<Base 64 Encoded Hash of Key>",
            "Algorithm": "SHA256",
            "Authorizations": [
                "metrics"
            ]
        }]
    }
}
```

This allows for multiple scenarios:
- Securing the /metrics endpoint from unauthorized access.
- Allowing finer grained preconfigured authorizations. For example, could have authorizations for each route, for different types of egress, etc.
- Ability to log the name of the key that was used for each access to a route.

jander-msft:
.NET Monitor does not currently support authorization scopes except for in the case of Entra ID authentication where a single scope is necessary to allow the use of .NET Monitor for each user. Additionally, there is no plan to add support for multiple API keys as there has not been any customer feedback for this. If fine-grained authorization is added and customers ask for multiple API keys, this can be revisited.

PR #320, opened 2021-05-17
Remove automatic HTTPS upgrade for custom metrics.

### LINKED ISSUE #212: Remove HTTPS upgrade for metrics
## Proposed Feature
When a user connects to http://[addr]:52325/ and receives a custom metrics entry, an https is currently performed. The https upgrade should not be done. Metrics is an unauthenticated endpoint and the https does not protect any data from being egressed, anyone who could man-in-the-middle the http endpoint could simply call the endpoint themselves.

jander-msft:
Automatically upgrading the metrics URLs to HTTPS when custom metrics are present can be confusing to users who are trying to configure custom metrics. The automatic HTTP upgrade doesn't really protect the data, especially since the data from metrics is not sensitive and authentication is not enforced on these URLs. If users do want to protect the data during transport, they can configure the upgrade manually via the command line, file system, or environment variables.

closes #212


documentation/configuration.md
```diff
@@ -237,18 +236,6 @@ Additional metrics providers and counter names to return from this route can be
 }
 ```
```
shirhatti, line 286:
```suggestion

> **Warning:** In the default configuration, custom metrics will be exposed along with all other metrics on an unauthenticated endpoint. If your metrics contains sensitive information, we recommend disabling the [metrics urls](#metrics-urls) and consuming metrics from the authenticated endpoint (`--urls`) instead.

```

jander-msft: I don't see a "metrics-urls" section. Did you intend to link to something else?

src/Tools/dotnet-monitor/LoggingExtensions.cs
```diff
@@ -114,33 +114,21 @@ internal static class LoggingExtensions
...
-        private static readonly Action<ILogger, string, string, Exception> _metricUrlUpdated =
-            LoggerMessage.Define<string, string>(
-                eventId: new EventId(19, "MetricUrlUpdated"),
-                logLevel: LogLevel.Warning,
-                formatString: "Updated {originalUrl} to {newUrl} due to custom metrics.");
-
         private static readonly Action<ILogger, string, Exception> _optionsValidationFalure =
             LoggerMessage.Define<string>(
-                eventId: new EventId(20, "OptionsValidationFailure"),
+                eventId: new EventId(18, "OptionsValidationFailure"),
```
shirhatti, line 119: We get a hallpass here since we haven't reached a stable release yet, but you really shouldn't change EventIds. That's like the public API surface for logs.

documentation/configuration.md
jander-msft: Nevermind, found it!

src/Tools/dotnet-monitor/LoggingExtensions.cs
jander-msft: Yup :-) it's the only reason why I'm changing them now.

documentation/configuration.md
wiktork: I would not recommend disabling metrics url, but manually upgrading it. Remember the metrics url has a more permissive binding but a more restricted set of actions.

shirhatti: Do we automatically auth upgraded endpoints?

shirhatti: If we do, I think that's a bad idea. If we don't, it doesn't help prevent disclosing sensitive information.
Either way, I got issues with it 😂

wiktork: We do not but I see what you are getting at.

wiktork (approved):

documentation/configuration.md
jander-msft: I logged another issue where we can allow adding auth to the metrics URLs (Prometheus does seem to support specifying the Authorization header) but we would likely want to be able to configuration the authentication separately from the default one. See https://github.com/dotnet/dotnet-monitor/issues/216

PR #326, opened 2021-05-19
Fix key-per-file to signal change token.

jander-msft:
This change fixes key-per-file configuration to signal the change token when data is reloaded. This is a workaround until https://github.com/dotnet/aspnetcore/issues/32836 is fixed. This change enables the use of dependency injection to acquire key-per-file options from configuration and perform other options operations, such as post configuration and validation, in future changes.

### DIFF
--- a/src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationHandler.cs
+++ b/src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationHandler.cs
@@ -31,6 +31,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor
 
         private readonly ApiAuthenticationOptions _apiAuthOptions;
 
+
         public ApiKeyAuthenticationHandler(
             IOptionsMonitor<ApiKeyAuthenticationHandlerOptions> options,
             IOptionsSnapshot<ApiAuthenticationOptions> apiAuthOptions,
@@ -39,6 +40,10 @@ namespace Microsoft.Diagnostics.Tools.Monitor
             ISystemClock clock)
             : base(options, loggerFactory, encoder, clock)
         {
+            // AuthenticationHandler<T> implementations are registered as transient services, so its okay to
+            // use IOptionsSnapshot to get the current value of the api authentication options. This also has
+            // the side effect of getting a snapshot of the options rather than getting options that can mutate
+            // (e.g. configuration reload) while attempting to authenticate a particular request.
             _apiAuthOptions = apiAuthOptions.Value;
         }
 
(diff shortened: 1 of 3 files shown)


src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationHandler.cs
```diff
@@ -34,9 +29,17 @@ internal sealed class ApiKeyAuthenticationHandler : AuthenticationHandler<ApiKey
...
+
+        public ApiKeyAuthenticationHandler(
+            IOptionsMonitor<ApiKeyAuthenticationHandlerOptions> options,
+            IOptionsSnapshot<ApiAuthenticationOptions> apiAuthOptions,
+            ILoggerFactory loggerFactory,
+            UrlEncoder encoder,
+            ISystemClock clock)
             : base(options, loggerFactory, encoder, clock)
         {
+            _apiAuthOptions = apiAuthOptions.Value;
```
wiktork, line 42: Can you add a comment regarding why you can get the value instead of getting the snasphot later? IIRC this is a transient service, but that's not immediately obvious.

jander-msft: Correct, I'll add comments to explain this.

src/Tools/dotnet-monitor/ConfigurationBuilderExtensions.cs
```diff
@@ -0,0 +1,94 @@
...
+                Debug.Assert(null != s_loadMethod);
+                if (null != s_loadMethod)
+                {
+                    if (source.ReloadOnChange && source.FileProvider != null)
+                    {
+                        _changeTokenRegistration = ChangeToken.OnChange(
+                            () => source.FileProvider.Watch("*"),
+                            () =>
+                            {
+                                Thread.Sleep(source.ReloadDelay);
```
wiktork, line 70: I take it there is no async version of ChangeToken.OnChange? Thread.sleep is probably not ideal here, although it shouldn't happen that often.

wiktork (approved):

src/Tools/dotnet-monitor/ConfigurationBuilderExtensions.cs
jander-msft: I'm not sure, but I'm guessing there isn't. This is the code that the underlying KeyPerFileConfigurationProvider uses for reload. I just replicated it here in order to plumb in the OnReload call.

jander-msft: Checked this and there is no async version.

PR #335, opened 2021-05-21
Log authentication configuration errors on startup, configuration change, and HTTP requests.

### LINKED ISSUE #74: Validate API HashKeys during configuration load/reload, not just during auth.

jander-msft:
How this works:
- `ApiKeyAuthenticationOptions` is updated to contain the validated hash algorithm, hash value, and validation errors.
- `ApiKeyAuthenticationPostConfigureOptions` fills in properties on the `ApiKeyAuthenticationOptions` instance based on the validated properties from `ApiAuthenticationOptions`.
- `ApiKeyAuthenticationOptionsChangeTokenSource` listens for changes to `ApiAuthenticationOptions` and signals that `ApiKeyAuthenticationOptions` has changed when the change occurs.
- `ApiKeyAuthenticationHostedService` logs the initial state of the API Key configuration, listens for changes to `ApiKeyAuthenticationOptions`, and logs the updated state on the change.
- `ApiKeyAuthenticationHandler` checks if there are any `ApiKeyAuthenticationOptions` validation errors, and if there are, logs them and fails authentication early.

Example console output:
```
22:40:30 warn: Microsoft.Diagnostics.Tools.Monitor.ApiKeyAuthenticationHostedService[21]
      ApiAuthentication settings are invalid: The ApiKeyHash field is required.
22:40:30 warn: Microsoft.Diagnostics.Tools.Monitor.ApiKeyAuthenticationHostedService[21]
      ApiAuthentication settings are invalid: The ApiKeyHashType field is required.
22:40:31 warn: Microsoft.Diagnostics.Tools.Monitor.Startup[14]
      WARNING: Authentication is enabled over insecure http transport. This can pose a security risk and is not intended for production environments.
22:40:31 warn: Microsoft.AspNetCore.Server.Kestrel[0]
      Overriding address(es) 'http://localhost:52323'. Binding to endpoints defined in UseKestrel() instead.
22:40:32 info: Microsoft.Hosting.Lifetime[0]
      Now listening on: http://localhost:52323
22:40:32 info: Microsoft.Hosting.Lifetime[0]
      Now listening on: http://localhost:52325
22:40:32 info: Microsoft.Hosting.Lifetime[0]
22:40:32 info: Microsoft.Hosting.Lifetime[0]
22:40:32 info: Microsoft.Hosting.Lifetime[0]
      Content root path: D:\src\dotnet\dotnet-monitor\artifacts\bin\dotnet-monitor\Debug\netcoreapp3.1\
22:40:52 info: Microsoft.Diagnostics.Tools.Monitor.ApiKeyAuthenticationHostedService[22]
      ApiAuthentication settings have changed.
22:40:52 warn: Microsoft.Diagnostics.Tools.Monitor.ApiKeyAuthenticationHostedService[21]
      ApiAuthentication settings are invalid: The field ApiKeyHash must be a string or array type with a minimum length of '64'.
22:40:52 warn: Microsoft.Diagnostics.Tools.Monitor.ApiKeyAuthenticationHostedService[21]
      ApiAuthentication settings are invalid: The ApiKeyHashType field is required.
22:41:21 info: Microsoft.Diagnostics.Tools.Monitor.ApiKeyAuthenticationHostedService[22]
      API key authentication settings have changed.
22:41:21 warn: Microsoft.Diagnostics.Tools.Monitor.ApiKeyAuthenticationHostedService[21]
      ApiAuthentication settings are invalid: The ApiKeyHashType field is required.
22:41:41 info: Microsoft.Diagnostics.Tools.Monitor.ApiKeyAuthenticationHostedService[22]
      API key authentication settings have changed.
```

Initially, dotnet-monitor does not have API Key authentication configured.
- At `22:40:52`, I updated ApiKeyHash with a value that is too small.
- At `22:41:21`, I updated ApiKeyHash with a valid hash value.
- At `22:41:41`, I updated ApiKeyHashType with the hash algorithm name.

closes #74

### DIFF
--- a/src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationHostedService.cs
+++ b/src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationHostedService.cs
@@ -1,0 +1,73 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using System;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    /// <summary>
+    /// Service that monitors API Key authentication options changes and logs issues with the specified options.
+    /// </summary>
+    internal class ApiKeyAuthenticationHostedService :
+        IHostedService,
+        IDisposable
+    {
+        private readonly ILogger<ApiKeyAuthenticationHostedService> _logger;
+        private readonly IOptionsMonitor<ApiKeyAuthenticationOptions> _options;
+
+        private IDisposable _changeRegistration;
+
+        public ApiKeyAuthenticationHostedService(
+            ILogger<ApiKeyAuthenticationHostedService> logger,
+            IOptionsMonitor<ApiKeyAuthenticationOptions> options
+            )
+        {
+            _logger = logger;
+            _options = options;
+        }
+
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            _changeRegistration = _options.OnChange(OnApiKeyAuthenticationOptionsChanged);
+
+            // Write out current validation state of options when starting the tool.
+            CheckApiKeyAuthenticationOptions(_options.CurrentValue);
+
+            return Task.CompletedTask;
+        }
+
+        public Task StopAsync(CancellationToken cancellationToken)
+        {
+            _changeRegistration.Dispose();
+
+            return Task.CompletedTask;
+        }
+
+        public void Dispose()
+        {
+            _changeRegistration.Dispose();
+        }
+
+        private void OnApiKeyAuthenticationOptionsChanged(ApiKeyAuthenticationOptions options)
+        {
+            _logger.ApiKeyAuthenticationOptionsChanged();
+
+            CheckApiKeyAuthenticationOptions(options);
+        }
+
+        private void CheckApiKeyAuthenticationOptions(ApiKeyAuthenticationOptions options)
+        {
+            if (options.ValidationErrors.Any())
+            {
+                _logger.ApiKeyValidationFailures(options.ValidationErrors);
+            }
+        }
+    }
+}
--- a/src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationOptionsChangeTokenSource.cs
+++ b/src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationOptionsChangeTokenSource.cs
@@ -1,0 +1,49 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Primitives;
+using System;
+using System.Threading;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    /// <summary>
+    /// Notifies that ApiKeyAuthenticationOptions changes when ApiAuthenticationOptions changes.
+    /// </summary>
+    internal sealed class ApiKeyAuthenticationOptionsChangeTokenSource :
+        IOptionsChangeTokenSource<ApiKeyAuthenticationOptions>,
+        IDisposable
+    {
+        private readonly IOptionsMonitor<ApiAuthenticationOptions> _optionsMonitor;
+        private readonly IDisposable _changeRegistration;
+
+        private ConfigurationReloadToken _reloadToken = new ConfigurationReloadToken();
+
+        public ApiKeyAuthenticationOptionsChangeTokenSource(
+            IOptionsMonitor<ApiAuthenticationOptions> optionsMonitor)
+        {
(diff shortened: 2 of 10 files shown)


kelltrick (approved):
LGTM, I'll rebase my work on your changes here.

src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationHostedService.cs
```diff
@@ -0,0 +1,73 @@
...
+            IOptionsMonitor<ApiKeyAuthenticationOptions> options
+            )
+        {
+            _logger = logger;
+            _options = options;
+        }
+
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            _changeRegistration = _options.OnChange(OnApiKeyAuthenticationOptionsChanged);
```
wiktork, line 38: I don't quite understand why we need a hosted service for this. Couldn't you
1) Check configuration after setting up the app to write out errors
2) At the same time register for change registration and then produce the same warnings?

jander-msft: Exactly where else would I do that though? I could do that in the command handler, but that feels like throwing this there because it's convenient. The Startup class doesn't have a lifetime notion so I could unregistered the change token. 

Would creating it as a normal singleton that the Startup class initialized be acceptable? I'll have to see if the change token registration can be disposed on service disposal.

wiktork: Assuming a normal singleton's IDispose works, then I think that would be a better way to go.

src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationOptionsChangeTokenSource.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        public void Dispose()
+        {
+            _changeRegistration.Dispose();
+        }
+
+        private void OnReload(ApiAuthenticationOptions options)
+        {
+            Interlocked.Exchange(ref _reloadToken, new ConfigurationReloadToken()).OnReload();
```
wiktork, line 46: So we can get multiple OnReload's on different threads? Does that mean we can have multiple OnChange operations happening at the same time?

src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationOptionsChangeTokenSource.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            _optionsMonitor = optionsMonitor;
+            _changeRegistration = _optionsMonitor.OnChange(OnReload);
+        }
+
+        public string Name => Options.DefaultName;
+
+        public IChangeToken GetChangeToken()
+        {
+            return _reloadToken;
```
wiktork, line 36: It's been a while since i've looked at the details for this, but you have a read/write access to _reloadToken guarded by Interlocked and a read access on GetChangeToken with no memory barriers. IIRC this can cause stale data on some threads if multiple threads are attempting to call GetChangeToken.

src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationOptionsChangeTokenSource.cs
jander-msft: That may be possible, but this is to make sure that multiple callers of GetChangeToken get the same token if they happen to call in when the options monitor is signaling the change. Implementation is basically the same as https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Configuration/src/ConfigurationProvider.cs

wiktork: I'm wondering if Interlocked was just used for convenience of assigning a field while returning the previous field. It would be odd for any of this to be multi-threaded.

wiktork (approved):

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
```diff
@@ -155,6 +155,13 @@ public static IHostBuilder CreateHostBuilder(IConsole console, string[] urls, st
                         AuthOptions authenticationOptions = new AuthOptions(noAuth ? KeyAuthenticationMode.NoAuth : KeyAuthenticationMode.StoredKey);
                         services.AddSingleton<IAuthOptions>(authenticationOptions);
 
+                        // Although this is only observing API key authentication changes, it does handle
+                        // the case when API key authentication is not enabled. This class could evolve
+                        // to observe other opitons in the future, at which point it might be good to
```
wiktork, line 160: nit: typo

src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationOptionsChangeTokenSource.cs
jander-msft: You're probably correct for any given chain of updates. I do see that it is a different thread for each change, but the entire change starting from the KeyPerFileConfigurationProvider change callback all the way to when the observer wants to log the validation errors is all a single-threaded single execution context.

src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationOptionsChangeTokenSource.cs
jander-msft: It seems that the entire execution is single threaded, so when the change token raises the change callback, all of the handlers can back in one right the other to get the new change token.

PR #345, opened 2021-05-24
Run PR on feature branches.

jander-msft:


dotnet-monitor.yml
```diff
@@ -7,6 +7,7 @@ pr:
     - main
     - release/*
     - internal/release/*
+    - feature/*
```
kelltrick, line 10: Do we need to do a corresponding update in GitHub to "protect" these branches (like in ADO) or are branches always protected?

kelltrick (approved):

dotnet-monitor.yml
jander-msft: I will add branch protection for these branches.

Issue #82, opened 2021-03-16
Generate ephemeral API keys


jander-msft:
Allow the tool to generate ephemeral API keys to make local easier. Enabled with command line switch and log a warning when enabled. Log key information to console.

PR #352, opened 2021-05-26
Ephemeral Keys

kelltrick:
# Implements #82 
- Add `--temp-apikey` option to generate an APIKey during startup
- Prints the temporary key to the console like this:
```
D:\repos\github\dotnet-monitor-kelltrick [dev/pafenelo/eph2 ≡ +0 ~9 -0 !]> D:\repos\github\dotnet-monitor-kelltrick\artifacts\bin\dotnet-monitor\Debug\netcoreapp3.1\dotnet-monitor.exe collect --temp-apikey
13:45:24 warn: Microsoft.Diagnostics.Tools.Monitor.Startup[23]
      Generated one-time-use ApiKey for dotnet-monitor; use the following header for authorization:
      Authorization: MonitorApiKey Un3LJob3MqrR69WmFFNiU6G7U6QqgF0cY/Tm4FWSjxI=
13:45:24 warn: Microsoft.AspNetCore.Server.Kestrel[0]
      Overriding address(es) 'https://localhost:52323'. Binding to endpoints defined in UseKestrel() instead.
```
- Add test hook to add commandline flag `--temp-apikey`, add test to exercise new test

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
@@ -186,14 +186,13 @@ namespace Microsoft.Diagnostics.Monitoring.UnitTests
         }
 
         /// <summary>
-        /// Tests that API key authentication can be configured correctly and
-        /// that the key can be rotated wihtout shutting down dotnet-monitor.
+        /// Tests that --temp-apikey flag can be used to generate a key.
         /// </summary>
         [Fact]
-        public async Task TempApiKeyAuthenticationSchemeTest()
+        public async Task TempApiKeyTest()
         {
             await using MonitorRunner toolRunner = new(_outputHelper);
-            toolRunner.TempApiKey = true;
+            toolRunner.UseTempApiKey = true;
 
             _outputHelper.WriteLine("Starting Monitor.");
             await toolRunner.StartAsync();
@@ -207,6 +206,49 @@ namespace Microsoft.Diagnostics.Monitoring.UnitTests
             // Check that /processes is authenticated
             var processes = await apiClient.GetProcessesAsync();
             Assert.NotNull(processes);
+
+            // Test that clearing the Authorization header will result in a 401
+            httpClient.DefaultRequestHeaders.Authorization = null;
+            var statusCodeException = await Assert.ThrowsAsync<ApiStatusCodeException>(
+                () => apiClient.GetProcessesAsync());
+            Assert.Equal(HttpStatusCode.Unauthorized, statusCodeException.StatusCode);
+        }
+
+        /// <summary>
+        /// Tests that --temp-apikey will override the ApiAuthentication configuration.
+        /// </summary>
+        [Fact]
+        public async Task TempApiKeyOverridesApiAuthenticationTest()
+        {
+            const string AlgorithmName = "SHA256";
+
+            await using MonitorRunner toolRunner = new(_outputHelper);
+            toolRunner.UseTempApiKey = true;
+
+            // Generate initial API key
+            byte[] apiKey = GenerateApiKey();
+
+            // Set API key via key-per-file
+            RootOptions options = new();
+            options.UseApiKey(AlgorithmName, apiKey);
+            toolRunner.WriteKeyPerValueConfiguration(options);
+
+            _outputHelper.WriteLine("Starting Monitor.");
+            await toolRunner.StartAsync();
+
+            using HttpClient httpClient = await toolRunner.CreateHttpClientDefaultAddressAsync(_httpClientFactory);
+            ApiClient apiClient = new(_outputHelper, httpClient);
+
+            // Check that /processes is authenticated when using the supplied temp key
+            var processes = await apiClient.GetProcessesAsync();
+            Assert.NotNull(processes);
+
+            // Test that setting the Authorization header for the supplied config will result in a 401
+            httpClient.DefaultRequestHeaders.Authorization =
+                new AuthenticationHeaderValue(ApiKeyScheme, Convert.ToBase64String(apiKey));
+            var statusCodeException = await Assert.ThrowsAsync<ApiStatusCodeException>(
+                () => apiClient.GetProcessesAsync());
+            Assert.Equal(HttpStatusCode.Unauthorized, statusCodeException.StatusCode);
         }
 
         /// <summary>
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Diagnostics.Monitoring.UnitTests
     [Collection(DefaultCollectionFixture.Name)]
     public class AuthenticationTests
     {
-        private const string ApiKeyScheme = "MonitorApiKey";
+        internal const string ApiKeyScheme = "MonitorApiKey";
 
         private readonly IHttpClientFactory _httpClientFactory;
         private readonly ITestOutputHelper _outputHelper;
@@ -182,6 +182,30 @@ namespace Microsoft.Diagnostics.Monitoring.UnitTests
 
             // Check that /processes does not challenge for authentication
             processes = await apiClient.GetProcessesAsync();
+            Assert.NotNull(processes);
+        }
+
+        /// <summary>
+        /// Tests that API key authentication can be configured correctly and
+        /// that the key can be rotated wihtout shutting down dotnet-monitor.
+        /// </summary>
+        [Fact]
+        public async Task TempApiKeyAuthenticationSchemeTest()
+        {
+            await using MonitorRunner toolRunner = new(_outputHelper);
+            toolRunner.TempApiKey = true;
+
+            _outputHelper.WriteLine("Starting Monitor.");
+            await toolRunner.StartAsync();
+
+            using HttpClient httpClient = await toolRunner.CreateHttpClientDefaultAddressAsync(_httpClientFactory);
+            ApiClient apiClient = new(_outputHelper, httpClient);
+
+            // Assert that the httpClient came back with a populated Authorization header
+            Assert.NotNull(httpClient.DefaultRequestHeaders.Authorization?.Parameter);
+
+            // Check that /processes is authenticated
+            var processes = await apiClient.GetProcessesAsync();
             Assert.NotNull(processes);
         }
 
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
@@ -194,7 +194,6 @@ namespace Microsoft.Diagnostics.Monitoring.UnitTests
             await using MonitorRunner toolRunner = new(_outputHelper);
             toolRunner.UseTempApiKey = true;
 
-            _outputHelper.WriteLine("Starting Monitor.");
             await toolRunner.StartAsync();
 
             using HttpClient httpClient = await toolRunner.CreateHttpClientDefaultAddressAsync(_httpClientFactory);
@@ -233,7 +232,6 @@ namespace Microsoft.Diagnostics.Monitoring.UnitTests
             options.UseApiKey(AlgorithmName, apiKey);
             toolRunner.WriteKeyPerValueConfiguration(options);
 
-            _outputHelper.WriteLine("Starting Monitor.");
             await toolRunner.StartAsync();
 
             using HttpClient httpClient = await toolRunner.CreateHttpClientDefaultAddressAsync(_httpClientFactory);
--- a/src/Tools/dotnet-monitor/Auth/AuthOptions.cs
+++ b/src/Tools/dotnet-monitor/Auth/AuthOptions.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor
         public bool EnableKeyAuth => (KeyAuthenticationMode == KeyAuthenticationMode.StoredKey) ||
                                      (KeyAuthenticationMode == KeyAuthenticationMode.TemporaryKey);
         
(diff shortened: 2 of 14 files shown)


kelltrick:
@shirhatti can you take a look at the 2 user-visible strings here?
1. The new command line parameter is `--temp-apikey`
2. The log message is: 
```
Generated one-time-use ApiKey for dotnet-monitor, use the following header for authorization:
Authorization: MonitorApiKey DDcvBFMwz5zTQZIWkMQIAfwBvn270PTwgrvK/HReNrA=
```

src/Tools/dotnet-monitor/LoggingExtensions.cs
```diff
@@ -144,6 +145,12 @@ internal static class LoggingExtensions
                 logLevel: LogLevel.Information,
                 formatString: nameof(ConfigurationKeys.ApiAuthentication) + " settings have changed.");
 
+        private static readonly Action<ILogger, string, string, Exception> _logTempKey =
+            LoggerMessage.Define<string, string>(
+                eventId: new EventId(23, "LogTempApiKey"),
+                logLevel: LogLevel.Information,
+                formatString: "Generated one-time-use ApiKey for dotnet-monitor, use the following header for authorization:{0}Authorization: MonitorApiKey {MonitorApiKey}");
```
jander-msft, line 151:
```suggestion
                formatString: "Generated one-time-use ApiKey for dotnet-monitor; use the following header for authorization:{0}Authorization: MonitorApiKey {MonitorApiKey}");
```

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
```diff
@@ -144,133 +150,156 @@ public static IHostBuilder CreateHostBuilder(IConsole console, string[] urls, st
...
+                })
+                .UseStartup<Startup>();
+            });
+
             return hostBuilder;
         }
 
+        private static void ConfigureTempApiHashKey(IConfigurationBuilder builder, AuthOptions authenticationOptions)
+        {
+            const string DefaultHashAlgorithm = "SHA512";
```
jander-msft, line 287: The generatekey command on dotnet-monitor generates a SHA256 key. I think these should be consistent. Additionally, I think SHA256 should be the default; if I recall correctly, there might be a bug in the ApiKeyAuthenticationHandler that assumes the data is only 32 bytes long.

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
```diff
@@ -144,133 +150,156 @@ public static IHostBuilder CreateHostBuilder(IConsole console, string[] urls, st
...
+            using HashAlgorithm algorithm = HashAlgorithm.Create(DefaultHashAlgorithm);
```
jander-msft, line 288: I know that this might seem trivial to have recreated this code here, but can this be shared amongst this configuration and the https://github.com/dotnet/dotnet-monitor/blob/main/src/Tools/dotnet-monitor/GenerateApiKeyCommandHandler.cs handler? Maybe add two static methods to the handler:
- `byte[] GenerateRandomKey()`: Generates a cryptographically random key.
- `string ComputeHexadecimalHash(byte[] data)`: Hashes the data and returns it as hexadecimal encoded.

These can then be used by the GenerateApiKeyCommandHandler, this configuration injection, and AuthOptions.ctor. This allows it to be consistent across the two usages and more easily updatable.

kelltrick (commented):
Addressed my own comment.

shirhatti:
>The log message is

Category/log level?

src/Tools/dotnet-monitor/Auth/AuthOptions.cs
```diff
@@ -14,10 +16,20 @@ internal sealed class AuthOptions : IAuthOptions
...
+        public byte[] GeneratedKey { get; }
 
         public AuthOptions(KeyAuthenticationMode mode)
         {
             KeyAuthenticationMode = mode;
+            
+            if (mode == KeyAuthenticationMode.TemporaryKey)
+            {
+                byte[] newKey = new byte[32];
+                RNGCryptoServiceProvider rngCsp = new RNGCryptoServiceProvider();
```
wiktork, line 29: What is the difference between RNGCryptoServiceProvider and RandomNumberGenerator? Note that GenerateApiKey command uses the latter. Should that be changed to use RNGCryptoServiceProvider?

wiktork: (It would also be nice to actually have the same code generate the key in both cases, should we ever change how we generate it in the first place)

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
```diff
@@ -144,133 +150,156 @@ public static IHostBuilder CreateHostBuilder(IConsole console, string[] urls, st
 
                     builder.AddKeyPerFileWithChangeTokenSupport(path, optional: true, reloadOnChange: true);
                     builder.AddEnvironmentVariables(ConfigPrefix);
+
+                    if (!noAuth && tempApiKey)
```
wiktork, line 154: nit: consider using authMode since you've already worked out exactly what mode you are in.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
```diff
@@ -185,6 +185,30 @@ public async Task ApiKeyAuthenticationSchemeTest()
             Assert.NotNull(processes);
         }
 
+        /// <summary>
+        /// Tests that API key authentication can be configured correctly and
+        /// that the key can be rotated wihtout shutting down dotnet-monitor.
+        /// </summary>
+        [Fact]
+        public async Task TempApiKeyAuthenticationSchemeTest()
```
jander-msft, line 193: Could you add another test that enables the temp API key usage and custom API key at the same time and validate that the custom API key will 401 whereas the temp API key will 200?

jander-msft, line 193: nit: TempApiKeyTest, since TempApiKey is not a scheme itself. Also, please update the test xml docs to reflect what the test is doing.

src/Tools/dotnet-monitor/LoggingExtensions.cs
```diff
@@ -144,6 +144,12 @@ internal static class LoggingExtensions
...
```
wiktork, line 151: Unnecessary {0}?

jander-msft (changes requested):

src/Tools/dotnet-monitor/Auth/AuthOptions.cs
kelltrick: I'm apparently wrong on this. In .NET framework, the standard is to always use `new RNGCryptoServiceProvider()` because `RandomNumberGenerator` can be a non-cryptographically secure algorithm. In .NET core the standard is to use `RandomNumberGenerator.Create()` and it will only ever give you a `RNGCryptoServiceProvider`.

src/Tools/dotnet-monitor/LoggingExtensions.cs
jander-msft: It's used to inject an Environment.NewLine into the string. See `LoggingExtensions.LogTempKey`.

src/Tools/dotnet-monitor/LoggingExtensions.cs
```diff
@@ -144,6 +144,12 @@ internal static class LoggingExtensions
                 logLevel: LogLevel.Information,
                 formatString: nameof(ConfigurationKeys.ApiAuthentication) + " settings have changed.");
 
+        private static readonly Action<ILogger, string, string, Exception> _logTempKey =
+            LoggerMessage.Define<string, string>(
+                eventId: new EventId(23, "LogTempApiKey"),
+                logLevel: LogLevel.Information,
```
jander-msft, line 150: Warning level

src/Tools/dotnet-monitor/LoggingExtensions.cs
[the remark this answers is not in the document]
kelltrick: I just extracted these parameters and grabbed the correct values

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
kelltrick: refactored to a single generator

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
kelltrick: @jander-msft , let me know if you're okay with the way I did this. I didn't exactly follow your instructions.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/AuthenticationTests.cs
```diff
@@ -185,6 +185,72 @@ public async Task ApiKeyAuthenticationSchemeTest()
...
+        /// <summary>
+        /// Tests that --temp-apikey flag can be used to generate a key.
+        /// </summary>
+        [Fact]
+        public async Task TempApiKeyTest()
+        {
+            await using MonitorRunner toolRunner = new(_outputHelper);
+            toolRunner.UseTempApiKey = true;
+
+            _outputHelper.WriteLine("Starting Monitor.");
```
jander-msft, line 197: nit: These "Starting Monitor" logs shouldn't be necessary. There are a whole bunch of logs that already detail how dotnet-monitor is started and when in the test it occurs.

src/Tools/dotnet-monitor/GenerateApiKeyCommandHandler.cs
```diff
@@ -21,24 +21,57 @@ namespace Microsoft.Diagnostics.Tools.Monitor
...
+            console.Out.WriteLine(FormattableString.Invariant($"ApiKeyHash: {newKey.HashValue}"));
+            console.Out.WriteLine(FormattableString.Invariant($"ApiKeyHashType: {newKey.HashAlgorithm}"));
+
+            return Task.FromResult(0);
+        }
+    }
+
+    internal sealed class GeneratedApiKey
+    {
+        private const int KeyLengthBytes = 32;
```
jander-msft, line 37: Because this class is parameterized to allow different hash algorithms, this shouldn't be a constant. I think you can use `hasher.HashSize / 8` instead so that if we use SHA512, it would automatically work.

kelltrick:
> Also not a fan of "Microsoft.Diagnostics.Tools.Monitor.Startup". It probably warrants it's own category (or sub-category)?

I can't see an easy way of creating this log message it's own ILogger so it displays something nicer. This is probably something waiting for customer feedback on.

src/Tools/dotnet-monitor/GenerateApiKeyCommandHandler.cs
```diff
@@ -21,24 +21,57 @@ namespace Microsoft.Diagnostics.Tools.Monitor
...
+
+            console.Out.WriteLine(FormattableString.Invariant($"Authorization: {Monitoring.RestServer.AuthConstants.ApiKeySchema} {newKey.MonitorApiKey}"));
+            console.Out.WriteLine(FormattableString.Invariant($"ApiKeyHash: {newKey.HashValue}"));
+            console.Out.WriteLine(FormattableString.Invariant($"ApiKeyHashType: {newKey.HashAlgorithm}"));
+
+            return Task.FromResult(0);
+        }
+    }
+
+    internal sealed class GeneratedApiKey
```
jander-msft, line 35: Can you put this in its own file.

src/Tools/dotnet-monitor/GenerateApiKeyCommandHandler.cs
kelltrick: No, this is a hard coded constant at [ApiKeyAuthenticationHandler.cs:66](https://github.com/dotnet/dotnet-monitor/blob/c6b8b937134cc3b732f8a6766e846a6348f5e251/src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationHandler.cs#L66). 
It probably makes sense to have `ApiKeyAuthenticationHandler` reference this from `GeneratedApiKey` or vice versa and have this be a single parameter.

jander-msft (approved):

src/Tools/dotnet-monitor/GenerateApiKeyCommandHandler.cs
jander-msft: Just because something else in dotnet-monitor is making a bad assumption doesn't mean you should code a different part to assume that same bad assumption.

The intent was to allow other algorithms to be use, which can have different key lengths. Hence why there is an option for specifying the algorithm used to hash the original key. Otherwise, we'd just bake it in to use SHA256.

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
jander-msft: Yup, perfectly fine with the refactor.

src/Tools/dotnet-monitor/GenerateApiKeyCommandHandler.cs
kelltrick: Just to be clear here, this value represents the length of the `MonitorApiKey` not the `HashValue` which would be determined by the algorithm, say `SHA512` vs `SHA256`. We assume the "private key" you'll use is always 32 bytes, and it just happens to be that we use SHA256 by default which also happens to provide a 32 byte output hash value.
There is no reason the `MonitorApiKey` always needs to be 32 bytes, but there also isn't a reason not to require it to be 32 bytes.

kelltrick: Although I think you're onto something, I tried this as `SHA512` and it gets rejected.

kelltrick: Ahh, I copy/pasted the `SHA256.Create()` without realizing it. I intended to make this `HashAlgorithm.Create(hashAlgorithm)`

Issue #360, opened 2021-05-27
Make generatekey take configurable MonitorApiKey length


kelltrick:
## Background and Motivation

If we choose to implement #359, we should modify `generatekey` command generate keys that use the new configuration.

## Proposed Feature

- Add `--key-length`/`-l` command line parameter to `generatekey` that will accept an integer between 32 and 2048.
- Add `--hash-algorithm`/`-h` command line parameter to `generatekey` that will accept a string specification of a Hash Algorithm such as `SHA256` or `SHA512`.

## Usage Examples

1. Getting a 128 byte MonitorApiKey with SHA512:
Input:
`dotnet-monitor.exe generatekey --key-length 128 --hash-algorithm SHA512`
Output:
```
Authorization: MonitorApiKey lmWOJfSJRXO/6Vge40O3KdW/VunK2btVg9VTvSiGUM7zZueihJhIjDhCi3M56/D1bxCf7YB0duCV89zmCYoeiu4RUEshS82zME5TEKInMukbaPZvT6euQyqCyHIqc69FMchdfFv1CkD1n2AnLh2XKqQOrJBfJPShkeUKo8C/Dks=
ApiKeyHash: 7F78E57A36A46C531451278DC100084EB4D093A9FA05F16E25CF122166239FC94837931795CE49582A3C3FA477B3EB3944010BEE943AC961C369DE80B02A8409
ApiKeyHashType: SHA512
```
2. Getting a 32 byte MonitorApiKey with SHA256:
Input:
`dotnet-monitor.exe generatekey -l 32 -h SHA256`
Output:
```
Authorization: MonitorApiKey h9cK6yaNDVKSxHpb4p/ccYpiYfi8oUEl74kj+SABxpQ=
ApiKeyHash: 61429B0382507E34D21522A05D3411D9B34D20601F4CE6848909959A2960360C
ApiKeyHashType: SHA256
```
3. Using the default parameters, getting a 32 byte MonitorApiKey with SHA256:
Input:
`dotnet-monitor.exe generatekey`
Output:
```
Authorization: MonitorApiKey +R2bQIb07IuTRdRBOivqSSO3stujIyKJSfReeQ02oww=
ApiKeyHash: 090BA3CADA14F6763EB16A02A55E129BAB6B841B931F425E31D54531EC5E8DCF
ApiKeyHashType: SHA256
```

Issue #359, opened 2021-05-26
Make MonitorApiKey length configurable


kelltrick:
## Background and Motivation

The MonitorApiKey muse be exactly 32 bytes long, as defined in [ApiKeyAuthenticationHandler.cs:66](https://github.com/dotnet/dotnet-monitor/blob/5f97a7154a1d2f6bc33fbd0a848cdb253e784cb1/src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationHandler.cs#L68). We only require the user to tell us what the hash of that key is during configuration, so we shouldn't care what length this value is.

## Proposed Feature

- Change `ApiKeyAuthenticationHandler.cs` to allow keys of any length between 32 bytes and 2048 bytes.
- Place an additional max cap of `((2048 * 8) / 6) + 2` bytes in the raw base64 encoded Authorization parameter.

## Usage Examples

1. Using a 128 byte MonitorApiKey with SHA512:
```
Authorization: MonitorApiKey lmWOJfSJRXO/6Vge40O3KdW/VunK2btVg9VTvSiGUM7zZueihJhIjDhCi3M56/D1bxCf7YB0duCV89zmCYoeiu4RUEshS82zME5TEKInMukbaPZvT6euQyqCyHIqc69FMchdfFv1CkD1n2AnLh2XKqQOrJBfJPShkeUKo8C/Dks=
```
```
"ApiAuthentication": {
  "ApiKeyHash": "7F78E57A36A46C531451278DC100084EB4D093A9FA05F16E25CF122166239FC94837931795CE49582A3C3FA477B3EB3944010BEE943AC961C369DE80B02A8409",
  "ApiKeyHashType": "SHA512"
},
```
2. Using a 32 byte MonitorApiKey with SHA256
```
Authorization: MonitorApiKey kRh/TVsGw/sxoP7eE2cLmxNZu+H+em1oWOfbF4902XI=
```
```
"ApiAuthentication": {
  "ApiKeyHash": "5395C7FD870DFA6F61BEBBA86134A7F9A0125D36677FBD091C7E5150ACC4E16D",
  "ApiKeyHashType": "SHA256"
},
```

jander-msft:
I would separate this into two issues:
1) allowing the Authorization header to contain a key that is longer than 32 bytes.
2) allowing the key generation to be configurable.

Because the first is probably a bug and the second is an enhancement.

PR #361, opened 2021-05-27
Work to implement configurable MonitorApiKey

kelltrick:
## Implement #359 : Make MonitorApiKey length configurable
- Re-writes part of `ApiKeyAuthenticationHandler` to accept `MonitorApiKey` values with length between 32 bytes and 2048 bytes.
## Implement #360 : Make generatekey take configurable MonitorApiKey length
- Add --key-length/-l command line parameter to generatekey that will accept an integer between 32 and 2048.
- Add --hash-algorithm/-h command line parameter to generatekey that will accept a string specification of a Hash Algorithm such as SHA256 or SHA512.


src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationHandler.cs
```diff
@@ -24,7 +24,10 @@ namespace Microsoft.Diagnostics.Tools.Monitor
     /// </summary>
     internal sealed class ApiKeyAuthenticationHandler : AuthenticationHandler<ApiKeyAuthenticationOptions>
     {
-        public const int ApiKeyNumBytes = 32;
+        public const int ApiKeyByteMinLength = 32;
+        public const int ApiKeyByteMaxLength = 2048;
+        // This is the max length to efficiently encode an ApiKey of the length ApiKeyByteMaxLength.
+        private const int ApiKeyBase64MaxLength = ((ApiKeyByteMaxLength * 8) / 6) + 2;
```
jander-msft, line 30: This doesn't seem quite right.

For example, 3 bytes encoded into base64 requires 4 characters. The calculation here says (3 * 8) / 6 + 2 = 6 characters.
| Byte Count | Encoded Length |
|---|---|
| 0 | 2 |
| 1 | 3 |
| 2 | 4 |
| 3 | 6 |
| 4 | 7 |

I think the algorithm is (byteLength + 2) / 3 * 4 giving:
| Byte Count | Encoded Length |
|---|---|
| 0 | 0 |
| 1 | 4 |
| 2 | 4 |
| 3 | 4 |
| 4 | 8 |

I think you can use System.Buffers.Text.Base64.GetMaxEncodedToUtf8Length to calculate the static value and not have to think about the algorithm :-)

src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationHandler.cs
```diff
@@ -63,11 +66,26 @@ protected override Task<AuthenticateResult> HandleAuthenticateAsync()
...
-            if (!Convert.TryFromBase64String(authHeader.Parameter, span, out int bytesWritten) || bytesWritten < 32)
+            if (authHeader.Parameter.Length > ApiKeyBase64MaxLength)
+            {
+                return Task.FromResult(AuthenticateResult.Fail("Invalid API key format."));
+            }
+
+            // Calculate the max length of the base64 encoded value,
+            // the rules to decode this are really complex and allow white space in the
+            // string which should be ignored, simply calculate the max it can be.
+            int byteLen = (authHeader.Parameter.Length + 1) * 3 / 4;
```
jander-msft, line 77: Similarly, use System.Buffers.Text.Base64.GetMaxDecodedFromUtf8Length so we don't have to worry about the algorithm.

src/Tools/dotnet-monitor/Auth/GeneratedApiKey.cs
```diff
@@ -23,15 +27,25 @@ private GeneratedApiKey(string monitorApiKey, string hashAlgorithm, string hashV
...
+            if (keyLength < ApiKeyAuthenticationHandler.ApiKeyByteMinLength || keyLength > ApiKeyAuthenticationHandler.ApiKeyByteMaxLength)
+            {
+                throw new ArgumentOutOfRangeException(nameof(keyLength));
+            }
+
             using RandomNumberGenerator rng = RandomNumberGenerator.Create();
             using HashAlgorithm hasher = System.Security.Cryptography.HashAlgorithm.Create(hashAlgorithm);
 
-            byte[] secret = new byte[ApiKeyAuthenticationHandler.ApiKeyNumBytes];
+            byte[] secret = new byte[keyLength];
```
jander-msft, line 48: Should the key length be at least as long as the hash length?

src/Tools/dotnet-monitor/Auth/HashAlgorithmChecker.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.Linq;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    class HashAlgorithmChecker
+    {
+        private static readonly string[] DisallowedHashAlgorithms = new string[]
```
jander-msft, line 8: RIPEMD160 and MAC3DES were dropped. Intentional?

src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationHandler.cs
kelltrick: I wish I knew about `System.Buffers.Text.Base64.GetMaxEncodedToUtf8Length` before doing this. I couldn't find any of this.

src/Tools/dotnet-monitor/Auth/GeneratedApiKey.cs
```diff
@@ -23,15 +27,25 @@ private GeneratedApiKey(string monitorApiKey, string hashAlgorithm, string hashV
...
         public static GeneratedApiKey Create()
         {
-            return GeneratedApiKey.Create("SHA256");
+            return GeneratedApiKey.Create(DefaultKeyLength, DefaultHashAlgorithm);
         }
 
-        private static GeneratedApiKey Create(string hashAlgorithm)
+        public static GeneratedApiKey Create(int keyLength, string hashAlgorithm)
         {
+            if (!HashAlgorithmChecker.IsAllowedAlgorithm(hashAlgorithm))
```
wiktork, line 35: Are these exceptions visible in the logs after configuration validation? If so, maybe they can be more verbose?

wiktork (approved):

src/Tools/dotnet-monitor/Auth/GeneratedApiKey.cs
kelltrick: I think it's simpler to have the "more than 32 bytes" restriction be constant and let the user pick the algorithm. While I don't think it makes sense to have a 32 byte key and a 64 byte hash output, I don't think it has a meaningful security implication so I'm leaning towards the simpler restriction.

src/Tools/dotnet-monitor/Auth/HashAlgorithmChecker.cs
kelltrick: Yes, this was discussed as PR feedback from the PR to implement temp api keys. I apparently forgot to "git add" the change when I committed my PR changes so I added that change into this PR.

src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationHandler.cs
kelltrick: Done

src/Tools/dotnet-monitor/Auth/HashAlgorithmChecker.cs
kelltrick: Done

src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationHandler.cs
kelltrick: Done

src/Tools/dotnet-monitor/Auth/ApiKeyAuthenticationHandler.cs
```diff
@@ -24,7 +24,10 @@ namespace Microsoft.Diagnostics.Tools.Monitor
     /// </summary>
     internal sealed class ApiKeyAuthenticationHandler : AuthenticationHandler<ApiKeyAuthenticationOptions>
     {
-        public const int ApiKeyNumBytes = 32;
+        public const int ApiKeyByteMinLength = 32;
+        public const int ApiKeyByteMaxLength = 2048;
+        // This is the max length to efficiently encode an ApiKey of the length ApiKeyByteMaxLength.
+        private readonly int ApiKeyBase64MaxLength = System.Buffers.Text.Base64.GetMaxEncodedToUtf8Length(ApiKeyByteMaxLength);
```
jander-msft, line 30: nit: Move `System.Buffers.Text` to using.

jander-msft (approved):

Issue #84, opened 2021-03-16
Make dotnet-monitor localizable


jander-msft:
Split out strings so that it is localizable.  Localized != Localizable.

PR #362, opened 2021-05-27
Convert strings to Resource Strings

kelltrick:
## Start #84 
- Convert `LoggingExtensions.cs` to resource strings
- Convert all descriptions in `Program.cs` to resource strings

### Open Question: What else should be converted to resource strings?
I believe we should not allow AuthenticationResult failure messages to be localized and then localize all other exceptions.
- [x] AuthenticationResult failure messages
- [x] Strings used to define error messages that end up in exceptions
- [x] Descriptions of config parameters specified in `appsettings.json`
- [x] Error messages in `LoggerExtensions`


src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -0,0 +1,311 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
wiktork, line 1: Nit: maybe Resources.resx, since we could add other things here in the future.

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -0,0 +1,311 @@
...
+    </xsd:element>
+  </xsd:schema>
+  <resheader name="resmimetype">
+    <value>text/microsoft-resx</value>
+  </resheader>
+  <resheader name="version">
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
```
wiktork, line 115: So is this just an artifact from the previous format, which netcore ignores, or will System.Windows.Forms actually be loaded into dotnet-monitor?

wiktork (approved):

jander-msft:
> I believe we should not allow AuthenticationResult failure messages to be localized and then localize all other exceptions.

I think the AuthenticationResult strings should be localized. They are just as user facing as all of the other log statements.

src/Tools/dotnet-monitor/Strings.resx
kelltrick: This seems to be there for backwards compatibility, .net core seems to ignore it and emits a reference to it's own `ResourceManager`:
```
.assembly extern System.Resources.ResourceManager
{
  .publickeytoken = (B0 3F 5F 7F 11 D5 0A 3A )                         // .?_....:
  .ver 4:1:2:0
}
```

src/Tools/dotnet-monitor/Strings.resx
kelltrick: I decided to call this `String` because my intention would be to put other resource in their own resx file for organization.

kelltrick:
> > I believe we should not allow AuthenticationResult failure messages to be localized and then localize all other exceptions.
> 
> I think the AuthenticationResult strings should be localized. They are just as user facing as all of the other log statements.

- I'm okay with localizing these too; I was considering this to be similar to how Azure responds to authentication failures in storage REST calls. The messages are always the same English string, which makes sense because they don't have any specific language to return with the call. 
- My only concern is that we may leak they language of the machine to an un-authenticated caller; probably not a real weakness, but something to consider.

jander-msft:
> > > I believe we should not allow AuthenticationResult failure messages to be localized and then localize all other exceptions.
> > 
> > 
> > I think the AuthenticationResult strings should be localized. They are just as user facing as all of the other log statements.
> 
> * I'm okay with localizing these too; I was considering this to be similar to how Azure responds to authentication failures in storage REST calls. The messages are always the same English string, which makes sense because they don't have any specific language to return with the call.
> * My only concern is that we may leak they language of the machine to an un-authenticated caller; probably not a real weakness, but something to consider.

IIRC from my perusal of the ASP.NET Core Authentication stack, the message is only used for logging through ILogger so I don't think it'll ever get back to the HTTP client unless we go out of our way to do it.

src/Microsoft.Diagnostics.Monitoring.WebApi/SharedStrings.resx
```diff
@@ -0,0 +1,236 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
jander-msft, line 1: nit: rather than "SharedStrings", because there can be multiple sharing groups, maybe call this OptionsDisplayStrings to more clearly show the intent of what this particular resource file is used. I'd also like to avoid it becoming a dumping ground of strings that are shared between all of the assemblies.

jander-msft (changes requested):

src/Microsoft.Diagnostics.Monitoring.WebApi/LoggingExtensions.cs
```diff
@@ -13,37 +13,37 @@ internal static class LoggingExtensions
...
                 eventId: new EventId(5, "WrittenToHttpStream"),
                 logLevel: LogLevel.Information,
-                formatString: "Written to HTTP stream.");
+                formatString: Strings.LogFormatString_WrittenToHttpStream);
 
         private static readonly Action<ILogger, int, int, Exception> _throttledEndpoint =
             LoggerMessage.Define<int, int>(
                  eventId: new EventId(6, "ThrottledEndpoint"),
                 logLevel: LogLevel.Warning,
-                formatString: "Request limit for endpoint reached. Limit: {limit}, oustanding requests: {requests}");
```
jander-msft, line 46: I know you did a whole bunch of work here to make the log statements localizable, but there will be an issue if we go to actually localize these. The parameters in the log statement string are positional (e.g. the `limit` parameter is filled from the first argument, the `requests` parameter is filled from the second argument). When this statement is localized, the parameters may be switched around depending on the language. This will cause arguments to not be correctly matched with parameters. So I'm wondering if we should (a) not localize log statements, and/or (b) consult with someone in corefx about localization + logging.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests.csproj
```diff
@@ -29,6 +29,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\Microsoft.Diagnostics.Monitoring.WebApi\Microsoft.Diagnostics.Monitoring.WebApi.csproj" />
```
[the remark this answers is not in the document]
kelltrick: This is the project that has `OptionsDisplyStrings` (aka `SharedStrings`). I couldn't get .net to embed 2 copies of the resources from a single resx without manually calling `resgen`, so I decided the additional project reference was the simplest way to make it work.

src/Microsoft.Diagnostics.Monitoring.WebApi/LoggingExtensions.cs
shirhatti: FYI @maryamariyan. Do we have established guidance for this question?

jander-msft (approved):

src/Microsoft.Diagnostics.Monitoring.WebApi/LoggingExtensions.cs
maryamariyan: Similar to APIs like [`TextWriter.WriteLine(string format, object? arg0, object? arg1)`](https://docs.microsoft.com/en-us/dotnet/api/system.io.textwriter.writeline?view=net-5.0#System_IO_TextWriter_WriteLine_System_String_System_Object_System_Object_), or [`string.Format(string format, object? arg0, object? arg1, object? arg2)`](https://docs.microsoft.com/en-us/dotnet/api/system.string.format?view=net-5.0) the action returning `LoggerMessage.Define<.., T1, .., TN, ..>` APIs format strings according to order in which they receive name hole values within the formatted string. 

Therefore, it is up to the consumer of these APIs to carefully decide how to use these APIs based on localization (perhaps to flip parameter orders given a specific localization).

/cc @tarekgh

jander-msft: > Similar to APIs like [`TextWriter.WriteLine(string format, object? arg0, object? arg1)`](https://docs.microsoft.com/en-us/dotnet/api/system.io.textwriter.writeline?view=net-5.0#System_IO_TextWriter_WriteLine_System_String_System_Object_System_Object_), or [`string.Format(string format, object? arg0, object? arg1, object? arg2)`](https://docs.microsoft.com/en-us/dotnet/api/system.string.format?view=net-5.0) the action returning `LoggerMessage.Define<.., T1, .., TN, ..>` APIs format strings according to order in which they receive name hole values within the formatted string.

I don't think that is necessarily correct. Those APIs fill the format according to the parameter index defined within the format string. Because they use parameter index, it doesn't matter how the format string arranges them. https://docs.microsoft.com/en-us/dotnet/api/system.io.textwriter.writeline?view=net-5.0#System_IO_TextWriter_WriteLine_System_String_System_Object_System_Object_System_Object_ clearly states that it uses indexes, not names, for the parameters.

On the other hand, it is encouraged that the format string for ILogger uses named parameters (from which their positions are inferred). Thus, if a format string is localized and the positions of those named parameters change within the format string, then they will match to the wrong positions.

I'm not certain if ILogger supports using parameter indexes (my guess that it doesn't in the same sense as the above APIs, probably treats them as names), but then you lose fidelity of the parameter values when consuming the logs (e.g. via console logs, saying parameter "0" is "John" is less information than saying parameter "firstName" is "John").

jander-msft: > Therefore, it is up to the consumer of these APIs to carefully decide how to use these APIs based on localization (perhaps to flip parameter orders given a specific localization).

I think it is unlikely that anyone will be able to take an arbitrarily localized string and to know at execution time what the new parameter order is and then dynamically adjust the parameter order. Localization (at least via resx) does not provide rich enough information of the parameter reordering.

I could hypothetically say "if Spanish, use this order; if Japanese, use this order" etc, but that is not scalable to every individual string and can easily change if some decides to relocalize the string in a slightly different order.

Given that, I think the conclusion is that localization is highly discouraged due to reordering of named parameters, but I would love to be wrong.

tarekgh: It is impossible for the logging APIs to format according to the resource language. The reorder of the string parts is not easy to handle and need to know how the rendering of this string can be later. I can give tons of examples of that but the easiest one is when using BiDi languages which the order of the string parts can change if your render the whole string from right to left or render it from right to left. 

> Given that, I think the conclusion is that localization is highly discouraged due to reordering of named parameters, but I would love to be wrong.

I wouldn't say that though. The callers of the logging can do some testing with different languages and can decide if it is good enough or not. The guidelines should be users of the logging APIs need to do some localization testing.

jander-msft: > It is impossible for the logging APIs to format according to the resource language.

As logging is today, yes.

However, it could support parameter indexes. I haven't thought it through, but my first pass would be to allow optional inclusion of the parameter index in the slots e.g. "The object has the color {0:color} and has the shape {1:shape}". Whether these two slots are rearranged in any particular language is then not a worry of either the developer nor the localizer. This allows explicit assignment of which parameter index to get the value from as well the name of the parameter when it is logged. I don't know what other capabilities are allowed for each slot (is format specifiers a thing for logger format strings?), but this would be my first naive attempt.

> The callers of the logging can do some testing with different languages and can decide if it is good enough or not. The guidelines should be users of the logging APIs need to do some localization testing.

Given the current functionality, assessing all of the localized strings for a non-trivial project would be difficult. I could write a test that extracts all of the localized content and checks whether the order of the slots are correct. But when localization team inserts a new update in the future that breaks that arrangement, my recourses are (a) to tell them "no, you can't localize it that way" but I don't think it's my place to dictate how a native language speaker expresses that content, or (b) special case each and every violation to rearrange the parameters into what the language expects on the fly at runtime and add that new arrangement to the test. I feel like both of those solutions are untenable, but that's for each team to assess or maybe there is another low manual maintenance solution that I haven't thought.

Given all of that, I would make the argument for the projects that I work on to never localize ILogger statements.

tarekgh: > However, it could support parameter indexes.

This still wouldn't be enough to ensure the final resulted text would be in the right form for different languages. There is extreme complexity which usually handled by the rendering engines.  Imagine for a second the localized strings and the parameters are ended up in different languages and need to fit together. Look at [Unicode BiDi Algorithm](https://unicode.org/reports/tr9/) as example to get the sense of the difficulty to layout the resulted string in the right form.

jander-msft: > > However, it could support parameter indexes.
> 
> This still wouldn't be enough to ensure the final resulted text would be in the right form for different languages.

How so? If using indexes, it should not be any different algorithmically than how `string.Format` does slot replacement.

I don't know what "rendering engines" have to do with this. All this is about is matching the correct arguments with the indexes as described by the format string. It does not have to have any knowledge of the language of the format string.

PR #393, opened 2021-06-03
Add unit tests for Process Filtering

wiktork:
- Also rename existing unit test project


src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/DefaultProcessConfigurationTests.cs
```diff
@@ -0,0 +1,155 @@
...
+            var filter = CreateFilterEntry(processKey);
+            ValidateProcessFilter(filter, DiagProcessFilterCriteria.ProcessName, processKey.ProcessName);
+
+            processKey = new ProcessKey(Guid.NewGuid());
+            filter = CreateFilterEntry(processKey);
+            ValidateProcessFilter(filter, DiagProcessFilterCriteria.RuntimeId, processKey.RuntimeInstanceCookie.Value.ToString("D"));
+
+            processKey = new ProcessKey(5);
+            filter = CreateFilterEntry(processKey);
+            ValidateProcessFilter(filter, DiagProcessFilterCriteria.ProcessId, processKey.ProcessId.Value.ToString(System.Globalization.CultureInfo.InvariantCulture));
```
jander-msft, line 34: nit: Add System.Globalization as a using to minimize the length of this statement.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/DefaultProcessConfigurationTests.cs
```diff
@@ -0,0 +1,155 @@
...
+                DiagProcessFilterCriteria.ProcessId,
+                DiagProcessFilterCriteria.RuntimeId,
+                DiagProcessFilterCriteria.ProcessName,
+                DiagProcessFilterCriteria.CommandLine
+            }.Cast<int>().ToArray();
+
+            Assert.Equal(expectedValues.Length, actualValues.Length);
+        }
+
+        private static void ValidateProcessFilter(DiagProcessFilterEntry filter,
```
jander-msft, line 120: nit: rename from `filter` to `actualFilter`. Test verification methods typically differentiate between `expected` and `actual` values.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/DefaultProcessConfigurationTests.cs
```diff
@@ -0,0 +1,155 @@
...
+            DiagProcessFilterCriteria expectedCriteria,
```
jander-msft, line 121: Can the expected parameters be listed first? This is typical of test verification methods for readability. I know that means no default value for `expectedMatchType` but you can add another method, e.g. `ValidateProcessFilterExact` that passes this argument if you don't want to write out the argument in every usage.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/DefaultProcessConfigurationTests.cs
```diff
@@ -0,0 +1,155 @@
...
+            //Sorted by unsigned integer value
+            var expectedValues = Enum.GetValues(typeof(DiagProcessFilterCriteria));
+
+            var actualValues = new[]
+            {
+                DiagProcessFilterCriteria.ProcessId,
+                DiagProcessFilterCriteria.RuntimeId,
+                DiagProcessFilterCriteria.ProcessName,
+                DiagProcessFilterCriteria.CommandLine
+            }.Cast<int>().ToArray();
```
jander-msft, line 115: Why is the `.Cast<int>()` needed? The values in the arrays are not compared, only the length of the arrays.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/DefaultProcessConfigurationTests.cs
```diff
@@ -0,0 +1,155 @@
...
+
+            processKey = new ProcessKey(5);
+            filter = CreateFilterEntry(processKey);
+            ValidateProcessFilter(filter, DiagProcessFilterCriteria.ProcessId, processKey.ProcessId.Value.ToString(System.Globalization.CultureInfo.InvariantCulture));
+        }
+
+        [Fact]
+        public void ConvertProcessConfig()
+        {
+            var filterDescriptorPid = new ProcessFilterDescriptor
```
jander-msft, line 40: Based on the implementation, it seems that the combination of ProcessFilterKey.ProcessId and ProcessFilterType.Contains is not really supported and is forcible transformed to Exact in [this code](https://github.com/dotnet/dotnet-monitor/blob/99e439e9f85e9cb9bc5b64faf1c83ea9c712ce88/src/Microsoft.Diagnostics.Monitoring.RestServer/DiagProcessFilter.cs#L64). Can you add another entry in this test (or a separate test) to validate that this behavior does not change?

jander-msft (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/DefaultProcessConfigurationTests.cs
wiktork: I actually intended to compare the values too.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/DefaultProcessConfigurationTests.cs
```diff
@@ -0,0 +1,178 @@
...
+        [Fact]
+        public void NewCriteriaTest()
+        {
+            //When new enumerations (such as Entrypoint) are added to DiagProcessFilterCriteria,
+            //this test will fail. This means new unit tests should be written for that criteria and this test can be updated.
+
+            //Sorted by unsigned integer value
+            var expectedValues = Enum.GetValues(typeof(DiagProcessFilterCriteria));
+
+            var actualValues = new[]
```
jander-msft, line 121: nit: These are the expected values and the ones from Enum.GetValues are the actual values.

wiktork: I could see that, but I guess I don't think of it that way in this case...the expected case is that we have test coverage for everything in the enum. The actual values represent what we've actually written tests for. I'll leave this alone for now.

jander-msft: What we expect is the enum to have exactly those four values. What it actually has is found in the result of Enum.GetValues. Hence, when we add another enum value, we still expected four items in the test and actually have five.

PR #425, opened 2021-06-12
Add operational style api for egressed artifacts

wiktork:
Note this follows a similar approach as the Azure Sdk (https://docs.microsoft.com/en-us/azure/azure-maps/creator-long-running-operation)

Some design considerations/possible changes
- Status and error return the same status code but slightly different structure. We could use our own 2xx code here to make it easier for callers to determine the state.
- There is no bounds on the operations we track. They stay around forever.
- 202 Accepted is a good return value but can be easily missed by the client. E.g. curl with egress without a -v just shows nothing.
- We could consider a redirect for the above as well.

Remaining items:
- [ ] Add more logging and comments
- [x] Consider a name pass over all the newly introduced types
- [x] Unit tests
- [x] Account for infinite duration tasks
- [ ] Update documentation


src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
```diff
@@ -228,7 +228,10 @@ public static IHostBuilder CreateHostBuilder(IConsole console, string[] urls, st
                     services.AddSingleton<IValidateOptions<DiagnosticPortOptions>, DiagnosticPortValidateOptions>();
 
                     services.AddSingleton<IEndpointInfoSource, FilteredEndpointInfoSource>();
+                    services.AddSingleton<IEgressOperationQueue, EgressOperationQueue>();
```
jander-msft, line 231: I think we should create a "add everything that the rest server implements" extension method and make all of these an implementation detail.

src/Microsoft.Diagnostics.Monitoring.RestServer/Operation/EgressOperationStore.cs
```diff
@@ -0,0 +1,163 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Threading;
+using System.Threading.Channels;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Monitoring.RestServer
+{
+    public interface IEgressOperationStore
```
jander-msft, line 16: I this interface really necessary? Isn't this an internal implementation detail of the server assembly?

src/Microsoft.Diagnostics.Monitoring.RestServer/Operation/EgressOperationStore.cs
```diff
@@ -0,0 +1,163 @@
...
+                    });
+            }
+
+            //Kick off work to attempt egress
+            await _taskQueue.QueueBackgroundWorkItemAsync(request);
+
+            return operationId;
+        }
+
+        public Task CancelOperation(Guid operationId)
```
jander-msft, line 72: This and the below methods should not be Task-returning methods if they aren't async, if the service is just an implementation within the server assembly.

src/Microsoft.Diagnostics.Monitoring.RestServer/Operation/EgressOperationQueue.cs
```diff
@@ -0,0 +1,62 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Channels;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Monitoring.RestServer
+{
+    internal interface IEgressOperationQueue
```
jander-msft, line 14: Is this interface needed? It looks like this is an implementation detail of the server assembly.

src/Microsoft.Diagnostics.Monitoring.RestServer/Operation/EgressOperationQueue.cs
```diff
@@ -0,0 +1,62 @@
...
+        public EgressOperationQueue()
+        {
+            var options = new BoundedChannelOptions(256)
+            {
+                FullMode = BoundedChannelFullMode.Wait
+            };
+            _queue = Channel.CreateBounded<EgressRequest>(options);
+        }
+
+        public ValueTask QueueBackgroundWorkItemAsync(
```
jander-msft, line 36: nit: EnqueueAsync

src/Microsoft.Diagnostics.Monitoring.RestServer/Controllers/OperationController.cs
```diff
@@ -0,0 +1,61 @@
...
+        [HttpGet("{operationId}")]
+        [ProducesWithProblemDetails(ContentTypes.ApplicationJson)]
+        [ProducesResponseType(typeof(Models.OperationStatusSuccess), StatusCodes.Status201Created)]
+        [ProducesResponseType(typeof(Models.OperationStatus), StatusCodes.Status200OK)]
+        public async Task<IActionResult> GetOperationStatus(Guid operationId)
+        {
+            return await this.InvokeService(async () =>
+            {
+                Models.OperationStatusBase status = await _operationStore.GetOperationStatus(operationId);
+                return new JsonResult(status)
```
jander-msft, line 42: Let's not assume JSON. Try using `ControllerBase.StatusCode(int, object)`

src/Microsoft.Diagnostics.Monitoring.RestServer/Controllers/OperationController.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        [HttpDelete("{operationId}")]
+        [ProducesWithProblemDetails(ContentTypes.ApplicationJson)]
+        [ProducesResponseType(typeof(void), StatusCodes.Status200OK)]
+        public async Task<IActionResult> CancelOperation(Guid operationId)
+        {
+            return await this.InvokeService(async () =>
+            {
+                await _operationStore.CancelOperation(operationId);
+                return Ok();
```
jander-msft, line 57: Can you add a comment that CancelOperation will throw an exception if the operation ID doesn't exist? Reading it as-is, one might think it would return (and thus return 200) regardless of the existence of the operation ID.

jander-msft (changes requested):

src/Microsoft.Diagnostics.Monitoring.RestServer/Operation/EgressOperationStore.cs
wiktork: It's not that I'm creating a public api here, this is just standard dependency injection into the controller. I can use the class directly, but it will need to remain public.

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationService.cs
```diff
@@ -0,0 +1,48 @@
...
+            {
+                var item = await _queue.DequeueAsync(stoppingToken);
+
+                //We have two stopping tokens, one per item that can be triggered via Delete
+                //and if we are stopping the service
+                using (CancellationTokenSource linkedTokenSource =
+                    CancellationTokenSource.CreateLinkedTokenSource(item.CancellationTokenSource.Token, stoppingToken))
+                {
+                    CancellationToken token = linkedTokenSource.Token;
+                    var result = await item.EgressOperation.ExecuteAsync(_serviceProvider, token);
```
wiktork, line 41: This approach will not work; it forces serialization of all requests, and indefinite requests will block all other requests from the queue.

src/Tests/test.runsettings
```diff
@@ -2,7 +2,7 @@
 <RunSettings>
```
jander-msft, line 2: Let's just delete this file. I don't think setting this environment variable helps at all.

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationService.cs
jander-msft: Maybe you can create another async method that effectively does a Task.WhenAny on the in-flight executions + one task from a completion source that is signaled when when a new operation is dequeued. If that last task is signaled, just loop back around and start a new Task.WhenAny. If any other task, complete the operation.

Another option is to add continuations to each of the tasks that completes the operation when it finishes.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EgressTests.cs
```diff
@@ -0,0 +1,138 @@
...
+                Assert.True(operationResult.Status == OperationState.Running);
+
+                HttpStatusCode deleteStatus = await apiClient.CancelEgressOperation(response.OperationUri).ConfigureAwait(false);
+                Assert.Equal(HttpStatusCode.OK, deleteStatus);
+                operationResult = await apiClient.GetOperationStatus(response.OperationUri).ConfigureAwait(false);
+                Assert.Equal(OperationState.Cancelled, operationResult.Status);
+            });
+        }
+
+        private async Task ExecuteEgressScenario(Func<ApiClient, AppRunner, Task> func)
```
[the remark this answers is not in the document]
wiktork: I didn't see an easy way to hook configuration for the tool. I can see if it's easy to add.

documentation/openapi.json
```diff
@@ -727,6 +745,67 @@
           }
         }
       }
+    },
+    "/operation/{operationId}": {
```
kkeirstead, line 749: For consistency with the new format, should this be something like "/operation?{operationId}" instead? Looking at the rest of the PR, I'm assuming this hasn't been merged with the main branch in a while since other endpoints are still using the old format, but just something I wanted to make sure I pointed out :)

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/OperationController.cs
```diff
@@ -0,0 +1,62 @@
...
+    [Route(ControllerName)]
+    [ApiController]
+    [HostRestriction]
+    [Authorize(Policy = AuthConstants.PolicyName)]
+    public class OperationController : ControllerBase
+    {
+        private readonly ILogger<OperationController> _logger;
+        private readonly EgressOperationStore _operationStore;
+
+        public const string ControllerName = "operation";
```
jander-msft, line 26: nit: Should this be "operations" since it represents the entire collection of operations?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HttpApi/ApiClient.cs
```diff
@@ -320,10 +320,37 @@ public async Task<string> GetMetricsAsync(CancellationToken token)
             throw await CreateUnexpectedStatusCodeExceptionAsync(response).ConfigureAwait(false);
         }
 
+        public async Task<OperationResponse> EgressTraceAsync(int processId, int durationSeconds, string egressProvider, CancellationToken token)
+        {
+            string uri = FormattableString.Invariant($"/trace/{processId}?egressProvider={egressProvider}&durationSeconds={durationSeconds}");
```
kkeirstead, line 325: Just another case of using the old format -> "/trace?pid={processId}&..."

wiktork: Yep, I didn't fix these up after merging.

documentation/openapi.json
wiktork: I actually intend to leave these as they are. I don't think these are the same types of requests as our diagnostic based requests, and follow REST semantics more closely. I.e. an operation has a unique id, where as a process is not (and there are several ways to specify a process).

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressResult.cs
```diff
@@ -6,14 +6,11 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
 {
     internal struct EgressResult
     {
-        public EgressResult(string name, string value)
+        public EgressResult(string value)
```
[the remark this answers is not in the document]
wiktork: I would like to keep it in case we end up with some more data here at some point.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -672,21 +682,40 @@ private static LogFormat ComputeLogFormat(IList<MediaTypeHeaderValue> acceptedHe
...
+                string newUrl = this.Url.Action(
+                    action: nameof(OperationController.GetOperationStatus),
+                    controller: OperationController.ControllerName, new { operationId = operationId },
+                    protocol: this.HttpContext.Request.Scheme, this.HttpContext.Request.Host.ToString());
+
+                return Accepted(newUrl);
+            }
+            catch (TooManyRequestsException)
+            {
+                return this.StatusCode(StatusCodes.Status429TooManyRequests);
```
wiktork, line 718: Note this does not work as expected for netcore 3.1 (it works fine for 5.0). It appears that StatusCodeResult gets converted to ObjectResult, and then there's a content type mismatch. This produces a 406.

wiktork: Thinking about this some more, I think we are going up against the declared api controller behavior, which is to produce a problemdetails. My tentative thinking is to get rid of the middleware altogether and let the exception throw and then get converted to an appropriate problemdetails.

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/EgressOperationStatus.cs
```diff
@@ -0,0 +1,63 @@
...
+        public DateTime CreatedDateTime { get; set; }
+
+        public OperationState Status { get; set; }
+    }
+
+    /// <summary>
+    /// Represents the state of a long running operation. Used for all types of results, including
+    /// successes and failures.
+    /// </summary>
+    public class OperationStatus
```
kkeirstead, line 33: I don't fully understand the relationship, so it might not be appropriate, but would it make sense for OperationStatus to inherit from OperationSummary (since the summary should always be a subset of the full status)? The duplication of the getters and setters between the classes caught my eye, so I thought I'd mention it. Also, relatedly, am I correct in assuming Id is the same as OperationId? In which case I'd think they should probably be named the same, and if not maybe a comment to point out the difference?

wiktork: I thought about doing this but I really wanted to enforce that a new object had to be made here. I'll correct the names and add some comments.

src/Microsoft.Diagnostics.Monitoring.WebApi/RequestThrottling/RequestLimitTracker.cs
```diff
@@ -0,0 +1,61 @@
...
+            _requestLimitTable.Add(Controllers.DiagController.ArtifactType_Trace, 3);
+
+            _logger = logger;
+        }
+
+        public IDisposable Increment(string key, out bool allowOperation)
+        {
+            if (!_requestLimitTable.TryGetValue(key, out int maxConcurrency))
+            {
+                throw new InvalidOperationException($"Unexpected limit key: {key}");
```
wiktork, line 43: nit: Localize

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EgressTests.cs
```diff
@@ -0,0 +1,209 @@
...
+                    Assert.Equal(OperationState.Cancelled, operationResult.Status);
+
+                    await appRunner.SendCommandAsync(TestAppScenarios.AsyncWait.Commands.Continue);
+                },
+                configureTool: (toolRunner) =>
+                {
+                    toolRunner.WriteKeyPerValueConfiguration(new RootOptions().AddFileSystemEgress(FileProviderName, _tempEgressPath.FullName));
+                });
+        }
+
```
wiktork, line 99: Add another test for listing all operations

src/Microsoft.Diagnostics.Monitoring.WebApi/ActionContextExtensions.cs
```diff
@@ -91,4 +83,43 @@ private static bool LogInformation(ILogger logger, Exception ex)
...
+        public static async Task InvokeAsync(this ActionContext context, Func<CancellationToken, Task> action, ILogger logger)
+        {
+            //We want to handle two scenarios:
+            //1) These functions are part of an ExecuteResultAsync implementation, in which case
+            //they don't need to return anything. (The delegate below handles this case).
+            //2) They are part of deferred processing and return a result.
+            Func<CancellationToken, Task<ExecutionResult<object>>> innerAction = async (CancellationToken token) =>
+            {
+                await action(token);
+                return new ExecutionResult<object>();
```
jander-msft, line 115: nit: Maybe create a static instance and return that so long as ExecutionResult is immutable.

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationStore.cs
```diff
@@ -0,0 +1,166 @@
...
+                    new EgressEntry
+                    {
+                        State = Models.OperationState.Running,
+                        EgressRequest = request,
+                        OperationId = operationId
+                    });
+            }
+
+            //Kick off work to attempt egress
+            await _taskQueue.EnqueueAsync(request);
```
jander-msft, line 68:
```suggestion
            await _taskQueue.EnqueueAsync(request).ConfigureAwait(false);
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationService.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            //We have two stopping tokens, one per item that can be triggered via Delete
+            //and if we are stopping the service
+            using (CancellationTokenSource linkedTokenSource =
+                CancellationTokenSource.CreateLinkedTokenSource(egressRequest.CancellationTokenSource.Token, stoppingToken))
+            {
+                CancellationToken token = linkedTokenSource.Token;
+                token.ThrowIfCancellationRequested();
+
+                var result = await egressRequest.EgressOperation.ExecuteAsync(_serviceProvider, token);
```
jander-msft, line 53:
```suggestion
                var result = await egressRequest.EgressOperation.ExecuteAsync(_serviceProvider, token).ConfigureAwait(false);
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationService.cs
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+        {
+            while (!stoppingToken.IsCancellationRequested)
+            {
+                EgressRequest egressRequest = await _queue.DequeueAsync(stoppingToken);
+
+                //Note we do not await these tasks, but we do limit how many can be executed at the same time
+                _ = Task.Run( async ()=>
```
jander-msft, line 36: No need to await.

```suggestion
                _ = Task.Run(() => ExecuteEgressOptions(egressResult, stoppingToken));
```

if you do keep await, add `.ConfigureAwait(false)`

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperation.cs
```diff
@@ -13,34 +13,34 @@
...
-            ILogger<EgressStreamResult> logger = context.HttpContext.RequestServices
+            ILogger<EgressOperation> logger = serviceProvider
                 .GetRequiredService<ILoggerFactory>()
-                .CreateLogger<EgressStreamResult>();
+                .CreateLogger<EgressOperation>();
 
             using var _ = logger.BeginScope(_scope);
 
-            await context.InvokeAsync(async (token) =>
+            return await ExecutionHelper.InvokeAsync(async (token) =>
```
jander-msft, line 41: Please add `.ConfigureAwait(false);`

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -672,24 +682,37 @@ private static LogFormat ComputeLogFormat(IList<MediaTypeHeaderValue> acceptedHe
...
-                    scope);
+                    scope),
+                    limitKey: artifactType);
             }
         }
 
+        private async Task<ActionResult> SendToEgress(EgressOperation egressStreamResult, string limitKey)
+        {
+            // Will throw TooManyRequestsException if there are too many concurrent operations.
+            Guid operationId = await _operationsStore.AddOperation(egressStreamResult, limitKey);
```
jander-msft, line 707:
```suggestion
            Guid operationId = await _operationsStore.AddOperation(egressStreamResult, limitKey).ConfigureAwait(false);
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -220,13 +225,13 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
                     scope.AddArtifactType(ArtifactType_Dump);
                     scope.AddEndpointInfo(processInfo.EndpointInfo);
 
-                    return new EgressStreamResult(
+                    return await SendToEgress(new EgressOperation(
```
jander-msft, line 228: Add `.ConfigureAwait(false)`

src/Microsoft.Diagnostics.Monitoring.WebApi/ActionContextExtensions.cs
```diff
@@ -91,4 +83,43 @@ private static bool LogInformation(ILogger logger, Exception ex)
...
+            };
+
+            ExecutionResult<object> result = await ExecutionHelper.InvokeAsync(innerAction, logger, context.HttpContext.RequestAborted);
```
jander-msft, line 118:
```suggestion
            ExecutionResult<object> result = await ExecutionHelper.InvokeAsync(innerAction, logger, context.HttpContext.RequestAborted).ConfigureAwait(false);
```

src/Microsoft.Diagnostics.Monitoring.WebApi/ActionContextExtensions.cs
```diff
@@ -91,4 +83,43 @@ private static bool LogInformation(ILogger logger, Exception ex)
...
+            if (result.ProblemDetails != null)
+            {
+                await context.ProblemAsync(new BadRequestObjectResult(result.ProblemDetails));
```
jander-msft, line 121:
```suggestion
                await context.ProblemAsync(new BadRequestObjectResult(result.ProblemDetails)).ConfigureAwait(false);
```

src/Microsoft.Diagnostics.Monitoring.WebApi/ActionContextExtensions.cs
```diff
@@ -14,68 +14,60 @@
...
+        public ProblemDetails ProblemDetails { get; set; }
+    }
+
+    internal static class ExecutionHelper
+    {
+        private static ExecutionResult<T> ExceptionToResult<T>(Exception ex) =>
+            new ExecutionResult<T>{ ProblemDetails = ex.ToProblemDetails((int)HttpStatusCode.BadRequest) };
+
+        public static async Task<ExecutionResult<T>> InvokeAsync<T>(Func<CancellationToken, Task<ExecutionResult<T>>> action, ILogger logger,
+            CancellationToken token)
```
jander-msft, line 30: nit: Is this an extra newline before the method body?

src/Microsoft.Diagnostics.Monitoring.WebApi/ActionContextExtensions.cs
```diff
@@ -14,68 +14,60 @@
 
 namespace Microsoft.Diagnostics.Monitoring.WebApi
 {
-    internal static class ActionContextExtensions
+    public sealed class ExecutionResult<T>
```
jander-msft, line 17: nit: It may be better to make this immutable. Also, maybe add two static methods for `static ExecutionResult<T> Success<T>(T result)` and `static ExecutionResult<T> Failed<T>(ProblemDetails details)`

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/EgressOperationStatus.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        public Guid OperationId { get; set; }
+
+        public DateTime CreatedDateTime { get; set; }
+
+        public OperationState Status { get; set; }
+    }
+
+    /// <summary>
+    /// Represents the state of a long running operation. Used for all types of results, including
```
jander-msft, line 30: Please remove newline so that openapi document doesn't have `\r\n` in it.

jander-msft (changes requested):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HttpApi/ApiClient.cs
```diff
@@ -304,10 +304,53 @@ public async Task<string> GetMetricsAsync(CancellationToken token)
...
+            return response;
+        }
+
+        public async Task<OperationResponse> EgressTraceAsync(int processId, int durationSeconds, string egressProvider, CancellationToken token)
+        {
+            string uri = FormattableString.Invariant($"/trace?pid={processId}&egressProvider={egressProvider}&durationSeconds={durationSeconds}");
+            using HttpRequestMessage request = new(HttpMethod.Get, uri);
+            using HttpResponseMessage response = await SendAndLogAsync(request, HttpCompletionOption.ResponseContentRead, token).ConfigureAwait(false);
+
+            return new OperationResponse(response.StatusCode, response.Headers.Location, await response.Content.ReadAsStringAsync().ConfigureAwait(false));
```
jander-msft, line 320: This method should be enforcing successful before returning, something like:

```cs
switch (response.StatusCode)
{
    case HttpStatusCode.OK:
        ValidateContentType(response, ContentTypes.ApplicationJson);
        return new OperationResponse(response.StatusCode, response.Headers.Location, await response.Content.ReadAsStringAsync().ConfigureAwait(false));
    case HttpStatusCode.BadRequest:
    case HttpStatusCode.TooManyRequests:
        ValidateContentType(response, ContentTypes.ApplicationProblemJson);
        throw await CreateValidationProblemDetailsExceptionAsync(response).ConfigureAwait(false);        
    case HttpStatusCode.Unauthorized:
        ThrowIfNotSuccess(response);
        break;
}

throw await CreateUnexpectedStatusCodeExceptionAsync(response).ConfigureAwait(false);
```

and the test callers should be catching the exceptions when those are expected (e.g. when testing the throttling).

All of the new ApiClient methods should be do this.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HttpApi/ApiClient.cs
```diff
@@ -304,10 +304,53 @@ public async Task<string> GetMetricsAsync(CancellationToken token)
...
+            response.EnsureSuccessStatusCode();
+
+            return response.StatusCode;
+        }
+
         private static async Task<T> ReadContentAsync<T>(HttpResponseMessage responseMessage)
         {
             using Stream contentStream = await responseMessage.Content.ReadAsStreamAsync().ConfigureAwait(false);
-            return await JsonSerializer.DeserializeAsync<T>(contentStream).ConfigureAwait(false);
+            return await JsonSerializer.DeserializeAsync<T>(contentStream, new JsonSerializerOptions { PropertyNameCaseInsensitive = true }).ConfigureAwait(false);
```
jander-msft, line 353: Why is this necessary? The models used by the tests should be exactly the same as used by the product, so there should be no disagreement of casing.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Options/OptionsExtensions.cs
```diff
@@ -97,10 +115,23 @@ private static void MapObject(object obj, string prefix, IDictionary<string, str
             {
                 if (!property.GetIndexParameters().Any())
                 {
-                    MapValue(
-                        property.GetValue(obj),
-                        FormattableString.Invariant($"{prefix}{property.Name}"),
-                        map);
+                    if (property.GetCustomAttribute<System.Text.Json.Serialization.JsonExtensionDataAttribute>() != null)
```
jander-msft, line 118: I don't think this change is necessary anymore due to the egress configuration refactor.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/OperationController.cs
jander-msft: You resolved this comment, but I don't see any action or follow up on it. Are you planning on making this change?

jander-msft (approved):
I have a few more things I'd like for you to change mostly in the tests, but I'm not going to block the PR anymore.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/OperationController.cs
wiktork: I think I misunderstood the scope of your comment: you want to rename the entire controller not just the field.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HttpApi/ApiClient.cs
wiktork: I will annotate the model to reflect the default serialization style instead of doing this.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/OperationController.cs
jander-msft: Correct, I would expect the routes to be `/operations` and `/operations/{id}` to insinuate that I'm listing all of the operations and that I'm getting a single operation by id from the collection of operations, respectively.

Issue #85, opened 2021-03-17
Request for always-on/low overhead collection mode


timmydo:
## Background and Motivation

I have a program that utilizes a lot of event source tracing. Currently we have a custom tool on Windows that listens to these events using ETW and logs them to rotated files (every 15 min or after a couple of GB). We'd like a cross-platform replacement.

## Proposed Feature

dotnet-monitor to have a mode that always collects the configured event sources. Perhaps it could launch without a web server and accept an EventPipeConfiguration from a file. Also accept a RollingFileSystemEgressProvider that provides a way to configure the max output file size, max output duration, and max files to keep.

## Usage Examples

`dotnet monitor log config.json`

config.json:

```
{ "EventPipeConfiguration":
  {"Providers": [{"Name":"Microsoft-Windows-DotNETRuntime","Keywords":"32769","EventLevel":4,"Arguments":null}]
  "RequestRundown":false,
  "BufferSizeInMB":256
  }
  "EgressProviders": [
{"Type": "RollingFileSystem", "DirectoryPath": "...", "IntermediateDirectory": "..."}]
}
```

wiktork:
Likely need to finish #69 first

PR #447, opened 2021-06-16
Disabling infinite traces due to their inability to complete; the valid trace duration range now goes from 1 second to 2147483647 seconds

### LINKED ISSUE #334: Trace endpoint should not allow indefinite durations
Alternatively, we should stop collecting gracefully during cancellations if duration is -1, so as to make the nettrace file usable.

kkeirstead:
Trace files cannot be opened when terminated before completion. As a result, traces that run indefinitely will never be accessible to users. The new range allows traces to run between 1 second and a signed integer's max value (equating roughly to 68 years). 

Closes #334

### DIFF
--- a/documentation/api/trace-custom.md
+++ b/documentation/api/trace-custom.md
@@ -39,7 +39,7 @@ The default host address for these routes is `https://localhost:52323`. This rou
 | `pid` | path | false | int | The ID of the process. |
 | `uid` | path | false | guid | A value that uniquely identifies a runtime instance within a process. |
 | `name` | path | false | string | The name of the process. |
-| `durationSeconds` | query | false | int | The duration of the trace operation in seconds. Default is `30`. Min is `-1` (indefinite duration). Max is `2147483647`. |
+| `durationSeconds` | query | false | int | The duration of the trace operation in seconds. Default is `30`. Min is '1`. Max is `2147483647`. |
 | `egressProvider` | query | false | string | If specified, uses the named egress provider for egressing the collected trace. When not specified, the trace is written to the HTTP response stream. See [Egress Providers](../egress.md) for more details. |
 
 See [ProcessIdentifier](definitions.md#ProcessIdentifier) for more details about the `pid`, `uid`, and `name` parameters.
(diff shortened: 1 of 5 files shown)


jander-msft:
While I agree with the sentiment that specifying an infinite duration does not allow for a complete nettrace file, I'm not convinced that removing this ability is the correct thing to do. For example, it might be possible to take the content stream of the /trace response with an infinite duration and pass it into an EventPipeEventSource for dynamic observability of interesting events (this is something we should test). Another possibility is that if the process shuts down gracefully, then the session will be completed correctly, but I'm not certain (maybe @josalem may know) and not sure if we should give this scenario much weight.

I'm thinking that we should document that an infinite duration will not yield a valid nettrace file (due to cancellation) and leave the functionality as-is if the response stream is usable for on-demand observation.

@shirhatti, @wiktork what do you think?

shirhatti:
> I'm thinking that we should document that an infinite duration will not yield a valid nettrace file (due to cancellation) and leave the functionality as-is if the response stream is usable for on-demand observation.

Agreed on not changing dotnet-monitor

I view that as a limitation of the tooling. The format is pretty resilient and we should be able to parse out everything except the last incomplete block.

The way I'd solve it, is add API to TraceEvent to allow parsing incomplete nettrace files 😄

@noahfalk for his thoughts?

shirhatti:
@josalem I was just looking at `dotnet trace convert` and if I'm not mistaken we already can handle incomplete nettraces?

josalem:
It'll handle incomplete traces but doesn't guarantee you have all the information needed to symbolize in those cases. A truncated trace most likely does not have rundown in it which makes symbolization difficult.

noahfalk:
> I view that as a limitation of the tooling. The format is pretty resilient and we should be able to parse out everything except the last incomplete block.

> The way I'd solve it, is add API to TraceEvent to allow parsing incomplete nettrace files 😄

👍 Agreed

> It'll handle incomplete traces but doesn't guarantee you have all the information needed to symbolize in those cases. A truncated trace most likely does not have rundown in it which makes symbolization difficult.

This is certainly a scenario limitation, at least for now. In the future (.NET 7?) I hope we can support emitting the symbols on demand rather than waiting for rundown to eliminate this issue.

wiktork:
We may need to do something better with VS handling partial .nettrace files too. Right now VS silently fails to load partial files.

Note that if we implement a rolling egress provider, such as with https://github.com/dotnet/dotnet-monitor/issues/85, this experience would be much better.

PR #463, opened 2021-06-18
Switch to Bearer auth

### LINKED ISSUE #247: Don't use non-standard auth scheme
Originally posted by @idg10  

This seems to be planning to use a non-standard scheme, `MonitorApiToken` for the `Authorization` HTTP header. According to the HTTP standard for authentication ([RFC7235](https://tools.ietf.org/html/rfc7235)) the scheme is meant to be one of those registered in IANA's [Hypertext Transfer Protocol (HTTP) Authentication Scheme Registry](http://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml). (The alternative mechanism you mention, `Negotiate`, _is_ registered with IANA. So the problem is just with the novel `MonitorApiToken` you've proposed.)

The first obvious question is: why aren't you just using `Bearer`? [RFC6750](https://tools.ietf.org/html/rfc6750) defines this scheme, and the operation of your proposed `MonitorApiToken` seems to be a reinvention of the existing `Bearer` token scheme but by another name.

There's a really good reason for using the existing standard `Bearer` scheme: bearer tokens have certain risks attached to them. Because there's no proof of ownership mechanism, it's really important to ensure they are suitably protected. Since `Bearer` is a well-defined standard scheme, HTTP infrastructure (proxies, caching, APIs) can be built with a baked-in understanding of the risks, making it possible to impose policies that prevent accidental leaking of data. (E.g., an egress point could detect the use of a `Bearer` header in an unencrypted request and block it.) But by defining your own non-standard bearer token scheme, you will lose any such protection. **This use of a non-standard security scheme introduces security risks**, and for no obvious benefit.

Second, if you were to attempt to get IANA to register your new scheme, it would almost certainly be rejected because it violates the third bullet point in [RFC7235's 'Considerations for New Authentication Schemes'](https://tools.ietf.org/html/rfc7235#section-5.1.2). You are in effect using the `token68` production in your header here. This is exactly what `Bearer` does, but the `Bearer` scheme is allowed to get away with this because, as that section of RFC7235 says:

> The "token68" notation was introduced for compatibility with existing authentication schemes

I.e., `Bearer` already worked this way, and it's regrettable, but too late to change, so the current spec makes a special case for it. But as it says:

> new schemes ought to use the auth-param syntax instead

Your proposed `MonitorApiToken` scheme is definitely "new" so it should use the auth-param syntax defined in [RFC7235's Challenge and Response section](https://tools.ietf.org/html/rfc7235#section-2.1).

### LINKED ISSUE #589: Add tests for generatekey
We should add unit tests for the generatekey command. Because it now has output formats and runs in multiple versions of dotnet, we should test the full dotnet version x output format matrix.

kelltrick:
Switch to `Bearer` auth scheme from `MonitorApiKey`.

We are also switching the way in which tokens are generated and validated. We will use JSON Web Tokens that are signed to validate requests. See [api-key-format](https://github.com/dotnet/dotnet-monitor/blob/2943cf17902e895e86573215af7380ae323f37bd/documentation/api-key-format.md) for details.

Closes #247 
Closes #589


jander-msft:
Did you investigate the impact of this change for when we decide to add OAuth support? I'm not certain how different Bearer authentication handlers are supported simultaneously. I'd like to know that it is possible to do before we switch this to Bearer and then may have to undo it later if we find some kind of collision.

kelltrick:
> Did you investigate the impact of this change for when we decide to add OAuth support? I'm not certain how different Bearer authentication handlers are supported simultaneously. I'd like to know that it is possible to do before we switch this to Bearer and then may have to undo it later if we find some kind of collision.

I'm testing how to register OAuth. I believe this can registered as middleware correctly such that both auth handlers get called and we can use either type of token.

wiktork:
I think this change should include some documentation around the fact that our "Bearer" token isn't a traditional token. For example, normally a leaked bearer token is likely to expire in a short time, whereas our token never expires. We should make sure users understand that this functions like an Api key.

jander-msft:
We might want to consider a more real bearer token, especially if we are going to do #216

shirhatti:
@kelltrick I assume this isn't getting merged in preview6?

kelltrick:
> @kelltrick I assume this isn't getting merged in preview6?

That is correct, this got booted to preview 7. We will need to do a security review for this which will take a while.

src/Tools/dotnet-monitor/GenerateApiKeyCommandHandler.cs
```diff
@@ -2,10 +2,19 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Microsoft.Diagnostics.Monitoring.WebApi;
+using Microsoft.IdentityModel.Tokens;
+using Microsoft.Net.Http.Headers;
 using System;
+using System.Collections.Generic;
 using System.CommandLine;
```
kelltrick, line 10: trim these

src/Tools/dotnet-monitor/LoggingExtensions.cs
```diff
@@ -115,11 +116,11 @@ internal static class LoggingExtensions
                 logLevel: LogLevel.Warning,
                 formatString: Strings.LogFormatString_ApiKeyValidationFailure);
 
-        private static readonly Action<ILogger, string, Exception> _apiKeyAuthenticationOptionsChanged =
+        private static readonly Action<ILogger, string, Exception> _apiKeyAuthenticationOptionsValidated =
             LoggerMessage.Define<string>(
-                eventId: new EventId(22, "ApiKeyAuthenticationOptionsChanged"),
+                eventId: new EventId(22, "ApiKeyAuthenticationOptionsValidated"),
```
kelltrick, line 121: Don't re-use this key

kelltrick:
I have addressed everything from the meeting except adding tests for the GenerateKey command. This is a bigger task because we need a way to parse generic console output coming from dotnet-monitor. I created #589 to track.

documentation/api-key-format.md
```diff
@@ -0,0 +1,69 @@
+# API Key Format
+API Keys or MonitorApiKeys used in `dotnet monitor` are JSON Web Tokens or JWTs as defined by [RFC 7519: JSON Web Token (JWT)](https://www.rfc-editor.org/rfc/rfc7519.html).
+> **Note:** Because the API Key is a `Bearer` token, it should treated as a secret and always transmitted over `TLS` or other protected protocol.
```
kkeirstead, line 3:
```suggestion
> **Note:** Because the API Key is a `Bearer` token, it should be treated as a secret and always transmitted over `TLS` or other protected protocol.
```

documentation/api-key-format.md
```diff
@@ -0,0 +1,69 @@
...
+
+It is possible to make your own API Keys for `dotnet monitor` by following the format as defined below. Although, it is recommended to use the `generatekey` command unless you have a specific reason to make your own key.
+
+## Token Format
+In-order to use a JWT for authentication with `dotnet monitor`, the token must follow certain constraints. These constraints will be validated by `dotnet monitor` at configuration load time, authentication time, and authorization time.
+
+For this example, lets consider the API Key given on the [Authentication page](authentication.md). This is the entire portion passed as a `Bearer` type in the `Authorization` HTTP header:
```
kkeirstead, line 10:
```suggestion
For this example, let's consider the API Key given on the [Authentication page](authentication.md). This is the entire portion passed as a `Bearer` type in the `Authorization` HTTP header:
```

src/Tools/dotnet-monitor/LoggingExtensions.cs
```diff
@@ -225,5 +235,19 @@ public static void DuplicateEgressProviderIgnored(this ILogger logger, string pr
         {
             _duplicateEgressProviderIgnored(logger, providerName, providerType, existingProviderType, null);
         }
+
+        public static void ApiKeyOptionsChanged(this ILogger logger, MonitorApiKeyConfiguration newOptions)
+        {
+            _apiKeyOptionsChanged(logger, nameof(MonitorApiKeyOptions), newOptions.Subject, JsonSerializer.Serialize(newOptions.PublicKey), null);
+        }
+        public static void ApiKeyAuthenticationOptionsValidated(this ILogger logger)
```
kkeirstead, line 243: nit: newline

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests.csproj
```diff
@@ -20,17 +20,22 @@
...
     <Compile Include="..\..\Tools\dotnet-monitor\Egress\AzureBlob\AzureBlobEgressProviderOptions.cs" Link="Options\AzureBlobEgressProviderOptions.cs" />
     <Compile Include="..\..\Tools\dotnet-monitor\Egress\FileSystem\FileSystemEgressProviderOptions.cs" Link="Options\FileSystemEgressProviderOptions.cs" />
     <Compile Include="..\..\Tools\dotnet-monitor\Egress\IEgressProviderCommonOptions.cs" Link="Options\IEgressProviderCommonOptions.cs" />
+    <Compile Include="..\..\Tools\dotnet-monitor\Auth\MonitorApiKeyOptions.cs" Link="Options\MonitorApiKeyOptions.cs" />
+    <Compile Include="..\..\Tools\dotnet-monitor\Auth\AuthenticationOptions.cs" Link="Options\AuthenticationOptions.cs" />
   </ItemGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.FileFormats" Version="$(MicrosoftFileFormatsVersion)" />
+    <PackageReference Include="Microsoft.IdentityModel.Tokens" Version="6.11.1" />
```
wiktork, line 37: nit: Change to variable.

src/Tools/dotnet-monitor/Auth/LiveJwtBearerHandler.cs
```diff
@@ -0,0 +1,36 @@
...
+            {
+                // This is required to get AuthenticationHandler<JwtBearerOptions> to reload options.
+                // Once InitializeAsync is called, the value of the JwtBearerOptions is stored in Options and updates are never taken.
+                base.InitializeAsync(Scheme, Context);
+            });
+        }
+
+        protected override Task<AuthenticateResult> HandleAuthenticateAsync()
+        {
+            return base.HandleAuthenticateAsync();
```
wiktork, line 33: nit: Why the override, when you just call the base class?

src/Tools/dotnet-monitor/Auth/MonitorApiKeyPostConfigure.cs
```diff
@@ -0,0 +1,130 @@
...
+
+            JsonWebKey jwk = null; 
+            if (!string.IsNullOrEmpty(jwkJson))
+            {
+                try
+                {
+                    jwk = JsonWebKey.Create(jwkJson);
+                }
+                // JsonWebKey will throw only throw ArgumentException or a derived class.
+                catch (ArgumentException)
```
wiktork, line 80: Did you not want to propagate any of the exception message text?

kelltrick: I suppose it's safe to propagate the message? I'm used to never doing that because it can expose unnecessary info.

src/Tools/dotnet-monitor/GenerateApiKeyCommandHandler.cs
```diff
@@ -18,39 +22,83 @@ namespace Microsoft.Diagnostics.Tools.Monitor
...
+                    }
+                    break;
             }
 
-            GeneratedApiKey newKey = GeneratedApiKey.Create(keyLength, hashAlgorithm);
-
-            console.Out.WriteLine(FormattableString.Invariant($"Authorization: {Monitoring.WebApi.AuthConstants.ApiKeySchema} {newKey.MonitorApiKey}"));
-            console.Out.WriteLine(FormattableString.Invariant($"ApiKeyHash: {newKey.HashValue}"));
-            console.Out.WriteLine(FormattableString.Invariant($"ApiKeyHashType: {newKey.HashAlgorithm}"));
+            Console.Out.Write(outputBldr);
```
wiktork, line 84: note this uses console not Console.

src/Tools/dotnet-monitor/Auth/GeneratedJwtKey.cs
```diff
@@ -0,0 +1,58 @@
...
+            Claim issClaim = new Claim("iss", AuthConstants.ApiKeyJwtIssuer);
+            Claim subClaim = new Claim("sub", subjectStr);
+            JwtPayload newPayload = new JwtPayload(new Claim[] { issClaim, subClaim });
+
+            JwtSecurityToken newToken = new JwtSecurityToken(newHeader, newPayload);
+
+            ECDsa pubDsa = ECDsa.Create(dsa.ExportParameters(includePrivateParameters: false));
+            ECDsaSecurityKey pubSecKey = new ECDsaSecurityKey(pubDsa);
+            JsonWebKey jwk = JsonWebKeyConverter.ConvertFromECDsaSecurityKey(pubSecKey);
+            string publicKeyJson = JsonSerializer.Serialize(jwk, new JsonSerializerOptions() { DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull });
```
jander-msft, line 49: I don't think this compiles for .NET Core 3.1

src/Tools/dotnet-monitor/GenerateApiKeyCommandHandler.cs
```diff
@@ -18,39 +22,84 @@ namespace Microsoft.Diagnostics.Tools.Monitor
...
-                    string.Format(
-                        CultureInfo.CurrentCulture,
-                        Strings.ErrorMessage_ParameterNotAllowedByteRange,
-                        nameof(hashAlgorithm),
-                        hashAlgorithm,
-                        ApiKeyAuthenticationHandler.ApiKeyByteMinLength,
-                        ApiKeyAuthenticationHandler.ApiKeyByteMaxLength));
-                return Task.FromResult(1);
+                case OutputFormat.Json:
+#if NET5_0_OR_GREATER
```
jander-msft, line 53: I've now seen variations of this at least three times. I'm thinking it might be goo to write an extension method for JsonSerializerOptions (e.g. `public static JsonSerializerOptions SkipWritingNull(this JsonSerializerOptions options)`) so then we don't have to condition the compilation and it might be more concise, e.g.:

```cs
JsonSerializer.Serialize(options, new JsonSerializerOptions() { WriteIndented = true }.SkipWritingNull());
```

or maybe a static method that just creates a new instance of JsonSerializerOptions with ignoring null by default e.g.

```cs
JsonSerializerOptions jsonOptions = CreateSkipWritingNull();
jsonOptions .WriteIndented = true;
JsonSerializer.Serialize(options, jsonOptions);
```

Basically, any solution that avoids scattering #if NET5_0_OR_GREATER everywhere for this is better.

src/Tools/dotnet-monitor/GenerateApiKeyCommandHandler.cs
```diff
@@ -18,39 +22,84 @@ namespace Microsoft.Diagnostics.Tools.Monitor
...
+                    break;
+                case OutputFormat.Text:
+                    outputBldr.AppendLine(FormattableString.Invariant($"Subject: {newJwt.Subject}"));
+                    outputBldr.AppendLine(FormattableString.Invariant($"PublicKey: {newJwt.PublicKey}"));
+                    break;
+                case OutputFormat.Cmd:
+                case OutputFormat.PowerShell:
+                case OutputFormat.Bash:
+                    IDictionary<string, string> optList = opts.ToEnvironmentConfiguration();
+                    foreach (KeyValuePair<string, string> kvp in optList)
```
jander-msft, line 76: nit: This is a great candidate for using a deconstructor:

```suggestion
                    foreach ((string name, string value) in optList)
```

src/Tools/dotnet-monitor/Program.cs
```diff
@@ -166,4 +158,12 @@ internal enum ConfigDisplayLevel
         Redacted,
         Full,
     }
+    internal enum OutputFormat
+    {
+        Json,
+        Text,
+        Cmd,
+        Bash,
```
jander-msft, line 166: This is probably better called Shell.

src/Tools/dotnet-monitor/Startup.cs
```diff
@@ -176,8 +176,8 @@ public void ConfigureServices(IServiceCollection services)
             app.UseAuthentication();
             app.UseAuthorization();
 
-            CorsConfiguration corsConfiguration = new CorsConfiguration();
-            Configuration.Bind(nameof(CorsConfiguration), corsConfiguration);
+            CorsConfigurationOptions corsConfiguration = new CorsConfigurationOptions();
+            Configuration.Bind(nameof(CorsConfigurationOptions), corsConfiguration);
```
jander-msft, line 180: This changes the root option name from "CorsConfiguration" to "CorsConfigurationOptions". Please use `ConfigurationKeys.CorsConfiguration` to preserve the existing name in configuration.

src/Tools/dotnet-monitor/dotnet-monitor.csproj
```diff
@@ -22,10 +22,12 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.Storage.Blobs" Version="$(AzureStorageBlobsVersion)" />
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="3.1.17" />
```
jander-msft, line 25: Please use the version variables from eng/Version.props

jander-msft (changes requested):

src/Tools/dotnet-monitor/Auth/JwtAlgorithmChecker.cs
```diff
@@ -0,0 +1,64 @@
...
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    internal static class JwtAlgorithmChecker
+    {
+        /// <summary>
+        /// This is the list of allowed algorithms for JWS that are public/private key based.
+        /// We also reject RSASSA-PSS because .net core does not have support for this algorithm.
+        /// </summary>
+        /// <remarks>
+        /// We enforce this list to prevent storage of private key infromation in configuration.
```
kkeirstead, line 15:
```suggestion
        /// We enforce this list to prevent storage of private key information in configuration.
```

src/Tools/dotnet-monitor/Auth/JwtBearerChangeTokenSource.cs
```diff
@@ -0,0 +1,54 @@
...
+        public JwtBearerChangeTokenSource(
+            IOptionsMonitor<MonitorApiKeyConfiguration> optionsMonitor)
+        {
+            _optionsMonitor = optionsMonitor;
+            _changeRegistration = _optionsMonitor.OnChange(OnReload);
+        }
+
+        /// <summary>
+        /// Returns Named config instance. <see cref="AuthenticationHander&lt;JwtBearerOptions>" /> expects
+        /// it's configuration to be named after the AuthenticationScheme it's using, not <see cref="Options.DefaultName"/>.
```
kkeirstead, line 35:
```suggestion
        /// its configuration to be named after the AuthenticationScheme it's using, not <see cref="Options.DefaultName"/>.
```

src/Tools/dotnet-monitor/Auth/LiveJwtBearerHandler.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.Extensions.Options;
+using System.Text.Encodings.Web;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authentication.JwtBearer;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    /// <summary>
+    /// This is a version of <see cref="JwtBearerHandler" /> that will refresh the 
+    /// stored JwtBearerOptionsoptions when the settings change.
```
kkeirstead, line 16: I'm assuming this should be either "JwtBearerOptions" or "JwtBearerOptions options"

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -67,21 +67,25 @@ public void Write(IConfiguration configuration, bool full)
...
                     _writer.WriteStartObject();
-                    ProcessChildSection(authSection, nameof(ApiAuthenticationOptions.ApiKeyHash), includeChildSections: false, redact: true);
-                    ProcessChildSection(authSection, nameof(ApiAuthenticationOptions.ApiKeyHashType), includeChildSections: false, redact: false);
-                    _writer.WriteEndObject();
+                    IConfigurationSection monitorApiKey = ProcessChildSection(auth, ConfigurationKeys.MonitorApiKey, includeChildSections: false);
+                    if (monitorApiKey != null)
+                    {
+                        ProcessChildSection(monitorApiKey, nameof(MonitorApiKeyOptions.Subject), includeChildSections: false, redact: false);
+                        // The PublicKey should only ever contain the public key, however we expect that accidents may occur and we should
+                        // redact this field in the event the JWK contians the private key information.
```
kkeirstead, line 84:
```suggestion
                        // redact this field in the event the JWK contains the private key information.
```

documentation/authentication.md
```diff
@@ -48,29 +56,31 @@ If you're running on Windows, you can save these settings to `%USERPROFILE%\.dot
...
+    "MonitorApiKey": {
+      "Subject": "e18fe6c1-1365-4b13-b3eb-0757c7af882c",
+      "PublicKey": "eyJBZGRpdGlvbmFsRGF0YSI6e30sIkNydiI6IlAtMzg0IiwiS2V5T3BzIjpbXSwiS3R5IjoiRUMiLCJYIjoic25vTkt3dDdjRWlFeVN3LVN6S3lwX3NzWkhnc09tcWQzcUd5dndhZjVCb3ZrRFJzblVwbUNfNXI0THFmS1Y1cyIsIlg1YyI6W10sIlkiOiJGNFp1ZmhKN2hhbTJTbmRTMnNKQVBMcWhsZS1uZUI2NmFPSnkzek1Vb0VmSU9IT0llRHNndGdCNk1iRC00Y182IiwiS2V5U2l6ZSI6Mzg0LCJIYXNQcml2YXRlS2V5IjpmYWxzZSwiQ3J5cHRvUHJvdmlkZXJGYWN0b3J5Ijp7IkNyeXB0b1Byb3ZpZGVyQ2FjaGUiOnt9LCJDYWNoZVNpZ25hdHVyZVByb3ZpZGVycyI6dHJ1ZSwiU2lnbmF0dXJlUHJvdmlkZXJPYmplY3RQb29sQ2FjaGVTaXplIjozMn19"
+    }
   }
 }
 ```
 
-Alternatively, you can use environment variables to specify the configuration.
+Alternatively, you can use environment variables to specify the configuration. Use `dotnet monitor generatekey --output sh` to get the format below.
```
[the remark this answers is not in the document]
kelltrick: changed the name mid-writing the doc, I updated this to `bash` which is what it should be

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/LoggingRunnerAdapter.cs
```diff
@@ -107,20 +107,25 @@ public async Task StartAsync(CancellationToken token)
             _standardOutputTask = ReadLinesAsync(_runner.StandardOutput, _standardOutputLines, ReceivedStandardOutputLine, _cancellation.Token);
         }
 
-        public async Task<int> WaitForExitAsync(CancellationToken token)
+        public async Task<int?> WaitForExitAsync(CancellationToken token)
```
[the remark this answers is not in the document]
kelltrick: I made this throw now.

documentation/api-key-format.md
```diff
@@ -0,0 +1,69 @@
+# API Key Format
+API Keys or MonitorApiKeys used in `dotnet monitor` are JSON Web Tokens or JWTs as defined by [RFC 7519: JSON Web Token (JWT)](https://www.rfc-editor.org/rfc/rfc7519.html).
+> **Note:** Because the API Key is a `Bearer` token, it should be treated as a secret and always transmitted over `TLS` or anther protected protocol.
+
+It is possible to make your own API Keys for `dotnet monitor` by following the format as defined below. Although, it is recommended to use the `generatekey` command unless you have a specific reason to make your own key.
+
+## Token Format
+In-order to use a JWT for authentication with `dotnet monitor`, the token must follow certain constraints. These constraints will be validated by `dotnet monitor` at configuration load time, authentication time, and authorization time.
+
+For this example, let's consider the API Key given on the [Authentication page](authentication.md). This is the entire portion passed as a `Bearer` type in the `Authorization` HTTP header:
```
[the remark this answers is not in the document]
kelltrick: I put "ffffffffffff" in every string to obscure the real values.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Runners/MonitorRunner.cs
```diff
@@ -103,6 +103,8 @@ internal sealed class MonitorRunner : IAsyncDisposable
         /// </summary>
         public bool DisableMetricsViaCommandLine { get; set; }
 
+        public Action<ConsoleLogEvent> ReceiveConsoleLogEvent { get; set; }
```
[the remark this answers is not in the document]
kelltrick: This is used by the PrivateKeyWarn test to ensure that we get a warning for including a private key in the publickey field

src/Tools/dotnet-monitor/Auth/GeneratedJwtKey.cs
kelltrick: This does not compile with `System.Text.Json` < 5.0. However, dotnet-monitor uses System.Text.Json version 5.0.0 netcoreapp3.0 assembly. This apparently is actually not a breaking change but is a breaking change in an independently versioning assembly.

src/Tools/dotnet-monitor/GenerateApiKeyCommandHandler.cs
kelltrick: This is my new favorite language sugar.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Options/OptionsExtensions.cs
```diff
@@ -78,91 +38,143 @@ public static RootOptions AddFileSystemEgress(this RootOptions options, string n
             return options;
         }
 
-        private static void MapDictionary(IDictionary dictionary, string prefix, IDictionary<string, string> map)
+        /// <summary>
+        /// Sets API Key authentication. Use this one by default
```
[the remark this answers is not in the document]
kelltrick: I updated the comment . This is the overload that should be used for everything unless specifically testing auth.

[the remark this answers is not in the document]
kelltrick: Updated

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Options/OptionsExtensions.cs
```diff
@@ -78,91 +38,143 @@ public static RootOptions AddFileSystemEgress(this RootOptions options, string n
...
-                {
-                    string keyString = Convert.ToString(key, CultureInfo.InvariantCulture);
-                    MapValue(
-                        value,
-                        FormattableString.Invariant($"{prefix}{keyString}"),
-                        map);
-                }
-            }
+            string subjectStr = subject.ToString("D");
+            const string ApiKeyJwtIssuer = "dotnet-monitor+MonitorApiKey";
```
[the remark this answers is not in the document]
kelltrick: Moved, I had the same through when I was going through this

src/Tools/dotnet-monitor/Auth/MonitorApiKeyConfiguration.cs
```diff
@@ -3,17 +3,22 @@
...
 
 namespace Microsoft.Diagnostics.Tools.Monitor
 {
-    internal sealed class ApiKeyAuthenticationOptions : AuthenticationSchemeOptions
+    /// <summary>
+    /// Represents internal version of <see cref="MonitorApiKeyOptions"/>.
+    /// This object contains the validation state of the object and the decoded
+    /// Json Web Key.
+    /// </summary>
+    internal class MonitorApiKeyConfiguration : AuthenticationSchemeOptions
```
jander-msft, line 18: This now feels like it it an intermediate options class that could just be folded into JwtBearerOptions (not literally put the properties on there, but in its IPostConfigureOptions and IValidateOptions implementations, you can do the same logic and then remove this options class). This would reduce the complexity of this code a tiny bit and probably remove one or two of the extra reloads.

jander-msft (changes requested):

src/Tools/dotnet-monitor/Auth/MonitorApiKeyConfiguration.cs
jander-msft: Maybe don't do this simplification now. We can try it after the rest of it gets in.

documentation/api-key-format.md
```diff
@@ -0,0 +1,73 @@
...
+### Header
+The header (decoded from the token above) must contain at least 2 elements: `alg` (or [Algorithm](https://www.rfc-editor.org/rfc/rfc7518.html#section-3.1)), and `typ` (or [Type](https://www.rfc-editor.org/rfc/rfc7519.html#section-5.1)). `dotnet monitor` expects the `typ` to always be `JWT` for a JSON Web Token. `dotnet monitor` supports 6 `alg` values: `ES256`, `ES384`, `ES512`, `RS256`, `RS384`, and `RS512`.
+
+```json
+{
+  "alg": "ES384",
+  "typ": "JWT"
+}
+```
+>**Note:** The `alg` requirement is designed to enforce `dotnet monitor` to use public/private key signed tokens. This allows the key that is stored in configuration (as `Authentication__MonitorApiKey__PublicKey`) to only contain public key information and thus does not need to be kept secret.
```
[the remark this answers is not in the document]
kelltrick: We do use the `__` syntax everywhere, so I just followed along.

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/LoggingRunnerAdapter.cs
```diff
@@ -108,42 +108,77 @@ public async Task StartAsync(CancellationToken token)
...
                     if (null == line)
+                    {
+                        readAborted = true;
                         break;
+                    }
 
                     lines.Add(line);
                     callback?.Invoke(line);
                 }
+                if (!readAborted && _finishReads)
```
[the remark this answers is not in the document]
kelltrick: This solves an issue where during shutdown, the rest of the stream can be dropped on the floor. It doesn't show up in the normal tests for `collect` because dotnet-monitor takes a while to shutdown and gives the `.ReadLineAsync()` operations a while to catch up but for `generatekey` it runs too fast and the log gets clipped.
The readAborted is the preserve the initial behavior of successfully cleaning up everything on `.DisposeAsync()`

src/Tools/dotnet-monitor/Program.cs
```diff
@@ -175,4 +167,12 @@ internal enum ConfigDisplayLevel
         Redacted,
         Full,
     }
+    internal enum OutputFormat
```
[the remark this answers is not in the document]
kelltrick: Also added a new line between `ConfigDisplayLevel` and `Program`

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/LoggingRunnerAdapter.cs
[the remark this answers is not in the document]
kelltrick: I added some comments explaining what this is doing.

kelltrick (commented):
GitHub, please let me comment again

src/Tools/dotnet-monitor/Auth/MonitorApiKeyConfigurationObserver.cs
```diff
@@ -31,31 +31,34 @@ IOptionsMonitor<ApiKeyAuthenticationOptions> options
...
+        private void CheckMonitorApiKeyOptions(MonitorApiKeyConfiguration options)
         {
             // ValidationErrors will be null if API key authentication is not enabled.
             if (null != options.ValidationErrors && options.ValidationErrors.Any())
             {
                 _logger.ApiKeyValidationFailures(options.ValidationErrors);
             }
+            else
+            {
+                _logger.ApiKeyAuthenticationOptionsValidated();
```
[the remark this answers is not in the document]
kelltrick: I will leave this event but remove the changed event; which was my intention.
Current: we always get `22:ApiKeyAuthenticationOptionsChanged` maybe then followed by `21:ApiKeyValidationFailure`(s).
Will become: either `26:ApiKeyAuthenticationOptionsValidated` *or* `21:ApiKeyValidationFailure`(s)

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/GenerateKeyTests.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public async Task GenerateKeyJson(bool explicitPassFormat)
+        {
+            const string jsonStartDelim = "Settings in Json format:";
+            string[] regexSpecs = new string[]
+            {
+                "^Authorization: Bearer [a-zA-Z0-9_-]{2,5000}\\.[a-zA-Z0-9_-]{2,5000}\\.[a-zA-Z0-9_-]{2,5000}$",
```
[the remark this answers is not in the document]
kelltrick: Done. I was also able to use this to validate that the tokens provided are valid against the config parameters now too.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunner.cs
```diff
@@ -5,67 +5,60 @@
...
+        protected readonly string _runnerTmpPath =
             Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString("D"));
 
         private bool _isDisposed;
 
         /// <summary>
         /// The path of the currently executing assembly.
         /// </summary>
-        private static string CurrentExecutingAssemblyPath =>
+        protected static string CurrentExecutingAssemblyPath =>
```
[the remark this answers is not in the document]
kelltrick: I went through all the fields and properties and made them as `private` as I could

jander-msft (changes requested):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunner.cs
```diff
@@ -3,71 +3,52 @@
...
-
-        private readonly ITestOutputHelper _outputHelper;
-
-        private readonly TaskCompletionSource<string> _readySource =
-            new(TaskCreationOptions.RunContinuationsAsynchronously);
-
-        private readonly DotNetRunner _runner = new();
-
-        private readonly string _runnerTmpPath =
+        protected readonly string _runnerTmpPath =
```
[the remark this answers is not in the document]
kelltrick: I think this commend is out of date, this was fixed in [c6022994cf8db8fb88eec2ef2af678554ed2855d](https://github.com/dotnet/dotnet-monitor/pull/463/commits/c6022994cf8db8fb88eec2ef2af678554ed2855d)

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunner.cs
```diff
@@ -240,54 +154,20 @@ public async Task StartAsync(CancellationToken token)
...
-            _readySource.TrySetCanceled(token);
         }
 
-        private void StandardOutputCallback(string line)
+        public virtual async Task WaitForExitAsync(CancellationToken token)
         {
-            ConsoleLogEvent logEvent = JsonSerializer.Deserialize<ConsoleLogEvent>(line);
-
-            switch (logEvent.Category)
+            await Task.Run(() => _runner.ExitedTask.Wait(token)).ConfigureAwait(false);
```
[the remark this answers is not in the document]
kelltrick: I'm not going to do this suggestion, this will cause a leak to occur in the framework whenever the process exits and the cancellation token isn't cancelled. [`DelayPromiseWithCancellation `](https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs#L5625) will fail to call `.Dispose()` on it's `CancellationTokenRegistration`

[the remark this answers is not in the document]
jander-msft: Oh, because of the infinite timeout. That's really easy to fix though... just use some absurdly large timeout value or a relatively smallish one and then do:

```cs
Task timeoutTask = Task.Delay(nonInfiniteTimeout, token);
Task completedTask = await Task.WhenAny(_runner.ExitedTask, timeoutTask).ConfigureAwait(false);
token.ThrowIfCancellationRequested();
if (completedTask == timeoutTask) throw new TimeoutException();
```

but the other suggestion still stands as being simpler:

```cs
await _runner.WaitForExitAsync(token).ConfigureAwait(false);
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunner.cs
```diff
@@ -240,54 +154,20 @@ public async Task StartAsync(CancellationToken token)
...
+            if (token.IsCancellationRequested)
```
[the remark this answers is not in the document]
kelltrick: `await Task.Run(() => _runner.ExitedTask.Wait(token)).ConfigureAwait(false);` won't throw if cancellation is requested. `.Wait()` will happily just return when cancellation occurs. So this check is required.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunner.cs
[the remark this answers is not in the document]
kelltrick: I figured out a way to make this work and work around the problem with the leak from an infinite waiting task. I also made it an extension method since it's pretty verbose.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunner.cs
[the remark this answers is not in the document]
kelltrick: Converted this phrase to use `.WithCancellation(token)` but it still doesn't throw, so this change isn't needed.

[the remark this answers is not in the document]
jander-msft: I see, `Task.Wait` returns a `bool` signifying whether the wait successfully completed or not.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorGenerateKeyRunner.cs
```diff
@@ -0,0 +1,235 @@
...
+        // Completion source containing the bearer token emitted by the generatekey command
+        private readonly TaskCompletionSource<string> _bearerTokenTaskSource =
+            new(TaskCreationOptions.RunContinuationsAsynchronously);
+        private readonly Regex _bearerTokenRegex = 
+            new Regex("^Authorization: Bearer (?<token>[a-zA-Z0-9_-]+\\.[a-zA-Z0-9_-]+\\.[a-zA-Z0-9_-]+)$", RegexOptions.Compiled);
+
+        // Completion source containing the format type emitted by the generatekey command
+        private readonly TaskCompletionSource<string> _formatHeaderSource =
+            new(TaskCreationOptions.RunContinuationsAsynchronously);
+        private bool _formatHeaderSeen = false;
```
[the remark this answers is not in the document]
kelltrick: 👍  I was looking for that property but didn't think to look off the `.Task` for it.

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/Microsoft.Diagnostics.Monitoring.ConfigurationSchema.csproj
```diff
@@ -3,10 +3,11 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net5.0</TargetFramework>
+    <DefineConstants>$(DefineConstants);SCHEMAGEN</DefineConstants>
```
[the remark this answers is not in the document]
kelltrick: Done.

src/Microsoft.Diagnostics.Monitoring.Options/RootOptions.cs
```diff
@@ -4,13 +4,19 @@
```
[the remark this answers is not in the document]
kelltrick: Done.

kelltrick:
@jander-msft the next round of changes should be ready now.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunner.cs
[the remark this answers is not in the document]
kelltrick: On second though, you're probably right, we should throw if the cancellation token gets fired.

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/TaskExtensions.cs
```diff
@@ -40,5 +41,28 @@ public static async Task<T> SafeAwait<T>(this Task<T> task, ITestOutputHelper ou
             }
             return fallbackValue;
         }
+
+        public static async Task WithCancellation(this Task task, CancellationToken token)
+        {
+            CancellationTokenSource localCancellationTokenSource = new();
```
jander-msft, line 47: This new CTS is not needed. You can create a linked one from just the `token` parameter and then cancel it in the finally.

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/TaskExtensions.cs
```diff
@@ -40,5 +41,28 @@ public static async Task<T> SafeAwait<T>(this Task<T> task, ITestOutputHelper ou
...
+            CancellationTokenSource combinedTokenSource = CancellationTokenSource.CreateLinkedTokenSource(token, localCancellationTokenSource.Token);
+
+            try
+            {
+                // Creating a Task.Delay with Infinite timeout is not intended to work with a cancellation token that never fires
+                // So we need to make a local "combined" token that we can cancel if the provided CancellationToken token is never canceled
+                // This allows the framework to properly cleanup the Task and it's associated token registrations
+                Task waitOnCancellation = Task.Delay(Timeout.Infinite, combinedTokenSource.Token);
+                await Task.WhenAny(task, waitOnCancellation).ConfigureAwait(false);
```
jander-msft, line 56: This will cause a TaskCancelledException to be thrown when either task is cancelled.
```suggestion
                await Task.WhenAny(task, waitOnCancellation).Unwrap().ConfigureAwait(false);
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunner.cs
```diff
@@ -240,54 +162,17 @@ public async Task StartAsync(CancellationToken token)
...
         }
 
-        private void CancelCompletionSources(CancellationToken token)
+        public virtual async Task WaitForExitAsync(CancellationToken token)
         {
-            _defaultAddressSource.TrySetCanceled(token);
-            _metricsAddressSource.TrySetCanceled(token);
-            _readySource.TrySetCanceled(token);
+            await RunnerExitedTask.WithCancellation(token).ConfigureAwait(false);
+            token.ThrowIfCancellationRequested();
```
jander-msft, line 170: With the suggested change in WithCancellation, this won't be necessary. It also doesn't really hurt.

jander-msft (approved):

PR #468, opened 2021-06-19
Add Support For Json Text Sequence When Acquiring Logs

### LINKED ISSUE #72: Add support for json text sequence
MIME Type: application/json-seq

https://tools.ietf.org/html/rfc7464

This is in addition to ndjson.

kkeirstead:
This is in addition to the existing NDJson and Text options. 

Uses the standard for beginning and end characters defined in: https://tools.ietf.org/html/rfc7464

This is an example of the new output (it looks very similar to the existing NDJson format); the rectangle character is 0x1E (record separator character) and entries are separated by 0x0A (line feed character).

Json_72_Example

Closes #72

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
@@ -42,6 +42,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
 
         private const Models.TraceProfile DefaultTraceProfiles = Models.TraceProfile.Cpu | Models.TraceProfile.Http | Models.TraceProfile.Metrics;
         private static readonly MediaTypeHeaderValue NdJsonHeader = new MediaTypeHeaderValue(ContentTypes.ApplicationNdJson);
+        private static readonly MediaTypeHeaderValue JsonHeader = new MediaTypeHeaderValue(ContentTypes.ApplicationJson);
         private static readonly MediaTypeHeaderValue EventStreamHeader = new MediaTypeHeaderValue(ContentTypes.TextEventStream);
 
         private readonly ILogger<DiagController> _logger;
@@ -360,7 +361,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
         /// <param name="level">The level of the logs to capture.</param>
         /// <param name="egressProvider">The egress provider to which the logs are saved.</param>
         [HttpGet("logs/{processKey?}", Name = nameof(CaptureLogs))]
-        [ProducesWithProblemDetails(ContentTypes.ApplicationNdJson, ContentTypes.TextEventStream)]
+        [ProducesWithProblemDetails(ContentTypes.ApplicationNdJson, ContentTypes.ApplicationJson, ContentTypes.TextEventStream)]
         [ProducesResponseType(typeof(void), StatusCodes.Status429TooManyRequests)]
         [ProducesResponseType(typeof(string), StatusCodes.Status200OK)]
         [RequestLimit(MaxConcurrency = 3)]
@@ -405,7 +406,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
         /// <param name="durationSeconds">The duration of the logs session (in seconds).</param>
         /// <param name="egressProvider">The egress provider to which the logs are saved.</param>
         [HttpPost("logs/{processKey?}", Name = nameof(CaptureLogsCustom))]
-        [ProducesWithProblemDetails(ContentTypes.ApplicationNdJson, ContentTypes.TextEventStream)]
+        [ProducesWithProblemDetails(ContentTypes.ApplicationNdJson, ContentTypes.ApplicationJson, ContentTypes.TextEventStream)]
         [ProducesResponseType(typeof(void), StatusCodes.Status429TooManyRequests)]
         [ProducesResponseType(typeof(string), StatusCodes.Status200OK)]
         [RequestLimit(MaxConcurrency = 3)]
@@ -483,7 +484,20 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
             }
 
             string fileName = FormattableString.Invariant($"{GetFileNameTimeStampUtcNow()}_{processInfo.EndpointInfo.ProcessId}.txt");
-            string contentType = format == LogFormat.EventStream ? ContentTypes.TextEventStream : ContentTypes.ApplicationNdJson;
+            string contentType = ContentTypes.TextEventStream;
+            
+            if (format == LogFormat.EventStream)
+            {
+                contentType = ContentTypes.TextEventStream;
+            }
+            else if (format == LogFormat.NDJson)
+            {
+                contentType = ContentTypes.ApplicationNdJson;
+            }
+            else if (format == LogFormat.Json)
+            {
+                contentType = ContentTypes.ApplicationJson;
+            }
 
             Func<Stream, CancellationToken, Task> action = async (outputStream, token) =>
             {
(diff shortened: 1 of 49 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -360,7 +361,7 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
         /// <param name="level">The level of the logs to capture.</param>
         /// <param name="egressProvider">The egress provider to which the logs are saved.</param>
         [HttpGet("logs/{processKey?}", Name = nameof(CaptureLogs))]
-        [ProducesWithProblemDetails(ContentTypes.ApplicationNdJson, ContentTypes.TextEventStream)]
+        [ProducesWithProblemDetails(ContentTypes.ApplicationNdJson, ContentTypes.ApplicationJson, ContentTypes.TextEventStream)]
```
jander-msft, line 364: The media type for JSON Sequence is "application/json-seq", not "application/json".

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -483,7 +484,20 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
...
+            
+            if (format == LogFormat.EventStream)
+            {
+                contentType = ContentTypes.TextEventStream;
+            }
+            else if (format == LogFormat.NDJson)
+            {
+                contentType = ContentTypes.ApplicationNdJson;
+            }
+            else if (format == LogFormat.Json)
```
jander-msft, line 497: Let's call this LogFormat.JsonSequence

src/Microsoft.Diagnostics.Monitoring.WebApi/StreamingLogger.cs
```diff
@@ -132,7 +143,17 @@ private void LogJson<TState>(LogLevel logLevel, EventId eventId, TState state, E
...
 
-            outputStream.WriteByte((byte)'\n');
+            if (jsonFormat == LogFormat.NDJson)
+            {
+                outputStream.WriteByte((byte)'\n');
+            }
+            else if (jsonFormat == LogFormat.Json)
+            {
+                const byte lineFeed = 0x0A;
+                outputStream.WriteByte(lineFeed);
```
jander-msft, line 153: Isn't this the same as '\n'?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LogsTests.cs
```diff
@@ -376,13 +400,14 @@ public Task LogsWildcardTest(DiagnosticPortConnectionMode mode)
                 appValidate: (runner, client) => ValidateResponseStream(
                     runner,
                     client.CaptureLogsAsync(runner.ProcessId, TestTimeouts.LogsDuration, logLevel),
-                    callback));
+                    callback, format));
```
jander-msft, line 403: Please place on next line.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LogsTests.cs
```diff
@@ -392,10 +417,10 @@ public Task LogsWildcardTest(DiagnosticPortConnectionMode mode)
                 appValidate: (runner, client) => ValidateResponseStream(
                     runner,
                     client.CaptureLogsAsync(runner.ProcessId, TestTimeouts.LogsDuration, configuration),
-                    callback));
+                    callback, format));
```
jander-msft, line 420: Please place on next line.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LogsTests.cs
```diff
@@ -437,8 +462,15 @@ private async Task ValidateResponseStream(AppRunner runner, Task<ResponseStreamH
 
             _outputHelper.WriteLine("Begin reading log entries.");
             string line;
+
             while (null != (line = await reader.ReadLineAsync()))
             {
+                if (format == LogFormat.Json)
+                {
+                    const char recordSeparator = '\u001E';
+                    line.Trim(recordSeparator);
```
jander-msft, line 471: Let's be a bit more specific about this: validate that the first character is the record separator and the second character is not the record separator; then you can TrimStart. This will ensure that the beginning is exactly as expected and preserve the end.

jander-msft (changes requested):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LogsTests.cs
kkeirstead: Would it also be useful to confirm that there is a second character? If not, I believe the line would be null once we TrimStart, in which case the while loop should break. Also, what behavior should we have if the validation fails (either the first character isn't the record separator or the second character is the record separator)? Should this be silently skipped over, should this be logged, or should we exit from the while loop due to the unexpected formatting?

jander-msft: Let's be diligent by not assuming anything is valid and verify it. So something like:

```
Assert.True(line.Length > 1);
Assert.Equal(recordSeperator, line[0]);
Assert.NotEqual(recordSeperator, line[1]);
line = line.TrimStart(recordSeperator);
```

The Assert.* methods will throw an exception and end the test if the assertion fails.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HttpApi/ApiClient.cs
```diff
@@ -199,9 +199,9 @@ private async Task<ResponseStreamHolder> CaptureDumpAsync(string processKey, Dum
         /// <summary>
         /// GET /logs/{pid}?level={logLevel}&durationSeconds={duration}
         /// </summary>
-        public Task<ResponseStreamHolder> CaptureLogsAsync(int pid, TimeSpan duration, LogLevel? logLevel, CancellationToken token)
+        public Task<ResponseStreamHolder> CaptureLogsAsync(int pid, TimeSpan duration, LogLevel? logLevel, CancellationToken token, string contentType = ContentTypes.ApplicationNDJson)
```
jander-msft, line 202: Couple of things here:
- Async methods typically place the `CancellationToken` parameter as the last parameter of the method.
- Don't give the `contentType` parameter a default value. Have the tests explicitly pass in the argument.
- Let's change the `contentType` parameter to `LogFormat logFormat` so that the conversion from the log format to the content type is kept internal to the `ApiClient`, thus it's not a concern of each test.

src/Microsoft.Diagnostics.Monitoring.WebApi/StreamingLogger.cs
```diff
@@ -67,27 +67,38 @@ public IDisposable BeginScope<TState>(TState state)
...
 
-        private void LogJson<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+        private void LogJson<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter, LogFormat jsonFormat = LogFormat.NDJson)
         {
             Stream outputStream = _outputStream;
 
+            
+            if (jsonFormat == LogFormat.JsonSequence)
+            {
+                const byte recordSeparator = 0x1E;
```
jander-msft, line 91: nit: Let's make this a constant on the class and rename to something like `JsonSequenceRecordSeparator`.

jander-msft (approved):

PR #496, opened 2021-06-24
Use queries instead of path segments

### LINKED ISSUE #66: Use queries instead of path segments
e.g. /logs?pid=1
/dump?uid={guid}

kkeirstead:
The table below provides examples of how the new format translates to the existing format. Note that this table does not show every possible configuration of these commands and is meant to demonstrate the general use of the updated queries.

| New Format | Existing Format | Notes |
| --------------| ----------------- | ------------- |
| /process       | /processes/process | Info about Default Process
| /processes   | /processes | Now includes is_default as true/false
| /env             | /processes/process/env | -
| /process?pid=12345       | /processes/12345 | Explicitly specify PID
| /process?uid=00000000-0000-0000-0000-000000000000 | /processes/00000000-0000-0000-0000-000000000000 | Explicitly specify UID
| /process?name=MyWebApp | /processes/MyWebApp | Explicitly specify name
| /logs?pid=12345&level=Information&durationSeconds=5 | /logs/12345?level=Information&durationSeconds=5 | -

If a user specifies multiple process identifiers (e.g. a PID and a UID), we check to ensure they are consistent (all pertaining to the same process). If so, this is a legal request; otherwise, the default process (if configured) will be used.

I've also updated the documentation, although it's possible I missed some lingering spots with the existing format, so please feel free to let me know if you notice any inconsistencies.

Closes #66


src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -528,6 +612,96 @@ private static TimeSpan ConvertSecondsToTimeSpan(int durationSeconds)
                 TimeSpan.FromSeconds(durationSeconds);
         }
 
+        private ProcessKey? GetProcessKeyFromIdentifier(
```
jander-msft, line 615: I'm not sure I understand what these methods are doing. Couldn't we just add a new `ProcessKey(int? pid, Guid? uid, string name)` ctor and just pass it into the relevant operation?

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/ProcessIdentifier.cs
```diff
@@ -21,5 +21,8 @@ public class ProcessIdentifier
 
         [JsonPropertyName("name")]
         public string Name { get; set; }
+
+        [JsonPropertyName("is_default")]
```
jander-msft, line 25: I think the rest of the HTTP surface does not use underscores, so let's call this `isDefault`.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HttpApi/ApiClient.cs
```diff
@@ -408,5 +427,25 @@ private static void AppendDuration(StringBuilder builder, TimeSpan duration)
                 builder.Append(Convert.ToInt32(duration.TotalSeconds).ToString(CultureInfo.InvariantCulture));
             }
         }
+
+        private string GetRequestPrefix(ProcessIdentifierFormat processIdentifierFormat)
```
jander-msft, line 431: I like this idea, but let's take it a few steps further. Instead of having each method pass in the identifier (which they were already doing) and this extra ProcessIdentifierFormat parameter, how about we do something like:

`private static string GetProcessQuery(int? pid = null, Guid? uid = null, string name = null)`

and this will return "pid={pid}", "uid={uid}", "name={name}", or throw an exception if none are specified.

I'll use GetProcessAsync as an example of how to use this. Change GetProcessAsync to be:

```diff
public Task<Models.ProcessInfo> GetProcessAsync(int pid, CancellationToken token)
{
-   return GetProcessAsync(pid.ToString(CultureInfo.InvariantCulture), token);
+   return GetProcessAsync(GetProcessQuery(pid: pid), token);
}

public Task<Models.ProcessInfo> GetProcessAsync(Guid uid, CancellationToken token)
{
-   return GetProcessAsync(uid.ToString("D"), token);
+   return GetProcessAsync(GetProcessQuery(uid: uid), token);
}

-private async Task<Models.ProcessInfo> GetProcessAsync(string processKey, CancellationToken token)
+private async Task<Models.ProcessInfo> GetProcessAsync(string processQuery, CancellationToken token)
```

This gets each of the public methods out of the business of knowing how to format each of the process identifier parameters and centralizes the query parameter construction (even if it is fairly simple).

jander-msft (changes requested):
Good start! Have some feedback for you.

Also, because ProcessKey is no longer part of the public surface area, we can make that internal and also remove the mapping in the OpenAPI generator for that type: https://github.com/dotnet/dotnet-monitor/blob/a4e462d6730b6af2538596f2d4436a9180b0e658/src/Tests/Microsoft.Diagnostics.Monitoring.OpenApiGen/Program.cs#L58

documentation/README.md
```diff
@@ -29,4 +29,4 @@ When running a dotnet application differences in diverse local and production en
 - [Troubleshooting](./troubleshooting.md)
 - [Clone, build, and test the repo](./building.md)
 - [Official Build Instructions](./official-build-instructions.md)
-- [Release Process](./release-process.md)
\ No newline at end of file
+- [Release Process](./release-process.md)
```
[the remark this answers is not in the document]
kkeirstead: Not off the top of my head; the only thing I can think of is that when I made some commits I remember seeing a note about some line endings being converted. I think that was in reference to my updates to openapi.json though.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
kkeirstead: The logic for these three methods is mostly just for comprehensively handling situations where: 1) users provide no identifiers, 2) users provide a single identifier, 3) users provide multiple connected identifiers, or 4) users provide multiple identifiers that are associated with different processes. `GetProcessKeyFromIdentifier` is mostly just a staging area for determining which fields were provided in the request, `GetProcessKey` determines which Process Key to use based on what was provided in the request, and `ProcessInfoIsEquivalent` is used to validate whether the PIDs/UIDs/Names provided by the user belong to the same process. I haven't been sure how much documentation to add in the code to avoid unnecessary/excessive comments, but I can put more documentation in if you think that would help.

jander-msft: I think it makes more sense to add the new ctor and then update https://github.com/dotnet/dotnet-monitor/blob/d4999f15c5cf80bfb98872599b0587a6e641938a/src/Microsoft.Diagnostics.Monitoring.WebApi/DiagProcessFilter.cs#L24 to add a filter for each non-null field on ProcessKey (instead of assuming there is only one valid field). This avoids repeated calls to the GetProcessAsync methods, which can be costly (time-wise) in the connect mode for dotnet-monitor.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -82,13 +95,22 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
...
         public Task<ActionResult<Models.ProcessInfo>> GetProcessInfo(
-            ProcessKey processKey)
+            [FromQuery]
+            int? pid = null,
+            [FromQuery]
+            Guid? uid = null,
+            [FromQuery]
+            string name = null)
         {
+            ProcessKey? processKey = (pid == null && uid == null && name == null) ? null : new ProcessKey(pid, uid, name);
```
jander-msft, line 112: While this is somewhat simple, I would suggest refactoring this into a method because (1) it is used multiple times, and (2) it'll be much easier to update when we add other filter criteria (e.g. entrypoint name).

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HttpApi/ApiClient.cs
```diff
@@ -408,5 +392,45 @@ private static void AppendDuration(StringBuilder builder, TimeSpan duration)
...
+
+            if (pid != null)
+            {
+                assembledQuery += "pid=" + pid.GetValueOrDefault().ToString(CultureInfo.InvariantCulture);
+                initialQuery = false;
+            }
+
+            if (uid != null)
+            {
+                if (initialQuery == false)
```
jander-msft, line 409: Remove this flag and just use `assembledQuery.Length > 0`

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HttpApi/ApiClient.cs
```diff
@@ -408,5 +392,45 @@ private static void AppendDuration(StringBuilder builder, TimeSpan duration)
...
         }
+
+        private static string GetProcessQuery(int? pid = null, Guid? uid = null, string name = null)
+        {
+            string assembledQuery = "";
+            bool initialQuery = true;
+
+            if (pid != null)
+            {
+                assembledQuery += "pid=" + pid.GetValueOrDefault().ToString(CultureInfo.InvariantCulture);
```
jander-msft, line 403: `pid.Value.ToString(...)`

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HttpApi/ApiClient.cs
```diff
@@ -408,5 +392,45 @@ private static void AppendDuration(StringBuilder builder, TimeSpan duration)
...
+            }
+
+            if (uid != null)
+            {
+                if (initialQuery == false)
+                {
+                    assembledQuery += "&";
+                }
+
+                assembledQuery += "uid=" + uid.GetValueOrDefault().ToString("D");
```
jander-msft, line 414: `uid.Value.Tostring(...)`

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HttpApi/ApiClient.cs
```diff
@@ -408,5 +392,45 @@ private static void AppendDuration(StringBuilder builder, TimeSpan duration)
...
+
+                assembledQuery += "name=" + name;
+            }
+
+            if (!assembledQuery.Equals(""))
+            {
+                return assembledQuery;
+            }
+
+            throw new Exception("One of PID, UID, or Name must be specified.");
```
jander-msft, line 433: Try to use more exceptions rather than Exception itself e.g. ArgumentException would be fitting here.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ProcessTests.cs
```diff
@@ -145,7 +145,137 @@ public async Task MultiProcessIdentificationTest(DiagnosticPortConnectionMode mo
...
+                }
+
+#if NET5_0_OR_GREATER
+                // CHECK 1: Get response for processes using PID, UID, and Name and check for consistency
+
+                List<ProcessInfo> processInfoQueriesCheck1 = new List<ProcessInfo>();
+
+                processInfoQueriesCheck1.Add(await apiClient.GetProcessAsync(pid: pids[0], uid: null, name: null, System.Threading.CancellationToken.None));
+                processInfoQueriesCheck1.Add(await apiClient.GetProcessAsync(pid: null, uid: uids[0], name: null, System.Threading.CancellationToken.None));
+                //processInfoQueriesCheck1.Add(await apiClient.GetProcessAsync(pid: null, uid: null, name: names[0], System.Threading.CancellationToken.None));
```
jander-msft, line 211: I don't think we'll ever only check by name in the unit tests as they are because the process is always `dotnet` regardless of which unit test is being run.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ProcessTests.cs
```diff
@@ -145,7 +145,137 @@ public async Task MultiProcessIdentificationTest(DiagnosticPortConnectionMode mo
...
+                AppRunner runner = new(_outputHelper, Assembly.GetExecutingAssembly(), appId: i + 1);
+                runner.ConnectionMode = appConnectionMode;
+                runner.DiagnosticPortPath = diagnosticPortPath;
+                runner.ScenarioName = TestAppScenarios.AsyncWait.Name;
+                runner.Environment[ExpectedEnvVarName] = Guid.NewGuid().ToString("D");
+
+                appRunners[i] = runner;
+            }
+
+            IEnumerable<ProcessIdentifier> identifiers;
```
jander-msft, line 185: This can go inside the ExceuteAsync annonymous function since it is not used outside.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ProcessTests.cs
```diff
@@ -145,7 +145,137 @@ public async Task MultiProcessIdentificationTest(DiagnosticPortConnectionMode mo
...
+                    uids.Add(processIdentifier.Uid);
+                    names.Add(processIdentifier.Name);
+                }
+
+#if NET5_0_OR_GREATER
+                // CHECK 1: Get response for processes using PID, UID, and Name and check for consistency
+
+                List<ProcessInfo> processInfoQueriesCheck1 = new List<ProcessInfo>();
+
+                processInfoQueriesCheck1.Add(await apiClient.GetProcessAsync(pid: pids[0], uid: null, name: null, System.Threading.CancellationToken.None));
```
jander-msft, line 209: Please do not use CancellationToken.None here. If there is ever a problem in dotnet-monitor or in the tests that cause an infinite hangup, the tests could run for hours until the build machine gives up.

Instead, add overloads into https://github.com/dotnet/dotnet-monitor/blob/main/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HttpApi/ApiClientExtensions.cs that specify timeouts and CancellationToken on your behalf.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ProcessTests.cs
```diff
@@ -145,7 +145,137 @@ public async Task MultiProcessIdentificationTest(DiagnosticPortConnectionMode mo
...
+
+                processInfoQueriesCheck2.Add(await apiClient.GetProcessAsync(pid: pids[0], uid: null, name: null, System.Threading.CancellationToken.None));
+                processInfoQueriesCheck2.Add(await apiClient.GetProcessAsync(pid: pids[0], uid: uids[0], name: null, System.Threading.CancellationToken.None));
+                processInfoQueriesCheck2.Add(await apiClient.GetProcessAsync(pid: pids[0], uid: uids[0], name: names[0], System.Threading.CancellationToken.None));
+
+                VerifyProcessInfoEquality(processInfoQueriesCheck2);
+
+                // CHECK 3: Get response for processes using PID and an unassociated (randomly generated) UID and ensure the proper exception is thrown
+
+                await Assert.ThrowsAsync<ValidationProblemDetailsException>(async () => await apiClient.GetProcessAsync(pid: pids[0], uid: Guid.NewGuid(), name: null, System.Threading.CancellationToken.None));
```
jander-msft, line 227: For each of these that we expect to throw ValidationProblemDetailsException, please assert the status code (I believe it is 400). You'll find examples on the other tests.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ProcessTests.cs
```diff
@@ -145,7 +145,137 @@ public async Task MultiProcessIdentificationTest(DiagnosticPortConnectionMode mo
...
+
+                // CHECK 4: Get response for processes using invalid PID, UID, or Name and ensure the proper exception is thrown
+
+                await Assert.ThrowsAsync<ValidationProblemDetailsException>(async () => await apiClient.GetProcessAsync(pid: -1, uid: null, name: null, System.Threading.CancellationToken.None));
+                await Assert.ThrowsAsync<ValidationProblemDetailsException>(async () => await apiClient.GetProcessAsync(pid: null, uid: Guid.NewGuid(), name: null, System.Threading.CancellationToken.None));
+                await Assert.ThrowsAsync<ValidationProblemDetailsException>(async () => await apiClient.GetProcessAsync(pid: null, uid: null, name: "", System.Threading.CancellationToken.None)); // Are we safe to assume that processes will not have an empty name?
```
jander-msft, line 233: > // Are we safe to assume that processes will not have an empty name?

Yes, I think that's a reasonable assumption. We are also unlike to ever attempt to launch a process with an empty name, even if it happens to be supported.

However, you should check in dotnet-monitor if ASP.NET Core translates that to an empty string or null string as an argument on the action method.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ProcessTests.cs
```diff
@@ -145,7 +145,137 @@ public async Task MultiProcessIdentificationTest(DiagnosticPortConnectionMode mo
...
+
+                // Verify each app instance is reported and shut them down.
+                int loopIter = 0;
```
jander-msft, line 236: Seems to be unused.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ProcessTests.cs
```diff
@@ -145,7 +145,137 @@ public async Task MultiProcessIdentificationTest(DiagnosticPortConnectionMode mo
...
+
+                foreach (ProcessIdentifier processIdentifier in identifiers)
+                {
+                    pids.Add(processIdentifier.Pid);
+                    uids.Add(processIdentifier.Uid);
+                    names.Add(processIdentifier.Name);
+                }
+
+#if NET5_0_OR_GREATER
+                // CHECK 1: Get response for processes using PID, UID, and Name and check for consistency
```
jander-msft, line 205: I feel like you can add all of these checks to the existing MultiProcessIdentificationTest rather than adding a new test. All of the new stuff you added is about verifying the identity of multiple processes.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ProcessTests.cs
```diff
@@ -145,7 +145,137 @@ public async Task MultiProcessIdentificationTest(DiagnosticPortConnectionMode mo
...
+            foreach (ProcessInfo processInfo in processInfos)
+            {
+                processInfoPIDs.Add(processInfo.Pid);
+                processInfoUIDs.Add(processInfo.Uid);
+                processInfoNames.Add(processInfo.Name);
+            }
+
+            Assert.Single(processInfoPIDs.Distinct());
+            Assert.Single(processInfoUIDs.Distinct());
+            Assert.Single(processInfoNames.Distinct());
```
jander-msft, line 274: nit: May specify a comparer (StringComparer.Ordinal) for this one.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ProcessTests.cs
```diff
@@ -145,7 +145,137 @@ public async Task MultiProcessIdentificationTest(DiagnosticPortConnectionMode mo
...
+                List<string> names = new List<string>();
+
+                foreach (ProcessIdentifier processIdentifier in identifiers)
+                {
+                    pids.Add(processIdentifier.Pid);
+                    uids.Add(processIdentifier.Uid);
+                    names.Add(processIdentifier.Name);
+                }
+
+#if NET5_0_OR_GREATER
```
jander-msft, line 204: The only thing in here that really needs to be macro'd out is the one that checks only by runtime instance identifier.

jander-msft (changes requested):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ProcessTests.cs
kkeirstead: I ended up keeping the other couple lines in the macro because they belong together as part of the check. Without that line, the test only evaluates PID against itself, thus it will always pass. Keeping it as a block (I hope) makes it more obvious to a maintainer that the test is only useful for NET5_0_OR_GREATER

documentation/README.md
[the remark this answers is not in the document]
jander-msft: It looks like you added an extra line to these files (which is fine) compared to what is in main but GitHub isn't showing the empty new lines for some reason.

jander-msft (approved):

PR #516, opened 2021-06-30
Skip uid-only check for failed uid values.

jander-msft:
Also run identifier checks for each process.


kkeirstead (approved):
Looks good; in hindsight it makes much more sense to test each process instead of just the initial one (I believe that was an artifact of when it was its own test since I had been setting appCount = 1). I wasn't aware of the potential issue with the UID being empty in the mentioned scenario, so that's good to know :)

PR #521, opened 2021-07-01
Retry process routes when name is "unknown"

jander-msft:
I created #523 to so that we can better mitigate this type of problem. We can never guarantee that we get all of the process information on the first call, but once we getting it, we should be able to cache it.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HttpApi/ApiClientExtensions.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HttpApi/ApiClientExtensions.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Diagnostics.Monitoring.UnitTests.HttpApi
         /// <summary>
         /// GET /processes with retry attempts
         /// </summary>
-        public static async Task<IEnumerable<ProcessIdentifier>> GetProcessesWithRetryAsync(this ApiClient client, ITestOutputHelper outputHelper, Func<ProcessIdentifier, bool> filter = null, int maxAttempts = 5)
+        public static async Task<IEnumerable<ProcessIdentifier>> GetProcessesWithRetryAsync(this ApiClient client, ITestOutputHelper outputHelper, int[] pidFilters, int maxAttempts = 5)
         {
             IList<ProcessIdentifier> identifiers = null;
 
@@ -50,9 +50,9 @@ namespace Microsoft.Diagnostics.Monitoring.UnitTests.HttpApi
 
                 identifiers = (await client.GetProcessesAsync()).ToList();
 
-                if (null != filter)
+                if (pidFilters.Length > 0)
                 {
-                    identifiers = identifiers.Where(filter).ToList();
+                    identifiers = identifiers.Where(i => pidFilters.Contains(i.Pid)).ToList();
                 }
 
                 // In .NET 5+, the process name comes from the command line from the ProcessInfo command, which can fail
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HttpApi/ApiClientExtensions.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HttpApi/ApiClientExtensions.cs
@@ -7,8 +7,11 @@ using Microsoft.Diagnostics.Monitoring.WebApi;
 using Microsoft.Extensions.Logging;
 using System;
 using System.Collections.Generic;
+using System.Linq;
+using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
+using Xunit.Abstractions;
 
 namespace Microsoft.Diagnostics.Monitoring.UnitTests.HttpApi
 {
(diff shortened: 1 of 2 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ProcessTests.cs
```diff
@@ -245,7 +245,15 @@ private void VerifyProcessInfoEquality(List<ProcessInfo> processInfos)
...
-            Assert.Single(processInfoNames.Distinct(StringComparer.Ordinal));
+            // In .NET 5+, the process name comes from the command line from the ProcessInfo command, which can fail
+            // or be abandoned if it takes too long to response. In .NET Core 3.1, the process name comes from the
+            // command line from issuing a very brief event source trace, which can also fail or be abandoned if it
+            // takes too long to finish. Additionally, for 'connect' mode, these values are recomputed for every http
+            // invocation, so a prior invocation could succeed whereas a subsequent one could fail. Much of this could
+            // be mitigated if process information could be cached between calls. In any of these scenarios, if the
+            // process name is failed to be determined, the http api will return "unknown". For testing purposes,
+            // filter out the "unknown" value before comparing the real process names. 
+            Assert.Single(processInfoNames.Except(new[] { "unknown" }, StringComparer.Ordinal).Distinct(StringComparer.Ordinal));
```
kkeirstead, line 256: I actually pulled and played around with the tests for a little while because they've been improved significantly compared to what I initially wrote, but I think I found an edge case where this _might_ be an issue. Basically, if every member of processInfoNames in VerifyProcessInfoEquality happens to be unknown due to this failure to respond quickly, the Assert.Single will fail since it'll be empty (all the unknown names would have been removed). Since I can't reliably recreate this situation and force processes to have the name unknown, I just switched "unknown" to "dotnet" and the test failed. I realize this is probably a pretty limited edge case, but I figured I should mention it.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ProcessTests.cs
```diff
@@ -245,7 +245,15 @@ private void VerifyProcessInfoEquality(List<ProcessInfo> processInfos)
 
             Assert.Single(processInfoPIDs.Distinct());
             Assert.Single(processInfoUIDs.Distinct());
-            Assert.Single(processInfoNames.Distinct(StringComparer.Ordinal));
+            // In .NET 5+, the process name comes from the command line from the ProcessInfo command, which can fail
+            // or be abandoned if it takes too long to response. In .NET Core 3.1, the process name comes from the
```
kkeirstead, line 249: Just a typo: response -> respond

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ProcessTests.cs
jander-msft: Absolutely correct. I thought about this case too, but I figure its okay to ignore for now. It would require retrying until we got non-"unknown" values for each of the calls. However, it's unlikely that all three will have "unknown" for the name. When we fix #523, then we only technically have to loop for the first call and then all subsequent calls should have the same values because they'll all be cached.

Because of how infrequent this should be an issue in the test with this workaround and we should fix the non-caching issue, I figure it's okay to ignore for now. We'll reassess if the test fails frequently. I've only been making these targeted workarounds because it's had a high frequency of failure.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ProcessTests.cs
jander-msft: Thanks! I've been doing too much HTTP and IPC work recently (lots of "response" everywhere).

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ProcessTests.cs
jander-msft: I added more to this comment explaining the condition under which it can fail still but that it is unlikely to fail.

jander-msft: This ended up "kicking the can down the road" such that when we go to do the `/process` invocation with the name parameter set to "unknown" it would basically fail each time. I ended up adding try logic to all of the positive lookups in order to get process information that have non-`unknown` process names.

kkeirstead (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HttpApi/ApiClientExtensions.cs
```diff
@@ -31,6 +34,45 @@ public static async Task<IEnumerable<ProcessIdentifier>> GetProcessesAsync(this
...
+            int attempt = 0;
+            while (attempt < maxAttempts)
+            {
+                attempt++;
+
+                outputHelper.WriteLine($"Attempt #{attempt} of {maxAttempts}: GET /processes");
+
+                identifiers = (await client.GetProcessesAsync()).ToList();
+
+                if (null != filter)
```
wiktork, line 53: Don't you want to put the filtering after the retries, since filtering by process name will be broken? (Although I didn't see any clients that do this)

wiktork (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HttpApi/ApiClientExtensions.cs
jander-msft: I wanted to filter to processes that we cared about rather than verifying that all of the processes we found have valid names. I can change the filter to just be on pid (or provide a list of pods instead of a filter) to force the intention rather that leaving it up to the tests.

Issue #529, opened 2021-07-02
Unable to read Kestrel output in Prometheus due to dash character in event name


webbsk:
### Description

Prometheus is unable to read the Kestrel output because of the dash being an illegal character.

This is what I am seeing when Prometheus attempts to read the Dotnet-Monitor metrics endpoint:
"invalid metric type "aspnetcore-server-kestrel_connections_per_second gauge"

This only started happening once I was able to successfully scrape the endpoint with Kestrel source enabled.

### Configuration

Only relevant configuration is probably that the Kestrel event provider is enabled.

### Regression?

Unknown.

### Other information

I'm going to see if I can use relabelling in the servicemonitor to workaround but the correct format should probably be outputted from Dotnet-Monitor (remove illegal characters from event names for Prometheus scraping).

webbsk:
The formatting for the DotNet providers must snap to the Prometheus format in order to be scraped without errors: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

webbsk:
You may want to do the full checking and formatting (removing all illegal characters from the name like in the link in the above comment) but that may be beyond the scope of this bug.

shirhatti:
The issue with our naive logic to drop `.` (and with your PR now `-`) from providers names is that they are no longer guaranteed to be unique. We may to escape the provider/counter names (I'm not a fan either🤮) before converting to snake case.

jander-msft:
> The issue with our naive logic to drop `.` (and with your PR now `-`) from providers names is that they are no longer guaranteed to be unique.

I have the same concern as well.

> We may to escape the provider/counter names (I'm not a fan either🤮) before converting to snake case.

We might need to convert every non-alphanumeric character to an underscore according to what is allows by the Prometheus spec.

webbsk:
Yes, I understand your concern about uniqueness but I think what I am focused on is this part from the Prometheus metrics names and label requirements: ` It may contain ASCII letters and digits, as well as underscores and colons. It must match the regex [a-zA-Z_:][a-zA-Z0-9_:]*.` How you decide to implement that on your side is up to you, my PR is just a quick way (and following your existing convention) to deal with the problem that you can utilize if you choose.

jander-msft:
Fixed by #531

PR #530, opened 2021-07-02
Replace the dash in the kestrel provider name to for Prometheus

webbsk:
Bug found [here](https://github.com/dotnet/dotnet-monitor/issues/529). Looking to solve this error when enabling Kestrel metrics:
`invalid metric type "aspnetcore-server-kestrel_connections_per_second gauge"`

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/MetricsStore.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/MetricsStore.cs
@@ -147,7 +147,7 @@ namespace Microsoft.Diagnostics.Monitoring
             }
 
             metricValue = value.ToString(CultureInfo.InvariantCulture);
-            return FormattableString.Invariant($"{metric.Provider.Replace(".", string.Empty).ToLowerInvariant()}_{metric.Name.Replace('-', '_')}{unitSuffix}");
+            return FormattableString.Invariant($"{metric.Provider.Replace(".", string.Empty).Replace('-', '_').ToLowerInvariant()}_{metric.Name.Replace('-', '_')}{unitSuffix}");
         }
 
         private static bool CompareMetrics(ICounterPayload first, ICounterPayload second)


src/Microsoft.Diagnostics.Monitoring.WebApi/MetricsStore.cs
```diff
@@ -147,7 +147,7 @@ private static string GetPrometheusMetric(ICounterPayload metric, out string met
             }
 
             metricValue = value.ToString(CultureInfo.InvariantCulture);
-            return FormattableString.Invariant($"{metric.Provider.Replace(".", string.Empty).ToLowerInvariant()}_{metric.Name.Replace('-', '_')}{unitSuffix}");
+            return FormattableString.Invariant($"{metric.Provider.Replace(".", string.Empty).Replace('-', '_').ToLowerInvariant()}_{metric.Name.Replace('-', '_')}{unitSuffix}");
```
shirhatti, line 150: nit: For the sake consistency.
`Microsoft.AspNetCore.Hosting`/`connections-duration` ➡️  `microsoftaspnetcorehosting_connections_duration`

`Microsoft-AspNetCore-Server-Kestrel`/`connection-queue-length` ➡️ `microsoftaspnetcoreserverkestrel_connection_queue_length`

```suggestion
            return FormattableString.Invariant($"{metric.Provider.Replace(".", string.Empty).Replace('-', string.Empty).ToLowerInvariant()}_{metric.Name.Replace('-', '_')}{unitSuffix}");
```

jander-msft:
@dotnet/dotnet-diag, what are the rules for what is allowed to be in an event source name? I've been trying to find that in documentation, but haven't come across it yet.

@dotnet/dotnet-monitor, we might need to convert more characters, especially  for custom event sources, depending on what the Prometheus spec allows.

src/Microsoft.Diagnostics.Monitoring.WebApi/MetricsStore.cs
wiktork: I'm not against this change, but it does alter the [namespace ](https://prometheus.io/docs/practices/naming/) of all the existing metrics. Anyone who has been consuming metrics before will see brand new ones.

webbsk: The namespace of existing namespaces will not be able to be parsed by Prometheus without making this change. As I mentioned in the bug, my understanding is that it must follow the format:  `It may contain ASCII letters and digits, as well as underscores and colons. It must match the regex [a-zA-Z_:][a-zA-Z0-9_:]*.` from https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

wiktork:
> depending on what the Prometheus spec allows.

Prom spec: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

wiktork:
> @dotnet/dotnet-diag, what are the rules for what is allowed to be in an event source name? I've been trying to find that in documentation, but haven't come across it yet.
> 
> @dotnet/dotnet-monitor, we might need to convert more characters, especially for custom event sources, depending on what the Prometheus spec allows.

There doesn't seem to be any restrictions on the EventSource name, outside of the fact it's case insensitive. This makes sense, given the name is converted to a Guid. I was able to emit the following custom counter:

```c#
EventSource source = new EventSource("!@#$%^&*()");

            EventCounter c1 = new EventCounter("TestCounter1", source);
            c1.DisplayName = "TestCounter1";
            c1.DisplayUnits = "b";

```

```
!@#$%^&*()_TestCounter2_bytes 4.6000000000000005 1625259959428
!@#$%^&*()_TestCounter2_bytes 4.6000000000000005 1625259969389
!@#$%^&*()_TestCounter2_bytes 4.6000000000000005 1625259979379
```

noahfalk:
> @dotnet/dotnet-diag, what are the rules for what is allowed to be in an event source name? I've been trying to find that in documentation, but haven't come across it yet.

The only restriction EventSource has is that the string can't be a null reference. By convention most people create names that are english letters + the '-' and '.' punctuation marks but there is nothing that enforces this.

src/Microsoft.Diagnostics.Monitoring.WebApi/MetricsStore.cs
webbsk: May I please ask what the resolution time frame may look like for this? Is it just that I should make the change Sourabh suggested? I see that the tests failed on my change but I'm unable to run them successfully locally without my change as well. I end up seeing errors like the following: ﻿
```
Microsoft.Diagnostics.Monitoring.UnitTests.EndpointInfoSourceTests.ServerSourceAddRemoveSingleConnectionTest
   Source: EndpointInfoSourceTests.cs line 108
   Duration: 5.1 sec

  Message: 
    System.InvalidOperationException : No process is associated with this object.

  Stack Trace: 
    Process.dll:token 0x6000130+0x18
    Process.dll:token 0x60000f9+0x8
    DotNetRunner.ForceClose() line 181
    LoggingRunnerAdapter.DisposeAsync() line 62
    AppRunner.DisposeAsync() line 102
    AppRunnerExtensions.ExecuteAsync(AppRunner runner, Func`1 func) line 44
    EndpointInfoSourceTests.ServerSourceAddRemoveSingleConnectionTest() line 120
    EndpointInfoSourceTests.ServerSourceAddRemoveSingleConnectionTest() line 139
    --- End of stack trace from previous location ---
```

wiktork: @webbsk I am currently working on a change to address #531 which will normalize metric names and providers. I expect to get it out for review in the next day or two.

webbsk: @wiktork That's great news! Thanks for the update! 
May I please ask - will this be available as another preview release, such as tag 6? Or should I switch back to the nightly images based off image 5 for now in order to pick up your changes?

jander-msft: About 24 hours after the change is merged (and if all of the builds succeed, but I try to stay on top of that), then it will be available in the nightly image.

However, this upcoming preview will have breaking changes compared to previous previews regarding the HTTP API surface. But if you are only using the `/metrics` route on the 52325 port, then I don't think any of those changes will affect you.

wiktork:
#543 should address this issue.

webbsk:
Closing this PR since wiktork has resolved this with #543

PR #540, opened 2021-07-06
Add /info Endpoint

### LINKED ISSUE #76: Add /info endpoint
This should provide dotnet-monitor version, and possibly configuration (w/o PII)
- Listening mode (+ port)
- other configuration knobs

kkeirstead:
Adding an info endpoint that shows the Dotnet Monitor version for Preview 6, with the potential for adding in more information in a later release. 

Closes #76


documentation/openapi.json
```diff
@@ -754,6 +754,36 @@
...
+          "400": {
+            "$ref": "#/components/responses/BadRequestResponse"
+          },
+          "401": {
+            "$ref": "#/components/responses/UnauthorizedResponse"
+          },
+          "200": {
+            "description": "Success",
+            "content": {
+              "application/json": {
```
kelltrick, line 774: This seems incorrect. I would expect this to be a new Item like `"#/components/schemas/InformationObject` and it should be a singleton not an array.

documentation/openapi.json
```diff
@@ -754,6 +754,36 @@
...
+          "Diag"
+        ],
+        "summary": "Gets the Dotnet Monitor Version (other configuration information may be added in the future).",
+        "operationId": "GetInfo",
+        "responses": {
+          "400": {
+            "$ref": "#/components/responses/BadRequestResponse"
+          },
+          "401": {
+            "$ref": "#/components/responses/UnauthorizedResponse"
```
kelltrick, line 769: I don't know where we landed, but this doesn't match with the declaration which seems to make this non-authorized. My assumption would be that this endpoint is not authenticated or authorized and would never return 401.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -520,6 +521,44 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
...
+                };
+                
+                _logger.WrittenToHttpStream();
+                return new ActionResult<Models.DotnetMonitorInfo>(dotnetMonitorInfo);
+            }, _logger);
+        }
+
+        private string GetDotnetMonitorVersion()
+        {
+            var assembly = Assembly.GetEntryAssembly() ?? Assembly.GetExecutingAssembly();
```
kelltrick, line 548: Hmm. I think we want this to always be one or the other. Probably return either `Assembly.GetExecutingAssembly()` or return null.

kelltrick (changes requested):
We should tweak a few things in this PR but the basics all look right to me.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -520,6 +521,44 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
             }, processKey, ArtifactType_Logs);
         }
 
+        /// <summary>
+        /// Gets the Dotnet Monitor Version (other configuration information may be added in the future).
+        /// </summary>
+        [HttpGet("info", Name = nameof(GetInfo))]
+        [ProducesWithProblemDetails(ContentTypes.ApplicationJson)]
+        [ProducesResponseType(typeof(IEnumerable<Models.ProcessIdentifier>), StatusCodes.Status200OK)]
```
jander-msft, line 529:
```suggestion
        [ProducesResponseType(typeof(Models.DotnetMonitorInfo), StatusCodes.Status200OK)]
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
jander-msft: The way Kyle has it is correct. This is exactly what System.CommandLine does, at least the version we are using for dotnet-monitor.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -520,6 +521,44 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
...
+                {
+                    Version = version
+                };
+                
+                _logger.WrittenToHttpStream();
+                return new ActionResult<Models.DotnetMonitorInfo>(dotnetMonitorInfo);
+            }, _logger);
+        }
+
+        private string GetDotnetMonitorVersion()
```
jander-msft, line 546:
```suggestion
        private static string GetDotnetMonitorVersion()
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -520,6 +521,44 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
...
 
+        /// <summary>
+        /// Gets the Dotnet Monitor Version (other configuration information may be added in the future).
+        /// </summary>
+        [HttpGet("info", Name = nameof(GetInfo))]
+        [ProducesWithProblemDetails(ContentTypes.ApplicationJson)]
+        [ProducesResponseType(typeof(IEnumerable<Models.ProcessIdentifier>), StatusCodes.Status200OK)]
+        public ActionResult<Models.DotnetMonitorInfo> GetInfo()
+        {
+            string version = GetDotnetMonitorVersion();
```
jander-msft, line 532: Please move this inside the anonymous method.

documentation/openapi.json
jander-msft: As it is, this route is authenticated because it is on the DiagController, which all of its actions are authenticated.

I'm not convinced that this should be unauthenticated. There may be more sensitive information on this endpoint that we wouldn't want arbitrary users to get. I don't think we would necessary put the authentication information on this route either, so it wouldn't be used to determine how to authenticate (avoiding a chicken-and-egg problem). So I think this declaration is correct.

@dotnet/dotnet-monitor, what do others think?

jander-msft (changes requested):

documentation/openapi.json
wiktork: Maybe we should consider putting /version as a separate thing that's not authenticated. That way the version information can always be retrieved without having to worry about leaking other data.

kkeirstead: Just to make sure I'm understanding correctly: would we have a /version endpoint that's unauthenticated (like /metrics), and then we would additionally have a /info endpoint that is authenticated (where we might put additional sensitive information in the future)? I imagine this _shouldn't_ be difficult to do, but I guess I'm not sure if it's overkill to have an endpoint reserved solely for showing the version. I just imagine in the future if there's other generic info we want to share that doesn't need to be authenticated that we may end up creating a plethora of these one-off endpoints.

I'm not sure if this is possible (or considered a good idea), but would it be possible to overload the /info endpoint such that it has different behavior for authenticated/non-authenticated requests? E.g. calling /info when not authenticated would show basic generic information, such as the version, and then calling /info when you are authenticated would provide you with more detailed, potentially 'secure' info (in addition to the generic information). Not sure if there would be too many implementation/security headaches involved in something like this, but I just want a more futureproof solution than lots of one-off endpoints for users to remember.

wiktork: - I think having the same endpoint behave differently for auth and non-auth would be difficult (we'd probably have to write custom authentication which we are moving away from). 

- I think it would be useful to get the version info without auth (since we may want the information for troubleshooting purposes). If we want to be flexible, would it be reasonable to add unauthenticated /info now and then if we do need secure information we add a separate /secureinfo or something like that?

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -520,6 +521,23 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
             }, processKey, ArtifactType_Logs);
         }
 
+        /// <summary>
+        /// Gets information about Dotnet-Monitor that should be secured behind authentication (currently a stub).
+        /// </summary>
+        [HttpGet("secureinfo", Name = nameof(GetSecureInfo))]
```
jander-msft, line 527: This isn't providing any value at this time, so don't add this route.

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/DotnetMonitorSecureInfo.cs
```diff
@@ -0,0 +1,15 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if UNITTEST
+namespace Microsoft.Diagnostics.Monitoring.UnitTests.Models
+#else
+namespace Microsoft.Diagnostics.Monitoring.WebApi.Models
+#endif
+{
+    public class DotnetMonitorSecureInfo
```
jander-msft, line 11: Please remove

jander-msft (changes requested):

documentation/openapi.json
```diff
@@ -985,6 +1012,24 @@
           }
         },
         "additionalProperties": false
+      },
+      "DotnetMonitorInfo": {
```
jander-msft, line 1016: You'll need to add documentation of this object to documentation\api\definitions.md

jander-msft (changes requested):

documentation/api/info.md
```diff
@@ -0,0 +1,50 @@
...
+
+## HTTP Route
+
+```http
+GET /info HTTP/1.1
+```
+
+## Host Address
+
+This route is available on all configured addresses.
```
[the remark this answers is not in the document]
kkeirstead: Oops, I forgot that everyone would automatically be notified when I pushed - this was just meant to be an intermediary change to save progress since I hadn't started doing any documentation. Sorry about that!

documentation/api/info.md
```diff
@@ -0,0 +1,51 @@
...
+| 200 OK | | Information about Dotnet-Monitor formatted as JSON.  | `application/json` |
+| 400 Bad Request | [ValidationProblemDetails](definitions.md#ValidationProblemDetails) | An error occurred due to invalid input. The response body describes the specific problem(s). | `application/problem+json` |
+| 401 Unauthorized | | Authentication is required to complete the request. See [Authentication](./../authentication.md) for further information. | |
+
+## Examples
+
+### Sample Request
+
+```http
+GET info HTTP/1.1
```
jander-msft, line 32:
```suggestion
GET /info HTTP/1.1
```

documentation/api/definitions.md
```diff
@@ -1,5 +1,15 @@
...
 
+## DotnetMonitorInfo
+
+Object describing diagnostic/automation information about the executing instance of Dotnet-Monitor.
+
+| Name | Type | Description |
+|---|---|---|
+| Version | string | The current version of Dotnet-Monitor. |
+| RuntimeVersion | string | The version of the dotnet runtime. |
+| ListeningMode | string | Indicates whether Dotnet-Monitor is in `Client` mode or `Listen` mode. |
```
jander-msft, line 11:
```suggestion
| DiagnosticPortMode | string | Indicates whether `dotnet montor` is in `connect` mode or `listen` mode. |
```

documentation/api/definitions.md
```diff
@@ -1,5 +1,15 @@
 # Definitions
 
+## DotnetMonitorInfo
+
+Object describing diagnostic/automation information about the executing instance of Dotnet-Monitor.
```
jander-msft, line 5:
```suggestion
Object describing diagnostic/automation information about the executing instance of `dotnet monitor`.
```

We try to stylize it as `dotnet monitor` in the documentation in order to promote the usage of launching it through the `dotnet` executable.

documentation/api/info.md
```diff
@@ -0,0 +1,51 @@
...
+Host: localhost:52323
+```
+
+### Sample Response
+
+```http
+HTTP/1.1 200 OK
+Content-Type: application/json
+
+{"Version":"5.0.0-dev.21362.1+813e40f07bcf3de83b3bb374e2736ad256d0c067","RuntimeVersion":"3.1.16","ListeningMode":"Connect"}
```
jander-msft, line 42:
```suggestion
{
  "Version": "5.0.0-preview.5.21302.5+a99ec90eb220fa3b57f089677d8d1e85d48fb3a8",
  "RuntimeVersion": "3.1.16",
  "DiagnosticPortMode": "Connect"
}
```

we try to line-separate each value for readability. Also, use a shipped version number as the example.

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/DotnetMonitorInfo.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// <summary>
+        /// The dotnet runtime version 
+        /// </summary>
+        [JsonPropertyName("runtimeVersion")]
+        public string RuntimeVersion { get; set; }
+
+        /// <summary>
+        /// Indicates whether Dotnet-Monitor is in Client mode or Listen mode
+        /// </summary>
+        [JsonPropertyName("listeningMode")]
```
jander-msft, line 30:
```suggestion
        [JsonPropertyName("diagnosticPortMode")]
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/DotnetMonitorInfo.cs
```diff
@@ -0,0 +1,33 @@
...
+        public string ListeningMode { get; set; }
```
jander-msft, line 31:
```suggestion
        public string DiagnosticPortMode { get; set; }
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/DotnetMonitorInfo.cs
```diff
@@ -0,0 +1,33 @@
...
+        public string Version { get; set; }
+
+        /// <summary>
+        /// The dotnet runtime version 
+        /// </summary>
+        [JsonPropertyName("runtimeVersion")]
+        public string RuntimeVersion { get; set; }
+
+        /// <summary>
+        /// Indicates whether Dotnet-Monitor is in Client mode or Listen mode
```
jander-msft, line 28:
```suggestion
        /// Indicates whether dotnet monitor is in 'connect' mode or 'listen' mode.
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/DotnetMonitorInfo.cs
```diff
@@ -0,0 +1,33 @@
...
+#if UNITTEST
+namespace Microsoft.Diagnostics.Monitoring.UnitTests.Models
+#else
+namespace Microsoft.Diagnostics.Monitoring.WebApi.Models
+#endif
+{
+    public class DotnetMonitorInfo
+    {
+        /// <summary>
+        /// The Dotnet-Monitor version
```
jander-msft, line 16:
```suggestion
        /// The dotnet monitor version 
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LogsTests.cs
```diff
@@ -44,14 +44,14 @@ public LogsTests(ITestOutputHelper outputHelper, ServiceProviderFixture serviceP
         /// Tests that all log events are collected if log level set to Trace.
         /// </summary>
         [Theory]
-        [InlineData(DiagnosticPortConnectionMode.Connect, LogFormat.JsonSequence)]
-        [InlineData(DiagnosticPortConnectionMode.Connect, LogFormat.NDJson)]
+        [InlineData(Options.DiagnosticPortConnectionMode.Connect, LogFormat.JsonSequence)]
```
jander-msft, line 47: Could you use `using Microsoft.Diagnostics.Monitoring.UnitTests.Options;` instead of clarifying the namespace for each of these? Even if there is ambiguity, it's probably a smaller subset to change the other types to be clarified or to use a type using e.g. `using Baz = Foo.Bar.Baz;`

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTests/EndpointInfoSourceTests.cs
```diff
@@ -4,7 +4,6 @@
 
 using Microsoft.Diagnostics.Monitoring.TestCommon;
 using Microsoft.Diagnostics.Monitoring.UnitTests.Runners;
-using Microsoft.Diagnostics.Monitoring.WebApi;
```
jander-msft, line 7: This needs to be kept in order to refer to the types that are being tested.

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/DotnetMonitorInfo.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// </summary>
+        [JsonPropertyName("runtimeVersion")]
+        public string RuntimeVersion { get; set; }
+
+        /// <summary>
+        /// Indicates whether Dotnet-Monitor is in Client mode or Listen mode
+        /// </summary>
+        [JsonPropertyName("listeningMode")]
+        public string ListeningMode { get; set; }
+    }
```
jander-msft, line 32: Let's also add `public string DiagnosticPortName { get; set; }`, which is the value of DiagnosticPortOptions.EndpointName. Without this information, knowing the connection mode is not terribly helpful.

documentation/api/info.md
```diff
@@ -0,0 +1,51 @@
+# Info - Get
+
+Gets information about the Dotnet-Monitor version, the runtime version, and the listening mode.
```
jander-msft, line 3:
```suggestion
Gets information about the `dotnet monitor` version, the runtime version, and the diagnostic port settings.
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/DotnetMonitorInfo.cs
jander-msft, line 31: I'm thinking that the type of this property should be DiagnosticPortConnectionMode, such that the OpenAPI document will then list the enum values as the expected returned values.

jander-msft (changes requested):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LogsTests.cs
kkeirstead: I believe for these I did it since LogFormat was in WebApi (so I had to allow WebApi to be included); I can change it to whatever you'd prefer though

documentation/api/info.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+### Sample Response
+
+```http
+HTTP/1.1 200 OK
+Content-Type: application/json
+
+{
+	"version":"5.0.0",
```
jander-msft, line 43: nit: Please replace the tabs with four spaces and put a space after each colon for each property.

jander-msft (approved):

kelltrick (approved):

PR #569, opened 2021-07-14
Add option to disable HTTP Response egress

### LINKED ISSUE #202: Add option to disable HTTP Response egress
Users of dotnet-monitor may want to disable HTTP Response egress of diagnostic artifacts and only allow other egress providers to be used. Can be used to prevent accidental disclosure of sensitive information

kkeirstead:
By using a --no-http-egress flag, a user is able to prevent the default HTTP response for any endpoint that allows for an egress provider (logs, traces, dumps, gcdumps). 

Closes #202


src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -623,6 +637,14 @@ private static string GetFileNameTimeStampUtcNow()
             return DateTime.UtcNow.ToString("yyyyMMdd_HHmmss");
         }
 
+        private void CheckForDisabledHTTPEgress(string egressProvider)
```
wiktork, line 640: nit: Http

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
```diff
@@ -113,23 +113,26 @@ public async Task<int> Start(CancellationToken token, IConsole console, string[]
...
 
-        public static IHostBuilder CreateHostBuilder(IConsole console, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, bool configOnly)
+        public static IHostBuilder CreateHostBuilder(IConsole console, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, bool noHTTPEgress, bool configOnly)
         {
             IHostBuilder hostBuilder = Host.CreateDefaultBuilder();
 
             KeyAuthenticationMode authMode = noAuth ? KeyAuthenticationMode.NoAuth : tempApiKey ? KeyAuthenticationMode.TemporaryKey : KeyAuthenticationMode.StoredKey;
             AuthOptions authenticationOptions = new AuthOptions(authMode);
 
+            EgressMode egressMode = noHTTPEgress ? EgressMode.HTTPDisabled : EgressMode.HTTPEnabled;
```
wiktork, line 133: While I'm ok with this approach, you also have some other choices here:
- an IActionFilterAsync. I haven't actually tried these but these are executed earlier in the pipeline than actions, so you may be able to stop execution based on settings/lack of egress providers.
- Middleware (such as the Throttling class). The big downside for middleware is that it doesn't do any of the nice error/exception processing that the mvc layer does for you.

wiktork: See https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/filters?view=aspnetcore-5.0#implementation

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        public void OnActionExecuted(ActionExecutedContext context)
+        {
+            // Not sure we need to do anything here...
+        }
+
+        public void OnActionExecuting(ActionExecutingContext context)
+        {
+            StringValues value;
+            bool egressProvider = context.HttpContext.Request.Query.TryGetValue("egressProvider", out value);
```
wiktork, line 56: Can you make sure Throttling middleware and this code use the same test for ensuring an egress provider is used?

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            return actionFilter;
+        }
+    }
+
+    public class EgressValidationUnhandledExceptionFilter : ActionFilterAttribute, IExceptionFilter
+    {
+        public void OnException(ExceptionContext context)
+        {
+            BadRequestObjectResult badRequestResult = new BadRequestObjectResult(context.Exception.Message);
```
wiktork, line 32: Not necessarily for this review, but we currently have a helper function that performs a similar translation (ActionContextExtensions). How do these two interact? Perhaps we can remove that helper entirely with an exception filter.

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
```diff
@@ -0,0 +1,67 @@
...
+        public void OnActionExecuting(ActionExecutingContext context)
+        {
+            StringValues value;
+            bool egressProvider = context.HttpContext.Request.Query.TryGetValue("egressProvider", out value);
+
+            if (egressProvider == false || value.ToString() == null)
+            {
+                if (_egressOutputOptions.EgressMode == EgressMode.HttpDisabled)
+                {
+                    throw new ArgumentException("Http Egress is currently disabled.");
```
wiktork, line 62: nit: Put errors into resx files

wiktork, line 62: Will this automatically show up in our logs for dotnet-monitor? If not, we should add an ILogger here as well.

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
jander-msft: > Perhaps we can remove that helper entirely with an exception filter.

Possibly, but we might lose logger scopes if we do that since it currently logs the exceptions in the `when` clause of the `catch` block in order to preserve the logger scopes. Something to keep in mind when IExceptionFilter is investigated.

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
kkeirstead: If I'm looking at the right thing, I do this a bit differently; instead of using ContainsKey, using TryGetValue gives us the ability to check for an empty string (I do this in the updated code). This protects us against users who use the query, "egressProvider=" because this would still produce the Http response if we only use ContainsKey.

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            StringValues value;
+
+            bool egressProviderGiven = context.HttpContext.Request.Query.TryGetValue(EgressQuery, out value);
+
+            if (!egressProviderGiven || value.ToString().Equals(""))
+            {
+                if (_egressOutputOptions.EgressMode == EgressMode.HttpDisabled)
+                {
+                    ILogger<EgressValidationActionFilter> logger = context.HttpContext.RequestServices
```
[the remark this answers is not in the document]
jander-msft: Actually, I think all of the logging should move to `EgressValidationUnhandledExceptionFilter`. This will keep the same pattern we have elsewhere (logging is done at the site of converting an exception to problem details), and you should be able to use DI directly rather than getting the logger factory and creating the logger yourself (update the `EgressValidationUnhandledExceptionFilter` constructor to have the ILogger parameter).

src/Microsoft.Diagnostics.Monitoring.WebApi/Strings.resx
```diff
@@ -121,6 +121,10 @@
     <value>Unable to get process environment.</value>
     <comment>Gets a string similar to "Unable to get process environment.".</comment>
   </data>
+  <data name="ErrorMessage_HttpEgressDisabled" xml:space="preserve">
+    <value>Http Egress was not enabled.</value>
```
jander-msft, line 125:
```suggestion
    <value>HTTP egress is not enabled.</value>
```

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
```diff
@@ -63,12 +63,12 @@ internal sealed class DiagnosticsMonitorCommandHandler
 
         private static readonly string UserSettingsPath = Path.Combine(UserConfigDirectoryPath, SettingsFileName);
 
-        public async Task<int> Start(CancellationToken token, IConsole console, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey)
+        public async Task<int> Start(CancellationToken token, IConsole console, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, bool noHTTPEgress)
```
jander-msft, line 66:
```suggestion
        public async Task<int> Start(CancellationToken token, IConsole console, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, bool noHttpEgress)
```

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
```diff
@@ -113,23 +113,26 @@ public async Task<int> Start(CancellationToken token, IConsole console, string[]
             return 0;
         }
 
-        public Task<int> ShowConfig(CancellationToken token, IConsole console, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, ConfigDisplayLevel level)
+        public Task<int> ShowConfig(CancellationToken token, IConsole console, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, bool noHTTPEgress, ConfigDisplayLevel level)
```
jander-msft, line 116:
```suggestion
        public Task<int> ShowConfig(CancellationToken token, IConsole console, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, bool noHttpEgress, ConfigDisplayLevel level)
```

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
```diff
@@ -113,23 +113,26 @@ public async Task<int> Start(CancellationToken token, IConsole console, string[]
...
+            IHost host = CreateHostBuilder(console, urls, metricUrls, metrics, diagnosticPort, noAuth, tempApiKey, noHTTPEgress, configOnly: true).Build();
             IConfiguration configuration = host.Services.GetRequiredService<IConfiguration>();
             using ConfigurationJsonWriter writer = new ConfigurationJsonWriter(Console.OpenStandardOutput());
             writer.Write(configuration, full: level == ConfigDisplayLevel.Full);
             
             return Task.FromResult(0);
         }
 
-        public static IHostBuilder CreateHostBuilder(IConsole console, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, bool configOnly)
+        public static IHostBuilder CreateHostBuilder(IConsole console, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, bool noHTTPEgress, bool configOnly)
```
jander-msft, line 126:
```suggestion
        public static IHostBuilder CreateHostBuilder(IConsole console, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, bool noHttpEgress, bool configOnly)
```

src/Microsoft.Diagnostics.Monitoring.WebApi/IEgressOutputOptions.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public enum EgressMode
+    {
+        HttpDisabled,
+        HttpEnabled
+    }
+
+    public interface IEgressOutputOptions
+    {
+        EgressMode EgressMode { get; }
```
jander-msft, line 17: This could be simplied to `bool IsHttpEgressEnabled` and then the enum can be removed. I don't know if we'll really have any other values on the enum other than enabled and disabled, so at that point its just replicating the behavior of a Boolean.

src/Tools/dotnet-monitor/Program.cs
```diff
@@ -106,6 +106,15 @@ class Program
                 Argument = new Argument<bool>(name: "noAuth", getDefaultValue: () => false)
             };
 
+        private static Option NoHTTPEgress() =>
```
jander-msft, line 109:
```suggestion
        private static Option NoHttpEgress() =>
```

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -275,6 +275,10 @@
     <value>Turn off authentication.</value>
     <comment>Gets the string to display in help that explains what the '--no-auth' option does.</comment>
   </data>
+  <data name="HelpDescription_OptionNoHTTPEgress" xml:space="preserve">
```
jander-msft, line 278:
```suggestion
  <data name="HelpDescription_OptionNoHttpEgress" xml:space="preserve">
```

jander-msft (changes requested):

src/Microsoft.Diagnostics.Monitoring.WebApi/Strings.resx
kkeirstead: Just so I know the convention, do we use Http in the context of a variable and HTTP in the context of documentation? Just so you don't need to keep manually fixing them all for me :)

jander-msft: Yep, when written for consumption via reading, it is "HTTP". When use in code, it is "Http". By convention, (if I recall the exact number), if an acronym is more than 3 letters, then the first is uppercased while the rest are lowercased (unless its the first word or a field of local variable, which then it is all lowercase).

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
```diff
@@ -58,17 +72,22 @@ public void OnActionExecuting(ActionExecutingContext context)
...
+        }
 
-                    throw new ArgumentException("Http Egress is currently disabled.");
+        public class EgressDisabledException : Exception
+        {
+            public override string Message
+            {
+                get
+                {
+                    return "Http Egress is currently disabled.";
```
kkeirstead, line 90: Should this also be using Strings.ErrorMessage_HttpEgressDisabled? I just noticed the discrepancy between this and what gets logged.

src/Tools/dotnet-monitor/Program.cs
```diff
@@ -26,6 +26,7 @@ class Program
                 HashAlgorithm(), KeyLength()
             };
 
+
```
kkeirstead, line 29: nit: remove newline

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressOutputOptions.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    internal sealed class EgressOutputOptions : IEgressOutputOptions
+    {
+        public bool IsHttpEgressEnabled { get; set; }
```
jander-msft, line 9: This should not have a setter.

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
```diff
@@ -0,0 +1,95 @@
...
+            EgressValidationActionFilter actionFilter = new EgressValidationActionFilter(egressOutputOptions);
+
+            return actionFilter;
+        }
+    }
+
+    [AttributeUsage(AttributeTargets.Class)]
+    public class EgressValidationUnhandledExceptionFilter : ActionFilterAttribute, IExceptionFilter
+    {
+        private ILogger _logger;
```
jander-msft, line 33:
```suggestion
        private readonly ILogger _logger;
```

src/Microsoft.Diagnostics.Monitoring.WebApi/IEgressOutputOptions.cs
```diff
@@ -0,0 +1,13 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    public interface IEgressOutputOptions
+    {
+        bool IsHttpEgressEnabled { get; set; }
```
jander-msft, line 11: Remove setter.

src/Microsoft.Diagnostics.Monitoring.WebApi/Strings.resx
```diff
@@ -216,7 +220,7 @@
 2. requests: The number of outstanding requests</comment>
   </data>
   <data name="LogFormatString_WrittenToHttpStream" xml:space="preserve">
-    <value>Written to HTTP stream.</value>
+    <value>Written to Http stream.</value>
```
jander-msft, line 223: Please revert this.

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
```diff
@@ -176,6 +178,13 @@ public static IHostBuilder CreateHostBuilder(IConsole console, string[] urls, st
 
                     services.AddSingleton<IAuthOptions>(authenticationOptions);
 
+                    services.AddSingleton<IEgressOutputOptions>(egressOptions);
+
+                    services.AddMvc(options =>
```
jander-msft, line 183: Forgot to mention this before, but this should be moved to the `Startup.ConfigureServices` method. We try to keep as much of the HTTP server configuration there as much as possible.

There's an existing call to `AddControllers` there that you can use the overloaded version of it to configure the filters on the MvcOptions instance.

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
```diff
@@ -0,0 +1,95 @@
...
+                context.Result = badRequestResult;
+
+                _logger.LogError(Strings.ErrorMessage_HttpEgressDisabled);
+            }
+        }
+    }
+
+    public class EgressValidationActionFilter : IActionFilter
+    {
+        private IEgressOutputOptions _egressOutputOptions;
```
jander-msft, line 56:
```suggestion
        private readonly IEgressOutputOptions _egressOutputOptions;
```

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
```diff
@@ -0,0 +1,95 @@
...
+        private const string EgressQuery = "egressprovider";
+
+        public EgressValidationActionFilter(IEgressOutputOptions egressOutputOptions)
+        {
+            _egressOutputOptions = egressOutputOptions;
+        }
+
+        public void OnActionExecuted(ActionExecutedContext context)
+        {
+            // Not sure we need to do anything here...
```
jander-msft, line 66: Nothing to do here; you can remove the comment.

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
```diff
@@ -0,0 +1,95 @@
...
+        public void OnException(ExceptionContext context)
+        {
+            if (context.Exception is EgressValidationActionFilter.EgressDisabledException egressException)
+            {
+                BadRequestObjectResult badRequestResult = new BadRequestObjectResult(egressException.Message);
+                badRequestResult.Value = ExceptionExtensions.ToProblemDetails(egressException, StatusCodes.Status400BadRequest);
+
+                context.Result = badRequestResult;
+
+                _logger.LogError(Strings.ErrorMessage_HttpEgressDisabled);
```
jander-msft, line 49: nit: Just log the message from the exception so that way if we change it, it is automatically logged here without needing to change it here too.

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
```diff
@@ -0,0 +1,95 @@
...
+            if (!egressProviderGiven || StringValues.IsNullOrEmpty(value))
+            {
+                if (!_egressOutputOptions.IsHttpEgressEnabled)
+                {
+                    throw new EgressDisabledException();
+                }
+            }
+        }
+
+        public class EgressDisabledException : Exception
```
jander-msft, line 84: I think this is okay, given that this exception is not going to be serialized. @wiktork, do you have an opinion on how this exception is implemented?

jander-msft (changes requested):

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
```diff
@@ -0,0 +1,94 @@
...
+using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Extensions.Logging;
+using System;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class)]
+    public sealed class EgressValidationAttribute : ActionFilterAttribute, IFilterFactory
```
jander-msft, line 16: Does this need to be public? see if you can make it internal.

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
```diff
@@ -0,0 +1,94 @@
...
+            var egressOutputOptions = serviceProvider.GetRequiredService<IEgressOutputOptions>();
+
+            EgressValidationActionFilter actionFilter = new EgressValidationActionFilter(egressOutputOptions);
+
+            return actionFilter;
+        }
+    }
+
+    [AttributeUsage(AttributeTargets.Class)]
+    public class EgressValidationUnhandledExceptionFilter : ActionFilterAttribute, IExceptionFilter
```
jander-msft, line 31: Does this need to be public? see if you can make it internal.

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
```diff
@@ -0,0 +1,94 @@
...
+                badRequestResult.Value = ExceptionExtensions.ToProblemDetails(egressException, StatusCodes.Status400BadRequest);
+
+                context.Result = badRequestResult;
+
+                _logger.LogError(egressException.Message);
+            }
+        }
+    }
+
+    public class EgressValidationActionFilter : IActionFilter
```
jander-msft, line 54: Does this need to be public? see if you can make it internal.

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            _logger = logger;
+        }
+
+        public void OnException(ExceptionContext context)
+        {
+            if (context.Exception is EgressValidationActionFilter.EgressDisabledException egressException)
+            {
+                BadRequestObjectResult badRequestResult = new BadRequestObjectResult(egressException.Message);
+                badRequestResult.Value = ExceptionExtensions.ToProblemDetails(egressException, StatusCodes.Status400BadRequest);
```
jander-msft, line 45: nit: ToProblemDetails is an extension method, so you typically should use it like `egressException.ToProblemDetails(StatusCodes.Status400BadRequest)`

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
```diff
@@ -0,0 +1,94 @@
...
+        public EgressValidationUnhandledExceptionFilter(ILogger<EgressValidationUnhandledExceptionFilter> logger)
+        {
+            _logger = logger;
+        }
+
+        public void OnException(ExceptionContext context)
+        {
+            if (context.Exception is EgressValidationActionFilter.EgressDisabledException egressException)
+            {
+                BadRequestObjectResult badRequestResult = new BadRequestObjectResult(egressException.Message);
```
jander-msft, line 44: I'd write this as:
```
context.Result = new BadRequestObjectResult(
    egressException.ToProblemDetails(StatusCodes.Status400BadRequest));
```

There's no need to instantiate the BadRequestObjectResult using the exception message and then overwriting it on the next line by setting the Value property. If those are inlined together, then you don't need an explicit badRequestResult local variable either.

Alternatively, use [this](https://github.com/dotnet/dotnet-monitor/blob/bf6844e907c40b617eab6ef9c807065603678d51/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagControllerExtensions.cs#L93) method (get ride of the `controller` parameter, it's not used anyway) since it does exactly what you want.

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
```diff
@@ -0,0 +1,94 @@
...
+                badRequestResult.Value = ExceptionExtensions.ToProblemDetails(egressException, StatusCodes.Status400BadRequest);
+
+                context.Result = badRequestResult;
```
jander-msft, line 47: Can you check that the HTTP response (using PostMan, curl, etc) report the correct content-type (application/problem+json)? If it doesn't, you might have to add `result.ContentTypes.Add(ContentTypes.ApplicationProblemJson);` to the result before returning.

jander-msft (approved):
Some more feedback, but I think it's good after you verify a few more things. Also, I would like to see a follow up PR to add a test that verifies that using this option does return 400s for non-egress calls and reports the correct content type.

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressValidationAttribute.cs
kkeirstead: Can confirm; Postman says "application/problem+json; charset=utf-8"

src/Microsoft.Diagnostics.Monitoring.WebApi/EgressOutputOptions.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    internal sealed class EgressOutputOptions : IEgressOutputOptions
```
kelltrick, line 7: I'm trying to standardize the naming of our options classes. Anything that is a direct object model to a user settable option, should end in `*Options`. Anything that is a class we use internally to store options or configuration values should be named `*Configuration`.
If I understand this correctly this should be `EgressOutputConfiguration` and `IEgressOutputConfiguration`

src/Tests/Microsoft.Diagnostics.Monitoring.OpenApiGen/Program.cs
```diff
@@ -50,6 +50,7 @@ public static void Main(string[] args)
                     diagnosticPort: null,
                     noAuth: false,
```
kelltrick, line 51: I wonder at what point we make this some kind of "bag of flags" to avoid having a method with 100 parameters. To be clear, I don't think this should be done now, but if we keep adding more of these we may want to find a better way.

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
```diff
@@ -176,6 +178,8 @@ public static IHostBuilder CreateHostBuilder(IConsole console, string[] urls, st
 
                     services.AddSingleton<IAuthOptions>(authenticationOptions);
 
+                    services.AddSingleton<IEgressOutputOptions>(egressOptions);
```
kelltrick, line 181: I think this should be moved in here: https://github.com/dotnet/dotnet-monitor/blob/d7c6fcb5e92632b3f4597b1a6bc6437486c00b6b/src/Tools/dotnet-monitor/ServiceCollectionExtensions.cs#L49 because it goes with egress options.

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -265,7 +265,7 @@
   </data>
   <data name="HelpDescription_OptionMetrics" xml:space="preserve">
     <value>Enable publishing of metrics</value>
-    <comment>Gets the string to display in help that explains what the '--metricls' option does.</comment>
+    <comment>Gets the string to display in help that explains what the '--metrics' option does.</comment>
```
kelltrick, line 268: thank you for fixing my typo.

kelltrick (approved):

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
jander-msft: It kind of does and kind of doesn't go with the egress system because 'http egress' is not a real egress, but to the user they probably see it the same. So I'm okay either way with it being where it is or where Patrick says to put it. We may refactor it back out later in the future depending on how we want to use the WebApi library.

src/Tests/Microsoft.Diagnostics.Monitoring.OpenApiGen/Program.cs
jander-msft: The System.CommandLine library mostly forces this pattern, but at lower levels (which I think CreateHostBuilder method can be considered at a lower level) we could move to using a settings object (a configuration object for setting up the configuration, if you will :smile:).

PR #598, opened 2021-07-21
Add info endpoint testing

kkeirstead:
Added in a basic test for the new info endpoint that checks whether the returned result has the expected values.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/InfoTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/InfoTests.cs
@@ -1,0 +1,69 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Diagnostics.Monitoring.TestCommon;
+using Microsoft.Diagnostics.Monitoring.UnitTests.Fixtures;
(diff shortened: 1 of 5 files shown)


jander-msft (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/InfoTests.cs
```diff
@@ -0,0 +1,69 @@
...
+                _outputHelper,
+                _httpClientFactory,
+                mode,
+                TestAppScenarios.AsyncWait.Name,
+                appValidate: async (runner, client) =>
+                {
+                    // GET /info
+                    DotnetMonitorInfo info = await client.GetInfoAsync();
+
+                    Assert.NotNull(info.Version); // Not sure of how to get Dotnet Monitor version from within tests...
```
wiktork, line 52: While this is not a great option, but do we uniformly version all the built assemblies? If so, then the test assembly version and the dotnet-monitor version are likely to be the same. Another option would be to load dotnet-monitor.dll into the test process but I don't think we should go that far.

kkeirstead: Are you saying that if we get the version the way we do in DiagController (GetDotnetMonitorVersion()) that it might be the same? That's what I originally experimented with, and the test failed saying that version was '5.0.0-dev...' vs '16.6.1'. Was this what you had in mind, or were you suggesting a different approach?

Issue #412, opened 2021-06-08
Add Kestrel configuration to dotnet-monitor configuration schema


wiktork:
Based on #23

PR #606, opened 2021-07-23
'config show' now outputs Logging info from appsettings.json

### LINKED ISSUE #481: Add "Logging" configuration section to `config show` command
It would be helpful to understand how dotnet-monitor is configured to log its operations when a user runs the `config show` command.

kkeirstead:
Note: I wasn't entirely sure about the scope of the issue based on what's written in #481, so if this is completely different than what was intended let me know. I essentially just latched onto the existing functionality for config show and added in Logging as one of the outputs. This is an example of the new output for `config show`

LoggingConfigShow

Closes #481

### DIFF
--- a/src/Tools/dotnet-monitor/ConfigurationKeys.cs
+++ b/src/Tools/dotnet-monitor/ConfigurationKeys.cs
@@ -19,5 +19,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor
         public const string Storage = nameof(Storage);
 
         public const string DefaultProcess = nameof(DefaultProcess);
+
+        public const string Logging = nameof(Logging);
     }
 }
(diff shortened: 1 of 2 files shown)


src/Tools/dotnet-monitor/ConfigurationKeys.cs
```diff
@@ -19,5 +19,7 @@ internal static class ConfigurationKeys
         public const string Storage = nameof(Storage);
 
         public const string DefaultProcess = nameof(DefaultProcess);
+
+        public const string Logging = nameof(Logging);
```
wiktork, line 23: Note that some already existing configuration sections have named constants. For example https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.hosting.webhostdefaults?view=aspnetcore-5.0. I don't know if Logging does, but it may be worthwhile to check.

wiktork:
Note there's also a related issue (https://github.com/dotnet/dotnet-monitor/issues/412). There are likely other sections (such as Kestrel) that we may want to add here.

wiktork (approved):

PR #637, opened 2021-07-28
Disable http response egress - testing

kkeirstead:
Added in two test cases to ensure the proper error ensues when using the --no-http-egress flag for dumps and logs.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EgressTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EgressTests.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Microsoft.AspNetCore.Http;
 using Microsoft.Diagnostics.Monitoring.TestCommon;
 using Microsoft.Diagnostics.Monitoring.UnitTests.Fixtures;
 using Microsoft.Diagnostics.Monitoring.UnitTests.HttpApi;
@@ -35,7 +36,6 @@ namespace Microsoft.Diagnostics.Monitoring.UnitTests
         private readonly DirectoryInfo _tempEgressPath;
 
         private const string FileProviderName = "files";
-        
 
         public EgressTests(ITestOutputHelper outputHelper, ServiceProviderFixture serviceProviderFixture)
         {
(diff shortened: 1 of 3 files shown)


jander-msft:
@kkeirstead, I think you included your commits from the prior change. I would recommend rebasing from `main` and skipping those commits so that only the test additions are in the PR. You can then force push the change to the remote branch.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EgressTests.cs
```diff
@@ -211,6 +211,39 @@ public async Task SharedConcurrencyLimitTest()
...
+
+                    // Dump Error Check
+                    ValidationProblemDetailsException validationProblemDetailsExceptionDumps = await Assert.ThrowsAsync<ValidationProblemDetailsException>(
+                        () => appClient.CaptureDumpAsync(appRunner.ProcessId, DumpType.Mini));
+                    Assert.Equal(HttpStatusCode.BadRequest, validationProblemDetailsExceptionDumps.StatusCode);
+                    Assert.Equal(StatusCodes.Status400BadRequest, validationProblemDetailsExceptionDumps.Details.Status);
+
+                    // Logs Error Check
+                    ValidationProblemDetailsException validationProblemDetailsExceptionLogs = await Assert.ThrowsAsync<ValidationProblemDetailsException>(
+                            () => appClient.CaptureLogsAsync(appRunner.ProcessId, TestTimeouts.LogsDuration, Extensions.Logging.LogLevel.None, WebApi.LogFormat.NDJson));
```
jander-msft, line 238: nit: Add `using`s at the top of this file for `Microsoft.Extensions.Logging` and `Microsoft.Diagnostics.Monitoring.WebApi` and then remove the namespace disambiguation here. It makes it a little harder to read given the length of the code line.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EgressTests.cs
```diff
@@ -211,6 +211,39 @@ public async Task SharedConcurrencyLimitTest()
...
+                appValidate: async (appRunner, appClient) =>
+                {
+                    ProcessInfo processInfo = await appClient.GetProcessAsync(appRunner.ProcessId);
+                    Assert.NotNull(processInfo);
+
+                    // Dump Error Check
+                    ValidationProblemDetailsException validationProblemDetailsExceptionDumps = await Assert.ThrowsAsync<ValidationProblemDetailsException>(
+                        () => appClient.CaptureDumpAsync(appRunner.ProcessId, DumpType.Mini));
+                    Assert.Equal(HttpStatusCode.BadRequest, validationProblemDetailsExceptionDumps.StatusCode);
+                    Assert.Equal(StatusCodes.Status400BadRequest, validationProblemDetailsExceptionDumps.Details.Status);
```
jander-msft, line 234: Given that the error message doesn't change between the endpoints or what is calling it, do you think you could add an assert for the expected error message? Maybe you can reference the resource string instead of embedding the literal message.

kkeirstead:
@jander-msft Yep, I noticed that and was able to merge to avoid all the files from appearing as changed but it kept that commit history. I'll see if I can make it look like less of a mess.

kkeirstead:
I had some difficulty with figuring out how to do the rebase, so I ended up taking a more manual approach to addressing the out-of-place commits. This should have the same intended effect of cleaning things up - just a different way of getting there.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EgressTests.cs
```diff
@@ -211,6 +214,43 @@ public async Task SharedConcurrencyLimitTest()
...
+                    // This should be identical to the error message found in Strings.resx
+                    Assert.Equal("HTTP egress is not enabled.", validationProblemDetailsExceptionDumps.Message);
```
jander-msft, line 239: Please make this a constant and move the comment to the constant.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EgressTests.cs
```diff
@@ -24,6 +27,7 @@
 using System.Threading.Tasks;
 using Xunit;
 using Xunit.Abstractions;
+using DiagnosticPortConnectionMode = Microsoft.Diagnostics.Monitoring.UnitTests.Options.DiagnosticPortConnectionMode;
```
jander-msft, line 30: Why is this necessary? The other existing tests in this file are already using this type.

jander-msft (changes requested):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EgressTests.cs
kkeirstead: Since you recommended I add the extension for WebApi, I started getting an error for DiagnosticPortConnectionMode that it was an ambiguous reference between Options and WebApi. We had added this when we had the same issue in LogsTests, so I reimplemented it here. I just pushed changes (without changing that), so if you have something else in mind let me know and I'll push out a new version.

kkeirstead:
Looks like the old commits came back with the way I made my changes; will push again shortly to resolve this.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EgressTests.cs
jander-msft: > Since you recommended I add the extension for WebApi, I started getting an error for DiagnosticPortConnectionMode that it was an ambiguous reference between Options and WebApi. We had added this when we had the same issue in LogsTests, so I reimplemented it here. I just pushed changes (without changing that), so if you have something else in mind let me know and I'll push out a new version.

I reread this a few times and didn't understand what you were referring to until I realized that "extension" is referring to the `using Microsoft.Diagnostics.Monitoring.WebApi;` statement. This makes more sense now.

jander-msft (approved):

PR #669, opened 2021-08-02
Add Logging configuration to dotnet-monitor configuration schema

### LINKED ISSUE #482: Add Logging configuration to dotnet-monitor configuration schema
The logging schema will end up looking something like https://github.com/SchemaStore/schemastore/blob/7330610cd202f61c2536535b0ecd0b4a99f6b388/src/schemas/json/appsettings.json#L292

The root `Logging` section has
- a `LogLevel` property which maps categories to log levels. There's no options object for this since it is loaded via custom IConfigureOptions implementation (see [LoggerFilterConfigureOptions.LoadRules](https://github.com/dotnet/runtime/blob/74905e767eb7b971636310965b66bf5dfd6a3545/src/libraries/Microsoft.Extensions.Logging.Configuration/src/LoggerFilterConfigureOptions.cs#L55))
- a `CaptureScopes` property as defined [here](https://github.com/dotnet/runtime/blob/74905e767eb7b971636310965b66bf5dfd6a3545/src/libraries/Microsoft.Extensions.Logging/src/LoggerFilterOptions.cs#L21)

The options for the `Console` section are defined [here]( https://github.com/dotnet/runtime/blob/release/5.0/src/libraries/Microsoft.Extensions.Logging.Console/src/ConsoleLoggerOptions.cs)
- Do not document the obsoleted options.
- This section has an additional property named `FormatterOptions` that is bound on-demand because its schema is different depending on the value of the `FormatterName` property. So there are a couple of ways to do this: (1) only document the common options, (2) document the union of all of the options and call out if they are specific to a formatter, (3) use [schema composition[(https://json-schema.org/understanding-json-schema/reference/combining.html) which might not work for IntelliSense in some tools, or (4) use [conditional schemas](https://json-schema.org/understanding-json-schema/reference/conditionals.html) which might not work for IntelliSense in some tools. This will require some investigation.
    - The common options are [ConsoleFormatterOptions](https://github.com/dotnet/runtime/blob/release/5.0/src/libraries/Microsoft.Extensions.Logging.Console/src/ConsoleFormatterOptions.cs)
    - The provider specific options are [JsonConsoleFormatterOptions](https://github.com/dotnet/runtime/blob/release/5.0/src/libraries/Microsoft.Extensions.Logging.Console/src/JsonConsoleFormatterOptions.cs) for the `json` formatter and [SimpleConsoleFormatterOptions](https://github.com/dotnet/runtime/blob/release/5.0/src/libraries/Microsoft.Extensions.Logging.Console/src/SimpleConsoleFormatterOptions.cs) for the `simple` formatter. The `systemd` formatter uses the common options object [ConsoleFormatterOptions](https://github.com/dotnet/runtime/blob/release/5.0/src/libraries/Microsoft.Extensions.Logging.Console/src/ConsoleFormatterOptions.cs).

All other logger providers, namely `Debug`, `EventLog` and `EventSource` only have a single property named `LogLevel` which has the same schema as the `LogLevel` property on the root `Logging` section.

kkeirstead:
Provides intellisense capabilities in schema.json when users are configuring Logging in their settings.json file.

Closes #482


src/Tools/dotnet-monitor/LogLevelThreshold.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Text.Json.Serialization;
+
+#if UNITTEST
+namespace Microsoft.Diagnostics.Monitoring.UnitTests.Options
+#else
+namespace Microsoft.Diagnostics.Tools.Monitor
+#endif
+{
+    [JsonConverter(typeof(JsonStringEnumConverter))]
+    public enum LogLevelThreshold
```
jander-msft, line 16: You should be able to replace this with the built-in LogLevel enum. Notice that the descriptions that you added to the individual fields do not appear in the schema.json file, so I don't thin defining it yourself adds much value. In the future, if we want that level of information (if we can add it), we can go into the schema graph after the fact and poke in the description values that we want.

jander-msft: If the converted is really needed, it (hopefully) can be added to `JsonSchemaGeneratorSettings.SerializerSettings.Converters` in `SchemaGenerator.GenerateSchema.

src/Tools/dotnet-monitor/ConsoleOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.ComponentModel;
+
+#if UNITTEST
+namespace Microsoft.Diagnostics.Monitoring.UnitTests.Options
+#else
+namespace Microsoft.Diagnostics.Tools.Monitor
+#endif
+{
+    internal sealed class ConsoleOptions
```
jander-msft, line 15:
```suggestion
    internal sealed class ConsoleLoggerOptions
```

src/Tools/dotnet-monitor/FormatterNameOptions.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Text.Json.Serialization;
+
+#if UNITTEST
+namespace Microsoft.Diagnostics.Monitoring.UnitTests.Options
+#else
+namespace Microsoft.Diagnostics.Tools.Monitor
+#endif
+{
+    [JsonConverter(typeof(JsonStringEnumConverter))]
+    public enum FormatterNameOptions
```
jander-msft, line 16:
```suggestion
    public enum ConsoleLoggerFormat
```

src/Tools/dotnet-monitor/FormatterOptions.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.ComponentModel;
+
+#if UNITTEST
+namespace Microsoft.Diagnostics.Monitoring.UnitTests.Options
+#else
+namespace Microsoft.Diagnostics.Tools.Monitor
+#endif
+{
+    internal sealed class FormatterOptions
```
jander-msft, line 15: The `simple` and `json` formatters have more options beyond these. These are just the common options. You could try using this as a base class, however it probably won't be understood by IntelliSense providers.

I think your other options are to (a) add all of the properties among all of the formatters into the same class and call out in the display strings to which formatters they apply, or (b) create a class for each formatter and use [implication](https://json-schema.org/understanding-json-schema/reference/conditionals.html#implication) to apply the schemas conditionally based on what value of FormatterName is. The latter gives a richer experience where IntelliSense is only offered for the properties that apply to the specified formatter name; it can also provide a default schema if the formatter name is not specified.

src/Microsoft.Diagnostics.Monitoring.WebApi/OptionsDisplayStrings.resx
```diff
@@ -193,6 +217,74 @@
...
+  </data>
+  <data name="DisplayAttributeDescription_LoggingOptions_EventSource" xml:space="preserve">
+    <value>Logging options for the EventSource provider.</value>
+    <comment>The description provided for the EventSource parameter on LoggingOptions</comment>
+  </data>
+  <data name="DisplayAttributeDescription_LoggingOptions_LogLevel" xml:space="preserve">
+    <value>Corresponding LogLevels to categories; applies to no specific provider.</value>
+    <comment>The description provided for the LogLevel parameter on LoggingOptions.</comment>
+  </data>
+  <data name="DisplayAttributeDescription_LoggingOptions_LogLevelOptions_LogLevel" xml:space="preserve">
```
jander-msft, line 256:
```suggestion
  <data name="DisplayAttributeDescriptions_LogLevelOptions_LogLevel" xml:space="preserve">
```

src/Tools/dotnet-monitor/FormatterOptions.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_ConsoleOptions_FormatterOptions_IncludeScopes))]
+        public bool IncludeScopes { get; set; }
```
jander-msft, line 20: This probably has a default value?

src/Microsoft.Diagnostics.Monitoring.WebApi/OptionsDisplayStrings.resx
```diff
@@ -157,6 +157,30 @@
     <value>Buffer size used when copying data from an egress callback returning a stream to the egress callback that is provided a stream to which data is written.</value>
     <comment>The description provided for the CopyBufferSize parameter on all egress provider options.</comment>
   </data>
+  <data name="DisplayAttributeDescription_ConsoleOptions_FormatterName" xml:space="preserve">
+    <value>Name of the log message formatter to use. Defaults to 'Simple'.</value>
+    <comment>The description provided for the FormatterName parameter on ConsoleOptions.</comment>
+  </data>
+  <data name="DisplayAttributeDescription_ConsoleOptions_FormatterOptions" xml:space="preserve">
```
jander-msft, line 164: All of the ones that are `DisplayAttributeDescription_ConsoleOptions_FormatterOptions`* probably should be `DisplayAttributeDescription_ConsoleFormatterOptions`*

jander-msft (changes requested):

src/Tools/dotnet-monitor/ConsoleOptions.cs
wiktork, line 15: Should we really be redefining these types? Wouldn't it be preferable to use existing types and add them into the schema generator unit test?

src/Tools/dotnet-monitor/FormatterOptions.cs
kkeirstead: This doesn't imply that it does for IncludeScopes (unless I'm just looking at the wrong page): https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.console.consoleformatteroptions?view=dotnet-plat-ext-5.0

documentation/schema.json
```diff
@@ -81,6 +81,17 @@
           "$ref": "#/definitions/ProcessFilterOptions"
         }
       ]
+    },
+    "Logging": {
```
wiktork, line 85: How does our schema compare to the "official" one? https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/appsettings.json

src/Microsoft.Diagnostics.Monitoring.WebApi/OptionsDisplayStrings.resx
kkeirstead: Does this also imply that I should rename the FormatterOptions class to be ConsoleFormatterOptions?

jander-msft: I think what's going to happen is that FormatterOptions is flattened into three new classes: ConsoleFormatterOptions, JsonFormatterOptions, and SimpleFormatterOptions.

src/Tools/dotnet-monitor/ConsoleOptions.cs
jander-msft: One issue that I encountered while reusing existing types for a different feature was that the schema generation will generate base schemas for the base types but I couldn't get VS Code IntelliSense to work with it; it's possible I did something wrong when trying it.

Also, we'd have to manually insert descriptions and default values for the entire structure in the schema generator project, which isn't ideal.

A minor benefit to the way it is now is that it allows tests to change logging levels through a strongly-type mechanism. They currently do it via environment variables with well-known names, which is fine to do because how logging is done isn't really going to change. A change in the category name or level would impact the tests regardless of how they are described.

wiktork: Then perhaps we can only compile them into the test code? We don't need these in the actual dotnet-monitor code.

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
```diff
@@ -29,11 +38,62 @@ public string GenerateSchema()
...
+
+            JsonSchemaProperty formatterOptionsProperty = new JsonSchemaProperty();
+            JsonSchemaProperty logToStandardErrorThresholdPropertyDefault = new JsonSchemaProperty();
+
+            JsonSchema formatterOptionsSchema = new JsonSchema();
+            formatterOptionsSchema.Reference = consoleFormatterOptions;
+
+            formatterOptionsProperty.OneOf.Add(formatterOptionsSchema);
+
+            consoleLoggerOptionsSchema.Properties.Add(nameof(ConsoleLoggerOptions.LogToStandardErrorThreshold), logToStandardErrorThresholdPropertyDefault);
```
jander-msft, line 92: This shouldn't be necessary, correct?

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
```diff
@@ -29,11 +38,62 @@ public string GenerateSchema()
...
+            consoleLoggerOptionsSchema.Properties.Add(nameof(ConsoleLoggerOptions.FormatterOptions), formatterOptionsProperty);
+            consoleLoggerOptionsSchema.AllowAdditionalProperties = false;
```
jander-msft, line 94: This shouldn't be necessary, correct? It's already in the ConsoleLoggerOptions schema.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/ConsoleLoggerFormat.cs
```diff
@@ -0,0 +1,27 @@
...
+using Microsoft.Diagnostics.Monitoring.WebApi;
+using System.ComponentModel.DataAnnotations;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
+{
+    [JsonConverter(typeof(JsonStringEnumConverter))]
+    public enum ConsoleLoggerFormat
+    {
+        [Display(
```
jander-msft, line 14: These display strings are not generated into the schema.json file. I would remove them so as to not imply that they are having some effect.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/SimpleConsoleFormatterOptions.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
+{
+    internal sealed class SimpleConsoleFormatterOptions
+    {
+#if NET5_0_OR_GREATER
```
jander-msft, line 19: Because dotnet-monitor is using the 5.0+ Logging infrastructure, we want to make sure this is documented in the schema.json file. The Microsoft.Diagnostics.Monitoring.ConfigurationSchema project is currently only building for netcoreapp3.1, so I think you'll need to update that project net50 so that it gets the Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests that targets .NET 5.

src/Tools/dotnet-monitor/ConsoleFormatterOptions.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
jander-msft, line 1: Remove since it was copied to the FunctionalTests project.

src/Tools/dotnet-monitor/ConsoleLoggerFormat.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
jander-msft, line 1: Remove since it was copied to the FunctionalTests project.

src/Tools/dotnet-monitor/ConsoleLoggerOptions.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
jander-msft, line 1: Remove since it was copied to the FunctionalTests project.

src/Tools/dotnet-monitor/JsonConsoleFormatterOptions.cs
```diff
@@ -0,0 +1,35 @@
+﻿using Microsoft.Diagnostics.Monitoring.WebApi;
```
jander-msft, line 1: Remove since it was copied to the FunctionalTests project.

src/Tools/dotnet-monitor/LogLevelOptions.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
jander-msft, line 1: Remove since it was copied to the FunctionalTests project.

src/Tools/dotnet-monitor/LoggingOptions.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
jander-msft, line 1: Remove since it was copied to the FunctionalTests project.

src/Tools/dotnet-monitor/SimpleConsoleFormatterOptions.cs
```diff
@@ -0,0 +1,45 @@
+﻿using Microsoft.Diagnostics.Monitoring.WebApi;
```
jander-msft, line 1: Remove since it was copied to the FunctionalTests project.

src/Microsoft.Diagnostics.Monitoring.WebApi/OptionsDisplayStrings.resx
```diff
@@ -157,6 +157,42 @@
     <value>Buffer size used when copying data from an egress callback returning a stream to the egress callback that is provided a stream to which data is written.</value>
     <comment>The description provided for the CopyBufferSize parameter on all egress provider options.</comment>
   </data>
+  <data name="DisplayAttributeDescription_ConsoleFormatterOptions_IncludeScopes" xml:space="preserve">
+    <value>Gets or sets whether scopes should be included or not.</value>
+    <comment>The description provider for the IncludesScopes parameter on FormatterOptions.</comment>
+  </data>
+  <data name="DisplayAttributeDescription_ConsoleFormatterOptions_TimestampFormat" xml:space="preserve">
+    <value>Gets or sets format string used to format timestamp in logging messages. Defaults to null.</value>
```
jander-msft, line 165:
```suggestion
    <value>Gets or sets the format string used to format the timestamp in logging messages. Defaults to null.</value>
```

src/Microsoft.Diagnostics.Monitoring.WebApi/OptionsDisplayStrings.resx
```diff
@@ -193,6 +229,38 @@
...
+  <data name="DisplayAttributeDescription_LoggingOptions_EventLog" xml:space="preserve">
+    <value>Logging options for the EventLog provider.</value>
+    <comment>The description provided for the EventLog parameter on LoggingOptions</comment>
+  </data>
+  <data name="DisplayAttributeDescription_LoggingOptions_EventSource" xml:space="preserve">
+    <value>Logging options for the EventSource provider.</value>
+    <comment>The description provided for the EventSource parameter on LoggingOptions</comment>
+  </data>
+  <data name="DisplayAttributeDescription_LoggingOptions_LogLevel" xml:space="preserve">
+    <value>Corresponding LogLevels to categories; applies to no specific provider.</value>
```
jander-msft, line 257:
```suggestion
    <value>A mapping of logger categories to log levels describing the threshold level at which messages are logged for each category; applies to no specific provider.</value>
```

src/Microsoft.Diagnostics.Monitoring.WebApi/OptionsDisplayStrings.resx
```diff
@@ -193,6 +229,38 @@
...
+    <comment>The description provided for the LogLevel parameter on LoggingOptions.</comment>
+  </data>
+  <data name="DisplayAttributeDescription_LogLevelOptions_LogLevel" xml:space="preserve">
+    <value>Corresponding LogLevels to categories; applies to the specified provider.</value>
```
jander-msft, line 261:
```suggestion
    <value>A mapping of logger categories to log levels describing the threshold level at which messages are logged for each category; applies to the specified provider.</value>
```

jander-msft (changes requested):

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
```diff
@@ -29,11 +38,62 @@ public string GenerateSchema()
                 kvp.Value.Default = JsonSchema.CreateAnySchema();
             }
 
+            JsonSchema jsonConsoleLoggerOptionsSchema = GenerateConsoleLoggerOptionsSchema(jsonConsoleFormatterOptions, ConsoleLoggerFormat.Json);
+            JsonSchema simpleConsoleLoggerOptionsSchema = GenerateConsoleLoggerOptionsSchema(simpleConsoleFormatterOptions, ConsoleLoggerFormat.Simple);
+            JsonSchema systemdConsoleLoggerOptionsSchema = GenerateConsoleLoggerOptionsSchema(systemdConsoleFormatterOptions, ConsoleLoggerFormat.Systemd);
+            JsonSchema defaultConsoleLoggerOptionsSchema = GenerateDefaultConsoleLoggerOptionsSchema(simpleConsoleFormatterOptions);
+
+            schema.Definitions[nameof(ConsoleLoggerOptions)].AnyOf.Add(jsonConsoleLoggerOptionsSchema);
```
jander-msft, line 46: Does this still work if it is switch from `AnyOf` to `OneOf`?

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
kkeirstead: I added this (and the other line about allowing additional properties) to essentially require that FormatterName not be provided. I wasn't sure how to require FormatterName not be present, so I just explicitly said which properties could be there and disallowed other properties. Feel free to let me know if you have a more elegant solution in mind :)

jander-msft: This seems to work:

```
"ConsoleLoggerOptions": {
    "type": "object",
    "properties": {
        "FormatterName": {
            "description": "Name of the log message formatter to use.",
            "oneOf": [
                {
                    "type": "null"
                },
                {
                    "$ref": "#/definitions/ConsoleLoggerFormat"
                }
            ]
        },
        "LogToStandardErrorThreshold": {
            "description": "The minimum level of messages that are written to Console.Error.",
            "oneOf": [
                {
                    "$ref": "#/definitions/LogLevel"
                }
            ]
        },
        "FormatterOptions": {
            "description": "Formatter options for the built-in console log formatter.",
            "oneOf": [
                {},
                {
                    "type": "null"
                }
            ]
        }
    },
    "oneOf": [
        {
            "required": [
                "FormatterName"
            ],
            "properties": {
                "FormatterName": {
                    "const": "Json"
                },
                "FormatterOptions": {
                    "oneOf": [
                        {
                            "$ref": "#/definitions/JsonConsoleFormatterOptions"
                        }
                    ]
                }
            }
        },
        {
            "required": [
                "FormatterName"
            ],
            "properties": {
                "FormatterName": {
                    "const": "Simple"
                },
                "FormatterOptions": {
                    "oneOf": [
                        {
                            "$ref": "#/definitions/SimpleConsoleFormatterOptions"
                        }
                    ]
                }
            }
        },
        {
            "required": [
                "FormatterName"
            ],
            "properties": {
                "FormatterName": {
                    "const": "Systemd"
                },
                "FormatterOptions": {
                    "oneOf": [
                        {
                            "$ref": "#/definitions/ConsoleFormatterOptions"
                        }
                    ]
                }
            }
        },
        {
            "properties": {
                "FormatterName": {
                    "type": "null"
                },
                "FormatterOptions": {
                    "oneOf": [
                        {
                            "$ref": "#/definitions/SimpleConsoleFormatterOptions"
                        }
                    ]
                }
            }
        }
    ]
}
```

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
jander-msft: I confirmed that `OneOf` does work.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/SimpleConsoleFormatterOptions.cs
kkeirstead: Just a bit confused about what is being requested by this comment. I see that the Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests linked under ConfigurationSchema's Projects has a path to the netcoreapp3.1 version (which I'm guessing we would want to instead be the net5.0 version?), but I'm not sure what action I'm supposed to be taking.

jander-msft: Change the Microsoft.Diagnostics.Monitoring.ConfigurationSchema project from netcoreapp3.1 to net50 so that it picks up the .NET 5.0 version of Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests so that ColorBehavior is written to the schema.json file.

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
```diff
@@ -29,11 +38,64 @@ public string GenerateSchema()
...
+            JsonSchemaProperty formatterNameProperty = new JsonSchemaProperty();
+            JsonSchemaProperty formatterOptionsProperty = new JsonSchemaProperty();
+
+            JsonSchema formatterOptionsSchema = new JsonSchema();
+            formatterOptionsSchema.Reference = consoleFormatterOptions;
+
+            formatterOptionsProperty.OneOf.Add(formatterOptionsSchema);
+
+            formatterNameProperty.ExtensionData = new Dictionary<string, object>();
+            formatterNameProperty.ExtensionData.Add("type", "null");
```
jander-msft, line 93: I don't think you need to use ExtensionData for this, just `formatterNameProperty.Type = JsonObjectType.Null;`

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/LogLevelOptions.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
+{
+    internal sealed class LogLevelOptions
+    {
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_LoggingOptions_LogLevel))]
+        [JsonConverter(typeof(JsonStringEnumConverter))]
+        public IDictionary<string, LogLevel> LogLevel { get; set; }
```
jander-msft, line 19:
```suggestion
        public IDictionary<string, LogLevel?> LogLevel { get; set; }
```

The LogLevel can be null here so that it falls back to the level set by the `Default` 'category'.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/LoggingOptions.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
+{
+    internal sealed class LoggingOptions
+    {
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_LoggingOptions_LogLevel))]
+        [JsonConverter(typeof(JsonStringEnumConverter))]
+        public IDictionary<string, LogLevel> LogLevel { get; set; }
```
jander-msft, line 20:
```suggestion
        public IDictionary<string, LogLevel?> LogLevel { get; set; }
```

The LogLevel can be null here so that it falls back to the level set by the `Default` 'category'.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/ConsoleLoggerFormat.cs
```diff
@@ -0,0 +1,18 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Diagnostics.Monitoring.WebApi;
+using System.ComponentModel.DataAnnotations;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
+{
+    [JsonConverter(typeof(JsonStringEnumConverter))]
+    public enum ConsoleLoggerFormat
```
jander-msft, line 12: nit: internal

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/ConsoleLoggerOptions.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using Microsoft.Diagnostics.Monitoring.WebApi;
+using System.ComponentModel.DataAnnotations;
+using System.ComponentModel;
+using Microsoft.Extensions.Logging;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
+{
+    internal sealed class ConsoleLoggerOptions
```
jander-msft, line 13: This is missing the LogLevel property.

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
```diff
@@ -29,11 +38,64 @@ public string GenerateSchema()
                 kvp.Value.Default = JsonSchema.CreateAnySchema();
             }
 
+            JsonSchema jsonConsoleLoggerOptionsSchema = GenerateConsoleLoggerOptionsSchema(jsonConsoleFormatterOptions, ConsoleLoggerFormat.Json);
```
jander-msft, line 41: A few lines above this comment, there is code to add an empty object schema for a handful of egress properties. You'll have to do the same for the following properties in order to get better defaults for object-typed properties:

- LoggingOptions.LogLevel
- LoggingOptions.Console
- LoggingOptions.EventLog
- LoggingOptions.Debug
- LoggingOptions.EventSource
- LogLevelOptions.LogLevel
- ConsoleLoggerOptions.FormatterOptions
- ConsoleLoggerOptions.LogLevel (when you add this to ConsoleLoggerOptions)

Otherwise, you get this which isn't helpful in determining what type of property it is:
nulldefault

jander-msft (changes requested):

src/Tools/dotnet-monitor/LogLevelThreshold.cs
jander-msft: > If the converted is really needed, it (hopefully) can be added to `JsonSchemaGeneratorSettings.SerializerSettings.Converters` in `SchemaGenerator.GenerateSchema.

Were you able to try this?

kkeirstead: I never ended up investigating that comment and just substituted in LogLevel for LogLevelThreshold. Just to understand your comment: instead of having `[JsonConverter(typeof(JsonStringEnumConverter))]` above each LogLevel method, is the idea here to achieve the same result within SchemaGenerator? Or am I misinterpreting what this is asking for?

jander-msft: You've added `[JsonConverter(typeof(JsonStringEnumConverter))]` to a lot of types and properties. The fact that the `LogLevel` enum type doesn't have this and yet still produces just names in the schema suggests this isn't needed on the enum types. Further, I think you can remove it from all properties and instead do `setttings.SerializerSettings.Converters.Add(new JsonStringEnumConverter())` in the SchemaGenerator (if it really is necessary).

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
kkeirstead: Yep, I had initially tried this but just used `null` instead of `JsonObjectType.Null` (and it didn't work). This works though and reads cleaner.

kkeirstead: This was giving me trouble in terms of defaulting to null when typing in `FormatterName` so the latest version switches to a different approach. It doesn't necessarily feel like as clean of a solution, but the behavior appears to be what we want.

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
```diff
@@ -11,29 +12,104 @@ internal sealed class SchemaGenerator
         public string GenerateSchema()
         {
             var settings = new JsonSchemaGeneratorSettings();
+
+            settings.SerializerSettings = new Newtonsoft.Json.JsonSerializerSettings();
```
jander-msft, line 16: nit: Use using statements if you can.

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
```diff
@@ -11,29 +12,104 @@ internal sealed class SchemaGenerator
...
+            JsonSchemaProperty formatterNameProperty = new JsonSchemaProperty();
+            JsonSchemaProperty formatterOptionsProperty = new JsonSchemaProperty();
+            
+            JsonSchema formatterOptionsSchema = new JsonSchema();
+            formatterOptionsSchema.Reference = consoleFormatterOptions;
+
+            formatterOptionsProperty.Reference = formatterOptionsSchema;
+
+            formatterNameProperty.ExtensionData = new Dictionary<string, object>();
+            formatterNameProperty.ExtensionData.Add("const", JsonObjectType.Null);
```
jander-msft, line 106: This writes out:
```json
"FormatterName": {
  "default": "Simple",
  "const": 8
},
```

which is not correct. Please remove this extension data.

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
```diff
@@ -11,29 +12,104 @@ internal sealed class SchemaGenerator
...
+
+            consoleLoggerOptionsSchema.Properties.Add(nameof(ConsoleLoggerOptions.FormatterName), formatterNameProperty);
+            consoleLoggerOptionsSchema.Properties.Add(nameof(ConsoleLoggerOptions.FormatterOptions), formatterOptionsProperty);
+            consoleLoggerOptionsSchema.Properties[nameof(ConsoleLoggerOptions.FormatterName)].Default = "Simple";
```
jander-msft, line 110: You already have access to this via the `formatterNameProperty` variable, so do this instead:
```
formatterNameProperty.Default = "Simple";
```

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
jander-msft: Please make this change.

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
kkeirstead: Looks like since I changed that at the same time as setting the formatterNameProperty to "Simple" I assumed it actually had played a role in addressing the issue (when it hadn't). This is reverted to the way it was before using `formatterNameProperty.Type = JsonObjectType.Null`

jander-msft (approved):

PR #679, opened 2021-08-03
Automatically flush trace data to Azure storage.

wiktork:
Some other considerations:
- We could push this logic out of the egress provider and into trace directly. We would need to break down CopyToAsync and periodically flush. I chose this approach since it ensures that anything that gets egressed to azure flushes on certain boundaries.
- The default buffer size in this PR and in the blob client is 4Mi. We could make this smaller and see data at an even more granular level. This would mean less memory usage, better data availability, but more frequent commit calls.

### DIFF
--- a/src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
+++ b/src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
@@ -22,9 +22,11 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Egress.AzureBlob
     /// <remarks>
     /// Blobs created through this provider will overwrite existing blobs if they have the same blob name.
     /// </remarks>
-    internal class AzureBlobEgressProvider :
+    internal partial class AzureBlobEgressProvider :
         EgressProvider<AzureBlobEgressProviderOptions>
     {
+        private int BlobStorageBufferSize = 4 * 1024 * 1024;
+
         public AzureBlobEgressProvider(ILogger<AzureBlobEgressProvider> logger)
             : base(logger)
         {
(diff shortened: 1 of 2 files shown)


jander-msft (approved):

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -75,12 +77,24 @@ public AzureBlobEgressProvider(ILogger<AzureBlobEgressProvider> logger)
...
+                {
+                    BufferSize = BlobStorageBufferSize,
+                };
+                using (Stream blobStream = await blobClient.OpenWriteAsync(overwrite: true, options: bloboptions, cancellationToken: token))
+                using (AutoFlushStream flushStream = new AutoFlushStream(blobStream, BlobStorageBufferSize))
                 {
+                    //Azure's stream from OpenWriteAsync will do the following
+                    //1. Write the data to a local buffer
+                    //2. Once that buffer is full, stage the data remotely (this data is not considered valid yet)
+                    //3. After 4Gi of data has been staged, the data will be commited. This can be forced earlier by flushing
```
kelltrick, line 91:
```suggestion
                    //3. After 4GiB of data has been staged, the data will be commited. This can be forced earlier by flushing
```

kelltrick (approved):
👍

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
wiktork: I've been using k8 nomenclature here: Mi, Gi, etc.

Issue #660, opened 2021-07-30
Create `CollectLogs` collection rule action


jander-msft:
- Create an action named `CollectLogs` that will invoke the logs pipeline to collect a logs artifact.
  - This action should output the egress path of the egressed logs so that other actions can consume the value.
- Create a strongly-typed options class, e.g. `CollectLogsActionOptions`, for the action with the following properties:
  - The `LogLevel` property is of type `LogLevel?` describes the default log level for log events that match entries in `FilterSpecs` without a log level. If not specified, the default is `Warning`.
  - The `FilterSpecs` property is of type `Dictionary<string, LogLevel?>?` and describes the custom filter specifications for capturing log levels that match the defined categories at the specified levels.
  - The `UseAppFilters` property is of type `bool?` and describes whether logs events that match the application-defined filters should be included in the logs. If not specified, the default is `true`.
  - The `Egress` property is of type `string` and describes which egress provider to use to egress the artifact. This egress provider must be described in the `Egress` root section of configuration.
- Create and register an `IValidateOptions<T>` instance for the action options.
  - The `LogLevel` property is optional.
  - The `FilterSpecs` property is optional.
  - The `UseAppFilters` property is optional.
  - The `Egress` property is required. Validate that the named egress provider does exist.

Issue #659, opened 2021-07-30
Create `CollectTrace` collection rule action


jander-msft:
- Create an action named `CollectTrace` that will invoke the trace pipeline to collect a nettrace artifact.
  - This action should output the egress path of the egressed nettrace so that other actions can consume the value.
- Create a strongly-typed options class, e.g. `CollectTraceActionOptions`, for the action with the following properties:
  - The `Profile` property is an optional enumeration and describes the name of a trace profile for which events are captured. Valid values are `Cpu`, `Http`, `Logs`, and `Metrics`. If not specified, the default is `Cpu, Http, Metrics`.
  - The `MetricsIntervalDuration` property is of type `TimeSpan?` and describes the duration of each collection interval when `Profile` property is `Metrics`. If not specified, the default value is 1 second.
  - The `Providers` property is of type `EventProvider[]` and describes the list of event providers from which events are captured.
    - The `ProviderName` property is of type `string` and describes the name of the event provider.
    - The `Keywords` property is of type `long?` and describes keywords to pass to the event provider. If not specified, the default is 0xFFFFFFFFFFFFFFFF (all keywords).
    - The `EventLevel` property is of type `EventLevel?` and describes the event level at which events are collected. If not specified, the default is `Verbose`.
    - The `Arguments` property is of type `Dictionary<string, string>?` and describes the named parameters and the argument values to pass to the event provider.
  - The `RequestRundown` property is of type `bool?` and describes if the target process should return rundown events in the trace. If not specified, the default is `true`.
  - The `BufferSizeMegabytes` property is of type `int?` and describes the buffer size of the event pipe to use in the target process. If not specified, the default is `256`.
  - The `Duration` property is of type `TimeSpan?` and describes the duration of the trace. If not specified, the default is 30 seconds.
  - The `Egress` property describes which egress provider to use to egress the artifact. This egress provider must be described in the `Egress` root section of configuration.
- The options properties are used in two parameter sets:
  - Profile Set: `Profile`, `MetricsIntervalSeconds`, `Duration`, `Egress`.
  - Advanced Set: `Providers`, `RequestRundown`, `BufferSizeMegabytes`, `Duration`, `Egress`.
- Create and register an `IValidateOptions<T>` instance for the action options.
  - Validate that only one parameter set can be satisfied.
  - Profile Set:
    - The `Profile` property is required.
    - The `MetricsIntervalDuration` property is optional. If specified, validate that it is greater than or equal to 1 second.
[...]

Issue #658, opened 2021-07-30
Create `CollectGCDump` collection rule action


jander-msft:
- Create an action named `CollectGCDump` that will invoke the GC dump pipeline to collect a gcdump artifact.
  - This action should output the egress path of the egressed gcdump so that other actions can consume the value.
- Create a strongly-typed options class, e.g. `CollectGCDumpActionOptions`, for the action with the following properties:
  - The `Egress` property describes which egress provider to use to egress the artifact. This egress provider must be described in the `Egress` root section of configuration.
- Create and register an `IValidateOptions<T>` instance for the action options.
  - The `Egress` property is required. Validate that the named egress provider does exist.

Issue #656, opened 2021-07-30
Create `AspNetResponseStatus` collection rule trigger


jander-msft:
- Create a trigger named `AspNetResponseStatus` that notifies when a number of HTTP responses with matching status codes exceeds a threshold during a specified sliding window of time.
- Create a strong-typed options class, e.g. `AspNetResponseStatusTriggerOptions`, that has the following properties:
  - The `StatusCodes` property is of type `string` and specifies the HTTP status codes to match. This value can be a single code, a range of codes, or a list of single codes and/or range of codes.
  - The `ResponseCount` property is of type `int` and specifies the threshold count for the trigger condition to be satisfied.
  - The `SlidingWindowDuration` property is of type `TimeSpan` and specifies the time window in which the number of responses (as specified by `ResponseCount`) with matching status codes must occur for the trigger condition to be satisfied.
  - The `IncludedPaths` property is of type `string` and specifies the list of request paths which should be monitored. If not specified, all paths are monitored (except for those specified in `ExcludedPaths`). The elements of the list are semi-colon delimited.
  - The `ExcludedPaths` property is of type `string` and specifies the list of request paths which should be ignored when monitoring. The elements of the list are semi-colon delimited.
- Create and register an `IValidateOptions<T>` for the response status trigger options:
  - The `StatusCodes` property is required. It can be a single status code (e.g. `400`), a range of codes (e.g. `400-499`), or a list of single codes and/or ranges of codes (e.g. `200-299;300;400-450;500`). All codes are required to be integers (there will not be any translation of status code names to status code values, or vice versa).
  - The `ResponseCount` property is required.
  - The `SlidingWindowDuration` property is required. It must be greater than or equal to 1 second. It cannot be infinite.
  - The `IncludedPaths` property is optional.
  - The `ExcludedPaths` property is optional.

Contingent on #647 investigation.

PR #688, opened 2021-08-05
Initial collection rule options, validation, and unit tests.

### LINKED ISSUE #625: Design collection rule configuration schema
Design the full configuration schema of collection rules and their underlying components. The schema should be conducive to supporting text completions when using the generated JSON schema for validation.

### LINKED ISSUE #626: Bind collection rule options from configuration
Add support for binding strongly-typed collection rule options from configuration.
- Create options class, e.g. `CollectionRuleOptions`, without any additional structure. Properties will be filled in later as different aspects of collection rules and triggers are brought online.
- The root property in configuration shall be named `CollectionRules` with a type of an dictionary of rule names mapping to the collection rule options type.
- Add `IConfigureOptions<T>` implementation for filling in collection rule options from configuration.
- Add empty `IValidateOptions<T>` implementation for validating individual aspects of a collection rule.
- Add `IOptionsChangeTokenSource<T>` implementation to notify when collection rules change.
- Reuse `EgressOptionsCache<T>` (rename it for generic use) to empty cache when change notification is raised.

Note: this should be similarly implemented compared to `EgressProviderConfigureNamedOptions<T>`, `EgressProviderValidateOptions<T>`, and `EgressOptionsCache<T>`. The custom configuration providers and change token sources are not necessary since there is only one type of collection rules options type, thus explicitly creating a `ConfigurationChangeTokenSource<T>` during service registration time with the `CollectionRules` section passed into it is acceptable.

### LINKED ISSUE #627: Bind process filters on collection rule options
- Add `Filters` property to collection rule options class with a type of an array of `ProcessFilterDescriptor`.
- Update IValidateOptions<T> for the collection rule options to validate the filters.

### LINKED ISSUE #628: Bind trigger options on collection rule options
- Create strongly-typed trigger options class, e.g. `CollectionRuleTriggerOptions`.
  - The `Type` property is of type `string` and is required.
  - The `Settings` property may need to be of type `object` since its schema is dependent on the `Type` property value. This property will need special handling in the IConfigureOptions/IPostConfigureOptions implementation to bind the correct concrete class to the data from configuration. The options type for each named trigger should be available via DI, so this shouldn't be too difficult to get.
- Add `Trigger` property to collection rule options class that is a scalar of the trigger options.
- Update `IValidateOptions<T>` for the collection rule options to validate the trigger options.
  - The `Trigger` property is required..
  - The `Trigger.Type` property is required.
  - The `Trigger.Settings` property is optional, but may be required by the trigger type. If specified, must be validated with the type of settings object that corresponds to `Trigger.Type`. Get `IValidateOptions<T>` instance for the settings type and invoke it.

### LINKED ISSUE #629: Bind action options on collection rule options
- Create strongly-typed action options class, e.g. `CollectionRuleActionOptions`
  - The `Type` property is a string.
  - The `Settings` property is likely of type `object` and will need special binding in the IConfigureOptions/IPostConfigureOptions implementation to bind the settings type from configuration that corresponds to the `Type` property. The options type for each named action should be available via DI, so this shouldn't be too difficult to get.
- Add `Actions` property of type of a array of the actions options.
- Update IValidateOptions for the collection rule options to validate the action rules.
  - The `Actions` property is optional. A rule without actions will still observe with the specified trigger. 
  - The `Actions[<index>].Type` property is required.
  - The `Actions[<index>].Settings` property is optional, but may be required by the action type. If specified, It must be validated with the type of settings object that corresponds to `Type`. Get `IValidateOptions<T>` for the settings type and invoke it.

### LINKED ISSUE #640: Bind limits options on collection rule options
- Create strongly-typed limits options class, e.g. `CollectionRuleLimitsOptions`
  - The `ActionCount` is of type `int?` and describes how many times an action list can be executed before it is throttled.  If not specified, a limit of 5 is used by default to prevent runaway collection rules.
  - The `ActionCountSlidingWindowDuration` property is of type `TimeSpan?` and describes the sliding window of time to consider whether action list execution should be throttled based on the number of times an action list was executed. If specified, the `ActionCount` limit is only considered for the action list execution that happen during that window of time. If not specified, the `ActionCount` limit is considered for the entire lifetime of the rule as it is applied per process.
  - The `RuleDuration` property is of type `TimeSpan?` and describes how long a rule is active after being applied per process. If not specified, the rule has an infinite duration (it executes until the target process exits).
- Add `Limits` property of type of scalar of the limits options class.
- No validation is necessary; all properties are optional scalar values.

jander-msft:
This change defines all of the options classes and properties of all of the aspects of the collection rules / triggers feature. Validation for the collection rule options and action options have been implemented and unit tests were added to verify the options validation for these types.

The validation for the triggers options is partially implemented (I plan to complete that in a subsequent change) but has been completely implemented for the EventCount trigger options.

The unit tests also validate that the custom binding of the trigger and actions options is done correctly since binding those uses a small custom algorithm.

To be done in subsequent PRs:
- Finish validation and unit tests for trigger options.
- Add trigger and action options to JSON schema document.
- Add descriptions for trigger and action option property so they are reflected in JSON schema document.

Partial completions for #630, #631, #632, #633, #634, #635, #648, #649, #656, #658, #659, #660

closes #625
closes #626
closes #627
closes #628
closes #629
closes #640

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/OptionsExtensions.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/OptionsExtensions.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Microsoft.Extensions.Configuration;
 using System;
 using System.Collections;
 using System.Collections.Generic;
@@ -15,13 +16,13 @@ using System.Text;
 using Microsoft.Diagnostics.Tools.Monitor;
 using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options;
 using Microsoft.Diagnostics.Tools.Monitor.Egress.FileSystem;
-using Microsoft.Extensions.Configuration;
 #endif
 
 namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
 {
     internal static class OptionsExtensions
     {
+        private const string EnvironmentVariablePrefix = "DotnetMonitor_";
         private const string KeySegmentSeparator = "__";
 
         /// <summary>
@@ -33,7 +34,7 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
         public static IDictionary<string, string> ToEnvironmentConfiguration(this RootOptions options)
         {
             Dictionary<string, string> variables = new(StringComparer.OrdinalIgnoreCase);
-            MapObject(options, DiagnosticsMonitorCommandHandler.ConfigPrefix, KeySegmentSeparator, variables);
+            MapObject(options, EnvironmentVariablePrefix, KeySegmentSeparator, variables);
             return variables;
         }
 
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/OptionsExtensions.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/OptionsExtensions.cs
@@ -15,12 +15,15 @@ using System.Text;
 using Microsoft.Diagnostics.Tools.Monitor;
 using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options;
 using Microsoft.Diagnostics.Tools.Monitor.Egress.FileSystem;
+using Microsoft.Extensions.Configuration;
 #endif
 
 namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
 {
     internal static class OptionsExtensions
     {
+        private const string KeySegmentSeparator = "__";
+
         /// <summary>
         /// Generates an environment variable map of the options.
         /// </summary>
@@ -30,14 +33,20 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
         public static IDictionary<string, string> ToEnvironmentConfiguration(this RootOptions options)
         {
             Dictionary<string, string> variables = new(StringComparer.OrdinalIgnoreCase);
-            MapObject(options, "DotNetMonitor_", "__", variables);
+            MapObject(options, DiagnosticsMonitorCommandHandler.ConfigPrefix, KeySegmentSeparator, variables);
             return variables;
         }
 
+        /// <summary>
+        /// Generates a map of options that can be passed directly to configuration via an in-memory collection.
+        /// </summary>
+        /// <remarks>
+        /// Each key is the configuration path; each value is the configuration path value.
+        /// </remarks>
         public static IDictionary<string, string> ToConfigurationValues(this RootOptions options)
         {
             Dictionary<string, string> variables = new(StringComparer.OrdinalIgnoreCase);
-            MapObject(options, string.Empty, ":", variables);
+            MapObject(options, string.Empty, ConfigurationPath.KeyDelimiter, variables);
             return variables;
         }
 
@@ -50,7 +59,7 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
         public static IDictionary<string, string> ToKeyPerFileConfiguration(this RootOptions options)
         {
             Dictionary<string, string> variables = new(StringComparer.OrdinalIgnoreCase);
-            MapObject(options, string.Empty, "__", variables);
+            MapObject(options, string.Empty, KeySegmentSeparator, variables);
             return variables;
         }
 
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/OptionsExtensions.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/OptionsExtensions.cs
@@ -11,6 +11,12 @@ using System.Reflection;
 using System.Security.Cryptography;
 using System.Text;
 
+#if !UNITTEST
+using Microsoft.Diagnostics.Tools.Monitor;
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options;
+using Microsoft.Diagnostics.Tools.Monitor.Egress.FileSystem;
+#endif
+
 namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
 {
     internal static class OptionsExtensions
@@ -24,7 +30,14 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
         public static IDictionary<string, string> ToEnvironmentConfiguration(this RootOptions options)
         {
             Dictionary<string, string> variables = new(StringComparer.OrdinalIgnoreCase);
-            MapObject(options, "DotNetMonitor_", variables);
+            MapObject(options, "DotNetMonitor_", "__", variables);
+            return variables;
+        }
+
+        public static IDictionary<string, string> ToConfigurationValues(this RootOptions options)
+        {
+            Dictionary<string, string> variables = new(StringComparer.OrdinalIgnoreCase);
+            MapObject(options, string.Empty, ":", variables);
             return variables;
         }
 
@@ -37,7 +50,7 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
         public static IDictionary<string, string> ToKeyPerFileConfiguration(this RootOptions options)
         {
             Dictionary<string, string> variables = new(StringComparer.OrdinalIgnoreCase);
-            MapObject(options, string.Empty, variables);
+            MapObject(options, string.Empty, "__", variables);
             return variables;
         }
 
@@ -78,7 +91,14 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
             return options;
         }
 
-        private static void MapDictionary(IDictionary dictionary, string prefix, IDictionary<string, string> map)
+        public static CollectionRuleOptions CreateCollectionRule(this RootOptions rootOptions, string name)
+        {
+            CollectionRuleOptions options = new();
+            rootOptions.CollectionRules.Add(name, options);
+            return options;
+        }
+
+        private static void MapDictionary(IDictionary dictionary, string prefix, string separator, IDictionary<string, string> map)
         {
             foreach (var key in dictionary.Keys)
             {
@@ -89,12 +109,13 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
                     MapValue(
                         value,
                         FormattableString.Invariant($"{prefix}{keyString}"),
+                        separator,
                         map);
                 }
             }
         }
 
-        private static void MapList(IList list, string prefix, IDictionary<string, string> map)
+        private static void MapList(IList list, string prefix, string separator, IDictionary<string, string> map)
         {
             for (int index = 0; index < list.Count; index++)
             {
@@ -104,12 +125,13 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
                     MapValue(
                         value,
                         FormattableString.Invariant($"{prefix}{index}"),
+                        separator,
                         map);
                 }
             }
         }
 
-        private static void MapObject(object obj, string prefix, IDictionary<string, string> map)
+        private static void MapObject(object obj, string prefix, string separator, IDictionary<string, string> map)
         {
             foreach (PropertyInfo property in obj.GetType().GetProperties(BindingFlags.Public | BindingFlags.Instance))
             {
@@ -118,17 +140,21 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
                     MapValue(
                         property.GetValue(obj),
                         FormattableString.Invariant($"{prefix}{property.Name}"),
+                        separator,
                         map);
                 }
             }
         }
 
-        private static void MapValue(object value, string valueName, IDictionary<string, string> map)
+        private static void MapValue(object value, string valueName, string separator, IDictionary<string, string> map)
         {
             if (null != value)
             {
                 Type valueType = value.GetType();
-                if (valueType.IsPrimitive || typeof(string) == valueType)
+                if (valueType.IsPrimitive ||
+                    valueType.IsEnum ||
+                    typeof(string) == valueType ||
+                    typeof(TimeSpan) == valueType)
                 {
                     map.Add(
                         valueName,
@@ -136,18 +162,18 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
                 }
                 else
                 {
-                    string prefix = FormattableString.Invariant($"{valueName}__");
+                    string prefix = FormattableString.Invariant($"{valueName}{separator}");
                     if (value is IDictionary dictionary)
                     {
-                        MapDictionary(dictionary, prefix, map);
+                        MapDictionary(dictionary, prefix, separator, map);
                     }
                     else if (value is IList list)
                     {
-                        MapList(list, prefix, map);
+                        MapList(list, prefix, separator, map);
                     }
                     else
                     {
-                        MapObject(value, prefix, map);
+                        MapObject(value, prefix, separator, map);
                     }
                 }
             }
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
@@ -65,7 +65,7 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
                 rootOptions =>
                 {
                     rootOptions.CreateCollectionRule(DefaultRuleName)
-                        .AddExecuteAction("echo");
+                        .AddExecuteAction("cmd.exe");
                 },
                 ex =>
                 {
@@ -666,7 +666,7 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
         [Fact]
         public Task CollectionRuleOptions_ExecuteAction_MinimumOptions()
         {
-            const string ExpectedExePath = "echo";
+            const string ExpectedExePath = "cmd.exe";
 
             return ValidateSuccess(
                 rootOptions =>
@@ -685,8 +685,8 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
         [Fact]
         public Task CollectionRuleOptions_ExecuteAction_RoundTrip()
         {
-            const string ExpectedExePath = "echo";
-            const string ExpectedArguments = "\"Hello\"";
+            const string ExpectedExePath = "cmd.exe";
+            const string ExpectedArguments = "/c \"echo Hello\"";
 
             return ValidateSuccess(
                 rootOptions =>
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
@@ -21,6 +21,7 @@ using Xunit.Abstractions;
 using Microsoft.Diagnostics.Monitoring.WebApi.Models;
 using Microsoft.Extensions.Logging;
 using System.Diagnostics.Tracing;
+using System.ComponentModel.DataAnnotations;
 
 namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
 {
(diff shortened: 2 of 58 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Options/CollectionRuleOptionsExtensions.cs
```diff
@@ -0,0 +1,270 @@
...
+        public static CollectTraceOptions VerifyCollectTraceAction(this CollectionRuleOptions ruleOptions, int actionIndex, IEnumerable<EventPipeProvider> providers, string expectedEgress)
+        {
+            CollectTraceOptions collectTraceOptions = ruleOptions.VerifyAction<CollectTraceOptions>(
+                actionIndex, KnownCollectionRuleActions.CollectTrace);
+
+            Assert.Equal(expectedEgress, collectTraceOptions.Egress);
+            Assert.NotNull(collectTraceOptions.Providers);
+            Assert.Equal(providers.Count(), collectTraceOptions.Providers.Count);
+
+            int index = 0;
```
kkeirstead, line 224: It looks like this isn't being incremented (and I'm assuming it should be...?)

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/CollectTraceOptions.Validate.cs
```diff
@@ -0,0 +1,70 @@
...
+            List<ValidationResult> results = new();
+
+            bool hasProfile = Profile.HasValue;
+            bool hasProviders = null != Providers && Providers.Any();
+
+            if (hasProfile)
+            {
+                if (hasProviders)
+                {
+                    // Both Profile and Providers cannot specified at the same time, otherwise
```
kkeirstead, line 31: nit: 'cannot specified' -> 'cannot be specified'

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/ExecuteOptions.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Linq;
+
+#if UNITTEST
+namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
+#else
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Actions
+#endif
+{
+    /// <summary>
+    /// Options for the Execute  action.
```
kkeirstead, line 17:
```suggestion
    /// Options for the Execute action.
```

src/Tools/dotnet-monitor/CollectionRules/Options/Triggers/EventCounterOptions.Validate.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Triggers
+{
+    partial class EventCounterOptions : IValidatableObject
+    {
+        IEnumerable<ValidationResult> IValidatableObject.Validate(ValidationContext validationContext)
+        {
+            List<ValidationResult> results = new();
+
+            if (!GreaterThan.HasValue && !LessThan.HasValue)
```
kkeirstead, line 16: It looks like GreaterThan and LessThan are both just doubles; should we be enforcing non-negative?

jander-msft: I don't think we have a need to enforce that. While I haven't seen any of the counters report negative numbers, I don't think there is any reason why they couldn't.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+      <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <Compile Include="..\Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests\Options\OptionsExtensions.cs" Link="Options\OptionsExtensions.cs" />
```
wiktork, line 8: Would it make sense to move OptionsExtensions to TestCommon?

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRuleConfigureNamedOptions.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            _actionOptionsProvider = actionOptionsProvider;
+            _collectionRulesSection = configuration.GetSection(nameof(ConfigurationKeys.CollectionRules));
+            _triggerOptionsProvider = triggerOptionsProvider;
+        }
+
+        public void Configure(string name, CollectionRuleOptions options)
+        {
+            IConfigurationSection ruleSection = _collectionRulesSection.GetSection(name);
+            Debug.Assert(ruleSection.Exists());
```
wiktork, line 34: Is this assert necessary?

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRuleTriggerOptionsProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public CollectionRuleTriggerOptionsProvider(IEnumerable<ICollectionRuleTriggerProvider> triggerProviders)
+        {
+            _triggerProviders = triggerProviders;
+        }
+
+        public bool TryGetOptionsType(string triggerType, out Type optionsType)
+        {
+            ICollectionRuleTriggerProvider triggerProvider = _triggerProviders
+                .FirstOrDefault(provider => string.Equals(triggerType, provider.TriggerType, StringComparison.Ordinal));
```
wiktork, line 24: is name uniqueness here as well as actiontype enforced at validation time?

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/ActionOptionsConstants.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
+#else
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Actions
+#endif
+{
+    // Constants for action options allowing reuse among multiple actions and for tests to verify ranges.
+    internal static class ActionOptionsConstants
+    {
+        public const int BufferSizeMegabytes_MaxValue = 1024;
+        public static readonly string BufferSizeMegabytes_MaxValue_String = BufferSizeMegabytes_MaxValue.ToString();
```
wiktork, line 15: nit: ToString(CultureInfo.InvariantCulture)

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests.csproj
jander-msft: It might, but none of the options classes are in that project. I'd probably have to move the links to those classes to that project.

In the long term, I'm almost thinking that everything should be refactored into more granular libraries and then only reference via project reference is allowed, because this link of source files is starting to get difficult to juggle.

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/CollectLogsOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        [EnumDataType(typeof(LogLevel))]
+        public LogLevel? LogLevel { get; set; }
+
+        public Dictionary<string, LogLevel?> FilterSpecs { get; set; }
+
+        public bool? UseAppFilters { get; set; }
+
+        [Range(typeof(TimeSpan), ActionOptionsConstants.Duration_MinValue, ActionOptionsConstants.Duration_MaxValue)]
+        public TimeSpan? Duration { get; set; }
```
wiktork, line 31: Do you think there's any value in making some of the more common action fields inherit from a base class (I'm assuming configuration can handle this). For example, actions that have egress and duration can share from a base option.

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRuleTriggerOptionsProvider.cs
jander-msft: I think I need to preprocess the list in the ctor, store the information as a dictionary for easier lookup, and log warnings for duplicates. Duplicates is basically not going to happen unless this opens up for extensibility in the future.

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/ExecuteOptions.cs
```diff
@@ -0,0 +1,32 @@
...
+    internal sealed class ExecuteOptions : IValidatableObject
+    {
+        [Required]
+        public string Path { get; set; }
+
+        public string Arguments { get; set; }
+
+        public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
+        {
+            return Enumerable.Empty<ValidationResult>();
```
wiktork, line 29: TODO?

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/CollectLogsOptions.cs
jander-msft: Configuration will handle it, however the schema.json file will largely not be usable for IntelliSense for the fields that are not on the deriving class. I'll try it again and check if that is still the case or if I messed something up.

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleLimitsOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_CollectionRuleLimitsOptions_ActionCount))]
+        public int? ActionCount { get; set; }
+
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_CollectionRuleLimitsOptions_ActionCountSlidingWindowDuration))]
+        [Range(typeof(TimeSpan), "00:00:01", "1.00:00:00")]
```
wiktork, line 28: nit: constants (which I believe were defined elsewhere)

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRuleConfigureNamedOptions.cs
jander-msft: It's helpful when debugging that the name you provided couldn't be found in the configuration.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Options/CollectionRuleOptionsExtensions.cs
jander-msft: Fixed

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/OptionsExtensions.cs
```diff
@@ -24,7 +30,14 @@ internal static class OptionsExtensions
...
             Dictionary<string, string> variables = new(StringComparer.OrdinalIgnoreCase);
-            MapObject(options, "DotNetMonitor_", variables);
+            MapObject(options, "DotNetMonitor_", "__", variables);
+            return variables;
+        }
+
+        public static IDictionary<string, string> ToConfigurationValues(this RootOptions options)
+        {
+            Dictionary<string, string> variables = new(StringComparer.OrdinalIgnoreCase);
+            MapObject(options, string.Empty, ":", variables);
```
wiktork, line 40: ConfigurationPath.KeyDelimiter

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/OptionsExtensions.cs
```diff
@@ -24,7 +30,14 @@ internal static class OptionsExtensions
         public static IDictionary<string, string> ToEnvironmentConfiguration(this RootOptions options)
         {
             Dictionary<string, string> variables = new(StringComparer.OrdinalIgnoreCase);
-            MapObject(options, "DotNetMonitor_", variables);
+            MapObject(options, "DotNetMonitor_", "__", variables);
```
wiktork, line 33: nit: const for "__"

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
```diff
@@ -0,0 +1,884 @@
...
+                    Assert.Single(failures);
+                    VerifyRequiredMessage(failures, 0, nameof(ExecuteOptions.Path));
+                });
+        }
+
+        private async Task Validate(
+            Action<RootOptions> setup,
+            Action<IOptionsMonitor<CollectionRuleOptions>> validate)
+        {
+            using IHost host = new HostBuilder()
```
wiktork, line 726: Should the host have a using? you are disposing/async disposing below

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
```diff
@@ -0,0 +1,884 @@
...
+                fieldName);
+
+            Assert.Equal(message, failures[index]);
+        }
+
+        private static void VerifyRangeMessage(string[] failures, int index, string fieldName, string min, string max)
+        {
+            string message = string.Format(
+                CultureInfo.InvariantCulture,
+                "The field {0} must be between {1} and {2}.",
```
wiktork, line 833: new RangeValidationAttribute().ErrorMessage or something simliar?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests.csproj
wiktork: If we do decide on this refactor I would do it in the main branch first.

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/CollectLogsOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+#endif
+{
+    /// <summary>
+    /// Options for the CollectLogs action.
+    /// </summary>
+    [DebuggerDisplay("CollectLogs")]
+    internal sealed partial class CollectLogsOptions
+    {
+        [EnumDataType(typeof(LogLevel))]
+        public LogLevel? LogLevel { get; set; }
```
wiktork, line 24: Will these be getting DisplayAttributes as well?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
```diff
@@ -0,0 +1,874 @@
...
+        }
+
+        [Fact]
+        public Task CollectionRuleOptions_NoTrigger()
+        {
+            return ValidateFailure(
+                rootOptions =>
+                {
+                    rootOptions.CreateCollectionRule(DefaultRuleName)
+                        .AddExecuteAction("echo");
```
wiktork, line 68: Sounds like we'll want a UseShellExecute settings on the Execute Action.

wiktork (approved):

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/CollectLogsOptions.cs
jander-msft: Yup, in subsequent change when updating the schema with the action and trigger specific options.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests.csproj
jander-msft: > If we do decide on this refactor I would do it in the main branch first.

https://github.com/dotnet/dotnet-monitor/pull/692

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Options/CollectionRuleOptionsExtensions.cs
```diff
@@ -0,0 +1,272 @@
...
+            actionOptions.Type = type;
+
+            options.Actions.Add(actionOptions);
+
+            return options;
+        }
+
+        public static CollectionRuleOptions AddCollectDumpAction(this CollectionRuleOptions options, DumpType type, string egress)
+        {
+            options.AddAction(KnownCollectionRuleActions.CollectDump, out CollectionRuleActionOptions actionOptions);
```
kkeirstead, line 35: Might not be worth calling out, but I noticed it read a bit funny to me so I thought I'd bring it up - we disregard the return value of `AddAction` and just rely on the fact that `options` was manipulated in `AddAction` by being passed by reference (at least that's my best guess). However, in other methods (such as `AddCollectLogsAction` on line 58), we actually do use the returned `options` value. If there's a functional reason for this behavior let me know; I just thought it was interesting that we sometimes chose to use the return value and other times discarded it and used the variable that had been changed by reference.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
jander-msft: I would prefer not to allow UseShellExecute in order to be able to capture stdout and stderr. I'll update the test examples to use cmd.exe instead of just calling "echo".

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Options/CollectionRuleOptionsExtensions.cs
jander-msft: The return type of the `AddAction` and `Add*Action` methods is `CollectionRuleOptions`, which is used to facilitate a builder pattern. The out parameter on these methods (if they have one) is the type of the action options that were added. The return type and the type of the out parameter are distinct.

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/CollectLogsOptions.cs
jander-msft: I'll take a look at this refactor when I go to add the action options to the JSON schema.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Options/CollectionRuleOptionsExtensions.cs
kkeirstead: I actually wasn't referring to the out parameter, so maybe I just completely misunderstood what I was looking at. When we call `options.AddAction(...)`, my understanding is that `options` gets updated in the method (for example, `options.Actions.Add(actionOptions)`. Instead of saying `options = options.AddAction(...)`, we don't use the returned value. However, AddCollectLogsAction and AddCollectTraceAction both use the returned `CollectionRuleOptions` in their return statements. Since `options` is passed by reference, functionally, I don't think it makes a difference; I was just wondering why we returned a value if we typically didn't use it.

jander-msft: That's the builder pattern.

For example:

```cs
rootOptions.CreateCollectionRule(DefaultRuleName)
    .SetStartupTrigger()
    .AddCollectDumpAction(ExpectedDumpType, ExpectedEgressProvider);
```

can also be written:

```cs
rootOptions.CreateCollectionRule(DefaultRuleName)
    .AddCollectDumpAction(ExpectedDumpType, ExpectedEgressProvider)
    .SetStartupTrigger();
```

I don't want this code to be opinionated on the order in which the rule is constructed, so all of the Add/Set methods return the CollectionRuleOptions to allow you to build it in any order.

PR #692, opened 2021-08-07
Move all options to new assembly instead of linking source files.

jander-msft:
These changes move all of the options classes used for binding configuration into a new assembly to reduce the amount of source files linked into multiple projects and reduce if-def'ing namespaces. Now all projects an consume the options classes in a uniform manner.

This change largely leaves the namespaces intact. A subsequent change will be made to reorganize the options into better and more consistent namespaces.


src/Microsoft.Diagnostics.Monitoring.Options/Microsoft.Diagnostics.Monitoring.Options.csproj
```diff
@@ -0,0 +1,38 @@
...
+    <InternalsVisibleTo Include="Microsoft.Diagnostics.Monitoring.ConfigurationSchema" />
+    <InternalsVisibleTo Include="Microsoft.Diagnostics.Monitoring.OpenApiGen" />
+    <InternalsVisibleTo Include="Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests" />
+    <InternalsVisibleTo Include="Microsoft.Diagnostics.Monitoring.WebApi" />
+    <InternalsVisibleTo Include="Microsoft.Diagnostics.Monitoring.WebApi.UnitTests" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.ComponentModel.Annotations" Version="5.0.0" />
+    <PackageReference Include="System.Text.Json" Version="4.7.2" />
```
wiktork, line 20: nit: push to variables

wiktork, line 20: Any reason not to take 5.* version here btw?

wiktork (approved):

src/Microsoft.Diagnostics.Monitoring.Options/Microsoft.Diagnostics.Monitoring.Options.csproj
jander-msft: There's a version downgrade somewhere if 5.* is used (I can't remember off the top of my head).

jander-msft: The netstandard2.0 version of the WebApi is referencing System.Text.Encodings.Web 4.7.2; System.Text.Json 5.0.0 wants System.Text.Encodigns.Web of at least 5.0.0.

I've now removed the dependency on System.Text.Json from the options assembly.

jander-msft:
With the latest change, I've moved all of the logging options to the schema generator project. If we feel that we need those options for the functional tests, then we can create another assembly (e.g. Microsoft.Diagnostics.Monitoring.Options.Schema) that encompasses the entire usable options surface and then have the functional tests and schema generator reference that assembly.

PR #700, opened 2021-08-10
Simplify model class namespaces.

jander-msft:
Since the functional tests project already has a project-to-project reference to the web api project, there is no need to include links to the model classes in the web api project and the namespaces do not need to be conditional anymore.


kkeirstead (approved):
From what I can tell, what's here seems to look good (and tests are passing). I can Approve this, but if you want to get another set of eyes on it before merging maybe @kelltrick can look over it as well.

hoyosjs (approved):

PR #704, opened 2021-08-10
Refactor schema gen to generically create discriminated subschemas

jander-msft:
This change allows adding more discriminated subschemas in the future without reimplementing the concept for each set of options. Also, generate schemas from type information and cache it in a resolver so that type schemas can be reused across multiple type generations.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
@@ -98,9 +98,9 @@ namespace Microsoft.Diagnostics.Monitoring.ConfigurationSchema
 
         private static JsonSchemaProperty AddDiscriminatedSubSchema(
             JsonSchema parentSchema,
-            string descriminatingPropertyName,
-            string descriminatingPropertyValue,
-            string descriminatedPropertyName,
+            string discriminatingPropertyName,
+            string discriminatingPropertyValue,
+            string discriminatedPropertyName,
             JsonSchema subSchema = null)
         {
             if (null == subSchema)
@@ -110,13 +110,13 @@ namespace Microsoft.Diagnostics.Monitoring.ConfigurationSchema
 
             JsonSchemaProperty descriminatingProperty = new JsonSchemaProperty();
             descriminatingProperty.ExtensionData = new Dictionary<string, object>();
-            descriminatingProperty.ExtensionData.Add("const", descriminatingPropertyValue);
+            descriminatingProperty.ExtensionData.Add("const", discriminatingPropertyValue);
 
-            subSchema.Properties.Add(descriminatingPropertyName, descriminatingProperty);
+            subSchema.Properties.Add(discriminatingPropertyName, descriminatingProperty);
 
             JsonSchemaProperty descriminatedProperty = new JsonSchemaProperty();
 
-            subSchema.Properties.Add(descriminatedPropertyName, descriminatedProperty);
+            subSchema.Properties.Add(discriminatedPropertyName, descriminatedProperty);
 
             parentSchema.OneOf.Add(subSchema);
 
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
@@ -17,26 +17,17 @@ namespace Microsoft.Diagnostics.Monitoring.ConfigurationSchema
     {
         public string GenerateSchema()
         {
-            var settings = new JsonSchemaGeneratorSettings();
+            var schema = new JsonSchema();
+            var context = new GenerationContext(schema);
+            context.SetRoot<RootOptions>();
 
-            settings.SerializerSettings = new JsonSerializerSettings();
-            settings.SerializerSettings.Converters.Add(new StringEnumConverter());
-
-            JsonSchema schema = JsonSchema.FromType<RootOptions>(settings);
             schema.Id = @"https://www.github.com/dotnet/dotnet-monitor";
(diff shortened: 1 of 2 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
```diff
@@ -57,64 +48,111 @@ public string GenerateSchema()
...
+
+            public GenerationContext(JsonSchema rootSchema)
+            {
+                Schema = rootSchema;
+
+                _settings = new JsonSchemaGeneratorSettings();
+                _settings.SerializerSettings = new JsonSerializerSettings();
+                _settings.SerializerSettings.Converters.Add(new StringEnumConverter());
+
+                _resolver = new JsonSchemaResolver(rootSchema, _settings);
```
kkeirstead, line 140: Supposedly JsonSchemaResolver is now considered obsolete; does this change anything?

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
```diff
@@ -57,64 +48,111 @@ public string GenerateSchema()
...
             consoleLoggerOptionsSchema.Properties.Add(nameof(ConsoleLoggerOptions.FormatterName), formatterNameProperty);
             consoleLoggerOptionsSchema.Properties.Add(nameof(ConsoleLoggerOptions.FormatterOptions), formatterOptionsProperty);
 
-            return consoleLoggerOptionsSchema;
+            context.Schema.Definitions[nameof(ConsoleLoggerOptions)].OneOf.Add(consoleLoggerOptionsSchema);
+        }
+
+        private static JsonSchemaProperty AddDiscriminatedSubSchema(
+            JsonSchema parentSchema,
+            string descriminatingPropertyName,
```
kkeirstead, line 101: I'm assuming these names should be `discriminating` instead of `descriminating`

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
jander-msft: Where do you see that it is obsolete? There's no Obsolete attribute on it, no compiler warnings/errors, and I haven't seen that in documentation.

kkeirstead: Maybe I'm looking at the wrong thing? This is what I found when I searched: https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_Schema_JsonSchemaResolver.htm (it says at the top `Note: This API is now obsolete`).

jander-msft: > Maybe I'm looking at the wrong thing? This is what I found when I searched: https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_Schema_JsonSchemaResolver.htm (it says at the top `Note: This API is now obsolete`).

The schema generation uses [NJsonSchema](https://github.com/RicoSuter/NJsonSchema), not Newtonsoft.Json. While NJsonSchema uses NewtonsoftJson as a dependency, this schema generation uses the NJsonSchema API.

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
```diff
@@ -57,64 +48,111 @@ public string GenerateSchema()
...
+        private static JsonSchemaProperty AddDiscriminatedSubSchema(
+            JsonSchema parentSchema,
+            string discriminatingPropertyName,
+            string discriminatingPropertyValue,
+            string discriminatedPropertyName,
+            JsonSchema subSchema = null)
+        {
+            if (null == subSchema)
+            {
+                subSchema = new JsonSchema();
```
kkeirstead, line 108: Not sure if this is actually a problem, but I noticed that you separated `FormatterName` as a RequiredProperty (line 69) from where you add `discriminatedPropertyName` as a property down below. In theory, if `subSchema` is null, then when we initialize `subSchema` we won't have `FormatterName` as a required property. Do you think this is a problem? Should we add `subSchema.RequiredProperties.Add(nameof(ConsoleLoggerOptions.FormatterName))` on line 120?

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
```diff
@@ -57,64 +48,111 @@ public string GenerateSchema()
...
+            }
+
+            JsonSchemaProperty descriminatingProperty = new JsonSchemaProperty();
```
kkeirstead, line 111: nit: Looks like you missed this `descriminating` (and the one on line 117)

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
jander-msft: I don't think that's going to be an issue, in fact, the "required" array on the sub schemas in this case might not actually be needed at all (for any of the given subschemas, if FormatterName is "const": "<somevalue>", then by virtue of the fact that it has a value then the requiredness of the property is fulfilled implicitly).

There will be other schemas where the properties in the parent schema are already defined to be required. It _might_ not be necessary to redeclare that those properties in the subschema are again required. My empirical testing has not shown that this is necessary.

kkeirstead (approved):
I think this is everything I noticed when going over the PR, so I can Approve this, but if you want to get another set of eyes on it before merging maybe @kelltrick can look over it as well.

PR #706, opened 2021-08-11
Generate action and trigger options in schema.

jander-msft:
- Generate action and trigger options in schema.
- Add display name descriptions to these options.
- Add defaults for the optional properties.
- Other small changes such as property renames and changing IncludePaths and ExcludePaths property to arrays.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
+++ b/src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
@@ -126,11 +126,11 @@
     <comment>The description provided for the ApiKeyHashType parameter on ApiAuthenticationOptions.</comment>
   </data>
   <data name="DisplayAttributeDescription_AspNetRequestCountOptions_ExcludePaths" xml:space="preserve">
-    <value>The list of request paths to ignore.</value>
+    <value>The list of request paths to ignore. A request path must exactly match one of the items in the list to be ignored.</value>
     <comment>The description provided for the ExcludePaths parameter on AspNetRequestCountOptions.</comment>
   </data>
   <data name="DisplayAttributeDescription_AspNetRequestCountOptions_IncludePaths" xml:space="preserve">
-    <value>The list of request paths to monitor. If not specified, all request paths are monitored. Request paths matching the ExcludePaths list will not be monitored.</value>
+    <value>The list of request paths to monitor. If not specified, all request paths are monitored.  A request path must exactly match one of the items in the list to be monitored. Request paths matching the ExcludePaths list will not be monitored.</value>
     <comment>The description provided for the IncludePaths parameter on AspNetRequestCountOptions.</comment>
   </data>
   <data name="DisplayAttributeDescription_AspNetRequestCountOptions_RequestCount" xml:space="preserve">
@@ -142,11 +142,11 @@
     <comment>The description provided for the SlidingWindowDuration parameter on AspNetRequestCountOptions.</comment>
   </data>
   <data name="DisplayAttributeDescription_AspNetRequestDurationOptions_ExcludePaths" xml:space="preserve">
-    <value>The list of request paths to ignore.</value>
+    <value>The list of request paths to ignore. A request path must exactly match one of the items in the list to be ignored.</value>
     <comment>The description provided for the ExcludePaths parameter on AspNetRequestDurationOptions.</comment>
   </data>
   <data name="DisplayAttributeDescription_AspNetRequestDurationOptions_IncludePaths" xml:space="preserve">
-    <value>The list of request paths to monitor. If not specified, all request paths are monitored. Request paths matching the ExcludePaths list will not be monitored.</value>
+    <value>The list of request paths to monitor. If not specified, all request paths are monitored. A request path must exactly match one of the items in the list to be monitored. Request paths matching the ExcludePaths list will not be monitored.</value>
     <comment>The description provided for the IncludePaths parameter on AspNetRequestDurationOptions.</comment>
   </data>
   <data name="DisplayAttributeDescription_AspNetRequestDurationOptions_RequestCount" xml:space="preserve">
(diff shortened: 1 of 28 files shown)


src/Tools/dotnet-monitor/CollectionRules/Options/Triggers/AspNetRequestCountOptionsDefaults.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Triggers
+{
+    internal static class AspNetRequestCountOptionsDefaults
+    {
+        public const string SlidingWindowDuration = "00:01:00";
```
kkeirstead, line 9: I noticed that we define this constant with the same value in four different places; would it make sense to have this as a shared default?

src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
```diff
@@ -125,6 +125,62 @@
     <value>Hash algorithm used to compute ApiKeyHash, typically 'SHA256'. 'SHA1' and 'MD5' are not allowed.</value>
     <comment>The description provided for the ApiKeyHashType parameter on ApiAuthenticationOptions.</comment>
   </data>
+  <data name="DisplayAttributeDescription_AspNetRequestCountOptions_ExcludePaths" xml:space="preserve">
+    <value>The list of request paths to ignore.</value>
+    <comment>The description provided for the ExcludePaths parameter on AspNetRequestCountOptions.</comment>
+  </data>
+  <data name="DisplayAttributeDescription_AspNetRequestCountOptions_IncludePaths" xml:space="preserve">
```
kelltrick, line 132: Does this support wild carding like `/path/to/*` or does `/path/to/` match all sub paths?, it might be worth it to make this callout in the description.

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/CollectTraceOptionsDefaults.cs
```diff
@@ -0,0 +1,14 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Actions
+{
+    internal static class CollectTraceOptionsDefaults
+    {
+        public const int MetricsIntervalSeconds = 1;
+        public const bool RequestRundown = true;
+        public const int BufferSizeMegabytes = 256;
```
kelltrick, line 11: This seems like a big constant given our recommended container limits. Does this buffer not end up in our memory?

src/Tools/dotnet-monitor/CollectionRules/Options/Triggers/AspNetResponseStatusOptions.cs
```diff
@@ -12,17 +14,34 @@ namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Triggers
     /// </summary>
     internal sealed class AspNetResponseStatusOptions
     {
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_AspNetResponseStatusOptions_StatusCodes))]
         [Required]
-        public string StatusCodes { get; set; }
+        [MinLength(1)]
```
kelltrick, line 21: Does it make sense to have path filters as a generic across all `AspNet*Options` triggers?

Maybe I'm thinking it would be worth it to have `AspNetPathFilters` which holds `IncludePaths` and `ExcludePaths` and that filter can be a trigger just by itself and can be included in each of these options classes.

kelltrick (approved):
👍

src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
jander-msft: The spec currently does not call this out either way.

The initial implementation should match the path exactly (give or take the trailing slash). Then we can add wildcards later.

We'll need to discuss the wildcard system before that's implemented or a library is pulled in. I'd like to see wildcards, but we'll want to keep it relatively simple rather going full regex.

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/CollectTraceOptionsDefaults.cs
jander-msft: This is the existing default when doing a trace, but I agree. I think we need to adjust these downward a bit.

The buffer ends up in the memory of the target application. It's the size of the buffer that events are put into just before the diagnostic server pulls them out and sends them over the event pipe. If an event source cannot put a new event into the buffer due to running out of room, the event is dropped relatively silently. This makes code that is listening for a start and stop event interesting because it has to guess whether the stop event just hasn't come through yet or was dropped.

src/Tools/dotnet-monitor/CollectionRules/Options/Triggers/AspNetResponseStatusOptions.cs
jander-msft: Nesting of options like that comes at making the options configuration more complex. For example, the request count trigger might end up looking like:

```json
{
  "RequestCount": 100,
  "PathFilters": {
    "Include": [..],
    "Exclude": [..]
  }
}

```

I think the reusability of the options is somewhat low and replicating these two properties across the options is not too difficult to do or maintain. I think the complexity of specifying this type of structure outweighs the reusability or compactness we get from an engineering perspective.

Another option is to have a base class that contains these properties, have all of the ASP.NET triggers derive from it, and then flatten the base schema into the deriving schema when generating the schema.

At the very least, would could add these onto a common interface that allows for some reusable utility code for matching the path to the filters.

src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
jander-msft: I'll update the description to say (for now) that the path must match exactly.

src/Tools/dotnet-monitor/CollectionRules/Options/Triggers/AspNetRequestCountOptionsDefaults.cs
jander-msft: I'll add another shared constant that these constants can reference. I'd like to maintain the aspects of each trigger and action separately as much as possible, but not at the expense of fixing an issue multiple times, as you pointed out.

src/Tools/dotnet-monitor/CollectionRules/Options/Triggers/AspNetResponseStatusOptions.cs
jander-msft: I implemented the interface for now. We can review the configuration surface area later and determine if we want to do this type of nesting of settings.

PR #710, opened 2021-08-11
Create `Execute` collection rule action 

### LINKED ISSUE #633: Create `Execute` collection rule action
- Create an action named `Execute` that will invoke an executable with arguments.
  - This action can use output values from prior actions as well as process information as variable replacement in the `Arguments` property to allow more fine-grain knowledge of prior action results (e.g. collect action egresses an artifact an outputs the location of the artifact; this location can be passed to the executable).
- Create a strongly-typed options class, e.g. `ExecuteActionOptions`, for the execute action with the following properties:
  - The `Path` property is of type `string` and represents the path of the executable to execute.
  - The `Arguments` is of type `string` and represents the arguments to pass to the executable
- Create and register an `IValidateOptions<T>` instance for the execute action options.
  - The `Path` property is required. Consider validating that the executable does exist (may not be possible if variable replacement is allowed in the `Path` property).

kkeirstead:
Since Justin already added in ExecuteOptions, this is mostly focused around IValidateOptions, an interface for actions, and the Execute action.

Closes #633


src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
```diff
@@ -0,0 +1,33 @@
...
+            ActionResponse executeResponse = new();
+
+            await Task.Run(() =>
+            {
+                string path = options.Path;
+                string arguments = options.Arguments;
+
+                Process process = Process.Start(path, arguments);
+
+                process.WaitForExit();
```
wiktork, line 22: I would do something like the implementation for AwaitExtensions.WaitForExitAsync, rather than blocking an entire tpl thread.

wiktork, line 22: Do we have a timeout we can plugin to the action?

src/Tools/dotnet-monitor/CollectionRules/ExecuteActionValidateOptions.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Actions;
+using Microsoft.Extensions.Options;
+using System.IO;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules
+{
+    internal sealed class ExecuteActionValidateOptions : IValidateOptions<ExecuteOptions>
```
jander-msft, line 7: This validator is not necessary as it is already covered by the Required attribute on the ExecuteOptions.Path property + the DataAnnotationValidatOptions implementation registered for these options.

jander-msft: Also, the path existence should probably be checked in the IAction<T>.Execute method so that configuration can pass validation when its loaded.

src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules
+{
+    internal sealed class ExecuteAction : IAction<ExecuteOptions>
+    {
+        public async Task<ActionResponse> Execute(ExecuteOptions options, CancellationToken token)
+        {
+            ActionResponse executeResponse = new();
+
+            await Task.Run(() =>
```
jander-msft, line 15: Wrapping this with Task.Run and immediately awaiting it only tosses the work to a thread pool thread. You might as well just do all of the work in the current thread and await for the process exit.

src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
```diff
@@ -0,0 +1,33 @@
...
+            {
+                string path = options.Path;
+                string arguments = options.Arguments;
+
+                Process process = Process.Start(path, arguments);
+
+                process.WaitForExit();
+
+                int exitCode = process.ExitCode;
```
jander-msft, line 24: A non-zero exit code should fail the action. Maybe we need another property (e.g. `IgnoreExitCode`) on ExecuteOptions class to state that the exit code should be ignored, but that behavior should be opt-in.

src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
```diff
@@ -0,0 +1,33 @@
...
+        public async Task<ActionResponse> Execute(ExecuteOptions options, CancellationToken token)
+        {
+            ActionResponse executeResponse = new();
+
+            await Task.Run(() =>
+            {
+                string path = options.Path;
+                string arguments = options.Arguments;
+
+                Process process = Process.Start(path, arguments);
```
jander-msft, line 20: We might want to capture stdout and stderr and possibly return those in the result, but I don't have a good scenario in mind where we need this. So maybe don't do this and just add a consideration comment in the code to that effect.

src/Tools/dotnet-monitor/CollectionRules/IAction.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules
+{
+    internal interface IAction<TOptions>
```
jander-msft, line 13:
```suggestion
    internal interface ICollectionRuleAction<TOptions>
```

to be more specific of which system this action belongs.

src/Tools/dotnet-monitor/CollectionRules/IAction.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        Task<ActionResponse> Execute(TOptions options, CancellationToken token);
```
jander-msft, line 15:
```suggestion
        Task<ActionResponse> ExecuteAsync(TOptions options, CancellationToken cancelltionToken);
```

src/Tools/dotnet-monitor/CollectionRules/IAction.cs
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    internal class ActionResponse
```
jander-msft, line 18: nit:

```suggestion
    internal struct ActionResult
```

Regarding rename: When I think of responses, I also think of there being a request. But there is no structural notion of a request.

Regarding class vs struct: I don't think anything is ever going to derive from this type and this type is likely only going to be used in locals (never saved off to the heap), so it's probably better for this to be a struct. @wiktork, what do you think?

src/Tools/dotnet-monitor/CollectionRules/IAction.cs
jander-msft, line 15: I think we're going to need to pass some notion of what the target process is to the action, especially for the collect actions. Maybe a `DiagnosticsClient client` parameter (or maybe some lower level notion that represents the connection to the target process).

src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
jander-msft: We have test code that does this already that you can copy and adapt to this: https://github.com/dotnet/dotnet-monitor/blob/cf94ef78c80aca5f19933add98150e14d5fd6d50/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/DotNetRunner.cs#L162

jander-msft (changes requested):

src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
kkeirstead: When you say "should fail the action," what did you have in mind? In the current state, the `ExitCode` could be processed by the caller, but if we don't want to enforce that responsibility on it, would we want to throw from `ExecuteAsync`? Or something else entirely? And in the event that a user has set `IgnoreExitCode` and another action is dependent on the result of the current action, what behavior should we expect?

src/Tools/dotnet-monitor/CollectionRules/ExecuteActionValidateOptions.cs
kkeirstead: What do you mean by "so that configuration can pass validation when its loaded?" My understanding of what you want is just to perform a check upon entering `ExecuteAsync` (and then failing the action if there's nothing to execute for the path), but I just want to make sure I'm understanding what you're asking for.

src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
jander-msft: > When you say "should fail the action," what did you have in mind?

The action should throw an exception with a specific message e.g. `The process exited with exit code {0}`.

> In the current state, the `ExitCode` could be processed by the caller, but if we don't want to enforce that responsibility on it, would we want to throw from `ExecuteAsync`?

The caller should not have an specific knowledge of any of the properties in the result.

> And in the event that a user has set `IgnoreExitCode` and another action is dependent on the result of the current action, what behavior should we expect?

I would expect that it still fills out the same properties in the result, but they might be null, string.empty, or default(T) depending on the type. So if this action is reporting `ExitCode`, `StandardOut`, and `StandardError` when the process normally exits, I'd still expect it to write out those properties; you'll still have an exit code, standard out, and standard error for a process with a non-zero exit code.

src/Tools/dotnet-monitor/CollectionRules/ExecuteActionValidateOptions.cs
jander-msft: > What do you mean by "so that configuration can pass validation when its loaded?"

Just because the path doesn't exist when the options are loaded from configuration doesn't mean that the path won't exist when you go to execute the action. For example, if we are to perform token replacement for the options of each action before the action is executed, then you cannot verify that the path exists when you load the options (because replacing the tokens may make the path exist). Another example is what if the process of one Execute action writes out a file to disk in a known location and the subsequent Execute action has that file specified in its options? That file doesn't exist at options load time but only after executing the first action.

> My understanding of what you want is just to perform a check upon entering `ExecuteAsync` (and then failing the action if there's nothing to execute for the path), but I just want to make sure I'm understanding what you're asking for.

Yes, check that the path exits before calling `Process.Start` on it and throw a reasonable exception e.g. `FileNotFoundException`.

src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
jander-msft: I wonder if all actions should have this and the execution pipeline can handle the timeout by raising cancellation on the cancellation token. We'd probably want this to be a configurable property on every action. We might need to think about the design on this one for a little bit to strike a balance on where this information is found in configuration and who is responsible for binding the information into some set of options (is there an overlay options that we bind for each action section? Do we force action options to implement an interface? etc).

kkeirstead: The latest version integrates an adaptation of this - I wasn't entirely sure about how much of it was specific to the DotNetRunner file and how much was needed for our use case, so if I should have included or removed other pieces just let me know.

src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
```diff
@@ -0,0 +1,78 @@
...
+        public Task<int> ExitedTask => _exitedSource.Task;
+
+        public async Task<ActionResult> ExecuteAsync(ExecuteOptions options, DiagnosticsClient client, CancellationToken cancellationToken)
+        {
+            string path = options.Path;
+            string arguments = options.Arguments;
+
+            if (!File.Exists(path))
+            {
+                throw new FileNotFoundException("A file could not be found at the provided path: " + path);
```
jander-msft, line 29: All of the strings (except the "ExitCode") should be in the Strings.resx file.

src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            int exitCode = process.ExitCode;
+
+            if (!options.IgnoreExitCode && exitCode != 0)
+            {
+                // Do we want to use a specific type of Exception here? My thought was that the proper type of Exception may be dependent on the exitCode's value
+                throw new Exception("The process exited with exit code " + exitCode.ToString());
+            }
+
+            ActionResult executeResponse = new();
```
jander-msft, line 55: nit: you can use initializer syntax here:

```cs
return new ActionResult()
{
    OutputValues = new Dictionary<string, string>(StringComparer.Ordinal)
    {
        { "ExitCode", exitCode.ToString() }
    }
};
```

src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            executeResponse.OutputValues = new Dictionary<string, string> { { "ExitCode", exitCode.ToString() } };
```
jander-msft, line 57: Use CultureInfo.InvariantCulture when calling ToString to make sure it remains in a neutral format (e.g. doesn't have commas or periods as separators).

src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
```diff
@@ -0,0 +1,78 @@
...
+                throw new InvalidOperationException($"Unable to start: {process.StartInfo.FileName} {process.StartInfo.Arguments}");
+            }
+
+            await WaitForExitAsync(process, cancellationToken).ConfigureAwait(false);
+
+            int exitCode = process.ExitCode;
+
+            if (!options.IgnoreExitCode && exitCode != 0)
+            {
+                // Do we want to use a specific type of Exception here? My thought was that the proper type of Exception may be dependent on the exitCode's value
```
jander-msft, line 51: Never use `Exception` directly. Always use a derived exception. I think `InvalidOperationException` is fine.

src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
```diff
@@ -0,0 +1,78 @@
+﻿using Microsoft.Diagnostics.NETCore.Client;
```
jander-msft, line 1: Add copyright header.

src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
```diff
@@ -0,0 +1,78 @@
...
+{
+    internal sealed class ExecuteAction : ICollectionRuleAction<ExecuteOptions>
+    {
+        // Completion source that is signaled when the process exits
+        private TaskCompletionSource<int> _exitedSource;
+
+        /// <summary>
+        /// Gets a task that completes with the process exit code when it exits.
+        /// </summary>
+        public Task<int> ExitedTask => _exitedSource.Task;
```
jander-msft, line 20: This needs to be removed too.

src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
jander-msft: Also, please add a Actions folder under src/Tools/dotnet-monitor/CollectionRules and move this file to that folder.

src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
```diff
@@ -0,0 +1,78 @@
+﻿using Microsoft.Diagnostics.NETCore.Client;
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Actions;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules
```
jander-msft, line 10:
```suggestion
namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Actions
```

jander-msft (changes requested):

src/Tools/dotnet-monitor/CollectionRules/ICollectionRuleAction.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules
+{
+    internal interface ICollectionRuleAction<TOptions>
+    {
+        Task<ActionResult> ExecuteAsync(TOptions options, DiagnosticsClient client, CancellationToken token);
+    }
+
+    internal struct ActionResult
+    {
+        internal Dictionary<string, string> OutputValues { get; set; }
```
wiktork, line 21: nit: OutputValues should be public.

src/Tools/dotnet-monitor/CollectionRules/IAction.cs
jander-msft: Yeah, I agree, maybe `CollectionRuleActionResult` to better differentiate.

src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
wiktork: Did we want to have a more constrained set of exceptions that actions can return (similar to pipelines) so that the calling infrastructure can have an easier time handling failures?

src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            if (!File.Exists(path))
+            {
+                throw new FileNotFoundException("A file could not be found at the provided path: " + path);
+            }
+
+            // May want to capture stdout and stderr and return as part of the result in the future
+            Process process = new Process();
+
+            process.StartInfo = new ProcessStartInfo(path, arguments);
```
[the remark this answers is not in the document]
kkeirstead: I think the latest change gets at what you're asking for - the only thing I didn't know was if there was something I needed to do to explicitly 'unsubscribe'.

src/Tools/dotnet-monitor/CollectionRules/Actions/ExecuteAction.cs
```diff
@@ -0,0 +1,89 @@
...
+            // May want to capture stdout and stderr and return as part of the result in the future
+            Process process = new Process();
+
+            process.StartInfo = new ProcessStartInfo(path, arguments);
+            process.EnableRaisingEvents = true;
+
+            // Completion source that is signaled when the process exits
+            TaskCompletionSource<int> exitedSource = new TaskCompletionSource<int>(TaskCreationOptions.RunContinuationsAsynchronously);
+
+            EventHandler exitedHandler = (s, e) => exitedSource.SetResult(process.ExitCode);
```
jander-msft, line 36:
```suggestion
            EventHandler exitedHandler = (s, e) => exitedSource.TrySetResult(process.ExitCode);
```

src/Tools/dotnet-monitor/CollectionRules/Actions/ExecuteAction.cs
```diff
@@ -0,0 +1,89 @@
...
+                    { "ExitCode", process.ExitCode.ToString(CultureInfo.InvariantCulture) }
+                }
+            };
+        }
+
+        public async Task WaitForExitAsync(Process process, Task<int> exitedTask, CancellationToken token)
+        {
+            if (!process.HasExited)
+            {
+                await Task.WhenAny(exitedTask, Task.Delay(Timeout.Infinite, token)).ConfigureAwait(false);
```
jander-msft, line 68: I ended up giving you bad advise on this. @kelltrick pointed out that the implementation of Task.Delay will register a callback on the CancellationToken but never unregister it if the CancellationToken is never cancelled and the timeout is infinite.

So doing something like this might be better:
```cs
TaskCompletionSource<object> cancellationTaskSource = new<object>();
using var _ = token.Register(() => cancellationTaskSource.TrySetCancelled(token));
await Task.WhenAny(exitedTask, cancellationTaskSource.Task).Unwrap().ConfigureAwait(false);
```

src/Tools/dotnet-monitor/CollectionRules/Actions/ExecuteAction.cs
```diff
@@ -0,0 +1,89 @@
...
+            {
+                if (!process.Start())
+                {
+                    throw new InvalidOperationException(string.Format(CultureInfo.CurrentCulture, Strings.ErrorMessage_UnableToStartProcess, process.StartInfo.FileName, process.StartInfo.Arguments));
+                }
+
+                await WaitForExitAsync(process, exitedSource.Task, cancellationToken).ConfigureAwait(false);
+            }
+            finally
+            {
```
jander-msft, line 51: Should also do `process.Exited -= exitedHandler;` here.

src/Tools/dotnet-monitor/CollectionRules/Actions/ExecuteAction.cs
```diff
@@ -0,0 +1,89 @@
...
+        public async Task<CollectionRuleActionResult> ExecuteAsync(ExecuteOptions options, DiagnosticsClient client, CancellationToken cancellationToken)
+        {
+            string path = options.Path;
+            string arguments = options.Arguments;
+            bool IgnoreExitCode = options.IgnoreExitCode.GetValueOrDefault(ExecuteOptionsDefaults.IgnoreExitCode);
+
+            ValidateFilePathValidity(path);
+
+            // May want to capture stdout and stderr and return as part of the result in the future
+            Process process = new Process();
```
jander-msft, line 28:
```suggestion
            using Process process = new Process();
```

jander-msft (changes requested):

src/Tools/dotnet-monitor/CollectionRules/Actions/ExecuteAction.cs
```diff
@@ -0,0 +1,89 @@
...
+            {
+                throw new FileNotFoundException(string.Format(CultureInfo.CurrentCulture, Strings.ErrorMessage_FileNotFound, path));
+            }
+        }
+
+        internal static void ValidateExitCode(bool ignoreExitCode, int exitCode)
+        {
+            if (!ignoreExitCode && exitCode != 0)
+            {
+                throw new InvalidOperationException(string.Format(CultureInfo.CurrentCulture, Strings.ErrorMessage_NonzeroExitCode, exitCode.ToString(CultureInfo.InvariantCulture)));
```
jander-msft, line 85:
```suggestion
                throw new InvalidOperationException(string.Format(CultureInfo.InvariantCulture, Strings.ErrorMessage_NonzeroExitCode, exitCode));
```

All of the `string.Format` calls here should use `CultureInfo.InvariantCulture`.

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/ExecuteOptionsDefaults.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if UNITTEST
+namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
+#else
+using System.ComponentModel;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Actions
```
jander-msft, line 10: Remove the conditional namespaces and just use this one. I think this is cause the build failures.

src/Tools/dotnet-monitor/CollectionRules/ExecuteAction.cs
jander-msft: In general, I would say this is a good idea. However, if we open the actions up to extensibility, then we need to make that exception type available publicly as well. Otherwise, we end up with a mix of the prebuilt actions throwing the special exception and dotnet-monitor treating that as "yes, the action failed, but that's necessarily a bug" whereas external actions would only be able to throw exceptions that are not this type and all of them would be considered buggy behavior.

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/ExecuteOptionsDefaults.cs
jander-msft: This file also needs to be linked into the SchemaGeneration project.

kkeirstead: Ah, it would seem this was a case of me not fetching/pulling before pushing. My local branch didn't have all the links in ConfigurationSchema. I'll push again once I confirm things are working locally with the latest changes.

src/Tools/dotnet-monitor/CollectionRules/ICollectionRuleAction.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
kkeirstead, line 1: Would it make sense for this to also go in the new Actions directory where we put `ExecuteAction.cs`? I realized I never clarified where this should go (or if it should just stay put).

jander-msft: I think moving it to the Actions directory makes sense.

src/Tools/dotnet-monitor/CollectionRules/Actions/ExecuteAction.cs
```diff
@@ -0,0 +1,92 @@
...
+                    throw new InvalidOperationException(string.Format(CultureInfo.InvariantCulture, Strings.ErrorMessage_UnableToStartProcess, process.StartInfo.FileName, process.StartInfo.Arguments));
+                }
+
+                await WaitForExitAsync(process, exitedSource.Task, cancellationToken).ConfigureAwait(false);
+            }
+            finally
+            {
+                process.Exited -= exitedHandler;
+
+                ValidateExitCode(IgnoreExitCode, process.ExitCode);
```
jander-msft, line 54: Move this just outside the finally block.

src/Tools/dotnet-monitor/CollectionRules/Actions/ExecuteAction.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.Globalization;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Actions
+{
+    internal sealed class ExecuteAction : ICollectionRuleAction<ExecuteOptions>
+    {
+        public async Task<CollectionRuleActionResult> ExecuteAsync(ExecuteOptions options, DiagnosticsClient client, CancellationToken cancellationToken)
```
kkeirstead, line 19: From discussion with Justin - switch client to IEndpointInfo

jander-msft (approved):

PR #726, opened 2021-08-17
Move *EndpointInfo implementations to dotnet-monitor.

jander-msft:
The collection rule system will hook directly into the ServerEndpointInfoSource (and eventually, the ClientEndpointInfoSource), so this class needs to move to the dotnet-monitor project. This means that the unit tests for these types need to move to a new unit test assembly for dotnet-monitor (this unit test assembly already exists in the feature/triggers branch, but I'm bringing it to the main branch). As part of this, I've refactored entry point assembly path calculation for the tests into a common helper since I needed to changed some of the ServerEndpointInfoSource tests into theories with each supported TFM.

Overview of changes:
- Move *EndpointInfo implementations to dotnet-monitor.
- Add dotnet-monitor unit test assembly and move *EndpointInfo tests to it.
- Consolidate calculating entry point assembly path into single utility method.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Microsoft.Diagnostics.Monitoring.WebApi.csproj
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Microsoft.Diagnostics.Monitoring.WebApi.csproj
@@ -42,6 +42,7 @@
   <ItemGroup>
     <InternalsVisibleTo Include="dotnet-monitor" />
     <InternalsVisibleTo Include="Microsoft.Diagnostics.Monitoring.OpenApiGen" />
+    <InternalsVisibleTo Include="Microsoft.Diagnostics.Monitoring.Tool.UnitTests" />
     <InternalsVisibleTo Include="Microsoft.Diagnostics.Monitoring.WebApi.UnitTests" />
   </ItemGroup>
 
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/AssemblyHelper.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/AssemblyHelper.cs
@@ -19,27 +19,13 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon
 
             if (tfm != TargetFrameworkMoniker.Current)
             {
-                string currentFolderName = GetTargetFrameworkMonikerFolderName(DotNetHost.BuiltTargetFrameworkMoniker);
-                string targetFolderName = GetTargetFrameworkMonikerFolderName(tfm);
+                string currentFolderName = DotNetHost.BuiltTargetFrameworkMoniker.ToFolderName();
+                string targetFolderName = tfm.ToFolderName();
 
                 assemblyPath = assemblyPath.Replace(currentFolderName, targetFolderName);
             }
 
             return assemblyPath;
-        }
-
-        private static string GetTargetFrameworkMonikerFolderName(TargetFrameworkMoniker moniker)
-        {
-            switch (moniker)
-            {
-                case TargetFrameworkMoniker.Net50:
-                    return "net5.0";
-                case TargetFrameworkMoniker.Net60:
-                    return "net6.0";
-                case TargetFrameworkMoniker.NetCoreApp31:
-                    return "netcoreapp3.1";
-            }
-            throw new ArgumentException($"Moniker '{moniker:G}' is not supported.");
         }
     }
 }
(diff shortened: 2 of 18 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/AssemblyHelper.cs
```diff
@@ -0,0 +1,45 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Reflection;
+
+namespace Microsoft.Diagnostics.Monitoring.TestCommon
+{
+    public static class AssemblyHelper
+    {
+        public static string GetAssemblyArtifactBinPath(
```
jander-msft, line 12: @kkeirstead, I'm add this new method that helps in calculating the entry point assembly path based on the current test assembly location. It's basically the same algorithm the rest of the tests had previously used, but consolidated and added an additional means of choosing which TFM (target framework moniker e.g. `netcoreapp3.1`, `net5.0`, or `net6.0`) folder to find the assembly. You could probably use this in your new test that you are working on. I'll try to get this merged in soon and merged to the feature/triggers branch.

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/AssemblyHelper.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        private static string GetTargetFrameworkMonikerFolderName(TargetFrameworkMoniker moniker)
+        {
+            switch (moniker)
+            {
+                case TargetFrameworkMoniker.Net50:
+                    return "net5.0";
+                case TargetFrameworkMoniker.Net60:
+                    return "net6.0";
```
kelltrick, line 38: These are the same as `$(TargetFramework)` in the `.targets` files. Why not generate these into an extension method in `DotNetHost.cs.template` and keep all the re-configuration necessary for a version update in the same place?

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/DotNetRunner.cs
```diff
@@ -188,5 +193,30 @@ public void ForceClose()
...
+        private string GetAspNetCoreFrameworkVersion()
+        {
+            if (TargetFrameworkMoniker.Current == TargetFramework)
+            {
+                return DotNetHost.CurrentAspNetCoreVersionString;
+            }
+            throw new InvalidOperationException($"Unsupported target framework moniker: {TargetFramework}");
+        }
+
+        private string GetNetCoreAppFrameworkVersion()
```
kelltrick, line 206: Same with this method. Move to an extension method in the template.

kelltrick (approved):
My suggestion here would be to move all the things necessary to add `net7.0` into as few files as possible. Ideally all in `TargetFrameworkMoniker.cs` and the `DotNetHost.cs.template` + `GenerateDotNetHost.targets`

src/Microsoft.Diagnostics.Monitoring.WebApi/Microsoft.Diagnostics.Monitoring.WebApi.csproj
```diff
@@ -42,6 +42,7 @@
   <ItemGroup>
```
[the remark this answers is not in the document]
jander-msft: There are scenarios for the separate WebAPI, but they are not clearly defined. We'll probably discuss this more with diagnostics folks in the coming months.

jander-msft:
> My suggestion here would be to move all the things necessary to add `net7.0` into as few files as possible. Ideally all in `TargetFrameworkMoniker.cs` and the `DotNetHost.cs.template` + `GenerateDotNetHost.targets`

I'll see what I can do. If it requires iteration over a collection of the supported TFMs, then it might be be doable in the non-trivial case without writing a source generator. At the very least, I should be able to consolidate all TFM related logic into one manually written file.

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/DotNetRunner.cs
jander-msft: I've moved all TFM conversion methods to a TargetFrameworkMonikerExtensions type in TargetFrameworkMoniker.cs.

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/AssemblyHelper.cs
jander-msft: I've moved this conversion logic to be an extension method in the TargetFrameworkMoniker.cs file.

jander-msft:
> My suggestion here would be to move all the things necessary to add `net7.0` into as few files as possible. Ideally all in `TargetFrameworkMoniker.cs` and the `DotNetHost.cs.template` + `GenerateDotNetHost.targets`

All conversion methods have been made into extension methods in the TargetFrameworkMoniker.cs file. So if you go to add a new enum value, it's more obvious what needs to be updated.

Also, I've moved the notion of the BuildTargetFrameworkMoniker directly into the DotNetHost.cs file and used preprocessing to determines it value. This will cause a build failure if a new TFM is added to the project but is forgotten to be added to the enum and updating this static property.

PR #736, opened 2021-08-18
Basic implementation and registration of actions and triggers

jander-msft:
Overview of changes:
- Initial interfaces for collection rule triggers.
- Implement empty action and trigger implementations.
- Implement startup trigger with special interface denoting it is a startup trigger.
- Rename action provider and trigger provider to descriptor.
- Register action and trigger implementations with descriptors.

These changes provide the remainder of the base set of interfaces and implementations for actions and triggers so that the rule system can be built on top of them. The actions and triggers are intentionally not implemented at this time, but the existence of their classes allow them to be composed together for the execution of a rule (coming in a later PR).


src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRuleActionOptionsProvider.cs
```diff
@@ -16,24 +16,24 @@ internal sealed class CollectionRuleActionOptionsProvider :
 
         public CollectionRuleActionOptionsProvider(
             ILogger<CollectionRuleActionOptionsProvider> logger,
-            IEnumerable<ICollectionRuleActionProvider> providers)
+            IEnumerable<ICollectionRuleActionDescriptor> providers)
```
jander-msft, line 19: nit: rename to `descriptor` and `descriptors` throughout file

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRuleTriggerOptionsProvider.cs
```diff
@@ -16,24 +16,24 @@ internal sealed class CollectionRuleTriggerOptionsProvider :
 
         public CollectionRuleTriggerOptionsProvider(
             ILogger<CollectionRuleTriggerOptionsProvider> logger,
-            IEnumerable<ICollectionRuleTriggerProvider> providers)
+            IEnumerable<ICollectionRuleTriggerDescriptor> providers)
```
jander-msft, line 19: nit: rename to `descriptor` and `descriptors` throughout file

src/Tools/dotnet-monitor/CollectionRules/Configuration/ICollectionRuleActionDescriptor.cs
```diff
@@ -6,9 +6,9 @@
 
 namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Configuration
 {
-    internal interface ICollectionRuleActionProvider
+    internal interface ICollectionRuleActionDescriptor
```
jander-msft, line 9: Add `Type ActionType { get; }`

src/Tools/dotnet-monitor/CollectionRules/Triggers/AspNetResponseStatusTriggerFactory.cs
```diff
@@ -0,0 +1,23 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Diagnostics.Monitoring.WebApi;
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Triggers;
+using System;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Triggers
+{
+    /// <summary>
+    /// Factory for creating a new AspNetReponseStatusTrigger trigger.
```
kkeirstead, line 12:
```suggestion
    /// Factory for creating a new AspNetResponseStatusTrigger trigger.
```

src/Tools/dotnet-monitor/CollectionRules/Triggers/AspNetResponseStatusTriggerFactory.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// </summary>
+    internal sealed class AspNetResponseStatusTriggerFactory :
+        ICollectionRuleTriggerFactory<AspNetResponseStatusOptions>
+    {
+        /// <inheritdoc/>
+        public ICollectionRuleTrigger Create(IEndpointInfo endpointInfo, AspNetResponseStatusOptions options)
+        {
+            throw new NotImplementedException("TODO: Implement AspNetReponseStatusTrigger.");
```
kkeirstead, line 20:
```suggestion
            throw new NotImplementedException("TODO: Implement AspNetResponseStatusTrigger.");
```

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRuleActionDescriptor.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Configuration
+{
+    internal sealed class CollectionRuleActionDescriptor<TAction, TOptions> :
```
jander-msft, line 9: Add `where TAction : ICollectionRuleAction<TOptions>`

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRuleTriggerDescriptor.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public Type FactoryType => typeof(TFactory);
+
+        public Type OptionsType { get; }
+
+        public string TriggerName { get; }
+    }
+
+    internal sealed class CollectionRuleTriggerProvider<TFactory, TOptions> :
+        CollectionRuleTriggerDescriptor<TFactory>
```
jander-msft, line 31: Add `where TFactory : ICollectionRuleTriggerFactory<TOptions>`

kelltrick (approved):
Reviewed in meeting.

src/Tools/dotnet-monitor/CollectionRules/Triggers/ICollectionRuleTrigger.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    /// <summary>
+    /// Interface implemented by collection rule triggers.
+    /// </summary>
+    internal interface ICollectionRuleTrigger
+    {
+        /// <summary>
+        /// Waits for the trigger condition to be satisfied.
+        /// </summary>
+        Task WaitForSatisfiedConditionAsync(CancellationToken cancellationToken);
```
jander-msft, line 18: Consider splitting initialization with the target process from monitoring the events from the target process.

jander-msft: Address this in a subsequent PR when the collection rule system is realized more.

wiktork, line 18: So are you expecting that in the WaitForSatisfiedConditionAsync would effectively be a listener for the callback from ITraceEventTrigger (at least in many trigger cases)?

wiktork, line 18: Given this interface, a trigger can never be satisfied more than once correct? We would need to recreate the trigger after taking the actions?

jander-msft: That's not necessarily correct. When called, a TaskCompletionSource could be registered for each call to the method. When the next satisfaction of the trigger condition occurs, then all TCS instances are completed.

I haven't really settled on this being the correct way to do it; we could just have a similar implementation as the pipelines where it is initialized with a callback and then there are start/run/stop/etc methods.

In any event, this interface is likely to change as we go to implement them. I think what's really going to drive the change is the need to say "take this callback and set yourself up" separately from "now it's okay to notify of condition satisfaction".

jander-msft: That's the initial intent, however this interface is like to change. Just wanted something to start while I work out the initialization of the collection rule pipeline/execution.

wiktork: Sounds good. I do think that the callback approach will be much easier to manage for the caller.

PR #738, opened 2021-08-19
Add release notes

### LINKED ISSUE #737: Add release notes to repo
Add release notes for next release to the git repo.
These will be used in the release pipeline to specify the github release notes for the next release.

kelltrick:
- Add `releaseNotes.current.md` which will be the new notes for the immediately upcoming release.
- Add `releaseNotes.Preview8.md` for all features that are checked in after the snap for the current release.
Closes #737


documentation/releaseNotes/releaseNotes.Preview8.md
```diff
@@ -0,0 +1,5 @@
+Today we are releasing the next official preview of the `dotnet-monitor` tool. This release includes:
```
jander-msft, line 1: Do we want to make another copy of this and call it `releasenotes.template.md` to make it ever so slightly easier to start fresh with the next fork?

jander-msft (approved):

documentation/releaseNotes/releaseNotes.Preview8.md
kelltrick: I was also thinking it would be a good idea to add a `Readme.md` to explain the rolling of these files.

jander-msft: Why is there a Preview 8 release notes if we haven't started working on Preview 8 yet? What I think makes more sense is that:
- `main` should only have the released release notes, the current release notes, and the template. 
- `release/*` should only have the the released release notes from the previous releases, the release notes of the current release is still called `releaseNotes.current.md`, and the template.

Then the process is:
- When a fork occurs, you just merge from `main` to `release/*` and there's nothing to fix in the `release/*` branch. In the `main` branch, copy contents of `releaseNotes.template.md` to `releaseNotes.current.md` to reset the current release notes; this should be done at the same time when bumping the version information in `eng/Version.props`.
- When a release finally goes out the door, all you do is copy `releaseNotes.current.md` from  the `release/*` branch back to the `main` branch but it is named whatever the release is e.g. `releaseNotes.v5.0.0-preview.6.21370.3.md`.

I don't think there needs to be an intermediate `releaseNotes.Preview#.md` file.

jander-msft (approved):

PR #742, opened 2021-08-19
Execute Action - Testing

kkeirstead:
Adding independent tests for the Execute Action to show that it works before it's integrated in with triggers.


Directory.Build.props
```diff
@@ -9,6 +9,7 @@
     <WarningLevel>4</WarningLevel>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <SuppressNETCoreSdkPreviewMessage>true</SuppressNETCoreSdkPreviewMessage>
+    <DisableImplicitNamespaceImports_DotNet>true</DisableImplicitNamespaceImports_DotNet>
```
jander-msft, line 12: I haven't seen this issue and to my knowledge the PR/CI builds haven't had this issue either. I think you might be using a .NET 6 SDK version that is too old. I ask that you revert this change and try to update the SDK installed on your machine to the latest public release: https://dotnet.microsoft.com/download/dotnet/6.0

src/Tests/Microsoft.Diagnostics.Monitoring.ExecuteActionApp/Program.cs
```diff
@@ -0,0 +1,38 @@
...
+
+            switch (testType)
+            {
+                case "ZeroExitCode":
+                    return 0;
+
+                case "NonzeroExitCode":
+                    return -1;
+
+                case "TokenCancellation":
```
jander-msft, line 25: Don't name it for the scenario that is being tested; name it for what the application does eg. `Sleep`.

src/Tests/Microsoft.Diagnostics.Monitoring.ExecuteActionApp/Program.cs
```diff
@@ -0,0 +1,38 @@
...
+                    Thread.Sleep(DelayMs);
```
jander-msft, line 26: Rather than use a constant and have the test assume a different constant that needs to be coordinated with this one, take the value as a parameter from the command line like you did for `TextFileOutput`.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExecuteActionTests.cs
```diff
@@ -0,0 +1,132 @@
...
+            ExecuteAction action = new();
+
+            ExecuteOptions options = new();
+
+            options.Path = DotNetHost.HostExePath;
+            options.Arguments = GenerateArgumentsString(new string[] { "ZeroExitCode" });
+
+            CollectionRuleActionResult result = await action.ExecuteAsync(options, null, CreateCancellationToken());
+
+            Assert.Equal("0", result.OutputValues["ExitCode"]);
```
jander-msft, line 37:
```suggestion
            Assert.NotNull(result);
            Assert.NotNull(result.OutputValues);
            Assert.True(result.OutputValues.TryGetValue("ExitCode", actualExitCode));
            Assert.Equal("0", actualExitCode);
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExecuteActionTests.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            ExecuteOptions options = new();
+
+            options.Path = DotNetHost.HostExePath;
+            options.Arguments = GenerateArgumentsString(new string[] { "TokenCancellation" }); ;
+
+            TaskCanceledException taskCanceledException = await Assert.ThrowsAsync<TaskCanceledException>(
+                () => action.ExecuteAsync(options, null, CreateCancellationToken()));
+
+            Assert.Contains(TaskCanceledMessage, taskCanceledException.Message);
```
jander-msft, line 69: This is not necessary. Testing that it is throwing TaskCanceledException is sufficient.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExecuteActionTests.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        [Fact]
+        public async Task ExecuteAction_TextFileOutput()
+        {
+            ExecuteAction action = new();
+
+            ExecuteOptions options = new();
+
+            DirectoryInfo outputDirectory = Directory.CreateDirectory(Path.Combine(Path.GetTempPath(), "ExecuteAction", Guid.NewGuid().ToString()));
+            string textFileOutputPath = outputDirectory.FullName + "\\file.txt";
```
jander-msft, line 80:
```suggestion
            string textFileOutputPath = Path.Combine(outputDirectory.FullName, "file.txt");
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExecuteActionTests.cs
jander-msft: Apply this same strategy to all of the other success tests. In fact, make this check a static method so that it can be reused e.g. `private static void ValidateActionResult(CollectionActoinResult result, int expectedExitCode)`

src/Tests/Microsoft.Diagnostics.Monitoring.ExecuteActionApp/Program.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.IO;
+using System.Threading;
+
+namespace Microsoft.Diagnostics.Monitoring.ExecuteActionApp
+{
+    internal class Program
+    {
+        public static int Main(string[] args)
+        {
+            string testType = args[0];
```
wiktork, line 14: Consider some argument validation to make it easier to figure out failures

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExecuteActionTests.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            options.Path = DotNetHost.HostExePath;
+            options.Arguments = GenerateArgumentsString(new string[] { "TextFileOutput", textFileOutputPath, testMessage });
+
+            CollectionRuleActionResult result = await action.ExecuteAsync(options, null, CreateCancellationToken());
+
+            Assert.Equal("0", result.OutputValues["ExitCode"]);
+            Assert.Equal(testMessage, File.ReadAllText(textFileOutputPath));
+
+            try
```
jander-msft, line 92: Most of the test should be wrapped in a try...finally where in the finally block you delete the directory.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExecuteActionTests.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            FileNotFoundException fileNotFoundException = await Assert.ThrowsAsync<FileNotFoundException>(
+                () => action.ExecuteAsync(options, null, CreateCancellationToken()));
+
+            Assert.Equal(string.Format(Strings.ErrorMessage_FileNotFound, uniquePathName), fileNotFoundException.Message);
+        }
+
+        private static string GenerateArgumentsString(string[] additionalArgs)
+        {
+            return Assembly.GetExecutingAssembly().Location.Replace(
```
jander-msft, line 121: Please use AssemblyHelper.GetAssemblyArtifactBinPath to get the path to the assembly.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExecuteActionTests.cs
```diff
@@ -0,0 +1,132 @@
...
+                Assembly.GetExecutingAssembly().GetName().Name,
+                "Microsoft.Diagnostics.Monitoring.ExecuteActionApp") + ' ' + string.Join(' ', additionalArgs);
+        }
+
+        private static CancellationToken CreateCancellationToken()
+        {
+            CancellationTokenSource cancellationTokenSource = new CancellationTokenSource(TokenTimeoutMs);
```
jander-msft, line 128: CTS needs to be disposed so that you don't linger timers and they don't fire off cancellation on tokens after everything is done. This probably means you need to move the CTS creation to each test method.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExecuteActionTests.cs
```diff
@@ -0,0 +1,132 @@
...
+using Microsoft.Diagnostics.Monitoring.TestCommon;
+using System.Threading;
+using System;
+using System.IO;
+using System.Diagnostics;
+using Microsoft.Diagnostics.Tools.Monitor;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    public sealed class ExecuteActionTests
```
jander-msft, line 19: Please add a test for IgnoreExitCode

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExecuteActionTests.cs
```diff
@@ -0,0 +1,132 @@
...
+    {
+        private const int TokenTimeoutMs = 2000; // Must be less than DelayMs in the ExecuteApp
```
jander-msft, line 21: This timeout needs to be much longer, more like 10 seconds, to account for PR/CI build slowness.

jander-msft (changes requested):

Directory.Build.props
kkeirstead: I'll add a comment before marking this as resolved; my best guess is that it had to with being on an outdated version of the dogfood VS

src/Tests/Microsoft.Diagnostics.Monitoring.ExecuteActionApp/Program.cs
kkeirstead: What did you have in mind? In theory, since for each test we're manually setting the args, shouldn't we know that they'll all be valid?

jander-msft: For example:
- Maybe the default case of the switch should throw an exception stating that the test type is not one of the known ones.
- Maybe extract the remaining args into a new array that the test types can look instead of looking at the original `args` parameter. This allows us to change the ordering of the arguments passed on the command line without having to rewrite all of the test cases.
- Each test type should valid that its args length is exactly what it should expect and throw an exception if it is different.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExecuteActionTests.cs
kkeirstead: Just a note: xUnit complains "Do not use Assert.NotNull() on value type 'CollectionRuleActionResult", so I'll omit that check

src/Tests/Microsoft.Diagnostics.Monitoring.ExecuteActionApp/Program.cs
```diff
@@ -0,0 +1,50 @@
...
+
+            string[] additionalArgs = args.Skip(1).ToArray();
+
+            switch (testType)
+            {
+                case "ZeroExitCode":
+                    Assert.Equal(1, args.Length);
```
jander-msft, line 24: Please assert the length of `additionalArgs` instead of the original  `args`.

src/Tests/Microsoft.Diagnostics.Monitoring.ExecuteActionApp/Program.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Microsoft.Diagnostics.Monitoring.ExecuteActionApp
+{
+    internal class Program
+    {
+        public static int Main(string[] args)
+        {
+            string testType = args[0];
+
+            string[] additionalArgs = args.Skip(1).ToArray();
```
jander-msft, line 19: nit: `testArgs` to make it clear what these are.

src/Tests/Microsoft.Diagnostics.Monitoring.ExecuteActionApp/Program.cs
```diff
@@ -0,0 +1,50 @@
...
+                    return 0;
+
+                case "NonzeroExitCode":
+                    Assert.Equal(1, args.Length);
+                    return -1;
+
+                case "Sleep":
+                    Assert.Equal(2, args.Length);
+                    string delayArg = additionalArgs[0];
+                    int delay = int.Parse(delayArg) + 1000; // Add a second delay to the token cancellation time
```
jander-msft, line 34: Make the app are basic as possible: don't make the app have assumptions about how it is tested. Just take the sleep time as-is and sleep on it. Have the test itself add whatever padding it needs.

src/Tests/Microsoft.Diagnostics.Monitoring.ExecuteActionApp/Program.cs
```diff
@@ -0,0 +1,50 @@
...
+
+                case "TextFileOutput":
+                    Assert.Equal(3, args.Length);
+                    string pathArg = additionalArgs[0];
+                    string contentsArg = additionalArgs[1];
+                    File.WriteAllText(pathArg, contentsArg);
+                    return 0;
+
+                default:
+                    throw new ArgumentException("Unknown provided test type.");
```
jander-msft, line 46:
```suggestion
                    throw new ArgumentException($"Unknown test type {testType}.");
```
so it's easier to know what the specific problem is.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExecuteActionTests.cs
```diff
@@ -0,0 +1,169 @@
...
+            CollectionRuleActionResult result = await action.ExecuteAsync(options, null, cancellationTokenSource.Token);
+
+            ValidateActionResult(result, "-1");
+        }
+
+        private static string GenerateArgumentsString(string[] additionalArgs)
+        {
+            Assembly currAssembly = Assembly.GetExecutingAssembly();
+
+            return AssemblyHelper.GetAssemblyArtifactBinPath(currAssembly, currAssembly.GetName().Name, TargetFrameworkMoniker.NetCoreApp31).Replace(
```
jander-msft, line 155:
```suggestion
            return AssemblyHelper.GetAssemblyArtifactBinPath(currAssembly, "Microsoft.Diagnostics.Monitoring.ExecuteActionApp", TargetFrameworkMoniker.NetCoreApp31)
```

will get you the path to the Microsoft.Diagnostics.Monitoring.ExecuteActionApp.dll file. You shouldn't have to do any string replacement yourself.

jander-msft (changes requested):

kkeirstead:
I noticed tests that return -1 as the exit code were failing unexpectedly (that didn't fail locally), claiming the exit code was 255. I changed the exit codes to 1 in the latest commit just to see if the tests will pass (and because fundamentally they still have the desired effect), but I'd be curious to know if anyone has any insight into this and whether it's an issue.

jander-msft:
> I noticed tests that return -1 as the exit code were failing unexpectedly (that didn't fail locally), claiming the exit code was 255. I changed the exit codes to 1 in the latest commit just to see if the tests will pass (and because fundamentally they still have the desired effect), but I'd be curious to know if anyone has any insight into this and whether it's an issue.

It might be that POSIX systems only allow for 8-bit exit codes from 0-255. So -1 becomes 255.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExecuteActionTests.cs
```diff
@@ -0,0 +1,169 @@
...
+            CollectionRuleActionResult result = await action.ExecuteAsync(options, null, cancellationTokenSource.Token);
+
+            ValidateActionResult(result, "1");
+        }
+
+        private static string GenerateArgumentsString(string[] additionalArgs)
+        {
+            Assembly currAssembly = Assembly.GetExecutingAssembly();
+
+            return AssemblyHelper.GetAssemblyArtifactBinPath(currAssembly, "Microsoft.Diagnostics.Monitoring.ExecuteActionApp", TargetFrameworkMoniker.NetCoreApp31)
```
jander-msft, line 156: nit: It might read better (and be easier to extend in the future) if this was something more like:

```cs
List<string> args = new();

// Entrypoint assembly
args.Add(AssemblyHelper.GetAssemblyArtifactBinPath(currAssembly, "Microsoft.Diagnostics.Monitoring.ExecuteActionApp", TargetFrameworkMoniker.NetCoreApp31));
// Entrypoint arguments
args.AddRange(additionalArgs)

return string.Join(' ', args);
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExecuteActionTests.cs
```diff
@@ -0,0 +1,169 @@
...
+
+        [Fact]
+        public async Task ExecuteAction_TokenCancellation()
+        {
+            ExecuteAction action = new();
+
+            ExecuteOptions options = new();
+
+            options.Path = DotNetHost.HostExePath;
+            options.Arguments = GenerateArgumentsString(new string[] { "Sleep", (TokenTimeoutMs + DelayMs).ToString() }); ;
```
jander-msft, line 67: This might work most of the time. If it's not stable enough, maybe we can change this to write out a file (like the TextFileOutput scenario) and then sleep for some large amount of time. Back in the test, we start the process, wait for the file to appear, and then immediately cancel the token. This would provide more determinism so that if the host happens to be really slow that process startup doesn't exceed the token timeout.

jander-msft (approved):

PR #743, opened 2021-08-20
Add callback interface to IEndpointInfoSource implementations

jander-msft:
Change the derived ServerEndpointInfoSource in tests to a callback system which will allow other systems to participate in IEndpointInfoSource changes. For example, other systems can now configure a newly discovered process within the system before the process's runtime is resumed.


kelltrick:
Does it make sense to add an entry to [/documentation/releaseNotes.md](https://github.com/dotnet/dotnet-monitor/blob/main/documentation/releaseNotes/releaseNotes.md)?

src/Tools/dotnet-monitor/EndpointInfo/EndpointInfoExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Diagnostics.Monitoring.WebApi;
+using System;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    internal static class EndpointInfoExtensions
+    {
+        public static string GetDebuggerDisplay(this IEndpointInfo endpointInfo)
```
kelltrick, line 12: This is mostly used outside of getting a Debugger Display attribute. Maybe a better name would be:
```cs
public static string GetFriendlyDisplayString(this IEndpointInfo endpointInfo)
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfoSourceTests.cs
```diff
@@ -243,20 +251,19 @@ private static void VerifyConnection(AppRunner runner, IEndpointInfo endpointInf
             Assert.NotNull(endpointInfo.Endpoint);
         }
 
-        private sealed class TestServerEndpointInfoSource : ServerEndpointInfoSource
+        private sealed class ServerEndpointInfoCallback : IEndpointInfoSourceCallbacks
         {
             private readonly ITestOutputHelper _outputHelper;
-            private readonly List<Tuple<AppRunner, TaskCompletionSource<EndpointInfo>>> _addedEndpointInfoSources = new();
+            private readonly List<Tuple<AppRunner, TaskCompletionSource<IEndpointInfo>>> _addedEndpointInfoSources = new();
```
kelltrick, line 257:
```suggestion
            private readonly List<(AppRunner runner, TaskCompletionSource<IEndpointInfo> completionSource)> _addedEndpointInfoSources = new();
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfoSourceTests.cs
```diff
@@ -269,15 +276,20 @@ public async Task<EndpointInfo> WaitForNewEndpointInfoAsync(AppRunner runner, Ti
...
+                return Task.CompletedTask;
+            }
+
+            public void OnAddedEndpointInfo(IEndpointInfo info)
             {
-                _outputHelper.WriteLine($"[Source] Added: {info.DebuggerDisplay}");
+                _outputHelper.WriteLine($"[Source] Added: {info.GetDebuggerDisplay()}");
                 
                 lock (_addedEndpointInfoSources)
                 {
```
kelltrick, line 295: I realize this is out of the scope of this PR but doesn't this (lines 298 and 308) throw an exception from modifying a collection while iterating it?
```cs
foreach (var sourceTuple in _addedEndpointInfoSources)
{
     _addedEndpointInfoSources.Remove(sourceTuple);
}
```

kelltrick (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfoSourceTests.cs
jander-msft: Yup, it should throw. It probably doesn't because (my guess is that) it happens to remove the last item in the list, so there isn't anything else to iterate. I'll fix it in this PR.

src/Tools/dotnet-monitor/EndpointInfo/EndpointInfoExtensions.cs
jander-msft: I've split the extension method to be inlined with their current usages.

PR #768, opened 2021-08-25
Triggers - Action List Executor

### LINKED ISSUE #639: Create action list executor class
A new class, e.g. `ActionListExecutor`, should be created to handle the following functionality:
- Discover action implementations (likely through ctor DI injection).
- Take a list of action options and a property bag in order to execute each action.
  - Each action is executed sequentially. This is done by finding the action implementation based on the provided action options, replacing any parameterized options with values from the property bag, and then executing the action implementation with the action options.
  - Each action has the ability to add to the property bag when the action completes successfully.
  - An action implementation that fails to execute successfully will fail the remainder of the action list.

This class can be a singleton service since there should not be a need to have per-action-list state stored within the class instance.

Having this functionality encapsulated into one class will allow more targeted testing of individual actions and the interaction of actions via the property bag without involving the rest of the rule collection system.

kkeirstead:
Created the action list executor; currently, this does not grab properties from the 'property bag' to allow previous actions to influence future ones. This does include basic testing using the execute action, but this should be fleshed out further in the future as more actions are added (and the property bag is integrated).

Closes #639


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionListExecutorTests.cs
```diff
@@ -0,0 +1,100 @@
...
+using Microsoft.Extensions.Options;
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Triggers;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    public sealed class ActionListExecutorTests
+    {
+        private const int TokenTimeoutMs = 10000;
+
+        private ActionListExecutor _executor;
```
jander-msft, line 25: Because you are now using `TestHostHelper.CreateCollectionRulesHost`, there is no need for this field. Please remove it and change the test methods to just use a local ActionListExecutor.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionListExecutorTests.cs
```diff
@@ -0,0 +1,100 @@
...
+                    .AddExecuteActionAppAction(new string[] { "ZeroExitCode" })
+                    .AddExecuteActionAppAction(new string[] { "ZeroExitCode" })
+                    .SetStartupTrigger();
+            }, async host =>
+            {
+                _executor = host.Services.GetService<ActionListExecutor>();
+
+                using CancellationTokenSource cancellationTokenSource = new CancellationTokenSource(TokenTimeoutMs);
+
+                await _executor.ExecuteActions(host.Services.GetRequiredService<IOptionsMonitor<CollectionRuleOptions>>().Get("Default").Actions, null, cancellationTokenSource.Token);
```
jander-msft, line 48: nit: this might be more readable if the arguments are on separate lines. Also, if you move getting the CollectionRuleOptions to another line, that would help too.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionListExecutorTests.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            _outputHelper = outputHelper;
+        }
+
+        [Fact]
+        public async Task ActionListExecutor_AllActionsSucceed()
+        {
+            await TestHostHelper.CreateCollectionRulesHost(_outputHelper, rootOptions =>
+            {
+                rootOptions.CreateCollectionRule("Default")
```
jander-msft, line 38: nit: Maybe make the rule name a constant that all of the tests share.

jander-msft (approved):

PR #776, opened 2021-08-26
Fix race condition in EndpointInfoSourceTests.

jander-msft:
There's a race condition between when the ServerEndpointInfoCallback is invoked due to a process connecting in reverse mode and the LoggingRunnerAdapter setting the process Id information just after starting the process. Sometimes the reverse server callback is quicker than the adapter, which will then fail to find the runner with a matching process ID.

I changed the code so that the LoggingRunnerAdapter has an additional ProcessIdTask property so that it can be awaited. the ServerEndpointInfoCallback has been updated to await the ProcessIdTask property of each registered runner and process each one as they complete. This guarantees that ServerEndpointInfoCallback will wait for the runner to fulfill the process ID or be cancelled first.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/AppRunner.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/AppRunner.cs
@@ -60,6 +60,8 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Runners
 
         public int ProcessId => _adapter.ProcessId;
 
+        public Task<int> ProcessIdTask => _adapter.ProcessIdTask;
+
         /// <summary>
         /// Name of the scenario to run in the application.
         /// </summary>
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/AppRunner.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/AppRunner.cs
@@ -58,8 +58,6 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Runners
 
         public int ExitCode => _adapter.ExitCode;
 
-        public int ProcessId => _adapter.ProcessId;
-
         public Task<int> ProcessIdTask => _adapter.ProcessIdTask;
 
         /// <summary>
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfoSourceTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfoSourceTests.cs
@@ -118,7 +118,8 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
 
             AppRunner runner = CreateAppRunner(transportName, appTfm);
 
-            Task newEndpointInfoTask = callback.WaitForNewEndpointInfoAsync(runner, CommonTestTimeouts.StartProcess);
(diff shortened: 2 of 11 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfoSourceTests.cs
```diff
@@ -254,70 +256,89 @@ private static void VerifyConnection(AppRunner runner, IEndpointInfo endpointInf
         private sealed class ServerEndpointInfoCallback : IEndpointInfoSourceCallbacks
         {
             private readonly ITestOutputHelper _outputHelper;
-            private readonly List<(AppRunner Runner, TaskCompletionSource<IEndpointInfo> CompletionSource)> _addedEndpointInfoSources = new();
+            private readonly SemaphoreSlim _completionEntriesSemaphore = new(1);
```
wiktork, line 259: nit: Can you add a comment to indicate you are using a semaphore so that you can cancel the lock?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfoSourceTests.cs
```diff
@@ -254,70 +256,89 @@ private static void VerifyConnection(AppRunner runner, IEndpointInfo endpointInf
...
-                        _outputHelper.WriteLine($"[Source] Checking App{runner.AppId}");
-                        try
+                        // Wait for any of the process ID tasks to complete.
+                        Task<int> completedTask = await Task.WhenAny(map.Keys);
+
+                        map.Remove(completedTask, out CompletionEntry entry);
+
+                        _outputHelper.WriteLine($"[Source] Checking App{entry.Runner.AppId}");
+
+                        if (completedTask.IsCompletedSuccessfully)
```
wiktork, line 320: Should we be logging failures if the task does not complete successfully?

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/AppRunner.cs
```diff
@@ -60,6 +60,8 @@ public sealed class AppRunner : IAsyncDisposable
 
         public int ProcessId => _adapter.ProcessId;
```
wiktork, line 61: Do we still use this property? Or would you expect all access to be done through the task?

wiktork (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/AppRunner.cs
jander-msft: There are 33 uses of this property still. I can try to go through all of them and convert them to using the task.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfoSourceTests.cs
jander-msft: No, there can still be other processes that were discovered by the ServerEndpointInfoSource that haven't been processed by a callback yet.

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/AppRunner.cs
jander-msft: I've removed the property and replaced with with accesses to the ProcessIdTask property.

PR #818, opened 2021-09-03
Consolidate task and completion source cancellation methods.

jander-msft:
These changes consolidate cancelling the await of tasks (without cancelling the task itself) and the cancelling of TaskCompletionSources into the WithCancellation extension method and TaskCompletionSourceWithCancellation class, respectively.


src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/TaskTestExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using System;
+using System.Threading.Tasks;
+using Xunit.Abstractions;
+
+namespace Microsoft.Diagnostics.Monitoring.TestCommon
+{
+    public static class TaskTestExtensions
+    {
+        public static async Task SafeAwait(this Task task, ITestOutputHelper outputHelper = null)
```
jander-msft, line 13: These methods did not change but were split out into their own extension method class.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfoSourceCallback.cs
```diff
@@ -0,0 +1,143 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit.Abstractions;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    internal sealed class EndpointInfoSourceCallback : IEndpointInfoSourceCallbacks
```
jander-msft, line 17: Hardly anything change in this except:
1) moved from being a private class with the EndpointInfoSourceTests class
2) uses the TaskCompletionSourceWithCancellation class within the CompletionEntry class.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfoSourceCallback.cs
```diff
@@ -0,0 +1,143 @@
...
+        /// </summary>
+        private readonly SemaphoreSlim _completionEntriesSemaphore = new(1);
+        private readonly List<CompletionEntry> _completionEntries = new();
+
+        public EndpointInfoSourceCallback(ITestOutputHelper outputHelper)
+        {
+            _outputHelper = outputHelper;
+        }
+
+        public async Task<IEndpointInfo> WaitForNewEndpointInfoAsync(AppRunner runner, TimeSpan timeout)
```
kkeirstead, line 33: I don't believe we're using this param (`timeout`). Should we be for the `timeoutCancellation`?

src/Tools/dotnet-monitor/TaskCompletionSourceWithCancellation.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    public sealed class TaskCompletionSourceWithCancellation<T> :
+        TaskCompletionSource<T>
+    {
+        public TaskCompletionSourceWithCancellation(TaskCreationOptions creationOptions)
+            : base(creationOptions)
+        {
+        }
+
+        public async Task<T> WithCancellation(CancellationToken token)
```
wiktork, line 22: Wouldn't this be preferable as an extension method rather than an inherited class?

jander-msft: The runtime has basically the same implementation apparently with a derived class: https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/Threading/Tasks/TaskCompletionSourceWithCancellation.cs

The one benefit is that they don't have extra allocations for the closure because it passes the TCS is the state parameter and makes use of the new Register overload in .NET 6.

However, I can still make it an extension method instead since we can't avoid the allocation.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfoSourceCallback.cs
jander-msft: Good catch; I had it correct in my other branch but the merge got slightly messed up.

wiktork (approved):

PR #824, opened 2021-09-03
Create CollectDump Collection Rule Action

### LINKED ISSUE #632: Create `CollectDump` collection rule action
- Create an action named `CollectDump` that will invoke the dump pipeline to collect a dump artifact.
  - This action should output the egress path of the egressed dump so that other actions can consume the value.
- Create a strongly-typed options class, e.g. `CollectDumpActionOptions`, for the action with the following properties:
  - The `Type` property is an enumeration and describes what type of dump to collect. Valid values are `Triage`, `Mini`, `Heap`, and `Full`.
  - The `Egress` property describes which egress provider to use to egress the artifact. This egress provider must be described in the `Egress` root section of configuration.
- Create and register an `IValidateOptions<T>` instance for the action options.
  - The `Type` property is optional. Default is `Heap`.
  - The `Egress` property is required. Validate that the named egress provider does exist.

kkeirstead:
Includes one basic test for file system egress; open to including more if other tests are desired.

Closes #632


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/DumpTests.cs
```diff
@@ -66,73 +58,7 @@ public Task DumpTest(DiagnosticPortConnectionMode mode, DumpType type)
                     using ResponseStreamHolder holder = await client.CaptureDumpAsync(processId, type);
                     Assert.NotNull(holder);
 
-                    byte[] headerBuffer = new byte[64];
```
kkeirstead, line 69: This has been moved to `IDumpTestInterface`

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectDumpAction.cs
```diff
@@ -3,19 +3,80 @@
...
+        }
+
+        public async Task<CollectionRuleActionResult> ExecuteAsync(CollectDumpOptions options, IEndpointInfo endpointInfo, CancellationToken token)
+        {
+            DumpType dumpType = options.Type.GetValueOrDefault(CollectDumpOptionsDefaults.Type);
+            string egressProvider = options.Egress;
+
+            IProcessInfo processInfo = await _diagnosticServices.GetProcessAsync(new ProcessKey(endpointInfo.ProcessId), token);
+
+            string dumpFileName = DiagController.GenerateDumpFileName();
```
wiktork, line 36: Avoid this kind of coupling. Consider a separate utility class for generating the name rather than making Dump actions depend on WebServer controllers.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectDumpAction.cs
```diff
@@ -3,19 +3,80 @@
...
+
+            string dumpFilePath = "";
```
wiktork, line 38: nit: string.Empty

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfoSourceTests.cs
```diff
@@ -214,7 +214,7 @@ public static IEnumerable<object[]> GetTfmsSupportingPortListener()
             yield return new object[] { TargetFrameworkMoniker.Net60 };
         }
 
-        private ServerEndpointInfoSource CreateServerSource(out string transportName, ServerEndpointInfoCallback callback = null)
+        internal ServerEndpointInfoSource CreateServerSource(out string transportName, ServerEndpointInfoCallback callback = null)
```
wiktork, line 217: If these are useful as utility methods, refactor them to a separate class that can be used by multiple test classes. Test classes should not be calling to each other. Then the utility class can have public methods that are clearly meant to be consumed by other code.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/IDumpTestInterface.cs
```diff
@@ -0,0 +1,109 @@
...
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests
+{
+    public interface IDumpTestInterface
```
wiktork, line 18: I think what you really want here is a static class.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectDumpAction.cs
```diff
@@ -3,19 +3,80 @@
...
+                throw new ArgumentException("No Egress Provider was supplied.");
+            }
+            else
+            {
+                KeyValueLogScope scope = DiagController.GetDumpScope(processInfo);
+
+                try
+                {
+                    dumpFilePath = await SendToEgress(new EgressOperation(
+                        token => _diagnosticServices.GetDump(processInfo, dumpType, token),
```
jander-msft, line 53: Rather than using the `IDiagnosticServices` instance, consider refactoring `DiagnosticServices.GetDump` into a separate service interface and implementation (e.g. `IDumpService`) with a `Task<Stream> DumpAsync(IEndpointInfo, DumpType, CancellationToken)` method. This would alleviate the need to include the `IDiagnosticServices` implementation and its dependencies in tests and you would no longer need to call `IDiagnosticServices.GetProcessAsync`.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/TestHostHelper.cs
```diff
@@ -67,11 +69,23 @@ internal static class TestHostHelper
...
+                        {ConfigurationPath.Combine(ConfigurationKeys.Storage, nameof(StorageOptions.DumpTempFolder)), StorageOptionsDefaults.DumpTempFolder }
+                    });
                 })
-                .ConfigureServices(services =>
+                .ConfigureServices((HostBuilderContext context, IServiceCollection services) =>
                 {
                     services.ConfigureCollectionRules();
                     services.ConfigureEgress();
+
+                    services.AddSingleton<EgressOperationQueue>();
```
wiktork, line 83: Do we really need these dependencies? You are egressing directly from what I can see.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/TestHostHelper.cs
```diff
@@ -67,11 +69,23 @@ internal static class TestHostHelper
                     outputHelper.WriteLine("End Configuration");
 
                     builder.AddInMemoryCollection(configurationValues);
+
+                    builder.AddInMemoryCollection(new Dictionary<string, string>
```
jander-msft, line 73: Let's convert `DiagnosticsMonitorCommandHandler.ConfigureStorageDefaults` to an extension method in the `ConfigurationBuilderExtensions` class rather than copying the code here.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectDumpAction.cs
```diff
@@ -3,19 +3,80 @@
...
+                // Also, I would move this to Strings.resx if we do keep it, but I decided to wait for feedback before doing that.
+                throw new ArgumentException("No Egress Provider was supplied.");
+            }
+            else
+            {
+                KeyValueLogScope scope = DiagController.GetDumpScope(processInfo);
+
+                try
+                {
+                    dumpFilePath = await SendToEgress(new EgressOperation(
```
wiktork, line 52: Since we are going to have to egress from many other actions, it would be very useful if we had an egress facade that could be used by every action.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectDumpAction.cs
```diff
@@ -3,19 +3,80 @@
...
+            return new CollectionRuleActionResult()
+            {
+                OutputValues = new Dictionary<string, string>(StringComparer.Ordinal)
+                {
+                    { "EgressPath", dumpFilePath }
+                }
+            };
+        }
+
+        private async Task<string> SendToEgress(EgressOperation egressStreamResult, CancellationToken token)
```
jander-msft, line 75: I'm not sure having this separate adds much value, so inline it with the caller.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectDumpAction.cs
```diff
@@ -3,19 +3,80 @@
...
         {
-            throw new NotImplementedException("TODO: Implement action");
+            var result = await egressStreamResult.ExecuteAsync(_serviceProvider, token);
```
jander-msft, line 77: Since the result of `EgressOperation.ExecuteAsync` is not obvious, please use the concrete type instead of `var`.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectDumpAction.cs
jander-msft, line 75:
```suggestion
        private async Task<string> SendToEgress(EgressOperation egressOperation, CancellationToken token)
```

jander-msft (changes requested):

kkeirstead:
Currently still dealing with additional merge conflicts that are causing test failures; will push again once resolved.

jander-msft (changes requested):

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    public class Utilities
```
wiktork, line 10: internal static class.

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        public const string ArtifactType_Dump = "dump";
+        public const string ArtifactType_GCDump = "gcdump";
+        public const string ArtifactType_Logs = "logs";
+        public const string ArtifactType_Trace = "trace";
+        public const string ArtifactType_Metrics = "collectmetrics";
+
+        internal static string GenerateDumpFileName()
```
wiktork, line 18: Imo, internal classes with public methods are much better for this:
- It is clear the methods are meant to be consumed by any caller but the class as a whole is an implementation detail
- It gives room for true internal methods. Things designed to be called by tests or glue code between classes that shouldn't be called by just anyone
- It makes it very easy to switch to a public api should that ever be desired.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectDumpActionTests.cs
```diff
@@ -0,0 +1,127 @@
...
+
+        [Theory]
+        [InlineData(DumpType.Full)]
+        [InlineData(DumpType.Mini)]
+        [InlineData(DumpType.Triage)]
+        [InlineData(DumpType.WithHeap)]
+        [InlineData(null)]
+        public async Task CollectDumpAction_Success(DumpType? dumpType)
+        {
+            DumpType ExpectedDumpType = (dumpType != null) ? dumpType.Value : CollectDumpOptionsDefaults.Type;
```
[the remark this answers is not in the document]
kkeirstead: If we use a collection rule (as suggested in another comment), then we use `AddCollectDumpAction` which takes a `DumpType` and an egress provider. It looks like the extension method uses `DumpType` instead of `DumpType?`, though I see no reason why it couldn't be changed to accept an optional value. Since `DumpType` cannot be `null`, do we want to change the extension method to take a `DumpType?` to avoid needing to do `DumpType ExpectedDumpType = dumpType.GetValueOrDefault(CollectDumpOptionsDefaults.Type);`?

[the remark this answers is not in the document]
jander-msft: Yeah, change the extension method parameter to be `DumpType? type = null` and move it to the end. We'll want to keep the `egress` parameter without a default value since it is required.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectDumpAction.cs
```diff
@@ -3,19 +3,76 @@
...
-            throw new NotImplementedException("TODO: Implement action");
+            DumpType dumpType = options.Type.GetValueOrDefault(CollectDumpOptionsDefaults.Type);
+            string egressProvider = options.Egress;
+
+            string dumpFileName = Utils.GenerateDumpFileName();
+
+            string dumpFilePath = string.Empty;
+
+            // Given our options validation, I believe this is probably redundant...should I remove it?
+            if (string.IsNullOrEmpty(egressProvider))
```
kkeirstead, line 39: Should this stay or go?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfoSourceTests.cs
```diff
@@ -5,29 +5,28 @@
...
-using Microsoft.Diagnostics.Tools.Monitor;
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
 using Xunit.Abstractions;
+using static Microsoft.Diagnostics.Monitoring.Tool.UnitTests.EndpointUtilities;
```
jander-msft, line 15: Same comment above about removing the static.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectDumpAction.cs
jander-msft: Rather than checking individual fields, validate the object as it was designed to be validated:
```cs
ValidationContext context = new(options, _serviceProvider, items: null);
Validator.ValidateObject(options, context, validateAllProperties: true);
```
This will throw an exception if any of the properties are invalid.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectDumpAction.cs
```diff
@@ -3,19 +3,76 @@
...
-        ICollectionRuleAction<CollectDumpOptions>
+    internal sealed class CollectDumpAction : ICollectionRuleAction<CollectDumpOptions>
     {
-        public Task<CollectionRuleActionResult> ExecuteAsync(CollectDumpOptions options, IEndpointInfo endpointInfo, CancellationToken token)
+        private readonly IDumpService _dumpService;
+        private readonly IServiceProvider _serviceProvider;
+
+        public CollectDumpAction(IServiceProvider serviceProvider)
+        {
+            _serviceProvider = serviceProvider;
```
jander-msft, line 25:
```suggestion
            _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
```

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectDumpAction.cs
```diff
@@ -3,19 +3,76 @@
...
+            _dumpService = serviceProvider.GetRequiredService<IDumpService>();
+        }
+
+        public async Task<CollectionRuleActionResult> ExecuteAsync(CollectDumpOptions options, IEndpointInfo endpointInfo, CancellationToken token)
```
jander-msft, line 29: Validate parameters are not null (CancellationToken is a value type, so no need to check that).

jander-msft (changes requested):

src/Microsoft.Diagnostics.Monitoring.WebApi/DumpService.cs
```diff
@@ -0,0 +1,87 @@
...
+    internal class DumpService : IDumpService
+    {
+        private readonly IOptionsMonitor<StorageOptions> _storageOptions;
+
+        public DumpService(IOptionsMonitor<StorageOptions> storageOptions)
+        {
+            _storageOptions = storageOptions;
+        }
+
+        public async Task<Stream> DumpAsync(IEndpointInfo endpointInfo, Models.DumpType mode, CancellationToken token)
```
jander-msft, line 26: Validate endpoint info is not null.

src/Microsoft.Diagnostics.Monitoring.WebApi/DumpService.cs
```diff
@@ -0,0 +1,87 @@
...
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    internal class DumpService : IDumpService
+    {
+        private readonly IOptionsMonitor<StorageOptions> _storageOptions;
+
+        public DumpService(IOptionsMonitor<StorageOptions> storageOptions)
+        {
+            _storageOptions = storageOptions;
```
jander-msft, line 23: Validate storageOptions is not null

src/Microsoft.Diagnostics.Monitoring.WebApi/DumpService.cs
```diff
@@ -0,0 +1,87 @@
...
+        public async Task<Stream> DumpAsync(IEndpointInfo endpointInfo, Models.DumpType mode, CancellationToken token)
+        {
+            string dumpFilePath = Path.Combine(_storageOptions.CurrentValue.DumpTempFolder, FormattableString.Invariant($"{Guid.NewGuid()}_{endpointInfo.ProcessId}"));
+            DumpType dumpType = MapDumpType(mode);
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                // Get the process
+                Process process = Process.GetProcessById(endpointInfo.ProcessId);
+                    await Dumper.CollectDumpAsync(process, dumpFilePath, dumpType);
```
jander-msft, line 35: Fix indent

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectDumpActionTests.cs
```diff
@@ -0,0 +1,112 @@
...
+                    rootOptions.CreateCollectionRule(DefaultRuleName)
+                        .AddCollectDumpAction(ExpectedEgressProvider, ExpectedDumpType)
+                        .SetStartupTrigger();
+                }, async host =>
+                {
+                    IOptionsMonitor<CollectionRuleOptions> ruleOptionsMonitor = host.Services.GetService<IOptionsMonitor<CollectionRuleOptions>>();
+                    CollectDumpOptions options = (CollectDumpOptions)ruleOptionsMonitor.Get(DefaultRuleName).Actions[0].Settings;
+
+                    ICollectionRuleActionProxy action;
+                    host.Services.GetService<ICollectionRuleActionOperations>().TryCreateAction(KnownCollectionRuleActions.CollectDump, out action);
```
jander-msft, line 70: Assert.True

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectDumpActionTests.cs
```diff
@@ -0,0 +1,112 @@
...
+
+                    await runner.ExecuteAsync(async () =>
+                    {
+                        IEndpointInfo endpointInfo = await newEndpointInfoTask;
+
+                        using CancellationTokenSource cancellationTokenSource = new CancellationTokenSource(CommonTestTimeouts.DumpTimeout);
+                        CollectionRuleActionResult result = await action.ExecuteAsync(options, endpointInfo, cancellationTokenSource.Token);
+
+                        Assert.NotNull(result.OutputValues);
+                        Assert.True(result.OutputValues.TryGetValue("EgressPath", out string egressPath));
```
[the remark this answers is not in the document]
kkeirstead: I can do this for now since this is just for Dump, but seeing as most of our other actions are also going to be returning a result that uses "EgressPath" as a key, I feel like this should live somewhere that's agnostic of any specific action.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectDumpAction.cs
```diff
@@ -3,19 +3,82 @@
...
 {
-    internal sealed class CollectDumpAction :
-        ICollectionRuleAction<CollectDumpOptions>
+    internal sealed class CollectDumpAction : ICollectionRuleAction<CollectDumpOptions>
     {
-        public Task<CollectionRuleActionResult> ExecuteAsync(CollectDumpOptions options, IEndpointInfo endpointInfo, CancellationToken token)
+        private readonly IDumpService _dumpService;
+        private readonly IServiceProvider _serviceProvider;
+
+        internal const string egressPath = "EgressPath";
```
jander-msft, line 24:
```suggestion
        internal const string EgressPathOutputValueName = "EgressPath";
```

jander-msft (approved):

PR #825, opened 2021-09-03
Implement collection rule pipeline, event counter trigger, and tests.

jander-msft:
These changes create a new pipeline for executing a collection rule. The current implementation handles both non-startup and startup triggers, all action types, and enforces the configured limits. Tests were implemented to test the behavior of the pipeline for the startup trigger, event counter trigger, and collection rule limits.

Aspects deferred to future changes:
- Tests for validating negative cases (cancellation, bad-acting triggers, bad-acting actions).
- Asynchronous action execution.
- Logging

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRules/Actions/CallbackAction.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRules/Actions/CallbackAction.cs
@@ -1,0 +1,137 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Diagnostics.Monitoring.TestCommon;
+using Microsoft.Diagnostics.Monitoring.WebApi;
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Actions;
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit.Abstractions;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests.CollectionRules.Actions
+{
+    internal sealed class CallbackAction : ICollectionRuleAction<object>
+    {
+        public static readonly string ActionName = nameof(CallbackAction);
+
+        private readonly CallbackActionCallbackService _service;
+
+        public CallbackAction(CallbackActionCallbackService service)
+        {
+            _service = service;
+        }
+
+        public async Task<CollectionRuleActionResult> ExecuteAsync(object options, IEndpointInfo endpointInfo, CancellationToken token)
+        {
+            await _service.NotifyListeners(token);
+
+            return new CollectionRuleActionResult();
+        }
+    }
+
+    internal sealed class CallbackActionCallbackService
+    {
+        private readonly List<CompletionEntry> _entries = new();
+        private readonly SemaphoreSlim _entriesSemaphore = new(1);
+        private readonly List<DateTime> _executionTimestamps = new();
+        private readonly ITestOutputHelper _outputHelper;
+
+        private int _nextId = 1;
+
+        public CallbackActionCallbackService(ITestOutputHelper outputHelper)
+        {
+            _outputHelper = outputHelper ?? throw new ArgumentNullException(nameof(outputHelper));
+        }
+
+        public async Task NotifyListeners(CancellationToken token)
+        {
+            await _entriesSemaphore.WaitAsync(token);
+            try
+            {
+                lock (_executionTimestamps)
+                {
+                    _executionTimestamps.Add(DateTime.Now);
+                }
+                
+                _outputHelper.WriteLine("[Callback] Completing {0} source(s).", _entries.Count);
+
+                foreach (var entry in _entries)
+                {
+                    entry.Complete();
+                }
+                _entries.Clear();
+            }
+            finally
+            {
+                _entriesSemaphore.Release();
+            }
+        }
+
+        public async Task WaitWithCancellationAsync(CancellationToken token)
+        {
+            int id = _nextId++;
+            string name = $"Callback{id}";
+
+            CompletionEntry entry = new(_outputHelper, name);
+
+            await _entriesSemaphore.WaitAsync(token);
+            try
+            {
+                _outputHelper.WriteLine("[Test] Registering {0}.", name);
+
+                _entries.Add(entry);
+            }
+            finally
+            {
+                _entriesSemaphore.Release();
+            }
+
+            await entry.WithCancellation(token);
+        }
+
+        public IReadOnlyCollection<DateTime> ExecutionTimestamps
+        {
+            get
+            {
+                lock (_executionTimestamps)
+                {
+                    return _executionTimestamps.AsReadOnly();
+                }
+            }
+        }
+
+        private sealed class CompletionEntry
+        {
+            private readonly TaskCompletionSource<object> _completionSource =
+                new(TaskCreationOptions.RunContinuationsAsynchronously);
+            private readonly string _name;
+            private readonly ITestOutputHelper _outputHelper;
+
+            public CompletionEntry(ITestOutputHelper outputHelper, string name)
+            {
+                _name = name;
+                _outputHelper = outputHelper;
+            }
+
+            public void Complete()
+            {
+                _outputHelper.WriteLine("[Callback] Begin completing {0}.", _name);
+                if (!_completionSource.TrySetResult(null))
+                {
+                    _outputHelper.WriteLine("[Callback] Unable to complete {0}.", _name);
+                }
+                _outputHelper.WriteLine("[Callback] End completing {0}.", _name);
+            }
+
+            public async Task WithCancellation(CancellationToken token)
+            {
+                _outputHelper.WriteLine("[Test] Begin waiting for {0} completion.", _name);
+                await _completionSource.WithCancellation(token);
+                _outputHelper.WriteLine("[Test] End waiting for {0} completion.", _name);
+            }
+        }
+    }
+}
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRules/Actions/CallbackAction.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRules/Actions/CallbackAction.cs
@@ -17,9 +17,9 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests.CollectionRules.Action
     {
         public static readonly string ActionName = nameof(CallbackAction);
 
-        private readonly CallbackActionCallbackService _service;
+        private readonly CallbackActionService _service;
 
-        public CallbackAction(CallbackActionCallbackService service)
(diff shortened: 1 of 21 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/SpinWaitScenario.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        public static Task<int> ExecuteAsync(CancellationToken token)
+        {
+            return ScenarioHelpers.RunScenarioAsync(async logger =>
+            {
+                await ScenarioHelpers.WaitForCommandAsync(TestAppScenarios.SpinWait.Commands.StartSpin, logger);
+
+                Task continueTask = Task.Run(() => ScenarioHelpers.WaitForCommandAsync(TestAppScenarios.SpinWait.Commands.StopSpin, logger));
+
+                while (!continueTask.IsCompleted)
```
wiktork, line 35: Nit: could you achieve the same thing with SpinWait.SpinUntil

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRules/Actions/CallbackAction.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        public async Task<CollectionRuleActionResult> ExecuteAsync(object options, IEndpointInfo endpointInfo, CancellationToken token)
+        {
+            await _service.NotifyListeners(token);
+
+            return new CollectionRuleActionResult();
+        }
+    }
+
+    internal sealed class CallbackActionCallbackService
```
wiktork, line 35: nit: maybe just CallbackActionService

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/SpinWaitScenario.cs
jander-msft: I'll have to see if the produces enough detectable CPU usage, but I'll give it a try.

jander-msft: The SpinWait class yields the time slice of the current thread to other threads after a short time, so we end up using very little CPU. I've switched to using Thread.SpinWait which will utilize CPU on the thread for a number of "iterations", which is processor speed dependent. I chose an arbitrarily large number such that it doesn't constantly check the completion of the task; an iteration value of 1,000,000 seems to be about 100-150 ms on my Surface Book 2 when running on battery.

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -0,0 +1,248 @@
...
+            int actionCountLimit = (_ruleOptions.Limits?.ActionCount).GetValueOrDefault(CollectionRuleLimitsOptionsDefaults.ActionCount);
+            Queue<DateTime> executionTimestamps = new(actionCountLimit);
+
+            // Start cancellation timer for graceful stop of the collection rule
+            // when the rule duration has been specified. Conditionally enable this
+            // based on if the rule has a duration limit.
+            TimeSpan? ruleDuration = _ruleOptions.Limits?.RuleDuration;
+            if (ruleDuration.HasValue)
+            {
+                durationCancellationSource.CancelAfter(ruleDuration.Value);
```
wiktork, line 95: Shouldn't we only create this if there is a rule duration?

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -0,0 +1,248 @@
...
+            catch (OperationCanceledException) when (durationCancellationSource.IsCancellationRequested)
+            {
+                // This exception is caused by the pipeline duration expiring.
+                // Handle it to allow pipeline to be in completed state.
+            }
+        }
+
+        protected override Task OnCleanup()
+        {
+            _startedSource.TrySetCanceled();
```
wiktork, line 231: iirc OnCleanup will always happen not necessarily due to a cancellation.

src/Tools/dotnet-monitor/CollectionRules/Triggers/EventPipeTriggerFactory.cs
```diff
@@ -0,0 +1,91 @@
...
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Triggers
+{
+    internal sealed class EventPipeTriggerFactory
+    {
+        /// <summary>
+        /// Creates a collection rule trigger sourced from the event pipe of the target
+        /// process represented by the specified endpoint.
+        /// </summary>
+        public ICollectionRuleTrigger Create<TSettings>(
```
wiktork, line 22: So we should be able to use this factory for all EventPipe based triggers correct?

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -0,0 +1,248 @@
...
+
+                        // Start the trigger.
+                        await trigger.StartAsync(linkedToken).ConfigureAwait(false);
+
+                        // The pipeline signals that it has started just after starting a non-startup trigger.
+                        // Instances with startup triggers signal start after having finished executing the action list.
+                        if (trigger is not ICollectionRuleStartupTrigger)
+                        {
+                            // Signal that the pipeline trigger is initialized.
+                            _startedSource.TrySetResult(null);
```
wiktork, line 127: Do we have to special case startup triggers here btw? Perhaps their callback could return right away and satisfy the trigger condition?

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -0,0 +1,248 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules
+{
+    internal class CollectionRulePipeline : Pipeline
+    {
+        // The executor of the action list for the collection rule.
+        private readonly ActionListExecutor _actionListExecutor;
+
+        // The endpiont that represents the process on which the collection rule is executed.
```
kkeirstead, line 23:
```suggestion
        // The endpoint that represents the process on which the collection rule is executed.
```

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -0,0 +1,248 @@
...
+        private readonly IEndpointInfo _endpointInfo;
+
+        // The rule description that determines the behavior of the pipeline.
+        private readonly CollectionRuleOptions _ruleOptions;
+        
+        // Task completion source for signalling when the pipeline has finished starting.
+        private readonly TaskCompletionSource<object> _startedSource =
+            new(TaskCreationOptions.RunContinuationsAsynchronously);
+        
+        // Operations for getting triger information.
```
kkeirstead, line 33:
```suggestion
        // Operations for getting trigger information.
```

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -0,0 +1,248 @@
...
+                        }
+
+                        // Wait for the trigger to be satisfied.
+                        await triggerSatisfiedSource.WithCancellation(linkedToken).ConfigureAwait(false);
+                    }
+                    finally
+                    {
+                        try
+                        {
+                            // Intentially not using the linkedToken. If the linkedToken was signaled
```
kkeirstead, line 137:
```suggestion
                            // Intentionally not using the linkedToken. If the linkedToken was signaled
```

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -0,0 +1,248 @@
...
+
+        /// <summary>
+        /// Runs the pipeline to completion.
+        /// </summary>
+        /// <remarks>
+        /// The pipeline will only successfully complete in the following scenarios:
+        /// (1) the trigger is a startup trigger and the action list successfully executes once.
+        /// (2) without a specified action count window duration, the number of action list executions equals the action count limit.
+        /// </remarks>
+        protected override async Task OnRun(CancellationToken token)
```
wiktork, line 71: I don't think this is necessary for this PR, but maybe this method needs to be refactored into smaller parts.

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
jander-msft: Yes, because a startup trigger signals that the runtime should be resumed after all actions have run (by default) and a non-startup trigger signals that the runtime should be resume just after the trigger has started. The signal that the runtime is able to be resumed is by completing startup of the pipeline so that (in a future change) the collection rule service will just say "start a pipeline instance for each rule on this process" and then immediately resume the runtime after each pipeline declares that it has started.

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
jander-msft: Yeah, I was thinking of doing that too. I'll think about if I can do it now or in a subsequent change.

src/Tools/dotnet-monitor/CollectionRules/Triggers/EventPipeTriggerFactory.cs
jander-msft: Correct

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
jander-msft: That's fine because this is using TrySetCanceled. If a caller creates the pipeline, calls StartAsync, and disposes the pipeline before StartAsync completes, then the pipeline needs to cancel the started source so that the StartAsync method is canceled.

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
jander-msft: Conditionally creating a linked token source that needs to be conditionally disposed requires some extra code and acrobatics to conditionally create it, set the cancellation time, and dispose it. The way I have it just creates the linked token source and only applies the cancellation due to duration if the rule has a specified duration. Maybe this can be cleaned up better with a refactor of the OnRun method into smaller components.

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -0,0 +1,248 @@
...
+                                await asyncDisposableTrigger.DisposeAsync().ConfigureAwait(false);
+                            }
+                            else if (trigger is IDisposable disposableTrigger)
+                            {
+                                disposableTrigger.Dispose();
+                            }
+                        }
+                    }
+
+                    DateTime currentTimestamp = DateTime.Now;
```
wiktork, line 155: UtcNow. This can actually do strange things like go back in time on daylight savings time etc.

wiktork (approved):

Issue #439, opened 2021-06-15
Tool startup frequently hangs on MacOS tests


jander-msft:
Example test output when this occurs:
```
[Monitor] Wrote user settings.
[Monitor] User Settings Path: /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/ae05dced-2e19-42a3-a4c0-3db7e84ff21f/UserConfig/settings.json
[Monitor] Path: /Users/runner/work/1/s/artifacts/bin/dotnet-monitor/Debug/netcoreapp3.1/dotnet-monitor.dll
[Monitor] Args: collect --urls http://127.0.0.1:0 --metricUrls http://127.0.0.1:0
[Monitor] Begin Environment:
[Monitor] - COMPlus_EnableDiagnostics = 0
[Monitor] - Logging__Console__FormatterName = json
[Monitor] - Logging__LogLevel__Microsoft.Diagnostics.Tools.Monitor.Startup = Debug
[Monitor] - DotnetMonitorTestSettings__SharedConfigDirectoryOverride = /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/ae05dced-2e19-42a3-a4c0-3db7e84ff21f/SharedConfig
[Monitor] - DotnetMonitorTestSettings__UserConfigDirectoryOverride = /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/ae05dced-2e19-42a3-a4c0-3db7e84ff21f/UserConfig
[Monitor] End Environment:
[Monitor] Starting...
[Monitor] Process ID: 3014
[Monitor] Stopping...
[Monitor] Waiting for exit...
[Monitor] Exit Code: 137
[Monitor] Warning: Exception thrown while awaiting task: System.IO.IOException: Operation canceled
 ---> System.Net.Sockets.SocketException (89): Operation canceled
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken) in System.Net.Sockets.dll:token 0x60002e2+0x8
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.GetResult(Int16 token) in System.Net.Sockets.dll:token 0x60002dc+0x2d
   at System.IO.Pipes.PipeStream.ReadAsyncCore(Memory`1 destination, CancellationToken cancellationToken) in System.IO.Pipes.dll:token 0x60000ea+0xd1
[...]

PR #864, opened 2021-09-17
Cancel waiting for unresponsive processes.

### LINKED ISSUE #399: Dotnet-monitor hangs for apps spawned in the background
When creating apps in the background and using the hosting process (e.g. `dotnet run &`), dotnet-monitor will hang when making rest calls. So far, this has been reproduced on WSL2 with Ubuntu 18.04 and 20.04.

jander-msft:
This change allows dotnet-monitor to skip waiting for unresponsive processes in `connect` mode. Prior to this change, any unresponsive process that dotnet-monitor attempts to communicate with will cause HTTP requests to hang indefinitely until canceled. The change gives a brief window in which processes must respond; otherwise, dotnet-monitor will skip the inclusion of those offending processes in the process list.

This was tested on my local machine using WSL2; the unresponsive processes are skipped whereas the responsive ones are reported via the `/processes` route.

I will run the CI build several times to check that this fixes the `/processes` hang issue that is seen in various tests. Build results:
- Build [20210916.5](https://dev.azure.com/dnceng/public/_build/results?buildId=1367731&view=results): Success
- Build [20210916.6](https://dev.azure.com/dnceng/public/_build/results?buildId=1367766&view=results): Success
- Build [20210916.7](https://dev.azure.com/dnceng/public/_build/results?buildId=1367829&view=results): Success
- Build [20210916.8](https://dev.azure.com/dnceng/public/_build/results?buildId=1367846&view=results): Success
- Build [20210916.9](https://dev.azure.com/dnceng/public/_build/results?buildId=1367934&view=results): Failed; one startup issue on Mac (#439), one pipeline agent lost communication, otherwise all passed.

closes #399

### DIFF
--- a/src/Tools/dotnet-monitor/EndpointInfo/ClientEndpointInfoSource.cs
+++ b/src/Tools/dotnet-monitor/EndpointInfo/ClientEndpointInfoSource.cs
@@ -4,6 +4,7 @@
 
 using Microsoft.Diagnostics.Monitoring.WebApi;
 using Microsoft.Diagnostics.NETCore.Client;
+using Microsoft.Extensions.Logging;
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -17,8 +18,16 @@ namespace Microsoft.Diagnostics.Tools.Monitor
     {
         // The amount of time to wait before abandoning the attempt to create an EndpointInfo from
         // the enumerated processes. This may happen if a runtime instance is unresponsive to
-        // diagnostic pipe commands.
-        private readonly TimeSpan AbandonProcessTimeout = TimeSpan.FromSeconds(1);
+        // diagnostic pipe commands. Give a generous amount of time, but not too long since a single
+        // unresponsive process will cause all HTTP requests to be delayed by the timeout period.
+        private static readonly TimeSpan AbandonProcessTimeout = TimeSpan.FromSeconds(3);
+
+        private readonly ILogger<ClientEndpointInfoSource> _logger;
+
+        public ClientEndpointInfoSource(ILogger<ClientEndpointInfoSource> logger)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
 
         public async Task<IEnumerable<IEndpointInfo>> GetEndpointInfoAsync(CancellationToken token)
         {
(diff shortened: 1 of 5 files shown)


src/Tools/dotnet-monitor/EndpointInfo/ClientEndpointInfoSource.cs
```diff
@@ -26,7 +38,14 @@ public async Task<IEnumerable<IEndpointInfo>> GetEndpointInfoAsync(CancellationT
                 {
                     try
                     {
-                        return await EndpointInfo.FromProcessIdAsync(pid, token);
```
wiktork, line 29: Shouldn't each process be given the same timeout? That way if one process hangs, it doesn't abandon reading every subsequent process

src/Tools/dotnet-monitor/EndpointInfo/ClientEndpointInfoSource.cs
```diff
@@ -15,8 +15,20 @@ namespace Microsoft.Diagnostics.Tools.Monitor
 {
     internal sealed class ClientEndpointInfoSource : IEndpointInfoSourceInternal
     {
+        // The amount of time to wait before abandoning the attempt to create an EndpointInfo from
+        // the enumerated processes. This may happen if a runtime instance is unresponsive to
+        // diagnostic pipe commands.
+        private readonly TimeSpan AbandonProcessTimeout = TimeSpan.FromSeconds(1);
```
wiktork, line 21: Can you try this out with a very large number of processes? Will 1 second be enough to schedule and finish all the associated tasks?

wiktork (approved):

src/Tools/dotnet-monitor/EndpointInfo/ClientEndpointInfoSource.cs
jander-msft: I can make the timeout more generous, something like 3 or 5 seconds. We don't want to make it too long because then a single unresponsive process will cause all HTTP requests to be delayed by the timeout period. Maybe I can add a warning logging statement when abandoning getting process information for processes to make it more obvious to the user why requests take a "long" time.

src/Tools/dotnet-monitor/EndpointInfo/ClientEndpointInfoSource.cs
jander-msft: The timeout _is_ the same for all of the processes. In reality, each process gets _at least_ the timeout's worth of time to finish. The alternative is that I create a CancellationTokenSource for each process and start it for each process, but that would create a timer for each process.

Each of these is run as a separate task so as long as the threadpool is large enough, all of them can run simultaneously.

PR #877, opened 2021-09-18
Merge IProcessInfo and IEndpointInfo together

### LINKED ISSUE #869: Push process information into lower-level data structures
The collection rule service currently does not filter the collection rules according to the defined filters. The data structures that the filtering algorithm uses are too high up in the stack for it to be implemented. The process information needs to be pushed further down the implementation stack first and then process filtering for collection rules can be implemented.

jander-msft:
This change merges the concepts of IProcessInfo and IEndpointInfo together so that all aspects of IProcessInfo can be used throughout dotnet-monitor. This was done in two phases as separate commits:
1. Merged IProcessInfo and implementation into existing IEndpointInfo and implementation.
2. Rename IEndpointInfo and implementation to IProcessInfo and corresponding implementation.

closes #869


jander-msft:
Closing in favor of a simpler refactor.

PR #879, opened 2021-09-18
Collect dumps for hanging tests

### LINKED ISSUE #620: Test Fundamental: Capture dumps of all test processes for long running tests
The test runs should be collecting dumps for all test processes when they run long in order to be able to diagnose why some scenarios hang for a long time.

jander-msft:
Collect hang dump for the test host when individual tests run longer than 3 minutes.

closes #620

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -16,6 +16,9 @@
   <PropertyGroup Label="Arcade">
     <UsingToolXliff>false</UsingToolXliff>
     <UsingToolNetFrameworkReferenceAssemblies>false</UsingToolNetFrameworkReferenceAssemblies>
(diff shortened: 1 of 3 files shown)


wiktork (approved):

kelltrick (approved):
Can we get this change pulled in? I'm having a lot of issues getting a successful build and this would help/

jander-msft:
> Can we get this change pulled in? I'm having a lot of issues getting a successful build and this would help/

As it is, it will fail the tests and produce dumps, but the dumps won't be uploaded as artifacts. I'm going to add (for internal builds only) that and then I can merge this in.

PR #882, opened 2021-09-20
Implement collection rule filters.

### LINKED ISSUE #870: Implement collection rule process filtering

jander-msft:
This change filters collection rules based on the filters specified in configuration. It includes a refactor of the DiagnosticServices+ProcessInfo class to a separate ProcessInfoImpl class as well as a few basic tests to verify that the rules are able to be conditionally applied.

closes #870

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/ProcessInfoImpl.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/ProcessInfoImpl.cs
@@ -1,0 +1,131 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Diagnostics.Monitoring.EventPipe;
+using Microsoft.Diagnostics.NETCore.Client;
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    /// Named "ProcessInfoImpl" to disambiguate from Microsoft.Diagnostics.NETCore.client ProcessInfo
+    /// class returned from issuing GetProcesssInfo command on diagnostic pipe.
+    internal sealed class ProcessInfoImpl : IProcessInfo
+    {
+        // The amount of time to wait before cancelling get additional process information (e.g. getting
+        // the process command line if the IEndpointInfo doesn't provide it).
+        public static readonly TimeSpan ExtendedProcessInfoTimeout = TimeSpan.FromMilliseconds(1000);
+
+        // String returned for a process field when its value could not be retrieved. This is the same
+        // value that is returned by the runtime when it could not determine the value for each of those fields.
+        private static readonly string ProcessFieldUnknownValue = "unknown";
+
+        // The value of the operating system field of the ProcessInfo result when the target process is running
+        // on a Windows operating system.
+        private const string ProcessOperatingSystemWindowsValue = "windows";
+
+        public ProcessInfoImpl(
+            IEndpointInfo endpointInfo,
+            string commandLine,
+            string processName)
+        {
+            EndpointInfo = endpointInfo;
+
+            // The GetProcessInfo command will return "unknown" for values for which it does
+            // not know the value, such as operating system and process architecture if the
+            // process is running on one that is not predefined. Mimic the same behavior here
+            // when the extra process information was not provided.
+            CommandLine = commandLine ?? ProcessFieldUnknownValue;
+            ProcessName = processName ?? ProcessFieldUnknownValue;
+        }
+
+        public static async Task<IProcessInfo> FromEndpointInfoAsync(IEndpointInfo endpointInfo)
+        {
+            using CancellationTokenSource extendedInfoCancellation = new(ExtendedProcessInfoTimeout);
+            return await FromEndpointInfoAsync(endpointInfo, extendedInfoCancellation.Token);
+        }
+
+        // Creates a ProcessInfo object from the IEndpointInfo. Attempts to get the command line using event pipe
+        // if the endpoint information doesn't provide it. The cancelation token can be used to timebox this fallback
+        // mechansim.
+        public static async Task<IProcessInfo> FromEndpointInfoAsync(IEndpointInfo endpointInfo, CancellationToken extendedInfoCancellationToken)
+        {
+            if (null == endpointInfo)
+            {
+                throw new ArgumentNullException(nameof(endpointInfo));
+            }
+
+            DiagnosticsClient client = new(endpointInfo.Endpoint);
+
+            string commandLine = endpointInfo.CommandLine;
+            if (string.IsNullOrEmpty(commandLine))
+            {
+                try
+                {
+                    EventProcessInfoPipelineSettings infoSettings = new()
+                    {
+                        Duration = Timeout.InfiniteTimeSpan,
+                    };
+
+                    await using EventProcessInfoPipeline pipeline = new(client, infoSettings,
+                        (cmdLine, token) => { commandLine = cmdLine; return Task.CompletedTask; });
+
+                    await pipeline.RunAsync(extendedInfoCancellationToken);
+                }
+                catch
+                {
+                }
+            }
+
+            string processName = null;
+            if (!string.IsNullOrEmpty(commandLine))
+            {
+                // Get the process name from the command line
+                bool isWindowsProcess = false;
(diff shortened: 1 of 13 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/ProcessInfoImpl.cs
```diff
@@ -0,0 +1,131 @@
...
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    /// Named "ProcessInfoImpl" to disambiguate from Microsoft.Diagnostics.NETCore.client ProcessInfo
+    /// class returned from issuing GetProcesssInfo command on diagnostic pipe.
```
kkeirstead, line 16:
```suggestion
    /// class returned from issuing GetProcessInfo command on diagnostic pipe.
```

src/Microsoft.Diagnostics.Monitoring.WebApi/ProcessInfoImpl.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        public static async Task<IProcessInfo> FromEndpointInfoAsync(IEndpointInfo endpointInfo)
+        {
+            using CancellationTokenSource extendedInfoCancellation = new(ExtendedProcessInfoTimeout);
+            return await FromEndpointInfoAsync(endpointInfo, extendedInfoCancellation.Token);
+        }
+
+        // Creates a ProcessInfo object from the IEndpointInfo. Attempts to get the command line using event pipe
+        // if the endpoint information doesn't provide it. The cancelation token can be used to timebox this fallback
+        // mechansim.
```
kkeirstead, line 54:
```suggestion
        // mechanism.
```

src/Microsoft.Diagnostics.Monitoring.WebApi/ProcessInfoImpl.cs
```diff
@@ -0,0 +1,131 @@
...
+            ProcessName = processName ?? ProcessFieldUnknownValue;
+        }
+
+        public static async Task<IProcessInfo> FromEndpointInfoAsync(IEndpointInfo endpointInfo)
+        {
+            using CancellationTokenSource extendedInfoCancellation = new(ExtendedProcessInfoTimeout);
+            return await FromEndpointInfoAsync(endpointInfo, extendedInfoCancellation.Token);
+        }
+
+        // Creates a ProcessInfo object from the IEndpointInfo. Attempts to get the command line using event pipe
```
kkeirstead, line 52:
```suggestion
        // Creates an IProcessInfo object from the IEndpointInfo. Attempts to get the command line using event pipe
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorCollectRunner.CollectionRule.cs
```diff
@@ -38,19 +38,57 @@ public async Task WaitForCollectionRuleCompleteAsync(string ruleName, Cancellati
             }
         }
 
+        public async Task WaitForCollectionRuleUnmatchedFiltersAsync(string ruleName, CancellationToken token)
```
kkeirstead, line 41: Just noticing the code reuse; should we have these two methods just call a common method that takes the `LoggingEventId` as a parameter? Just seems like something we'd have to update twice since it looks like everything else in these methods is identical.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorCollectRunner.CollectionRule.cs
```diff
@@ -38,19 +38,57 @@ public async Task WaitForCollectionRuleCompleteAsync(string ruleName, Cancellati
...
-                        break;
                     case LoggingEventIds.CollectionRuleCompleted:
+                    case LoggingEventIds.CollectionRuleUnmatchedFilters:
+                    case LoggingEventIds.CollectionRuleStarted:
                         CompleteCollectionRuleCallbacks(key);
                         break;
+                    case LoggingEventIds.CollectionRuleFailed:
+                        FailCollectionRuleCallbacks(key, logEvent.Exception);
+                        break;
                 }
```
kkeirstead, line 92: Would it make sense for us to have a default here that throws an `ArgumentException`? You would know better than me, but in the event that there was a different `LoggingEventId` added in the future, that would prevent it from just falling through this switch statement undetected if we forgot to update it.

jander-msft: There a lot of other collection rule events that we aren't using in the tests. We would have to handle all of those here if we were to throw ArgumentException for unhandled events. So I don't think there's much value in doing that since we don't care about those events. The only reason why you'd want to add the event ID here is to have some kind of callback notification for that particular event, and we don't necessarily want to do that for all events unless we have to.

src/Microsoft.Diagnostics.Monitoring.WebApi/ProcessInfoImpl.cs
```diff
@@ -0,0 +1,131 @@
...
+        public static async Task<IProcessInfo> FromEndpointInfoAsync(IEndpointInfo endpointInfo, CancellationToken extendedInfoCancellationToken)
+        {
+            if (null == endpointInfo)
+            {
+                throw new ArgumentNullException(nameof(endpointInfo));
+            }
+
+            DiagnosticsClient client = new(endpointInfo.Endpoint);
+
+            string commandLine = endpointInfo.CommandLine;
```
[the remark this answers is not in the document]
josalem: That's odd. No, I'm not aware of an issue like this. This is only happening while the .net5 app is suspended?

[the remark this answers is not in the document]
jander-msft: Correct. .NET 5, non-Windows, suspended. This build demonstrates the symptoms: https://dev.azure.com/dnceng/public/_build/results?buildId=1372672&view=results

I can produce another build with more verbosing logging that demonstrates the issue precisely. I can also try adjusting the GetProcessInfoTests in the dotnet/diagnostics repo to demonstrate it more simply, but that's running on .NET 6 (which this does not repro), if I change them back to .NET 5.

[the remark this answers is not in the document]
jander-msft: I'm disabling the affected test for .NET 5 on non-Windows and logged an issue to track and investigate this later: https://github.com/dotnet/dotnet-monitor/issues/885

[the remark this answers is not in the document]
jander-msft: After disabling the test for .NET 5 on non-Windows, it started failing on .NET 6 for non-Windows. Also, I've updated the GetProcessInfoTests in dotnet/diagnostics to check that the values before and after consistent, which seems to be the case. So this is looking more like a dotnet-monitor bug than a runtime bug. I'm going to keep investigating this and update this thread with new info when its available.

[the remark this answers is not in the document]
jander-msft: Okay, now I have nearly definitive proof that the runtime (likely in both .NET 5 and .NET 6) only reports the process name for the command line field on the ProcessInfo reponse on non-Windows platforms. See this [build](https://dev.azure.com/dnceng/internal/_build/results?buildId=1374341&view=results), which is the same as the changes in https://github.com/dotnet/diagnostics/pull/2603, except that it writes out the command line at the end of the test and intentionally fails the test. It shows that is has a value of something like `/__w/1/s/.dotnet/dotnet`. This is occurring before and shortly after resuming the runtime. So this is likely another race condition where the command line arguments aren't available until sometime in the future after resuming the runtime. This does not repro on Windows.

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
```diff
@@ -146,12 +146,21 @@ public async ValueTask DisposeAsync()
 
                 if (null != options.Filters)
                 {
+                    IProcessInfo processInfo = await ProcessInfoImpl.FromEndpointInfoAsync(endpointInfo);
```
jander-msft, line 149: This should be moved into the ApplyRules method and then pass in the IProcessInfo instance since it will be the same amongst all of the rule instances for each process.

wiktork (approved):

PR #883, opened 2021-09-20
Create `CollectGCDump` Collection Rule Action

### LINKED ISSUE #658: Create `CollectGCDump` collection rule action

kkeirstead:
There's a fair number of similarities between this and the `CollectDump` Action. Most significant edits are either reorganization of existing functionality or an adjustment of code used for that action to fit this one. Once all the actions have been finished, I think there may be some benefit in trying to refactor similar code in the actions/tests into reusable methods, but for the sake of working on the actions simultaneously I've refrained from doing some of that for the time being.

Side note: I had originally attempted to rebase my changes based off main at Justin's request, but ended up having issues several times that broke my code. I ultimately merged with main, which appears to have worked fine; **however, if there are issues and a rebase is necessary, we can close this PR and I can take another stab at rebasing.**

Closes #658


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectGCDumpActionTests.cs
```diff
@@ -0,0 +1,127 @@
...
+                {
+                    uniqueEgressDirectory?.Delete(recursive: true);
+                }
+                catch
+                {
+                }
+            }
+        }
+
+        private static async Task ValidateGCDump(Stream gcdumpStream)
```
kkeirstead, line 104: This is loosely based off of the validation we currently do on Dumps. I don't know much about the composition of GCDumps, so this might be a poor approach, but I ended up cracking them open to look at the bytes instead of just checking that a file existed without examining its contents.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectionRuleActionConstants.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Actions
+{
+    internal class CollectionRuleActionConstants
+    {
+        public const string EgressPathOutputValueName = "EgressPath";
```
kkeirstead, line 9: I had mentioned this to Justin in the PR for `CollectDump`, but I wanted to reuse this variable between all the actions. If we want it to live somewhere else, we could do that, but I figured we didn't want everyone either 1) calling into the `CollectDump` action to get its value, or 2) copy-pasting this variable for each action.

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/CommonTestTimeouts.cs
```diff
@@ -23,6 +23,14 @@ public static class CommonTestTimeouts
         /// </summary>
         public static readonly TimeSpan WaitForExit = TimeSpan.FromSeconds(15);
 
+        /// <summary>
+        /// Default timeout for gcdump collection.
+        /// </summary>
+        /// <remarks>
+        /// GCDumps can be quite large and take a significant amount of time to transfer.
+        /// </remarks>
+        public static readonly TimeSpan GCDumpTimeout = TimeSpan.FromMinutes(3);
```
kkeirstead, line 32: This was arbitrarily chosen based off `DumpTimeout`; if someone with more intuition has a better value in mind, please feel free to let me know.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectGCDumpActionTests.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            byte[] buffer = new byte[24];
+
+            const string knownHeaderText = "!FastSerialization.1";
+
+            // Read enough to deserialize known header text.
+            int read;
+            int total = 0;
+            using CancellationTokenSource cancellation = new(CommonTestTimeouts.GCDumpTimeout);
+            while (total < buffer.Length && 0 != (read = await gcdumpStream.ReadAsync(buffer, total, buffer.Length - total, cancellation.Token)))
```
jander-msft, line 114: Since this is the same as in the dump utility, can you refactor this into an extension method? See https://github.com/dotnet/diagnostics/blob/main/src/Microsoft.Diagnostics.NETCore.Client/StreamExtensions.cs for an example.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectGCDumpActionTests.cs
```diff
@@ -0,0 +1,127 @@
...
+            {
+                total += read;
+            }
+
+            byte[] subarray = new byte[knownHeaderText.Length];
+            Array.Copy(buffer, 4, subarray, 0, subarray.Length); // The first header text character begins at the 4th index
```
jander-msft, line 120: The first 4 bytes are probably the character length of `!FastSerialization.1`.

jander-msft, line 120: Instead of extracting this to a new array, just use the `Encoding.GetString(byte[], int, int)` overload.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectGCDumpActionTests.cs
```diff
@@ -0,0 +1,127 @@
...
+                {
+                }
+            }
+        }
+
+        private static async Task ValidateGCDump(Stream gcdumpStream)
+        {
+            byte[] buffer = new byte[24];
+
+            const string knownHeaderText = "!FastSerialization.1";
```
jander-msft, line 108: Rather than doing the deserialization yourself, you can use `FastSerialization.Deserializer` type from `Microsoft.Diagnostics.FastSerialization` assembly. Just instantiate this type with the stream and it will parse this header out and throw an exception if that is not found.

Normally, I would recommend using the `GCHeapDump` type from `Microsoft.Diagnostics.Monitoring.EventPipe` assembly to do this, but I don't think the test assembly has the internal visibility to use it.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectGCDumpAction.cs
```diff
@@ -2,20 +2,71 @@
...
+            }
+
+            string egress = options.Egress;
+
+            ValidationContext context = new(options, _serviceProvider, items: null);
+            Validator.ValidateObject(options, context, validateAllProperties: true);
+
+            string gcdumpFileName = Utils.GenerateGCDumpFileName(endpointInfo);
+
+            string gcdumpFilePath = string.Empty;
```
jander-msft, line 45: nit: inline this with the `result.Result.Value` line.

jander-msft (changes requested):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectGCDumpActionTests.cs
kkeirstead: I'm a bit confused about how this fits in with the other changes; assuming the use of the newly created extension method and Encoding's GetString method, what is this responsible for doing?

jander-msft: Either do the other changes or do this. I was say do this one and ignore the others since it does exactly what you are doing. It probably requires the entire gcdump file to be available, which shouldn't be a problem because you've egressed it to the local file system.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectGCDumpAction.cs
```diff
@@ -2,20 +2,67 @@
...
+            }
+
+            if (endpointInfo == null)
+            {
+                throw new ArgumentNullException(nameof(endpointInfo));
+            }
+
+            string egress = options.Egress;
+
+            ValidationContext context = new(options, _serviceProvider, items: null);
```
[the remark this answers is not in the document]
jander-msft: They do, so as a wholly integrated system such as dotnet-monitor, yes, we don't need to do this. But from a "public-ish facing methods should always validate their parameters", this should still technically be done. Either way, I'm okay.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectGCDumpActionTests.cs
```diff
@@ -0,0 +1,102 @@
...
+using Xunit;
+using Xunit.Abstractions;
+using System.Text;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    public sealed class CollectGCDumpActionTests
+    {
+        private const string ExpectedEgressProvider = "TmpEgressProvider";
+        private const string DefaultRuleName = "Default";
```
jander-msft, line 28: nit: I typically change this name per group of tests to make sure that the collection rule system doesn't depend on any specific name, especially easy to guess ones like `Default`. Maybe rename to `GCDumpTestRule`.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectGCDumpActionTests.cs
```diff
@@ -0,0 +1,102 @@
...
+                    Assert.NotNull(gcdumpStream);
+
+                    ValidateGCDump(gcdumpStream);
+
+                    await runner.SendCommandAsync(TestAppScenarios.AsyncWait.Commands.Continue);
+                });
+            });
+        }
+
+        private static async void ValidateGCDump(Stream gcdumpStream)
```
jander-msft, line 88: Do not use `async void` without a really good reason to do so. This should be `async Task`.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectGCDumpActionTests.cs
```diff
@@ -0,0 +1,102 @@
...
+                    CollectionRuleActionResult result = await action.ExecuteAsync(options, endpointInfo, cancellationTokenSource.Token);
+
+                    Assert.NotNull(result.OutputValues);
+                    Assert.True(result.OutputValues.TryGetValue(CollectionRuleActionConstants.EgressPathOutputValueName, out string egressPath));
+                    Assert.True(File.Exists(egressPath));
+
+                    using FileStream gcdumpStream = new(egressPath, FileMode.Open, FileAccess.Read);
+                    Assert.NotNull(gcdumpStream);
+
+                    ValidateGCDump(gcdumpStream);
```
jander-msft, line 81: After changing to `async Task`, await this call.

jander-msft (changes requested):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectGCDumpActionTests.cs
kkeirstead: Oops, good catch, that was an artifact from when I was playing around with trying FastSerialization and removed the `async` and I forgot to change it back.

PR #884, opened 2021-09-20
Add support for Asp.net triggers

wiktork:
Depends on https://github.com/dotnet/diagnostics/pull/2592

### DIFF
--- a/src/Tools/dotnet-monitor/CollectionRules/Triggers/AspNetResponseStatusTriggerFactory.cs
+++ b/src/Tools/dotnet-monitor/CollectionRules/Triggers/AspNetResponseStatusTriggerFactory.cs
@@ -2,9 +2,14 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Microsoft.Diagnostics.Monitoring.EventPipe;
+using Microsoft.Diagnostics.Monitoring.EventPipe.Triggers;
+using Microsoft.Diagnostics.Monitoring.EventPipe.Triggers.AspNet;
 using Microsoft.Diagnostics.Monitoring.WebApi;
 using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Triggers;
 using System;
+using System.Globalization;
+using System.Linq;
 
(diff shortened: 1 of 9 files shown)


src/Tools/dotnet-monitor/CollectionRules/Triggers/AspNetResponseStatusTriggerFactory.cs
```diff
@@ -14,10 +19,45 @@ namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Triggers
...
+                StatusCodes = options.StatusCodes.Select(ParseRange).ToArray(),
+                SlidingWindowDuration = options.SlidingWindowDuration ?? TimeSpan.Parse(AspNetResponseStatusOptionsDefaults.SlidingWindowDuration, CultureInfo.InvariantCulture),
+            };
+
+            var aspnetTriggerSourceConfiguration = new AspNetTriggerSourceConfiguration();
+
+            return _eventPipeTriggerFactory.Create(endpointInfo, aspnetTriggerSourceConfiguration, _traceEventTriggerFactory, settings, callback);
+        }
+
+        private static StatusCodeRange ParseRange(string range)
```
kelltrick, line 51: I think this is a good start but I imagine we may want to expand this syntax to support something like: `200,202-204`

kelltrick (approved):
👍

src/Tools/dotnet-monitor/CollectionRules/Triggers/AspNetResponseStatusTriggerFactory.cs
jander-msft: Each item in the StatusCodes property above represents either a single code or a code range. An individual item should not have a combination of them. So this is permissable:

```json
[ "200", "202-204" ]
```

but not:

```json
[ "200,202-204" ]
```

jander-msft (approved):

src/Tools/dotnet-monitor/CollectionRules/Triggers/AspNetResponseStatusTriggerFactory.cs
wiktork: I will leave it as is for now, but if there's strong favor for having the payload be one string I can change it.

PR #894, opened 2021-09-21
Refactor actions to allow for async start and completion.

jander-msft:
These changes refactor the collection rule actions to have separate start and completion notions as well as use the factory pattern for creating the actions. This will allow (in a future change) to asynchronously start actions in the action list and synchronously wait for actions to complete when necessary.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectDumpActionTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectDumpActionTests.cs
@@ -91,14 +91,7 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
                         }
                         finally
                         {
-                            if (action is IAsyncDisposable asyncDisposableAction)
-                            {
-                                await asyncDisposableAction.DisposeAsync();
-                            }
-                            else if (action is IDisposable disposableAction)
-                            {
-                                disposableAction.Dispose();
-                            }
+                            await DisposableHelper.DisposeAsync(action);
                         }
 
                         Assert.NotNull(result.OutputValues);
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectDumpActionTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectDumpActionTests.cs
@@ -63,8 +63,8 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
                     IOptionsMonitor<CollectionRuleOptions> ruleOptionsMonitor = host.Services.GetService<IOptionsMonitor<CollectionRuleOptions>>();
                     CollectDumpOptions options = (CollectDumpOptions)ruleOptionsMonitor.Get(DefaultRuleName).Actions[0].Settings;
 
-                    ICollectionRuleActionProxy action;
-                    Assert.True(host.Services.GetService<ICollectionRuleActionOperations>().TryCreateAction(KnownCollectionRuleActions.CollectDump, out action));
+                    ICollectionRuleActionFactoryProxy factory;
+                    Assert.True(host.Services.GetService<ICollectionRuleActionOperations>().TryCreateFactory(KnownCollectionRuleActions.CollectDump, out factory));
 
                     EndpointInfoSourceCallback callback = new(_outputHelper);
                     await using var source = _endpointUtilities.CreateServerSource(out string transportName, callback);
@@ -78,11 +78,31 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
                     {
                         IEndpointInfo endpointInfo = await newEndpointInfoTask;
 
+                        ICollectionRuleAction action = factory.Create(endpointInfo, options);
+
                         using CancellationTokenSource cancellationTokenSource = new CancellationTokenSource(CommonTestTimeouts.DumpTimeout);
-                        CollectionRuleActionResult result = await action.ExecuteAsync(options, endpointInfo, cancellationTokenSource.Token);
+
+                        CollectionRuleActionResult result;
+                        try
+                        {
+                            await action.StartAsync(cancellationTokenSource.Token);
+
+                            result = await action.WaitForCompletionAsync(cancellationTokenSource.Token);
+                        }
+                        finally
+                        {
+                            if (action is IAsyncDisposable asyncDisposableAction)
+                            {
+                                await asyncDisposableAction.DisposeAsync();
+                            }
+                            else if (action is IDisposable disposableAction)
+                            {
+                                disposableAction.Dispose();
+                            }
+                        }
 
                         Assert.NotNull(result.OutputValues);
-                        Assert.True(result.OutputValues.TryGetValue(CollectDumpAction.EgressPathOutputValueName, out string egressPath));
+                        Assert.True(result.OutputValues.TryGetValue(CollectDumpActionFactory.EgressPathOutputValueName, out string egressPath));
                         Assert.True(File.Exists(egressPath));
 
                         using FileStream dumpStream = new(egressPath, FileMode.Open, FileAccess.Read);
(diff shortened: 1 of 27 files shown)


jander-msft:
I will merge and refactor https://github.com/dotnet/dotnet-monitor/pull/883 once it is completed.

src/Tools/dotnet-monitor/CollectionRules/Actions/ICollectionRuleAction.cs
```diff
@@ -2,19 +2,17 @@
...
-using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Actions
 {
-    internal interface ICollectionRuleAction<TOptions>
+    internal interface ICollectionRuleAction
```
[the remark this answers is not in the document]
jander-msft: Even if this method did exist, it would likely be used as:

```cs
using IDisposable _ = token.Register(() => action.Cancel(token));
```

and the code would likely still be conditionally disposing the action.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectDumpActionTests.cs
```diff
@@ -78,8 +78,28 @@ public async Task CollectDumpAction_Success(DumpType? dumpType)
...
+                        CollectionRuleActionResult result;
+                        try
+                        {
+                            await action.StartAsync(cancellationTokenSource.Token);
+
+                            result = await action.WaitForCompletionAsync(cancellationTokenSource.Token);
+                        }
+                        finally
+                        {
+                            if (action is IAsyncDisposable asyncDisposableAction)
```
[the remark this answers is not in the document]
wiktork: It would be nice to single source this, although I don't know about having extension methods on object

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectGCDumpAction.cs
```diff
@@ -2,67 +2,114 @@
...
-                throw new ArgumentNullException(nameof(endpointInfo));
+                _endpointInfo = endpointInfo ?? throw new ArgumentNullException(nameof(endpointInfo));
+                _options = options ?? throw new ArgumentNullException(nameof(options));
+                _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
             }
 
-            string egress = options.Egress;
+            public async ValueTask DisposeAsync()
+            {
+                _disposalTokenSource.SafeCancel();
```
wiktork, line 59: Would it make sense to push some of this shared infrastructure around cleanup into a base class for actions?

src/Tools/dotnet-monitor/CollectionRules/Actions/ExecuteAction.cs
```diff
@@ -16,76 +17,137 @@
...
             }
-            finally
+
+            public async Task StartAsync(CancellationToken token)
             {
-                process.Exited -= exitedHandler;
+                TaskCompletionSource<object> startCompleteSource = new(TaskCreationOptions.RunContinuationsAsynchronously);
+
+                CancellationToken disposalToken = _disposalTokenSource.Token;
+                _completionTask = Task.Run(() => ExecuteAsync(startCompleteSource, disposalToken), disposalToken);
```
wiktork, line 64: Do we not want to link the StartAsync token with the disposal token? In general, we don't really have IsDisposed state checking.

jander-msft: If those are linked together and passed into ExecuteAsync, then something could call StartAsync, its Task completes, and then the caller cancels the cancellation token. That would cause the execution of the action to be canceled.

It's my firm believe that the cancellation token that is passed into an async method, unless it is clearly obviously stated and documented otherwise, should only cancel the invocation of the method. Once the method completes, the token that was passed to it should not have any other affect on what that method did.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectGCDumpAction.cs
jander-msft: Should I go further than that and push all of DisposeAsync, StartAsync, and WaitForCompletion into a base class and have ExecuteAsync as an abstract method?

wiktork: I think that makes sense, there's a lot of shared functionality across these. We can always have multiple hierarchies if we find actions that greatly deviate in functionality.

src/Tools/dotnet-monitor/CollectionRules/Actions/ICollectionRuleAction.cs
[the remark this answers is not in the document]
wiktork: DisposeAsync essentially provides this functionality right?

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectDumpAction.cs
```diff
@@ -16,67 +16,127 @@
...
+                _endpointInfo = endpointInfo ?? throw new ArgumentNullException(nameof(endpointInfo));
+                _options = options ?? throw new ArgumentNullException(nameof(options));
+                _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
+                _dumpService = serviceProvider.GetRequiredService<IDumpService>();
+            }
 
-            KeyValueLogScope scope = Utils.CreateArtifactScope(Utils.ArtifactType_Dump, endpointInfo);
+            public async ValueTask DisposeAsync()
+            {
+                _disposalTokenSource.SafeCancel();
```
wiktork, line 64: Consider a _isDisposed flag that sets a disposed state and is used here as well as preventing calls to StartAync/WaitForCompletionAsync after disposal already happens.

src/Tools/dotnet-monitor/CollectionRules/Actions/ICollectionRuleAction.cs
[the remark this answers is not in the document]
jander-msft: Correct.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectionRuleActionBase.cs
```diff
@@ -0,0 +1,112 @@
...
+            // (which uses null in tests) will require this when needing to get process information for
+            // the actions property bag used for token replacement.
+            //EndpointInfo = endpointInfo ?? throw new ArgumentNullException(nameof(endpointInfo));
+            EndpointInfo = endpointInfo;
+            Options = options ?? throw new ArgumentNullException(nameof(options));
+        }
+
+        public async ValueTask DisposeAsync()
+        {
+            lock (_disposalTokenSource)
```
wiktork, line 38: nit: If you want to both remove the locking here and make sure that ThrowIfDisposed has the most up to date value for _isDisposed, use Interlocked.CompareExchange

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectionRuleActionBase.cs
```diff
@@ -0,0 +1,112 @@
...
+                source.TrySetCanceled(token);
+            }
+            else
+            {
+                source.TrySetException(ex);
+            }
+            return false;
+        }
+
+        private void ThrowIfDisposed()
```
wiktork, line 100: Consider using protected, so that derived classes can also use this check on new methods.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectionRuleActionBase.cs
```diff
@@ -0,0 +1,112 @@
...
+            // TODO: Allow null endpointInfo to allow tests to pass, but this should be provided by
+            // tests since it will be required by all aspects in the future. For example, the ActionListExecutor
+            // (which uses null in tests) will require this when needing to get process information for
+            // the actions property bag used for token replacement.
+            //EndpointInfo = endpointInfo ?? throw new ArgumentNullException(nameof(endpointInfo));
+            EndpointInfo = endpointInfo;
+            Options = options ?? throw new ArgumentNullException(nameof(options));
+        }
+
+        public async ValueTask DisposeAsync()
```
wiktork, line 36: nit: Not something we have to do now, but typically you would have another template method here to optionally override if derived classes needed to also dispose something.

wiktork (approved):

PR #900, opened 2021-09-22
Refactor collection rule startup into containers.

jander-msft:
This change refactors collection rule startup and running into a separate container class which each instance is associated with a specific process. This will allow easy incorporation of future changes, such as:
- Changing triggers and actions to be scoped to individual processes.
- Detecting when a process shuts down and stopping the rules associated with the process.
- Restarting rules for each previously discovered process when configuration changes.

### DIFF
--- a/src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
+++ b/src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
@@ -34,14 +34,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules
 
         public async ValueTask DisposeAsync()
         {
-            lock (_containers)
-            {
-                if (_disposed)
-                {
-                    return;
-                }
-                _disposed = true;
-            }
+            _disposed = true;
 
             foreach (CollectionRuleContainer container in _containers)
             {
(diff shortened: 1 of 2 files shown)


src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
```diff
@@ -3,50 +3,38 @@
...
             _provider = provider ?? throw new ArgumentNullException(nameof(provider));
-            _optionsMonitor = optionsMonitor ?? throw new ArgumentNullException(nameof(optionsMonitor));
-            _triggerOperations = triggerOperations ?? throw new ArgumentNullException(nameof(triggerOperations));
+            _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
         }
 
         public async ValueTask DisposeAsync()
         {
-            lock (_disposalTokenSource)
+            lock (_containers)
```
wiktork, line 37: nit: locking on containers here is a bit misleading since access to containers is otherwise unguarded.

wiktork (approved):

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
jander-msft: Yeah, it's not so much as to protect access to the _containers object, but to prevent simultaneous disposals from attempting to dispose everything.

jander-msft: I'm going to remove the lock and flag-check from here and just allow it to dispose each container. The application host is only going to call DisposeAsync once.

PR #909, opened 2021-09-24
Create CollectTrace Collection Rule Action

kkeirstead:
This was originally designed before Justin's PR to make the actions async; however, it's been adapted to fit in with Execute, CollectDump, and CollectGCDump. Includes test cases that do basic checking for the nettrace header (similar to what was used for GC Dumps).

Closes #688


src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities.cs
```diff
@@ -46,6 +55,78 @@ public static KeyValueLogScope CreateArtifactScope(string artifactType, IEndpoin
             return scope;
         }
 
+        public static AggregateSourceConfiguration GetTraceConfiguration(Models.TraceProfile profile, int metricsIntervalSeconds)
```
wiktork, line 58: https://github.com/dotnet/dotnet-monitor/pull/883#discussion_r713195397

kkeirstead: Yep, I meant to comment on that PR and apparently didn't; my plan is after I get all the actions in (so this PR and then the following one for Logs) to do some refactoring to improve organization and potentially improve the code reuse between the tests.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectTraceActionTests.cs
```diff
@@ -0,0 +1,159 @@
...
+                using FileStream traceStream = new(egressPath, FileMode.Open, FileAccess.Read);
+                Assert.NotNull(traceStream);
+
+                await ValidateTrace(traceStream);
+
+                await runner.SendCommandAsync(TestAppScenarios.AsyncWait.Commands.Continue);
+            });
+        }
+
+        private static async Task ValidateTrace(Stream traceStream)
```
jander-msft, line 140: The better way to do this is to use an EventPipeEventSource and verify that it can produce some event callback. See https://github.com/dotnet/diagnostics/blob/d425906a41dc8e2df0cf22766716feebe4634a5d/src/tests/Microsoft.Diagnostics.Monitoring.EventPipe/EventTracePipelineUnitTests.cs#L54 as an example.

jander-msft (changes requested):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectDumpActionTests.cs
```diff
@@ -24,9 +24,8 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
 {
     public sealed class CollectDumpActionTests
     {
-        private const string TempEgressDirectory = "/tmp";
         private const string ExpectedEgressProvider = "TmpEgressProvider";
-        private const string DefaultRuleName = "Default";
+        private const string DefaultRuleName = "DumpTestRule";
```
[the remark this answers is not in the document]
kkeirstead: Sorry, I had noticed it was out of date compared to what we had done with GC Dump so I had adjusted them in here since I figured it didn't need its own PR. That being said, I can pull these changes out so they go in their own PR. Yes, I just gave it a more descriptive Default Rule Name and switched to using TemporaryDirectory

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectTraceActionTests.cs
kkeirstead: I'm not entirely sure what to do here, so I'm going to try and clarify a few quick things. Also, if my understanding in 2 is completely off, just let me know :)

1. Do we want to completely take out the existing check for the headers, or are we supplementing it with this?
2. My understanding is that in the code you linked, line 61 gets run when there's an event callback (which is what we want to verify exists); do we just want to assert that we hit that line (potentially setting a boolean to true there that gets checked outside the block), or are we actually doing any validation on the objects in the trace itself?

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities.cs
```diff
@@ -46,6 +55,78 @@ public static KeyValueLogScope CreateArtifactScope(string artifactType, IEndpoin
...
+                    ));
+            }
+
+            return new EventPipeProviderSourceConfiguration(
+                providers: providers.ToArray(),
+                requestRundown: requestRundown,
+                bufferSizeInMB: bufferSizeInMB);
+        }
+
+        public static async Task GetTraceAction(IEndpointInfo endpointInfo, MonitoringSourceConfiguration configuration, TimeSpan duration, Stream outputStream, CancellationToken token)
```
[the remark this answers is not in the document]
kkeirstead: One small thing here - this gets called from within `DiagController`, which doesn't have the `startCompletionSource`. My current implementation just has `DiagController` passing null for that parameter and then this method doing a null check before `startCompletionSource.TrySetResult(null)`, but I wasn't sure if you'd consider that messy (since `token` is the last parameter, we can't give `startCompletionSource` a default null value, so I'm explicitly passing `null` for now)

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectTraceActionTests.cs
jander-msft: 1. Whole replacement. EventPipeEventSource will check that the format is as expected.
2. Set a boolean that any event was captured is fine and then assert that is is true after the Process call returns.

In the future, we can verify that we got the expected types of events if we really feel that is necessary.

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities.cs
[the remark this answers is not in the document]
jander-msft: That's absolutely fine to do. The HTTP requests don't care when it started, only that it finishes.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectDumpActionTests.cs
```diff
@@ -24,8 +24,9 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
 {
     public sealed class CollectDumpActionTests
     {
+        private const string TempEgressDirectory = "/tmp";
         private const string ExpectedEgressProvider = "TmpEgressProvider";
-        private const string DefaultRuleName = "DumpTestRule";
+        private const string DefaultRuleName = "Default";
```
kkeirstead: This was reverted to be in-line with what is currently in Main.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectTraceActionTests.cs
```diff
@@ -0,0 +1,159 @@
...
+                }
+
+                Assert.NotNull(result.OutputValues);
+                Assert.True(result.OutputValues.TryGetValue(CollectionRuleActionConstants.EgressPathOutputValueName, out string egressPath));
+                Assert.True(File.Exists(egressPath));
+
+                using FileStream traceStream = new(egressPath, FileMode.Open, FileAccess.Read);
+                Assert.NotNull(traceStream);
+
+                await ValidateTrace(traceStream, cancellationTokenSource.Token); // Do we want to be using a different token here? Does our timeout matter for this?
```
jander-msft, line 134:
```
                await ValidateTrace(traceStream, cancellationTokenSource.Token); // Do we want to be using a different token here? Does our timeout matter for this?
```

Yes, because capturing the trace would have taken a chunk of the time already. Rather than reusing the token source here, inside the ValidateTrace method, you can create a new CancellationTokenSource with a reasonable timeout e.g. 30 seconds. The validation shouldn't take more than a few seconds, if that, since the file stream should represent a completed trace.

jander-msft (approved):

PR #913, opened 2021-09-27
Allow action completion to be deferred.

jander-msft:
This change allows action completion to be deferred by default such that all actions are startup asynchronously (but in order), signal startup has completed, and then wait for completion on all deferred action completions. Actions can opt into synchronous completion by setting the WaitForCompletion property to true e.g.

```json
{
  "Type": "CollectTrace",
  "Settings": { ... },
  "WaitForCompletion": true
}
```

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionListExecutorTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionListExecutorTests.cs
@@ -15,6 +15,7 @@ using Xunit.Abstractions;
 using Microsoft.Extensions.Options;
 using Microsoft.Diagnostics.Tools.Monitor.CollectionRules;
 using Microsoft.Extensions.Logging;
+using System;
 
 namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
 {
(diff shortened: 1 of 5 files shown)


src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -226,5 +229,15 @@ public new Task StopAsync(CancellationToken token)
         {
             return base.StopAsync(token);
         }
+
+        // Ensures that the start callback is only invoked once.
+        private void InvokeStartCallback()
+        {
+            if (_invokedStartCallback)
```
kkeirstead, line 236: Should this be negated? As in, if we haven't invoked the callback, we do it (and set it to true) so we can only invoke it once?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionListExecutorTests.cs
```diff
@@ -75,23 +92,51 @@ public async Task ActionListExecutor_SecondActionFail()
...
+
                 CollectionRuleActionExecutionException actionExecutionException = await Assert.ThrowsAsync<CollectionRuleActionExecutionException>(
-                    () => executor.ExecuteActions(context, cancellationTokenSource.Token));
+                    () => executor.ExecuteActions(context, startCallback, cancellationTokenSource.Token));
 
                 Assert.Equal(1, actionExecutionException.ActionIndex);
 
                 Assert.Equal(string.Format(Strings.ErrorMessage_NonzeroExitCode, "1"), actionExecutionException.Message);
+
+                if (waitForCompletion)
```
kkeirstead, line 105: I realize it's really simple, so up to you, but for code reuse should this be put into a method? Just thinking that if we bulk up `ActionListExecutorTests` in the future we're going to have these lines repeated in even more places.

src/Tools/dotnet-monitor/CollectionRules/ActionListExecutor.cs
```diff
@@ -33,47 +35,100 @@ public ActionListExecutor(ILogger<CollectionRuleService> logger, ICollectionRule
...
+                            throw new InvalidOperationException(Strings.ErrorMessage_CouldNotMapToAction);
+                        }
+
+                        ICollectionRuleAction action = factory.Create(context.EndpointInfo, actionOption.Settings);
+
+                        try
+                        {
+                            await action.StartAsync(cancellationToken);
+
+                            // Check if the action completion should be awaited synchronously (in respected to
```
kkeirstead, line 68:
```suggestion
                            // Check if the action completion should be awaited synchronously (in respect to
```

src/Tools/dotnet-monitor/CollectionRules/ActionListExecutor.cs
```diff
@@ -33,47 +35,100 @@ public ActionListExecutor(ILogger<CollectionRuleService> logger, ICollectionRule
...
+                            // starting the next action). If not, add a deferred entry so that it can be completed
+                            // add the end of the execution list.
```
kkeirstead, line 70: I'm guessing this should be "at" or "after" the end of the execution list, but I may be misinterpreting the comment

src/Tools/dotnet-monitor/CollectionRules/ActionListExecutor.cs
```diff
@@ -33,47 +35,100 @@ public ActionListExecutor(ILogger<CollectionRuleService> logger, ICollectionRule
...
 
-                    ICollectionRuleAction action = factory.Create(context.EndpointInfo, actionOption.Settings);
+                    ++actionIndex;
+                }
+
+                // Notify that all actions have started
+                startCallback?.Invoke();
 
+                // Wait for any actions whose completion has been deferred.
+                for (int i = 0; i < deferredCompletions.Count; i++)
```
kkeirstead, line 102: More-so a thought than asking for a change, but I was curious about why we don't use a while loop here (e.g. `while (deferredCompletions.Count > 0`). Then we wouldn't have to play the game with incrementing and decrementing `i`, and when we assign `deferredCompletion` we could always use `deferredCompletions[0]`. In my head at least that feels more intuitive, but this also works.

wiktork: nit: Alternatively consider a queue

wiktork (approved):

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
jander-msft: Good catch; I'll adjust the tests to make sure this does in fact work after fixing it.

src/Tools/dotnet-monitor/CollectionRules/ActionListExecutor.cs
jander-msft: Either way, but yet, you are correct; I'll go with "after".

PR #914, opened 2021-09-27
Add support for publishing release to azure storage account.

kelltrick:


eng/PrepareRelease.yml
```diff
@@ -0,0 +1,43 @@
...
+      inputs:
+        targetType: filePath
+        filePath: '$(Build.Repository.LocalPath)/eng/release/Scripts/AcquireBuild.ps1'
+        arguments: '-BarBuildId "$(BARBuildId)" -AzdoToken "$(dn-bot-dotnet-all-scopes)" -MaestroToken "$(MaestroAccessToken)" -GitHubToken "$(BotAccount-dotnet-bot-repo-PAT)" -DownloadTargetPath "$(System.DefaultWorkingDirectory)\ReleaseTarget" -SasSuffixes "$(dotnetclichecksumsmsrc-dotnet-read-list-sas-token),$(dotnetclimsrc-read-sas-token)" -ReleaseVersion "$(Build.BuildNumber)"'
+        workingDirectory: '$(Build.Repository.LocalPath)'
+    - script: 'dotnet.cmd run --project $(Build.Repository.LocalPath)\eng\release\DiagnosticsReleaseTool\DiagnosticsReleaseTool.csproj -c Release -- prepare-release --input-drop-path "$(System.DefaultWorkingDirectory)\ReleaseTarget" --tool-manifest "$(Build.Repository.LocalPath)\eng\release\tool-list.json" --staging-directory "$(System.DefaultWorkingDirectory)\ReleaseStaging" --release-name "$(Build.BuildNumber)" --account-name "$(dotnet-diagnostics-storage-accountname)" --account-key "$(dotnet-diagnostics-storage-accountkey)" --sas-valid-days "$(dotnet-diagnostics-storage-retentiondays)" -v Ture'
```
jander-msft, line 36: Should be `-v True` instead of `-v Ture`

eng/PrepareRelease.yml
```diff
@@ -0,0 +1,43 @@
+stages:
+- stage: PrepareReleaseStage
```
jander-msft, line 2: This stage seems to run unconditionally, so long as the publish_using_darc stage runs; but that stage always runs regardless of whether the build is published to a channel or not. I feel this should only be run on release branches, or at the very least only on official branches; it probably shouldn't be run on private branches.

Not a requirement (it's not a big deal if my private branch publishes some files to the blob storage account... maybe), but something to think about.

eng/PrepareRelease.yml
jander-msft, line 36: Because this project is run on every build, I think it should be built as part of the normal build output so that someone doesn't accidentally regress its ability to build.

eng/PrepareRelease.yml
jander-msft, line 2: Because this is now tied into the product build, it has to produce the release assets correctly. Otherwise, you'd have to fix the publish issue and queue another build of the entire product. Consider a design where the product build produces the minimal assets and then have a separate yaml pipeline that can be run separately which takes the output of the product build. This would allow us, for example, to make minor changes to the release pipeline in the `main` branch yet publish bits from a release branch.

eng/PrepareRelease.yml
kelltrick: damn it, I swore I committed this typo fix

eng/PrepareRelease.yml
kelltrick: This was an intentional choice, my worry is that we will want to make a breaking change to the assets between say 6.0 and 7.0 and we won't be able to publish 6.0.x because it points at the 7.0 release tool. We may end up in a world where we need to port around a few changes to spin a 6.0.x build if we have new publishing guidelines. We also have the escape hatch of making a separate pipeline that takes as parameters the source build and main publishing tool and can "patch-up" an existing build if we need to.
My goal here was to make this as self-contained as possible to the branch and move as much as possible up into the build pipeline so as little as possible can go wrong during the "release".

eng/PrepareRelease.yml
kelltrick: I'm making this run `--version` on builds from branches that aren't `release/*`.

eng/PrepareRelease.yml
kelltrick: I'm making this run only from the `release/*` branches.

eng/release/DiagnosticsReleaseTool/Common/AzureBlobPublisher.cs
```diff
@@ -0,0 +1,267 @@
...
+                    {
+                        newClient = (await serviceClient.CreateBlobContainerAsync(_containerName, PublicAccessType.None, metadata: null, ct)).Value;
+                        _logger.LogInformation($"Created new contianer: {_containerName}.");
+                    }
+                    // 409 is the error code for an existing blob container already existing
+                    catch (RequestFailedException rfe) when (rfe.Status == 409)
+                    {
+                        _logger.LogInformation($"Found existing contianer: {_containerName}.");
+                        // Got an existing container
+                        newClient = serviceClient.GetBlobContainerClient(_containerName);
```
jander-msft, line 173: nit: Should this be done first before trying to create the container? I think there's an ExistsAsync method you can use to check that it truly exists before falling back to CreateBlobClientAsync

eng/release/DiagnosticsReleaseTool/Common/AzureBlobPublisher.cs
```diff
@@ -0,0 +1,267 @@
...
+                                PolicyStartsOn = new DateTimeOffset(DateTime.UtcNow),
+                                PolicyExpiresOn = new DateTimeOffset(DateTime.UtcNow.AddDays(_sasValidDays)),
+                            }
+                        };
+                        _logger.LogInformation($"Writing download access policy: {AccessPolicyDownloadId} to {_containerName}.");
+                        await newClient.SetAccessPolicyAsync(PublicAccessType.None, new BlobSignedIdentifier[] { downloadPolicyIdentifier }, cancellationToken: ct);
+                    }
+                    catch (Exception ex)
+                    {
+                        _logger.LogWarning(ex, $"Failed to write access policy for {_containerName}, retrying with new name.");
```
jander-msft, line 201: nit: This implies that a new name will be generated, but I don't see that behavior. It looks like it'll try again with the same container name.

jander-msft (approved):
See if @hoyosjs wants to take a look before merging.

eng/release/DiagnosticsReleaseTool/Common/AzureBlobPublisher.cs
```diff
@@ -0,0 +1,267 @@
...
+        private readonly int _sasValidDays;
+        private readonly ILogger _logger;
+
+        private BlobContainerClient _client;
+
+        private Uri AccountBlobUri
+        {
+            get
+            {
+                return new Uri($"https://{_accountName}.blob.core.windows.net");
```
wiktork, line 37: nit: FormattableString.Invariant

eng/release/DiagnosticsReleaseTool/Common/AzureBlobPublisher.cs
```diff
@@ -0,0 +1,267 @@
...
+                    await blobClient.UploadAsync(srcStream, overwrite: true, ct);
+
+                    BlobSasBuilder sasBuilder = new BlobSasBuilder()
+                    {
+                        BlobContainerName = client.Name,
+                        BlobName = blobClient.Name,
+                        Identifier = AccessPolicyDownloadId,
+                        Protocol = SasProtocol.Https,
+                    };
+                    Uri accessUri = blobClient.GenerateSasUri(sasBuilder);
```
wiktork, line 115: Did you want to be explicit on the expiration here?

eng/release/DiagnosticsReleaseTool/Common/AzureBlobPublisher.cs
```diff
@@ -0,0 +1,267 @@
...
+                        // Add the new (or update existing) "download" policy to the container
+                        // This is used to mint the SAS tokens without an expiration policy
+                        // Expiration can be added later by modifying this policy
+                        BlobSignedIdentifier downloadPolicyIdentifier = new BlobSignedIdentifier()
+                        {
+                            Id = AccessPolicyDownloadId,
+                            AccessPolicy = new BlobAccessPolicy()
+                            {
+                                Permissions = "r",
+                                PolicyStartsOn = new DateTimeOffset(DateTime.UtcNow),
```
wiktork, line 192: nit: UtcNow is evaluated twice. Not really meaningful in this scenario.

wiktork, line 192: Note that due to clock skew, if you set this policy and immediately attempt to access the resource it may fail. Consider using UtcNow - 1 min or something similar, unless this isn't really a concern here.

eng/release/DiagnosticsReleaseTool/DiagnosticsReleaseCommandLine.cs
```diff
@@ -73,12 +82,32 @@ static async Task<int> Main(string[] args)
...
+            new Option<string>(
+                aliases: new[] { "-c", "--container-name" },
+                description: "Storage account container name where the files will be uploaded.",
+                getDefaultValue: () => "dotnet-monitor");
+
+        private static Option<int> AzureStorageSasExpirationOption() =>
+            new Option<int>(
+                aliases: new[] { "--sas-valid-days" },
+                description: "Number of days to allow access to the blobs via the provided SAS URIs.",
+                getDefaultValue: () => 732); // default to just over 2 years
```
wiktork, line 111: I haven't worked with sas tokens that are based on policies, but normal sas tokens are tied to one of the account keys. Regularly scheduled key rotations would invalidate all the sas tokens (which normally isn't an issue since the tokens last much much less than 2 years). Would such rotation not affect these sas tokens?

wiktork (approved):

eng/release/DiagnosticsReleaseTool/Common/AzureBlobPublisher.cs
kelltrick: The ExistsAsync doesn't exist in this API (I'm using the blob client not the storage account client).

eng/release/DiagnosticsReleaseTool/Common/AzureBlobPublisher.cs
kelltrick: It's settable explicitly in the pipeline.

eng/release/DiagnosticsReleaseTool/DiagnosticsReleaseCommandLine.cs
kelltrick: That would affect these keys as well, the idea is that you don't need to do a key rotation at all because you can edit the policy to revoke access to an individual asset.

eng/release/DiagnosticsReleaseTool/Common/AzureBlobPublisher.cs
jander-msft:
```cs
newClient = serviceClient.GetBlobContainerClient(_containerName);
if (!await newClient.ExistsAsync(...))
{
    newClient = await (serviceClient.CreateBlobContainerAsync(_containerName, PublicAccessType.None, metadata: null, ct)).Value;
}
```
https://docs.microsoft.com/en-us/dotnet/api/azure.storage.blobs.blobserviceclient.getblobcontainerclient?view=azure-dotnet
https://docs.microsoft.com/en-us/dotnet/api/azure.storage.blobs.blobcontainerclient.existsasync?view=azure-dotnet#Azure_Storage_Blobs_BlobContainerClient_ExistsAsync_System_Threading_CancellationToken_
https://docs.microsoft.com/en-us/dotnet/api/azure.storage.blobs.blobserviceclient.createblobcontainerasync?view=azure-dotnet

eng/release/DiagnosticsReleaseTool/Common/AzureBlobPublisher.cs
kelltrick: I added a 15 minute clock skew compensation, this is the standard that Azure recommends.

kelltrick: fixed

eng/release/DiagnosticsReleaseTool/Common/AzureBlobPublisher.cs
kelltrick: Ahh, I didn't realize I the client creation didn't result in an API call to azure.

eng/release/DiagnosticsReleaseTool/DiagnosticsReleaseCommandLine.cs
```diff
@@ -73,12 +82,32 @@ static async Task<int> Main(string[] args)
...
+
+        private static Option<string> AzureStorageContainerNameOption() =>
+            new Option<string>(
+                aliases: new[] { "-c", "--container-name" },
+                description: "Storage account container name where the files will be uploaded.",
+                getDefaultValue: () => "dotnet-monitor");
+
+        private static Option<int> AzureStorageSasExpirationOption() =>
+            new Option<int>(
+                aliases: new[] { "--sas-valid-days" },
```
hoyosjs, line 109: 2 years feels a little long for how the intended functionality of this.

hoyosjs (approved):

eng/PrepareRelease.yml
```diff
@@ -0,0 +1,53 @@
...
+        workingDirectory: '$(Build.Repository.LocalPath)'
+      condition: and(succeeded(), eq(variables.isRelease, 'true'))
+    - script: 'dotnet.cmd run --project $(Build.Repository.LocalPath)\eng\release\DiagnosticsReleaseTool\DiagnosticsReleaseTool.csproj -c Release -- --version'
+      workingDirectory: '$(Build.Repository.LocalPath)\'
+      displayName: 'Build Manifest generation and asset publishing tool'
+      condition: and(succeeded(), eq(variables.isRelease, 'false'))
+    - script: 'dotnet.cmd run --project $(Build.Repository.LocalPath)\eng\release\DiagnosticsReleaseTool\DiagnosticsReleaseTool.csproj -c Release -- prepare-release --input-drop-path "$(System.DefaultWorkingDirectory)\ReleaseTarget" --tool-manifest "$(Build.Repository.LocalPath)\eng\release\tool-list.json" --staging-directory "$(System.DefaultWorkingDirectory)\ReleaseStaging" --release-name "$(Build.BuildNumber)" --account-name "$(dotnet-diagnostics-storage-accountname)" --account-key "$(dotnetstage-storage-key)" --sas-valid-days "$(dotnet-diagnostics-storage-retentiondays)" -v True'
```
hoyosjs, line 44: nitL The standard is to download/stage on `System.ArtifactsDirectory`

eng/release/DiagnosticsReleaseTool/DiagnosticsReleaseCommandLine.cs
```diff
@@ -1,11 +1,12 @@
...
+﻿using System;
+using System.CommandLine;
 using System.CommandLine.Builder;
 using System.CommandLine.Invocation;
 using System.CommandLine.Parsing;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
-
+using Azure.Storage.Blobs.Models;
```
hoyosjs, line 9: I don't think this gets used on this layer.

eng/release/DiagnosticsReleaseTool/DiagnosticsReleaseCommandLine.cs
```diff
@@ -73,12 +82,32 @@ static async Task<int> Main(string[] args)
...
+                description: "Storage account key, in base 64 format.")
+            {
+                IsRequired = true,
+            };
+
+        private static Option<string> AzureStorageContainerNameOption() =>
+            new Option<string>(
+                aliases: new[] { "-c", "--container-name" },
+                description: "Storage account container name where the files will be uploaded.",
+                getDefaultValue: () => "dotnet-monitor");
```
hoyosjs, line 105: Let me ask if they are ok with the second container getting created

eng/PrepareRelease.yml
```diff
@@ -0,0 +1,53 @@
+stages:
+- stage: PrepareReleaseStage
+  variables:
+    isRelease: $[in(variables['Build.SourceBranch'], 'refs/heads/release/')]
```
hoyosjs, line 4: Should be
```suggestion
    isRelease: $[startsWith(variables['Build.SourceBranch'], 'refs/heads/release/')]

```

eng/PrepareRelease.yml
```diff
@@ -0,0 +1,53 @@
...
+  displayName: Release Preparation
+  dependsOn:
+  - publish_using_darc
+  jobs:
+  - job: PrepareReleaseJob
+    displayName: Prepare release with Darc
+    pool: 
+      vmImage: windows-latest
+    variables:
+    - group: DotNet-Diagnostics-Storage
```
hoyosjs, line 14: These are only meant to be mapped in internal please. Can we split this file in two sections too please?:
```yml
# bunch of metadata and stage and job defs.
steps:
- ${{ if in(variables['Build.Reason'], 'PullRequest') }}:
  - # simple build case.
- ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest'), not(in(variables['Build.Reason'], 'PullRequest')) }}:
  -  real publish
```

eng/release/DiagnosticsReleaseTool/Common/AzureBlobPublisher.cs
```diff
@@ -0,0 +1,264 @@
...
+
+                    BlobSasBuilder sasBuilder = new BlobSasBuilder()
+                    {
+                        BlobContainerName = client.Name,
+                        BlobName = blobClient.Name,
+                        Identifier = AccessPolicyDownloadId,
+                        Protocol = SasProtocol.Https,
+                    };
+                    Uri accessUri = blobClient.GenerateSasUri(sasBuilder);
+
```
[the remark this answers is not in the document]
kelltrick: yes, the access policy controls read/write, read or write etc access

eng/PrepareRelease.yml
```diff
@@ -0,0 +1,49 @@
...
+  displayName: Release Preparation
+  dependsOn:
+  - publish_using_darc
+  jobs:
+  - job: PrepareReleaseJob
+    displayName: Prepare release with Darc
+    pool: 
+      vmImage: windows-latest
+    variables:
+    - ${{ if in(variables['Build.Reason'], 'PullRequest') }}:
```
hoyosjs, line 12:
```suggestion
    - ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest'), startsWith(variables['Build.SourceBranch'], 'refs/heads/release/')) }}:

```

kelltrick: done manually

PR #915, opened 2021-09-27
Migrate to .NET 6

jander-msft:
This change migrates dotnet-monitor to run only on the .NET 6 (or higher) runtime.

Overview of changes:
- Add net6.0 TFMs and fix build.
- Update tests to support running net6.0 dotnet-monitor.
- Remove netcoreapp3.1 TFM from dotnet-monitor and tests.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -55,7 +55,6 @@
     <MicrosoftAspNetCoreServerKestrelCoreVersion>2.1.7</MicrosoftAspNetCoreServerKestrelCoreVersion>
     <MicrosoftBclHashCodeVersion>1.1.0</MicrosoftBclHashCodeVersion>
     <MicrosoftExtensionsConfigurationAbstractionsVersion>5.0.0</MicrosoftExtensionsConfigurationAbstractionsVersion>
-    <MicrosoftExtensionsConfigurationKeyPerFileVersion>5.0.2</MicrosoftExtensionsConfigurationKeyPerFileVersion>
     <MicrosoftExtensionsLoggingAbstractionsVersion>5.0.0</MicrosoftExtensionsLoggingAbstractionsVersion>
     <MicrosoftExtensionsLoggingConsoleVersion>5.0.0</MicrosoftExtensionsLoggingConsoleVersion>
     <MicrosoftExtensionsLoggingEventSourceVersion>5.0.1</MicrosoftExtensionsLoggingEventSourceVersion>
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/InfoTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/InfoTests.cs
@@ -51,7 +51,8 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests
                     DotnetMonitorInfo info = await client.GetInfoAsync();
 
                     Assert.NotNull(info.Version); // Not sure of how to get Dotnet Monitor version from within tests...
-                    Assert.Equal(Environment.Version.ToString(), info.RuntimeVersion);
+                    Assert.True(Version.TryParse(info.RuntimeVersion, out Version runtimeVersion), "Unable to parse version from RuntimeVersion property.");
+                    Assert.True(runtimeVersion.Major >= 6, "RuntimVersion.Major is not greater than or equal to 6.");
                     Assert.Equal(mode, info.DiagnosticPortMode);
 
                     if (mode == DiagnosticPortConnectionMode.Connect)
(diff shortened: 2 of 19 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/InfoTests.cs
```diff
@@ -51,8 +51,9 @@ public Task InfoEndpointValidationTest(DiagnosticPortConnectionMode mode)
                     DotnetMonitorInfo info = await client.GetInfoAsync();
 
                     Assert.NotNull(info.Version); // Not sure of how to get Dotnet Monitor version from within tests...
-                    Assert.Equal(info.RuntimeVersion, Environment.Version.ToString());
-                    Assert.Equal(info.DiagnosticPortMode, mode);
+                    Assert.True(Version.TryParse(info.RuntimeVersion, out Version runtimeVersion), "Unable to parse version from RuntimeVersion property.");
+                    Assert.True(runtimeVersion.Major >= 6, "RuntimVersion.Major is not greater than or equal to 6.");
```
kkeirstead, line 55:
```suggestion
                    Assert.True(runtimeVersion.Major >= 6, "RuntimeVersion.Major is not greater than or equal to 6.");
```

src/Tools/dotnet-monitor/Auth/UserAuthorizationHandler.cs
```diff
@@ -42,6 +43,12 @@ protected override Task HandleRequirementAsync(AuthorizationHandlerContext conte
                     (context.User.Identity.AuthenticationType == AuthConstants.KerberosSchema) ||
                     (context.User.Identity.AuthenticationType == AuthConstants.NegotiateSchema))
             {
+                // Only supported on Windows
+                if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
jander-msft, line 47: Required because platform-specific APIs are statically checked since .NET 5.

src/Tools/dotnet-monitor/Startup.cs
```diff
@@ -46,7 +46,6 @@ public void ConfigureServices(IServiceCollection services)
                 // Allow serialization of enum values into strings rather than numbers.
                 options.JsonSerializerOptions.Converters.Add(new JsonStringEnumConverter());
             })
-            .SetCompatibilityVersion(CompatibilityVersion.Latest)
```
jander-msft, line 49: API was deprecated and generated a build warning-as-error. This API actually didn't do anything since it is a no-op since ASP.NET Core 3.0; probably a relic from when the project started as a ASP.NET Core 2.1 application.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunner.cs
```diff
@@ -72,6 +72,7 @@ public MonitorRunner(ITestOutputHelper outputHelper)
             // the correct ASP.NET Core version (which can be different than the .NET
             // version, especially for prereleases).
             _runner.FrameworkReference = DotNetFrameworkReference.Microsoft_AspNetCore_App;
+            _runner.TargetFramework = TargetFrameworkMoniker.Net60;
```
jander-msft, line 75: Rather than attempting to use the current TFM with dotnet-monitor (since that won't work if the test application is running as 3.1 or 5.0), always have dotnet-monitor run on .NET 6. We can adjust this later when testing roll-forward for .NET 7+.

eng/Versions.props
```diff
@@ -55,7 +55,6 @@
     <MicrosoftAspNetCoreServerKestrelCoreVersion>2.1.7</MicrosoftAspNetCoreServerKestrelCoreVersion>
     <MicrosoftBclHashCodeVersion>1.1.0</MicrosoftBclHashCodeVersion>
     <MicrosoftExtensionsConfigurationAbstractionsVersion>5.0.0</MicrosoftExtensionsConfigurationAbstractionsVersion>
-    <MicrosoftExtensionsConfigurationKeyPerFileVersion>5.0.2</MicrosoftExtensionsConfigurationKeyPerFileVersion>
     <MicrosoftExtensionsLoggingAbstractionsVersion>5.0.0</MicrosoftExtensionsLoggingAbstractionsVersion>
```
wiktork, line 58: Will these 5.0 packages automatically pick up higher versions?

jander-msft: Correct, they will automatically go to 6.0.0 in dotnet-monitor.

src/Tools/dotnet-monitor/ConfigurationBuilderExtensions.cs
```diff
@@ -4,101 +4,18 @@
...
-            KeyPerFileConfigurationSource,
-            IConfigurationSource
-        {
-            IConfigurationProvider IConfigurationSource.Build(IConfigurationBuilder builder)
-            {
-                return new KeyPerFileConfigurationProviderWithChangeTokenSupport(this);
-            }
-        }
-
-        private class KeyPerFileConfigurationProviderWithChangeTokenSupport :
```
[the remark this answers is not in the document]
jander-msft: Yes, it is supported. It was actually a bug in the original release that didn't get fixed until 6.0.0: https://github.com/dotnet/aspnetcore/issues/32836

wiktork (approved):

src/Tools/dotnet-monitor/ConfigurationBuilderExtensions.cs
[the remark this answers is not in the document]
jander-msft: FYI, the functional tests validate this via authentication tests (when rotating the API key) and I've manually validated it on an AKS sample.

PR #937, opened 2021-10-01
Restart collection rules when configuration changes.

### LINKED ISSUE #657: Add change token source for collection rule options when egress options change
Because several collection rule actions depend on egress providers, must implement an `IOptionsChangeTokenSource<T>` for the collection rule options that notifies that the options are invalidated when egress configuration changes. Scope this token source to only notify if collect actions are actively used by a collection rule to prevent egress changes from tearing down all active collection rules when the collect action is not used. (Although, how likely is it that a collection rule doesn't use a collect action?)

jander-msft:
This change plumbs through the ability to stop and restart collection rules on all processes when the collection rule configuration changes.

~~TODO: Add functional test that verifies that a rule change is actually reflected in the tool and can be observed.~~ Done

closes #657

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/CollectionRuleTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/CollectionRuleTests.cs
@@ -8,11 +8,13 @@ using Microsoft.Diagnostics.Monitoring.TestCommon.Runners;
 using Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests.Fixtures;
 using Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests.Runners;
 using Microsoft.Diagnostics.Monitoring.WebApi;
+using Microsoft.Diagnostics.Tools.Monitor;
 using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Triggers;
 using Microsoft.Extensions.DependencyInjection;
 using System;
 using System.IO;
 using System.Net.Http;
+using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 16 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/CollectionRuleTests.cs
```diff
@@ -247,6 +249,67 @@ public async Task CollectionRule_ProcessNameFilterNoMatchTest(DiagnosticPortConn
...
         }
 
+        /// <summary>
+        /// Validates that a change in the collection rule configuration is detected and applied correctly.
+        /// </summary>
+        [Theory]
+        [InlineData(DiagnosticPortConnectionMode.Listen)]
+        public async Task CollectionRule_ConfigurationChangeTest(DiagnosticPortConnectionMode mode)
+        {
+            string firstRuleName = "FirstRule";
```
kkeirstead, line 259: Should these be constants?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/TestTimeouts.cs
```diff
@@ -38,5 +38,10 @@ internal static class TestTimeouts
         /// Timeout for waiting for a collection rule to be filtered.
         /// </summary>
         public static readonly TimeSpan CollectionRuleFilteredTimeout = TimeSpan.FromSeconds(10);
+
+        /// <summary>
+        /// Timeout for waiting for a collection rule to be filtered.
```
kkeirstead, line 43: I think this comment got copy-pasted and needs to be updated.

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleContainer.cs
```diff
@@ -54,11 +54,64 @@ public async ValueTask DisposeAsync()
         {
             if (DisposableHelper.CanDispose(ref _disposalState))
             {
-                _disposalTokenSource.SafeCancel();
+                await StopRulesCore(CancellationToken.None);
```
kkeirstead, line 57: Do we want a token here or it fine to pass None?

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleContainer.cs
```diff
@@ -54,11 +54,64 @@ public async ValueTask DisposeAsync()
...
+            }
+        }
+
+        public async Task StartRulesAsync(
```
wiktork, line 61: I have no issues with this approach, but would it be easier to just destroy and recreate the entire rule container?

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleContainer.cs
```diff
@@ -54,11 +54,64 @@ public async ValueTask DisposeAsync()
...
+            IReadOnlyCollection<string> ruleNames,
+            CancellationToken token)
+        {
+            DisposableHelper.ThrowIfDisposed<CollectionRuleContainer>(ref _disposalState);
```
wiktork, line 65: Consider an assert or failure if we ever call Start twice in a row instead of Start/Stop.

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
```diff
@@ -55,42 +72,109 @@ public async ValueTask DisposeAsync()
...
 
-                IProcessInfo processInfo = await ProcessInfoImpl.FromEndpointInfoAsync(endpointInfo);
+                _logger.CollectionRuleConfigurationChanged();
 
-                CollectionRuleContainer container = new(
-                    _serviceProvider,
-                    _logger,
-                    processInfo);
-                _containers.Add(container);
+                // Get a copy of the container list to avoid having to
```
wiktork, line 116: So the scenario here is if we are restarting the existing set of containers, another one may be added correct?

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleContainer.cs
jander-msft: To do that, the service would need to keep a list of IEndpointInfo/IProcessInfo objects in order to recreate the containers. 

Arguably, this might be necessary anyway since the service/containers/pipelines have no way of detecting when the target process has exited; either the container or the service needs to poll the process periodically and then shutdown the rules when the process has exited. I was thinking that maybe the container could do this polling and then notify the service when the process is gone; this way, all of the process specific logic is contained in the container.

I was hoping to tackle that in the next change.

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
jander-msft: Yes, because configuration change can happen async to a process being discovered, I wanted some guards are modifying the collection of containers.

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
```diff
@@ -30,17 +33,31 @@ CollectionRulesConfigurationProvider provider
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
             _provider = provider ?? throw new ArgumentNullException(nameof(provider));
             _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
+
+            CancellationToken disposalToken = _disposalSource.Token;
+            _reapplyRulesTask = Task.Run(() => ReapplyRulesAsync(disposalToken), disposalToken).SafeAwait();
```
wiktork, line 38: Consider pushing the execution of this task into IHostedService instead of the ctor here.

wiktork (approved):

jander-msft:
The build failures in the latest build are a few tests that couldn't start dotnet-monitor on MacOS, which is an existing issue: #439 Bypassing successful PR validation since there are no other failures cause by this change.

PR #944, opened 2021-10-01
Use a singular option for all counter intervals

wiktork:
- [x] Prometheus metrics
- [x] EventCounter trigger
- [x] Trace
- [x] Custom trace
- [x] Live metrics
- [x] Custom live metrics
- [x] trace action
- [x] custom trace action

Not done:

- [ ] Asp.net duration heartbeat. Hardcoded in diagnostics repo
- [x] Make sure unit tests work
- [ ] Documentation changes

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.Options/GlobalCounterOptions.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.Options/GlobalCounterOptions.cs
@@ -2,14 +2,16 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Text;
+
 namespace Microsoft.Diagnostics.Monitoring.WebApi
 {
-    internal class MetricsOptionsDefaults
+    public class GlobalCounterOptions
     {
-        public const bool Enabled = true;
-
-        public const int MetricCount = 3;
-
-        public const bool IncludeDefaultProviders = true;
+        [Range(1, 3600 * 24)]
+        public int IntervalSeconds { get; set; }
     }
 }
--- a/src/Microsoft.Diagnostics.Monitoring.Options/GlobalCounterOptions.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.Options/GlobalCounterOptions.cs
@@ -12,6 +12,12 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
     public class GlobalCounterOptions
     {
         [Range(1, 3600 * 24)]
-        public int IntervalSeconds { get; set; }
+        public int? IntervalSeconds { get; set; }
+    }
+
+    internal static class GlobalCounterOptionsExtensions
+    {
+        public static int GetIntervalSeconds(this GlobalCounterOptions options) =>
+            options.IntervalSeconds.GetValueOrDefault(GlobalCounterOptionsDefaults.IntervalSeconds);
     }
 }
--- a/src/Microsoft.Diagnostics.Monitoring.Options/GlobalCounterOptions.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.Options/GlobalCounterOptions.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.ComponentModel.DataAnnotations;
 using System.Text;
 
@@ -11,7 +12,14 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
 {
     public class GlobalCounterOptions
     {
-        [Range(1, 3600 * 24)]
+        public const int IntervalMinSeconds = 1;
+        public const int IntervalMaxSeconds = 60 * 60 * 24; // One day
+
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_GlobalCounterOptions_IntervalSeconds))]
+        [Range(IntervalMinSeconds, IntervalMaxSeconds)]
+        [DefaultValue(GlobalCounterOptionsDefaults.IntervalSeconds)]
         public int? IntervalSeconds { get; set; }
     }
 
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.Metrics.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.Metrics.cs
@@ -28,7 +28,6 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
         /// <param name="uid">The Runtime instance cookie used to identify the target process.</param>
         /// <param name="name">Process name used to identify the target process.</param>
         /// <param name="durationSeconds">The duration of the metrics session (in seconds).</param>
-        /// <param name="metricsIntervalSeconds">The reporting interval (in seconds) for event counters.</param>
         /// <param name="egressProvider">The egress provider to which the metrics are saved.</param>
         [HttpGet("livemetrics", Name = nameof(CaptureMetrics))]
         [ProducesWithProblemDetails(ContentTypes.ApplicationJsonSequence)]
@@ -45,8 +44,6 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
             string name = null,
             [FromQuery][Range(-1, int.MaxValue)]
             int durationSeconds = 30,
-            [FromQuery][Range(1, int.MaxValue)]
-            int metricsIntervalSeconds = 5,
             [FromQuery]
             string egressProvider = null)
         {
@@ -60,9 +57,9 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
                 {
                     var client = new DiagnosticsClient(processInfo.EndpointInfo.Endpoint);
                     EventPipeCounterPipelineSettings settings = EventCounterSettingsFactory.CreateSettings(
+                        _counterOptions.CurrentValue,
                         includeDefaults: true,
-                        durationSeconds: durationSeconds,
-                        refreshInterval: metricsIntervalSeconds);
+                        durationSeconds: durationSeconds);
 
                     await using EventCounterPipeline eventCounterPipeline = new EventCounterPipeline(client,
                         settings,
@@ -89,7 +86,6 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
         /// <param name="uid">The Runtime instance cookie used to identify the target process.</param>
         /// <param name="name">Process name used to identify the target process.</param>
         /// <param name="durationSeconds">The duration of the metrics session (in seconds).</param>
-        /// <param name="metricsIntervalSeconds">The reporting interval (in seconds) for event counters.</param>
         /// <param name="egressProvider">The egress provider to which the metrics are saved.</param>
         [HttpPost("livemetrics", Name = nameof(CaptureMetricsCustom))]
         [ProducesWithProblemDetails(ContentTypes.ApplicationJsonSequence)]
@@ -108,8 +104,6 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
             string name = null,
             [FromQuery][Range(-1, int.MaxValue)]
             int durationSeconds = 30,
-            [FromQuery][Range(1, int.MaxValue)]
-            int metricsIntervalSeconds = 5,
             [FromQuery]
             string egressProvider = null)
         {
@@ -123,8 +117,8 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
                 {
                     var client = new DiagnosticsClient(processInfo.EndpointInfo.Endpoint);
                     EventPipeCounterPipelineSettings settings = EventCounterSettingsFactory.CreateSettings(
+                        _counterOptions.CurrentValue,
                         durationSeconds,
-                        metricsIntervalSeconds,
                         configuration);
 
                     await using EventCounterPipeline eventCounterPipeline = new EventCounterPipeline(client,
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
@@ -313,7 +313,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
             {
                 TimeSpan duration = Utilities.ConvertSecondsToTimeSpan(durationSeconds);
 
-                var aggregateConfiguration = Utilities.GetTraceConfiguration(profile, _counterOptions.CurrentValue.IntervalSeconds);
+                var aggregateConfiguration = Utilities.GetTraceConfiguration(profile, _counterOptions.CurrentValue.GetIntervalSeconds());
 
                 return StartTrace(processInfo, aggregateConfiguration, duration, egressProvider);
             }, processKey, Utilities.ArtifactType_Trace);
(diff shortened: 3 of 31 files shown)


src/Microsoft.Diagnostics.Monitoring.Options/GlobalCounterOptions.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Text;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    public class GlobalCounterOptions
+    {
+        [Range(1, 3600 * 24)]
+        public int IntervalSeconds { get; set; }
```
wiktork, line 15: Add to showconfig

src/Microsoft.Diagnostics.Monitoring.Options/GlobalCounterOptions.cs
```diff
@@ -0,0 +1,17 @@
...
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Text;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    public class GlobalCounterOptions
+    {
+        [Range(1, 3600 * 24)]
```
wiktork, line 14: Add display attributes

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.Metrics.cs
```diff
@@ -123,8 +117,8 @@ partial class DiagController
                 {
                     var client = new DiagnosticsClient(processInfo.EndpointInfo.Endpoint);
                     EventPipeCounterPipelineSettings settings = EventCounterSettingsFactory.CreateSettings(
+                        _counterOptions.CurrentValue,
```
jander-msft, line 120: Parameters are mixed up: duration is first, interval is second.

jander-msft: Nevermind, you rearranged parameters.

src/Microsoft.Diagnostics.Monitoring.Options/GlobalCounterOptions.cs
jander-msft, line 15: This should be nullabe.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -355,6 +355,15 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
 
             return InvokeForProcess(processInfo =>
             {
+                foreach(Models.EventPipeProvider provider in configuration.Providers)
+                {
+                    if (!CounterValidator.ValidateProviders(_counterOptions.CurrentValue,
+                        provider, out string errorMessage))
+                    {
+                        throw new InvalidOperationException(errorMessage);
```
jander-msft, line 363: MonitoringException or ValidationException

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
```diff
@@ -567,15 +558,12 @@ public Task CollectionRuleOptions_CollectTraceAction_PropertyValidation()
...
                 },
                 ex =>
                 {
                     string[] failures = ex.Failures.ToArray();
-                    Assert.Equal(5, failures.Length);
+                    Assert.Equal(4, failures.Length);
                     VerifyEnumDataTypeMessage<TraceProfile>(failures, 0, nameof(CollectTraceOptions.Profile));
-                    VerifyRangeMessage<int>(failures, 1, nameof(CollectTraceOptions.MetricsIntervalSeconds),
-                        ActionOptionsConstants.MetricsIntervalSeconds_MinValue_String, ActionOptionsConstants.MetricsIntervalSeconds_MaxValue_String);
                     VerifyRangeMessage<int>(failures, 2, nameof(CollectTraceOptions.BufferSizeMegabytes),
```
jander-msft, line 567: You'll need to adjust the indexes of the remain validation entries.

src/Microsoft.Diagnostics.Monitoring.Options/GlobalCounterOptions.cs
```diff
@@ -0,0 +1,17 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Text;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    public class GlobalCounterOptions
```
jander-msft, line 12: This needs to be added to https://github.com/dotnet/dotnet-monitor/blob/main/src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/CollectTraceOptions.Validate.cs
```diff
@@ -50,6 +54,13 @@ IEnumerable<ValidationResult> IValidatableObject.Validate(ValidationContext vali
 
                     Validator.TryValidateObject(provider, providerContext, results, validateAllProperties: true);
 
+                    IOptionsMonitor<GlobalCounterOptions> counterOptions = validationContext.GetRequiredService<IOptionsMonitor<GlobalCounterOptions>>();
+                    if (!CounterValidator.ValidateProviders(counterOptions.CurrentValue,
+                        provider, out string errorMessage))
+                    {
+                        results.Add(new ValidationResult(errorMessage, new[] { providerContext.MemberName }));
```
jander-msft, line 61:
```suggestion
                        results.Add(new ValidationResult(errorMessage, new[] { nameof(EventPipeProvider.Arguments) }));
```

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
```diff
@@ -236,6 +237,8 @@ public static IHostBuilder CreateHostBuilder(IConsole console, string[] urls, st
                     services.Configure<DiagnosticPortOptions>(context.Configuration.GetSection(ConfigurationKeys.DiagnosticPort));
                     services.AddSingleton<IValidateOptions<DiagnosticPortOptions>, DiagnosticPortValidateOptions>();
 
+                    services.ConfigureGlobalCounter(context.Configuration);
```
jander-msft, line 240: Why isn't this just `services.Configure<GlobalCounterOptions>(context.Configuration.GetSection(ConfigurationKeys.GlobalCounter)`? Especially if you are going to make the IntervalSeconds property nullable, you are going to have to use GetValueOrDefault anyway when accessing it.

jander-msft: Nevermind, I miss read it.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectTraceAction.cs
```diff
@@ -42,11 +44,13 @@ private sealed class CollectTraceAction :
             CollectionRuleActionBase<CollectTraceOptions>
         {
             private readonly IServiceProvider _serviceProvider;
+            private readonly IOptionsMonitor<GlobalCounterOptions> _counterOptions;
 
             public CollectTraceAction(IServiceProvider serviceProvider, IEndpointInfo endpointInfo, CollectTraceOptions options)
                 : base(endpointInfo, options)
             {
                 _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
+                _counterOptions = _serviceProvider.GetRequiredService<IOptionsMonitor<GlobalCounterOptions>>();
```
jander-msft, line 53: This will need to be configured on the configuration builder in the TestHostHelper: https://github.com/dotnet/dotnet-monitor/blob/42b94c9b77f6c30b5641c2d6bc2819ab4b8646ed/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/TestHostHelper.cs#L76

src/Microsoft.Diagnostics.Monitoring.WebApi/Validation/CounterValidator.cs
```diff
@@ -0,0 +1,37 @@
...
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Globalization;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi.Validation
+{
+    internal static class CounterValidator
+    {
+        public static bool ValidateProviders(GlobalCounterOptions counterOptions,
```
wiktork, line 15: nit: ValidateProvider

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
```diff
@@ -146,7 +146,6 @@ public Task CollectionRuleOptions_EventCounterTrigger_RoundTrip()
             const double ExpectedGreaterThan = 0.5;
             const double ExpectedLessThan = 0.75;
             TimeSpan ExpectedDuration = TimeSpan.FromSeconds(30);
-            const int ExpectedFrequency = 5;
```
wiktork, line 149: Note that metricsRefreshInterval is still in the test api, but it's simply ignored.

src/Microsoft.Diagnostics.Monitoring.Options/GlobalCounterOptions.cs
```diff
@@ -0,0 +1,31 @@
...
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_GlobalCounterOptions_IntervalSeconds))]
+        [Range(IntervalMinSeconds, IntervalMaxSeconds)]
+        [DefaultValue(GlobalCounterOptionsDefaults.IntervalSeconds)]
+        public int? IntervalSeconds { get; set; }
+    }
+
+    internal static class GlobalCounterOptionsExtensions
+    {
+        public static int GetIntervalSeconds(this GlobalCounterOptions options) =>
```
jander-msft, line 28: I like this better than our current scattering usage of GetValueOrDefault everywhere. Separately, I think we should do this with all of the nullable options properties.

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
```diff
@@ -313,12 +316,19 @@ private static void ConfigureMetricsEndpoint(IConfigurationBuilder builder, bool
             {
                 {ConfigurationPath.Combine(ConfigurationKeys.Metrics, nameof(MetricsOptions.Endpoints)), string.Join(';', metricEndpoints)},
                 {ConfigurationPath.Combine(ConfigurationKeys.Metrics, nameof(MetricsOptions.Enabled)), enableMetrics.ToString()},
-                {ConfigurationPath.Combine(ConfigurationKeys.Metrics, nameof(MetricsOptions.UpdateIntervalSeconds)), MetricsOptionsDefaults.UpdateIntervalSeconds.ToString()},
                 {ConfigurationPath.Combine(ConfigurationKeys.Metrics, nameof(MetricsOptions.MetricCount)), MetricsOptionsDefaults.MetricCount.ToString()},
                 {ConfigurationPath.Combine(ConfigurationKeys.Metrics, nameof(MetricsOptions.IncludeDefaultProviders)), MetricsOptionsDefaults.IncludeDefaultProviders.ToString()}
             });
         }
 
+        private static void ConfigureGlobalMetrics(IConfigurationBuilder builder)
```
jander-msft, line 324: This isn't really necessary because we are using the extension method you've added for the IntervalSeconds property.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
jander-msft: That can be fixed after these are merged.

jander-msft (approved):

jander-msft:
The only hesitation that I have about this change is that it is using `IOptionsMonitor<T>`, which allows the configuration to reflect updates from the binding sources. This means that the GlobalCounter:IntervalSeconds can be updated at any time and the new value will be used for new operations. Post preview, I think we need to consider caching the value per process when accessing it the first time for each process; this allows consistency throughout the lifetime of the process.

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
wiktork: I would still rather keep this. For example dotnet-monitor config show will actually produce a value here

wiktork:
> The only hesitation that I have about this change is that it is using `IOptionsMonitor<T>`, which allows the configuration to reflect updates from the binding sources. This means that the GlobalCounter:IntervalSeconds can be updated at any time and the new value will be used for new operations. Post preview, I think we need to consider caching the value per process when accessing it the first time for each process; this allows consistency throughout the lifetime of the process.

This is a reasonable concern. I think we can also address this by making sure things that depend on the interval properly reset their state when the configuration changes.

jander-msft:
> > The only hesitation that I have about this change is that it is using `IOptionsMonitor<T>`, which allows the configuration to reflect updates from the binding sources. This means that the GlobalCounter:IntervalSeconds can be updated at any time and the new value will be used for new operations. Post preview, I think we need to consider caching the value per process when accessing it the first time for each process; this allows consistency throughout the lifetime of the process.
> 
> 
> 
> This is a reasonable concern. I think we can also address this by making sure things that depend on the interval properly reset their state when the configuration changes.

I think the least acceptable scenario with this type of behavior would be cancelling a trace from the `/trace` route. But I'm not necessarily opposed to it.

src/Microsoft.Diagnostics.Monitoring.WebApi/Validation/CounterValidator.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        public static bool ValidateProvider(GlobalCounterOptions counterOptions,
+            EventPipeProvider provider,
+            out string errorMessage)
+        {
+            errorMessage = null;
+
+            if (provider.Arguments?.TryGetValue("EventCounterIntervalSec", out string intervalValue) == true)
+            {
+                if (int.TryParse(intervalValue, out int intervalSeconds) &&
```
jander-msft, line 23: The event counters infrastructure in the runtime accepts values that are parsable as `float`. I think we should do the same here (`float.TryParse`) so that a non-whole integer interval (e.g. 0.5) in the provider configuration doesn't get past this validation. Later, we can update the GlobalCounter:IntervalSeconds property and underlying pipelines to take a `float` instead of an `int` to be consistent with the runtime.

jander-msft (approved):

src/Tools/dotnet-monitor/DiagnosticsMonitorCommandHandler.cs
jander-msft: I actually don't know if I like that behavior though. I think it's also inconsistent as to what has default values as in-memory collections.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
```diff
@@ -183,21 +180,18 @@ public Task CollectionRuleOptions_EventCounterTrigger_PropertyValidation()
                         .SetEventCounterTrigger(out EventCounterOptions eventCounterOptions);
 
                     eventCounterOptions.SlidingWindowDuration = TimeSpan.FromSeconds(-1);
-                    eventCounterOptions.Frequency = -1;
                 },
                 ex =>
                 {
                     string[] failures = ex.Failures.ToArray();
                     // Property validation failures will short-circuit the remainder of the validation
                     // rules, thus only observe 4 errors when one might expect 5 (the fifth being that
```
kkeirstead, line 188: nit: Outdated comment

PR #954, opened 2021-10-04
Collection rule documentation for triggers and actions.

jander-msft:
This is the first pass at collection rule documentation, which includes documentation for each of the triggers and actions. In a separate PR, I will fill out of the filters, limits, and overall documentation of how collection rules work.

### DIFF
--- a/documentation/collectionrules.md
+++ b/documentation/collectionrules.md
@@ -1,3 +1,3 @@
 # Collection Rules
 
-`dotnet monitor` can be [configured](./configuration.md#Collection-Rule-Configuration) to automatically collection diagnostic artifacts based on conditions within the discovered processes.
+`dotnet monitor` can be [configured](./configuration.md#Collection-Rule-Configuration) to automatically collect diagnostic artifacts based on conditions within the discovered processes.
--- a/documentation/collectionrules.md
+++ b/documentation/collectionrules.md
@@ -1,1 +1,3 @@
 # Collection Rules
+
+`dotnet monitor` can be [configured](./configuration.md#Collection-Rule-Configuration) to automatically collection diagnostic artifacts based on conditions within the discovered processes.
--- a/documentation/configuration.md
+++ b/documentation/configuration.md
@@ -417,7 +417,7 @@ A trigger that has its condition satisfied when the number of HTTP requests is a
 | `IncludePaths` | string[] | false | The list of request path patterns to monitor. If not specified, all request paths are considered. If specified, only request paths matching one of the patterns in this list will be considered. Request paths matching a pattern in the `ExcludePaths` list will be ignored. | `null` | | |
 | `ExcludePaths` | string[] | false | The list of request path patterns to ignore. Request paths matching a pattern in this list will be ignored. | `null` | | |
 
-The `IncludePaths` and `ExcludePaths` support wildcards and globbing.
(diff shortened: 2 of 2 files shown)


documentation/collectionrules.md
```diff
@@ -1 +1,3 @@
-# Collection Rules
\ No newline at end of file
+# Collection Rules
+
+`dotnet monitor` can be [configured](./configuration.md#Collection-Rule-Configuration) to automatically collection diagnostic artifacts based on conditions within the discovered processes.
```
wiktork, line 3: This is probably a great spot to mention that in Preview 8 we only support triggers in reverse server mode, if you haven't already.

documentation/configuration.md
```diff
@@ -393,3 +393,239 @@ In addition to enabling custom providers, `dotnet monitor` also allows you to di
...
+}
+```
+
+#### `EventCounter` Trigger
+
+A trigger that has its condition satisfied when the value of a counter falls above, below, or between the described threshold values for a duration of time.
+
+##### Properties
+
+| Name | Type | Required | Description | Default Value | Min Value | Max Value |
```
wiktork, line 498: Not necessarily here, but we should probably have a link to the well-known .net counters docs.

documentation/configuration.md
```diff
@@ -393,3 +393,239 @@ In addition to enabling custom providers, `dotnet monitor` also allows you to di
...
+  "ProviderName": "System.Runtime",
+  "CounterName": "cpu-usage",
+  "GreaterThan": 70,
+  "SlidingWindowDuration": "00:00:10"
+}
+```
+
+#### ASP.NET Request Path Wildcards and Globbing
+
+The `IncludePaths` and `ExcludePaths` properties of the ASP.NET triggers allow for wildcards and globbing so that every included or excluded path does not necessarily need to be explicitly specified.
```
wiktork, line 521: It may be useful to indicate that exclude paths supersede include paths

wiktork (approved):

documentation/collectionrules.md
jander-msft: Yep, I'm going to describe the behavioral aspects here in the next PR.

kkeirstead, line 3:
```suggestion
`dotnet monitor` can be [configured](./configuration.md#Collection-Rule-Configuration) to automatically collect diagnostic artifacts based on conditions within the discovered processes.
```

documentation/configuration.md
```diff
@@ -393,3 +393,241 @@ In addition to enabling custom providers, `dotnet monitor` also allows you to di
...
+#### `AspNetRequestCount` Trigger
+
+A trigger that has its condition satisfied when the number of HTTP requests is above the described threshold level for a sliding window of time. The request paths can be filtered according to the described patterns.
+
+##### Properties
+
+| Name | Type | Required | Description | Default Value | Min Value | Max Value |
+|---|---|---|---|---|---|---|
+| `RequestCount` | int | true | The threshold of the number of requests that start within the sliding window of time. | | | |
+| `SlidingWindowDuration` | TimeSpan? | false | The sliding time window in which the number of requests are counted. | `"00:01:00"` (one minute) | `"1.00:00:00"` (1 day) | `"00:00:01"` (one second) |
```
kkeirstead, line 416: Min and Max are swapped (this also applies to all of the other triggers).

PR #975, opened 2021-10-07
Refactor ServerEndpointInfoSource to poll state of endpoint connections.

jander-msft:
This change refactors the ServerEndpointInfoSource into a background service implementation that polls the state of the endpoint connections periodically in order to notify when endpoint connections are no longer valid. The previous ability of getting a current snapshot of the endpoint list is preserved.

In a future change, I would like to merge the ClientEndpointInfoSource into this class, which would actively poll for new processes and place them into the polling lifetime algorithm that is in this change, in order to unify monitoring and callback mechanisms for both connect an listen modes.


src/Tools/dotnet-monitor/EndpointInfo/ServerEndpointInfoSource.cs
```diff
@@ -248,33 +138,130 @@ private async Task ResumeAndQueueEndpointInfo(IpcEndpointInfo info, Cancellation
...
+                _activeEndpointsSemaphore.Release();
+            }
+
+            if (endpointsToRemove.Count > 0)
+            {
+                lock (_pendingRemovalEndpoints)
+                {
+                    foreach (IEndpointInfo endpoint in endpointsToRemove)
+                    {
+                        _pendingRemovalEndpoints.Enqueue(endpoint);
```
wiktork, line 233: Rather than having a lock, queue, and semaphore for _pendingRemovalEndpoints, would it be easier to just make this a Channel that you dequeue from when data is available?

src/Tools/dotnet-monitor/EndpointInfo/ServerEndpointInfoSource.cs
```diff
@@ -248,33 +138,130 @@ private async Task ResumeAndQueueEndpointInfo(IpcEndpointInfo info, Cancellation
...
         }
 
-        private void VerifyNotDisposed()
+        private async Task MonitorEndpointsAsync(CancellationToken token)
         {
-            if (_disposed)
+            while (!token.IsCancellationRequested)
             {
-                throw new ObjectDisposedException(nameof(ServerEndpointInfoSource));
+                await Task.Delay(TimeSpan.FromSeconds(3));
```
wiktork, line 156: nit: move to constant.

src/Tools/dotnet-monitor/EndpointInfo/ServerEndpointInfoSource.cs
```diff
@@ -248,33 +162,119 @@ private async Task ResumeAndQueueEndpointInfo(IpcEndpointInfo info, Cancellation
...
+                }
+            }
+            finally
+            {
+                _activeEndpointsSemaphore.Release();
+            }
+
+            if (endpointsToRemove.Count > 0)
+            {
+                foreach (IEndpointInfo endpoint in endpointsToRemove)
```
wiktork, line 247: We are pruning on a timer and every time we get all the endpoints. Maybe do the writing inside the semaphore to avoid multiple writers?

wiktork (approved):

PR #987, opened 2021-10-11
Triggers - Actions Refactoring

kkeirstead:
Through the development of the Actions, it became apparent that some code could be shared or better organized (as well as other miscellaneous improvements). Now that all of the initial Actions are complete, this is meant to make some of those improvements.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/TestConditions.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/TestConditions.cs
@@ -2,9 +2,10 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Microsoft.Diagnostics.Monitoring.TestCommon;
 using System.Runtime.InteropServices;
 
-namespace Microsoft.Diagnostics.Monitoring.TestCommon
+namespace Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests
 {
     internal static class TestConditions
     {
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectDumpActionTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectDumpActionTests.cs
@@ -32,10 +32,16 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
             _endpointUtilities = new(_outputHelper);
         }
 
-        [ConditionalTheory(typeof(TestConditions), nameof(TestConditions.IsDumpSupported))]
+        [Theory]
         [MemberData(nameof(ActionTestsHelper.GetTfmsAndDumpTypes), MemberType = typeof(ActionTestsHelper))]
         public async Task CollectDumpAction_Success(TargetFrameworkMoniker tfm, DumpType? dumpType)
         {
+            // MacOS dumps inconsistently segfault the runtime on .NET 5: https://github.com/dotnet/dotnet-monitor/issues/174
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX) && tfm == TargetFrameworkMoniker.Net50)
+            {
+                return;
+            }
+
             using TemporaryDirectory tempDirectory = new(_outputHelper);
 
             await TestHostHelper.CreateCollectionRulesHost(_outputHelper, rootOptions =>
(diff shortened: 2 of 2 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.ExecuteActionApp/Microsoft.Diagnostics.Monitoring.ExecuteActionApp.csproj
```diff
@@ -1,10 +1,14 @@
...
-    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>net6.0</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="xunit.assert" Version="$(XunitAssertVersion)" />
   </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.Diagnostics.Monitoring.Tool.UnitTests\Microsoft.Diagnostics.Monitoring.Tool.UnitTests.csproj" />
```
jander-msft, line 12: This is causing the failures due to the mismatch xunit.assert dependencies. The Microsoft.Diagnostics.Monitoring.Tool.UnitTests assembly references 2.4.2.0 (even though we ask it to build against 2.4.1.0); it's probably upgraded because it is known to be a unit test project. However, the ExecuteActionApp is not a test project, so it'll use whatever version we tell it to use, which is 2.4.1.0.

I think you should remove this ProjectReference and either move whatever common information that is needed to TestCommon or just link it in (the latter is probably easier).

src/Tests/Microsoft.Diagnostics.Monitoring.ExecuteActionApp/Microsoft.Diagnostics.Monitoring.ExecuteActionApp.csproj
```diff
@@ -1,10 +1,15 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>net6.0</TargetFrameworks>
```
jander-msft, line 4: You need to update https://github.com/dotnet/dotnet-monitor/blob/6b1993792fdb0a00e9bfe06790b60a8cf20a2c3d/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExecuteActionTestHelper.cs#L17 to use `Net60` instead of `NetCoreApp31`.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests.csproj
```diff
@@ -14,4 +14,8 @@
     <ProjectReference Include="..\Microsoft.Diagnostics.Monitoring.TestCommon\Microsoft.Diagnostics.Monitoring.TestCommon.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <InternalsVisibleTo Include="Microsoft.Diagnostics.Monitoring.ExecuteActionApp" />
```
jander-msft, line 18: This isn't necessary anymore.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionTestsHelper.cs
```diff
@@ -0,0 +1,101 @@
...
+        internal static string ValidateEgressPath(CollectionRuleActionResult result)
+        {
+            Assert.NotNull(result.OutputValues);
+            Assert.True(result.OutputValues.TryGetValue(CollectionRuleActionConstants.EgressPathOutputValueName, out string egressPath));
+            Assert.True(File.Exists(egressPath));
+
+            return egressPath;
+        }
+
+        internal static T GetOptions<T>(string ruleName, IHost host)
```
jander-msft, line 74: nit: GetActionOptions, to be more precise about what options it is getting.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionTestsHelper.cs
```diff
@@ -0,0 +1,101 @@
...
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    internal static class ActionTestsHelper
+    {
+        public static TargetFrameworkMoniker[] tfmsToTest = new TargetFrameworkMoniker[] { TargetFrameworkMoniker.Net50, TargetFrameworkMoniker.Net60 };
```
jander-msft, line 24: Although collection rules are only supported on .NET 5+, the actions themselves do not have this limitation. Technically, you should be able to add NetCoreApp31 and all of the actions should work.

kkeirstead: I ended up not adding this one in; I was getting test failures (Task Cancellations) on every actions test with 3.1, though occasionally that type of thing happens on my local machine but isn't actually an issue. If you definitely want it in, just let me know and I'll push out another update including it.

PR #988, opened 2021-10-11
Remove collection rules when process is no longer detected.

### LINKED ISSUE #962: Stop collection rules when process shuts down

jander-msft:
The change removes collection rules associated with a process when that process is no longer detected (e.g. it exited, crashed, etc). A test has been added to verify that removal is detectable.

closes #962

### DIFF
--- a/src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
+++ b/src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
@@ -18,6 +18,10 @@ namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules
 {
     internal class CollectionRuleService : BackgroundService, IAsyncDisposable
     {
+        // The number of items that the pending removal channel will hold before forcing
+        // the writer to wait for capacity to be available.
+        private const int PendingRemovalChannelCapacity = 1000;
+
         private readonly Dictionary<IEndpointInfo, CollectionRuleContainer> _containersMap = new();
         private readonly ChannelReader<CollectionRuleContainer> _containersToRemoveReader;
         private readonly ChannelWriter<CollectionRuleContainer> _containersToRemoveWriter;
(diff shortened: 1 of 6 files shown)


src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
```diff
@@ -32,16 +36,27 @@ CollectionRulesConfigurationProvider provider
...
+            BoundedChannelOptions containersToRemoveChannelOptions = new(1000)
+            {
+                SingleReader = true,
+                SingleWriter = true
+            };
+
+            Channel<CollectionRuleContainer> containersToRemoveChannel =
+                Channel.CreateBounded<CollectionRuleContainer>(containersToRemoveChannelOptions);
+            _containersToRemoveReader = containersToRemoveChannel.Reader;
+            _containersToRemoveWriter = containersToRemoveChannel.Writer;
```
wiktork, line 49: nit: Consider completing the writer during disposal

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
```diff
@@ -114,13 +152,13 @@ protected override async Task ExecuteAsync(CancellationToken token)
                 // Get a copy of the container list to avoid having to
                 // lock the entire list during stop and restart of all containers.
                 CollectionRuleContainer[] containers;
-                lock (_containers)
+                lock (_containersMap)
                 {
-                    containers = _containers.ToArray();
+                    containers = _containersMap.Values.ToArray();
```
wiktork, line 157: I don't think this is an issue but the order in which we operate on the containers won't be stable.

wiktork (approved):

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
jander-msft: Thought I did that but maybe I didn't push it. Thanks for catching that.

PR #996, opened 2021-10-13
Add dependencies between actions, based on tokens.

wiktork:
TODO:

- [x] Add additional unit tests, in particular for logging error cases
- [x] More manual testing around execute actions that depend on results from pre-resume actions
- [x] Fix up WaitForCompletion tests, since start callback behavior has changed

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionDependencyAnalyzerTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionDependencyAnalyzerTests.cs
@@ -2,23 +2,22 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Microsoft.Diagnostics.Monitoring.TestCommon.Options;
+using Microsoft.Diagnostics.Monitoring.Tool.UnitTests.CollectionRules.Actions;
+using Microsoft.Diagnostics.Tools.Monitor;
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules;
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Actions;
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
-using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Actions;
-using System.Threading;
-using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options;
-using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Exceptions;
-using Microsoft.Diagnostics.Tools.Monitor;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Diagnostics.Monitoring.TestCommon.Options;
 using Xunit.Abstractions;
-using Microsoft.Extensions.Options;
-using Microsoft.Diagnostics.Tools.Monitor.CollectionRules;
-using Microsoft.Extensions.Logging;
-using System;
-using Microsoft.Diagnostics.Monitoring.Tool.UnitTests.CollectionRules.Actions;
-using System.Linq;
-using System.Collections.Generic;
 
 namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
 {
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionListExecutorTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionListExecutorTests.cs
@@ -169,10 +169,12 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
 
             await TestHostHelper.CreateCollectionRulesHost(_outputHelper, rootOptions =>
             {
-                CollectionRuleOptions options = rootOptions.CreateCollectionRule(DefaultRuleName);
-                options.AddPassThroughAction("a1", "a1input1", "a1input2", "a1input3");
-                a2Settings = (PassThroughOptions)options.AddPassThroughAction("a2", a2input1, a2input2, a2input3).Actions.Last().Settings;
-                options.SetStartupTrigger();
+                CollectionRuleOptions options = rootOptions.CreateCollectionRule(DefaultRuleName)
+                    .AddPassThroughAction("a1", "a1input1", "a1input2", "a1input3")
+                    .AddPassThroughAction("a2", a2input1, a2input2, a2input3)
+                    .SetStartupTrigger();
+
+                a2Settings = (PassThroughOptions)options.Actions.Last().Settings;
             }, async host =>
             {
                 ActionListExecutor executor = host.Services.GetService<ActionListExecutor>();
@@ -233,7 +235,6 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
 
         private static void VerifyStartCallbackCount(bool waitForCompletion, int callbackCount)
         {
-
             //Currently, any attempt to wait on completion will automatically trigger the start callback.
             //This is necessary to ensure that the process is resumed prior to completing artifact collection.
             Assert.Equal(1, callbackCount);


src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
```diff
@@ -0,0 +1,269 @@
...
+                        continue;
+                    }
+                    if (!results.OutputValues.TryGetValue(actionDependency.ResultName, out string result))
+                    {
+                        _ruleContext.Logger.InvalidResultReference(actionToken);
+                        continue;
+                    }
+                    newValue = newValue.Replace(actionToken, result, StringComparison.OrdinalIgnoreCase);
+                }
+                property.Value.Property.SetValue(settings, newValue);
```
jander-msft, line 148: I'm almost certain that you should not do this. The options are cached as part of a CollectionRuleOptions instance in the IOptionsCache<CollectionRuleOtions> service. If the same rule is applied to more than one process at the same time (very likely) then there will be multi-threading issues when setting and resetting these property values.

It's probably better (maybe even easier) to make a shallow clone of the options instance, change the values on the new instance, and return that new instance to be passed into the action.

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
```diff
@@ -0,0 +1,269 @@
...
+                return false;
+            }
+
+            action = dependencyOptions;
+            actionResultName = resultName;
+            return true;
+        }
+
+        private IEnumerable<PropertyInfo> GetPropertiesFromSettings(CollectionRuleActionOptions options) =>
+            options.Settings.GetType()
```
jander-msft, line 264: Technically, actions are not required to have a settings instance (for example, user does not specify settings, this using all of the default values for each of the properties). However, none of the actions so far allow for a null settings property. At least add a consideration explaining why this is okay for now and the potential way to fix it.

src/Tools/dotnet-monitor/LoggingEventIds.cs
```diff
@@ -55,5 +55,8 @@ internal static class LoggingEventIds
         public const int CollectionRulesStopped = 45;
         public const int CollectionRuleCancelled = 46;
         public const int DiagnosticRequestFailed = 47;
+        public const int InvalidToken = 48;
```
jander-msft, line 58: nit: InvalidActionReferenceToken

jander-msft (commented):
A few items to think about before I do a full review tomorrow.

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
```diff
@@ -0,0 +1,269 @@
...
+            string[] parts = actionReference.Split(Separator);
+            if (parts.Length != 2)
+            {
+                _ruleContext.Logger.InvalidActionReference(actionReference);
+                return false;
+            }
+
+            string name = parts[0];
+            //We only check previous actions for our dependencies.
+            CollectionRuleActionOptions dependencyOptions = _ruleContext.Options.Actions.Take(actionIndex)
```
jander-msft, line 242: Rather than doing this, validation should be updated to enforce that all action names are unique with a collection rule.

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
```diff
@@ -0,0 +1,269 @@
...
+
+                int foundIndex = 0;
+                int startIndex = 0;
+
+                PropertyDependency propertyDependency = null;
+
+                while ((foundIndex = newValue.IndexOf(ActionReferencePrefix, startIndex, StringComparison.OrdinalIgnoreCase)) >= 0)
+                {
+                    int suffixIndex = newValue.IndexOf(SubstitutionSuffix, foundIndex, StringComparison.OrdinalIgnoreCase);
+                    if (suffixIndex == -1)
```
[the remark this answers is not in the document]
jander-msft: An example is if you are using Bash to execute some inline command to remove all Docker containers:

`docker container stop $(docker container ls -aq)`

While this is `$(docker` and not `$(Actions.`, it is fairly close, and I wouldn't presume that no one has named an executable that starts with `Actions.`.

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
wiktork: Since we execute actions in sequence, I do not want to create a reference to a future action. Name uniqueness does make sense though.

src/Tools/dotnet-monitor/CollectionRules/ActionListExecutor.cs
```diff
@@ -52,14 +61,31 @@ public ActionListExecutor(ILogger<CollectionRuleService> logger, ICollectionRule
 
                     try
                     {
+                        IList<CollectionRuleActionOptions> actionDependencies = dependencyAnalyzer.GetActionDependencies(actionIndex);
+                        foreach (CollectionRuleActionOptions actionDependency in actionDependencies)
+                        {
+                            for (int i = 0; i < deferredCompletions.Count; i++)
+                            {
+                                ActionCompletionEntry deferredCompletion = deferredCompletions[i];
+                                if (deferredCompletion.Options.Name?.Equals(actionDependency.Name, StringComparison.OrdinalIgnoreCase) == true)
```
jander-msft, line 70: using string.Equals static method to avoid null checks and nullable lifting.

src/Tools/dotnet-monitor/CollectionRules/ActionListExecutor.cs
```diff
@@ -129,6 +141,42 @@ public ActionListExecutor(ILogger<CollectionRuleService> logger, ICollectionRule
...
+                context.ActionResults.Add(actionOption.Name, results);
+            }
+
+            _logger.CollectionRuleActionCompleted(context.Name, actionOption.Type);
+        }
+
+        private async Task WaitForCompletion(CollectionRuleContext context,
+            Action startCallback, ActionCompletionEntry entry, CancellationToken cancellationToken)
+        {
+            try
```
jander-msft, line 166: The try...catch...finally should be put into the other WaitForCompletion method so that end of the ExecuteActions method that waits for all of the remaining completes can handle exceptions, log them, and wrap them as necessary.

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleContext.cs
```diff
@@ -31,5 +33,12 @@ public CollectionRuleContext(string name, CollectionRuleOptions options, IEndpoi
         public CollectionRuleOptions Options { get; }
 
         public string Name { get; }
+
+        /// <summary>
+        /// Note that we only reference named actions. As such, unnamed actions will not save their results for other actions
+        /// to consume.
+        /// </summary>
+        public IDictionary<string, CollectionRuleActionResult> ActionResults { get; } =
```
jander-msft, line 41: I'm not a fan of putting this on the CollectionRuleContext. These values only represent the output of the actions of a specific action list invocation, not for the entire lifetime of the rule. Additionally, everything on this context is largely just for reference and not for mutation.

If you keep this here, you absolutely need to fix the fact that these results are not reset between runs of the action list inside of the CollectionRulePipeline. Again, I don't like the idea of the pipeline having to need to know how to set up the state of the action list execution.

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
```diff
@@ -0,0 +1,255 @@
...
+    /// - Note that property substitution uses the results of the dependency analysis, rather than recalculating the tokens.
+    /// Dependencies are organized as follows:
+    /// Action -> (Property, OriginalValue) -> (PreviousActionReference, PreviousActionResultReference)
+    /// </remarks>
+    internal sealed class ActionOptionsDependencyAnalyzer
+    {
+        private const string SubstitutionPrefix = "$(";
+        private const string SubstitutionSuffix = ")";
+        private const string Separator = ".";
+        private const string ActionReferencePrefix = SubstitutionPrefix + "Actions.";
```
jander-msft, line 37:
```suggestion
        private const string ActionReferencePrefix = SubstitutionPrefix + "Actions" + Separator;
```

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
```diff
@@ -0,0 +1,255 @@
...
+            if (parts.Length != 2)
+            {
+                _ruleContext.Logger.InvalidActionReference(actionReference);
+                return false;
+            }
+
+            string name = parts[0];
+            //We only check previous actions for our dependencies.
+            CollectionRuleActionOptions dependencyOptions = _ruleContext.Options.Actions.Take(actionIndex)
+                .FirstOrDefault(a => a.Name.Equals(name, StringComparison.OrdinalIgnoreCase));
```
jander-msft, line 224: Action name is not required; use static string.Equals method instead of instance method.

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
```diff
@@ -0,0 +1,255 @@
...
+            actionResultName = null;
+
+            string[] parts = actionReference.Split(Separator);
+            if (parts.Length != 2)
+            {
+                _ruleContext.Logger.InvalidActionReference(actionReference);
+                return false;
+            }
+
+            string name = parts[0];
```
jander-msft, line 221: Check that name is not null or empty and fail early.

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
```diff
@@ -0,0 +1,255 @@
...
+                    propertyDependency.ActionDependencies.Add(dependency);
+                }
+            }
+        }
+
+        private PropertyDependency GetOrCreateDependency(int actionIndex, PropertyInfo propertyInfo, string originalValue)
+        {
+            if (!_dependencies.TryGetValue(actionIndex, out Dictionary<string, PropertyDependency> properties))
+            {
+                properties = new Dictionary<string, PropertyDependency>();
```
jander-msft, line 197: Add StringComparer.Ordinal to the ctor.

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleContext.cs
```diff
@@ -31,5 +33,12 @@ public CollectionRuleContext(string name, CollectionRuleOptions options, IEndpoi
...
+            new Dictionary<string, CollectionRuleActionResult>();
```
jander-msft, line 42: Add StringComparer.Ordinal to ctor.

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
```diff
@@ -0,0 +1,255 @@
...
+        }
+
+        public ActionOptionsDependencyAnalyzer(CollectionRuleContext context)
+        {
+            _ruleContext = context ?? throw new ArgumentNullException(nameof(context));
+        }
+
+        public IList<CollectionRuleActionOptions> GetActionDependencies(int actionIndex)
+        {
+            EnsureDependencies();
```
jander-msft, line 82: What's the reason for doing this on each method call and not just once in the ActionOptionsDependencyAnalyzer ctor?

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
```diff
@@ -0,0 +1,255 @@
...
+            }
+
+            return settings;
+        }
+
+        private void EnsureDependencies()
+        {
+            if (_dependencies == null)
+            {
+                _dependencies = new Dictionary<int, Dictionary<string, PropertyDependency>>();
```
jander-msft, line 140: Add _ruleContext.Options.Actions.Count to the ctor to reduce number of reallocations inside the dictionary.

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
```diff
@@ -0,0 +1,255 @@
...
+
+        public IList<CollectionRuleActionOptions> GetActionDependencies(int actionIndex)
+        {
+            EnsureDependencies();
+
+            if (_dependencies.TryGetValue(actionIndex, out Dictionary<string, PropertyDependency> properties))
+            {
+                return properties
+                    .SelectMany(p => p.Value.ActionDependencies)
+                    .DistinctBy(a => a.Action.Name, StringComparer.OrdinalIgnoreCase)
```
jander-msft, line 88: Action names should be compared ordinal.

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
```diff
@@ -0,0 +1,255 @@
...
+            {
+                settings = cloneable.Clone();
+            }
+            else
+            {
+                _ruleContext.Logger.InvalidSettings(settings.GetType().FullName);
+                return settings;
+            }
+
+            foreach (KeyValuePair<string, PropertyDependency> property in properties)
```
jander-msft, line 112: nit: you can use decomposition to make it more clear what each part is:

```suggestion
            foreach ((string propertyName, PropertyDependency dependency) in properties)
```

jander-msft (changes requested):
First pass; wanted to get this feedback out while I finish the remainder.

src/Tools/dotnet-monitor/LoggingEventIds.cs
```diff
@@ -55,5 +55,9 @@ internal static class LoggingEventIds
         public const int CollectionRulesStopped = 45;
         public const int CollectionRuleCancelled = 46;
         public const int DiagnosticRequestFailed = 47;
+        public const int InvalidActionReferenceToken = 48;
+        public const int InvalidActionReference = 49;
+        public const int InvalidResultReference = 50;
+        public const int InvalidSettings = 51;
```
jander-msft, line 61: This is too generic; maybe "ActionSettingsTokenizationNotSupported".

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -511,6 +511,18 @@
...
 0 Format Parameters</comment>
   </data>
+  <data name="LogFormatString_InvalidActionReference" xml:space="preserve">
+    <value>Invalid action reference '{actionReference}'.</value>
+  </data>
+  <data name="LogFormatString_InvalidResultReference" xml:space="preserve">
+    <value>Invalid result refrence '{actionToken}'.</value>
+  </data>
+  <data name="LogFormatString_InvalidSettings" xml:space="preserve">
+    <value>Settings '{settingsType}' must implement ICloneable.</value>
```
jander-msft, line 521:
```suggestion
    <value>Collection rule action '{actionType}' options must implement ICloneable to support property tokenization.</value>
```

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
wiktork: 1) I don't want to put a lot of work into the ctor, since that's usually unexpected by the caller.
2) I don't want to have a separate method that needs to be called prior to the methods that actually extract information.

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
```diff
@@ -0,0 +1,255 @@
...
+                    {
+                        _ruleContext.Logger.InvalidResultReference(actionToken);
+                        continue;
+                    }
+                    if (!results.OutputValues.TryGetValue(actionDependency.ResultName, out string result))
+                    {
+                        _ruleContext.Logger.InvalidResultReference(actionToken);
+                        continue;
+                    }
+                    newValue = newValue.Replace(actionToken, result, StringComparison.OrdinalIgnoreCase);
```
jander-msft, line 128: Because this is done iteratively, this could result in dynamic expansion in a limited sense so long as two or more action results are references initially and the value of the first action result contains the token of the subsequent action result. This is probably okay because its not recursive but could be exponential.

For example, if the following actions exist:
```json
[{
  "Name": "Action1",
  "Type": "ActionType1"
},{
  "Name": "Action2",
  "Type": "ActionType2"
},{
  "Type": "ActionType3",
  "Settings": {
    "SomeProperty": "$(Actions.Action1.Output1) $(Actions.Action2.Output1)"
  }
}]
```

If ActionType1 outputs a result called Output1 with a value of "$(Actions.Action2.Output1)" and ActionType2 outputs a result called Output1 with a value of "Hello", then the SomeProperty property of Action3 will have a value of "Hello Hello" passed into the action.

I don't think if this is something we need to prevent. If we do, we could replace all of the tokens in the property value with placeholders and then fill the placeholders individually rather than just doing the iterative find-and-replace.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionListExecutorTests.cs
```diff
@@ -151,6 +153,57 @@ private async Task ActionListExecutor_FirstActionFail(bool waitForCompletion)
...
+                CollectionRuleContext context = new(DefaultRuleName, ruleOptions, null, logger);
+
+                int callbackCount = 0;
+                Action startCallback = () => callbackCount++;
+
+                await executor.ExecuteActions(context, startCallback, cancellationTokenSource.Token);
+
+                Assert.Equal(1, callbackCount);
+                Assert.Equal(2, context.ActionResults.Count);
+                Assert.Equal(3, context.ActionResults["a2"].OutputValues.Count);
```
jander-msft, line 189:
```suggestion
                Assert.True(context.ActionsResults.TryGetValue("a2", out CollectionRuleActionResult a2Result));
                Assert.Equal(3, a2Result.OutputValues.Count);
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionListExecutorTests.cs
```diff
@@ -151,6 +153,57 @@ private async Task ActionListExecutor_FirstActionFail(bool waitForCompletion)
...
+                //Verify that all options were reverted after execution
+                Assert.Equal(a2input1, a2Settings.Input1);
+                Assert.Equal(a2input2, a2Settings.Input2);
+                Assert.Equal(a2input3, a2Settings.Input3);
+
+                Assert.Equal("a1input1 with a1input2", context.ActionResults["a2"].OutputValues["Output1"]);
+                Assert.Equal("a1input2", context.ActionResults["a2"].OutputValues["Output2"]);
+                Assert.Equal("Output a1input3", context.ActionResults["a2"].OutputValues["Output3"]);
+
+                return;
```
jander-msft, line 200: Not necessary

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionListExecutorTests.cs
```diff
@@ -151,6 +153,57 @@ private async Task ActionListExecutor_FirstActionFail(bool waitForCompletion)
...
+                Assert.Equal(1, callbackCount);
+                Assert.Equal(2, context.ActionResults.Count);
+                Assert.Equal(3, context.ActionResults["a2"].OutputValues.Count);
+
+                //Verify that all options were reverted after execution
+                Assert.Equal(a2input1, a2Settings.Input1);
+                Assert.Equal(a2input2, a2Settings.Input2);
+                Assert.Equal(a2input3, a2Settings.Input3);
+
+                Assert.Equal("a1input1 with a1input2", context.ActionResults["a2"].OutputValues["Output1"]);
```
jander-msft, line 196:
```suggestion
                Assert.True(a2Result.OutputValues.TryGetValue("Output1", out string a2Output1));
                Assert.Equal("a1input1 with a1input2", a2Output1);
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionDependencyAnalyzerTests.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Linq;
+
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    public sealed class ActionDependencyAnalyzerTests
+    {
+        private readonly ITestOutputHelper _outputHelper;
+        private static readonly TimeSpan TimeoutMs = TimeSpan.FromMilliseconds(500);
+        private const string DefaultRuleName = "Default";
```
jander-msft, line 29: nit: Don't use "Default" as the rule name; choose something not too guessable so that the product functionality doesn't accidentally take a dependency on the "Default" name.

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleContext.cs
jander-msft: Alternatively, return this dictionary from the ExecuteActions method rather than stashing on the context.

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
wiktork: I decided to fix this by tracking each token location. That also allows me to do potentially less allocations on the final result.

PR #1000, opened 2021-10-13
Replace text/event-stream with text/plain for logs.

### LINKED ISSUE #71: Enhance /logs call
Default to different levels based on prefix: Microsoft warning, Default informational
- Allow override (post method). This can set custom levels for each category.
- Use text/plain format instead of eventstream

jander-msft:
Overview of changes:
- Replace text/event-stream with text/plain for logs.
- Reformat log entry output to same as simple console logger.
- Update documentation.

Example output:
```
2021-10-13 18:22:57Z fail: WebApplication15.Pages.IndexModel[5][MyEventId]
      => ConnectionId:0HMCEEJBBP01Q => RequestId:0HMCEEJBBP01Q:00000002, RequestPath:/ => ConnectionId:0HMCEEJBBP01R => ActionId:6dc10cda-b2b7-4e77-afa3-a07519e30346, ActionName:/Index
      An error: Some error message!
      System.InvalidOperationException: Some exception
         at WebApplication15.Pages.IndexModel.OnGet() in C:\Users\username\source\WebApplication15\WebApplication15\Pages\Index.cshtml.cs:line 28
```

closes #71

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/StreamingLogger.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/StreamingLogger.cs
@@ -47,8 +47,18 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
         private readonly LogFormat _logFormat;
         private readonly LogLevel? _logLevel;
 
-        private static readonly int LogTextPaddingLength = GetLogLevelString(LogLevel.Information).Length + ": ".Length;
-        private static readonly string LogTextPadding = new string(' ', LogTextPaddingLength);
+        // This is the padding used for each non-header line of the logs output when formatted as plain text.
+        // Since the plain text output adheres to the same format as simple console formatter, the same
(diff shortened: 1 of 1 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/StreamingLogger.cs
```diff
@@ -47,6 +47,9 @@ public sealed class StreamingLogger : ILogger
         private readonly LogFormat _logFormat;
         private readonly LogLevel? _logLevel;
 
+        private static readonly int LogTextPaddingLength = GetLogLevelString(LogLevel.Information).Length + ": ".Length;
```
wiktork, line 50: Maybe put a comment? I'm guessing you want the longest possible log level padding?

jander-msft: It's the same padding that json console logger uses. I'll add a comment about it though.

PR #1006, opened 2021-10-14
Do not write BOM for logs.

jander-msft:
Currently, each log entry within the `/logs` output has a BOM at the beginning of the entry, which is not allowed. This change removes the BOM from being emitted.

I've tested that the following routes return their respective data with BOM using curl that writes the output directly to disk and using a hex editor to check the delimiters:
- `/metrics`
- `/logs`
- `/logs` with `Accept: application/json-seq`
- `/logs` with `Accept: application/x-ndjson`

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/EncodingCache.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/EncodingCache.cs
@@ -8,7 +8,8 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
 {
     internal static class EncodingCache
     {
-        public static readonly Encoding UTF8NoBOM =
-            new UTF8Encoding(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: true);
+        // Encode UTF8 without BOM and write "?" as fallback replacement.
+        public static readonly Encoding UTF8NoBOMNoThrow =
+            new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
     }
 }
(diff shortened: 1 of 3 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/EncodingCache.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Text;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    internal class EncodingCache
```
wiktork, line 9: nit: static

src/Microsoft.Diagnostics.Monitoring.WebApi/EncodingCache.cs
```diff
@@ -0,0 +1,14 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Text;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    internal static class EncodingCache
+    {
+        public static readonly Encoding UTF8NoBOM =
+            new UTF8Encoding(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: true);
```
wiktork, line 12: Do you know the practical impact of the second parameter? For example, can someone emit a log statement that contains invalid bytes and we end up crashing as a result when we attempt to encode the payload?

wiktork (approved):

PR #1103, opened 2021-11-04
Do not attempt to prune processes during tracing

wiktork:
It appears that the event pipe is not responsive during rundown, which causes our pruning mechanism to remove processes (similar to dumps). Possibly related to https://github.com/dotnet/runtime/issues/1892.

- The pruning algorithm should probably do multiple attempts over a period of time to see if the process is really to be removed.
- The real fix should make rundown more configurable. We are defaulting to true in some cases we may not need.
- There may be other affected artifacts (gcdump/logs).
- We should converge the OperationTrackerService and the RequestLimitTracker at some point.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfoSourceTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfoSourceTests.cs
@@ -161,9 +161,9 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
         public async Task ServerSourceNoPruneDuringDumpTest(TargetFrameworkMoniker appTfm)
         {
             EndpointInfoSourceCallback callback = new(_outputHelper);
-            var operatonTrackerService = new OperationTrackerService();
-            MockDumpService dumpService = new(operatonTrackerService);
-            await using ServerSourceHolder sourceHolder = await _endpointUtilities.StartServerAsync(callback, dumpService, operatonTrackerService);
+            var operationTrackerService = new OperationTrackerService();
+            MockDumpService dumpService = new(operationTrackerService);
+            await using ServerSourceHolder sourceHolder = await _endpointUtilities.StartServerAsync(callback, dumpService, operationTrackerService);
 
             AppRunner runner = _endpointUtilities.CreateAppRunner(sourceHolder.TransportName, appTfm);
 
@@ -244,7 +244,7 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
                 }
                 finally
                 {
-                    operationRegistration.Dispose();
+                    operationRegistration?.Dispose();
                 }
             }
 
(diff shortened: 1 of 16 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfoSourceTests.cs
```diff
@@ -161,8 +161,9 @@ public async Task ServerSourceAddRemoveMultipleConnectionTest(TargetFrameworkMon
         public async Task ServerSourceNoPruneDuringDumpTest(TargetFrameworkMoniker appTfm)
         {
             EndpointInfoSourceCallback callback = new(_outputHelper);
-            MockDumpService dumpService = new();
-            await using ServerSourceHolder sourceHolder = await _endpointUtilities.StartServerAsync(callback, dumpService);
+            var operatonTrackerService = new OperationTrackerService();
```
kkeirstead, line 164: nit: operationTrackerService (instead of operatonTrackerService)

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/TestHostHelper.cs
```diff
@@ -86,9 +86,10 @@ internal static class TestHostHelper
                 .ConfigureServices((HostBuilderContext context, IServiceCollection services) =>
                 {
                     services.ConfigureGlobalCounter(context.Configuration);
+                    services.AddSingleton<OperationTrackerService>();
                     services.ConfigureCollectionRules();
                     services.ConfigureEgress();
-
+;
```
kkeirstead, line 92: nit: Think this is just supposed to be an empty line

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfoSourceTests.cs
```diff
@@ -226,30 +227,27 @@ private sealed class MockDumpService : IDumpService
...
+                IDisposable operationRegistration = null;
                 try
                 {
+                    operationRegistration = _operationTrackerService.Register(endpointInfo);
                     return await _dumpCompletionSource.Task;
                 }
                 finally
                 {
-                    Interlocked.Decrement(ref _operationCount);
+                    operationRegistration.Dispose();
```
jander-msft, line 247:
```suggestion
                    operationRegistration?.Dispose();
```

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectTraceAction.cs
```diff
@@ -83,7 +85,18 @@ public CollectTraceAction(IServiceProvider serviceProvider, IEndpointInfo endpoi
                 KeyValueLogScope scope = Utils.CreateArtifactScope(Utils.ArtifactType_Trace, EndpointInfo);
 
                 EgressOperation egressOperation = new EgressOperation(
-                    (outputStream, token) => TraceUtilities.CaptureTraceAsync(startCompletionSource, EndpointInfo, configuration, duration, outputStream, token),
+                    async (outputStream, token) =>
+                    {
+                        IDisposable operationRegistration = _operationTrackerService.Register(EndpointInfo);
```
jander-msft, line 90: nit: Because this is unconditionally created, you could use a `using IDisposable registration = ` and not use a try...finally. I guess when compiled, it does the same thing.

src/Tools/dotnet-monitor/DumpServiceEndpointInfoSourceCallback.cs
```diff
@@ -9,13 +9,14 @@
 
 namespace Microsoft.Diagnostics.Tools.Monitor
 {
+    //TODO Rename this
     internal sealed class DumpServiceEndpointInfoSourceCallback : IEndpointInfoSourceCallbacks
```
jander-msft, line 13: Rename to `OperationTrackerServiceEndpointInfoSourceCallback`

jander-msft (approved):

PR #1110, opened 2021-11-05
Documentation - Collection Rule Examples

kkeirstead:
I wrote up a few samples for people to use as a reference point when getting started with triggers. I don't have great intuition for which artifacts make sense under which conditions, so feel free to recommend adjustments that you feel make more sense for out-of-the-box use. My thinking is that people will make their own changes, but it would be good for the baseline scenarios to make sense.


documentation/collectionruleexamples.md
```diff
@@ -0,0 +1,223 @@
...
+
+The following examples provide sample scenarios for using a collection rule. These templates can be copied directly into your configuration file with minimal adjustments to work with your application, or they can be adjusted for your specific use-case. [Learn more about configuring collection rules](collectionrules.md).
+
+## Collect GCDump - Startup (`Startup` Trigger)
+
+### JSON
+
+```json
+{
+  "GCDumpOnStartup": {
```
wiktork, line 11: One possibly useful scenario for startup is tracing assembly loads. See https://docs.microsoft.com/en-us/dotnet/fundamentals/diagnostics/runtime-loader-binder-events.

A good example for gcdump might be an event counter trigger against the heap size counter (https://docs.microsoft.com/en-us/dotnet/core/diagnostics/available-counters)

documentation/collectionruleexamples.md
```diff
@@ -0,0 +1,223 @@
...
+
+```json
+{
+  "BadResponseStatus": {
+    "Trigger": {
+      "Type": "AspNetResponseStatus",
+      "Settings": {
+        "ResponseCount": 5,
+        "StatusCodes": [
+          "400-499"
```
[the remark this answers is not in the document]
kkeirstead: Yep, they do, but when I tested this it actually required 7 404s to trigger it (which is in-line with what I found in the bug, since it took 3 404s to trigger when the `ResponseCount` was supposed to be 1)

documentation/collectionruleexamples.md
```diff
@@ -0,0 +1,223 @@
...
+        "Settings": {
+          "Egress": "artifacts",
+          "Type": "Mini"
+        },
+        "WaitForCompletion": true
+      },
+      {
+        "Type": "Execute",
+        "Settings": {
+          "Path": "C:\\Program Files\\Microsoft Visual Studio\\2022\\Preview\\Common7\\IDE\\devenv.exe",
```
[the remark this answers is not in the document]
kkeirstead: Assuming you've tried this scenario before, how do you do it? I tried setting my path to my dotnet-dump.exe `"C:\\Users\\kkeirstead\\.dotnet\\tools\\dotnet-dump.exe"` and my args to analyze the dump `"analyze \"$(Actions.MyDump.EgressPath)\""`, but this doesn't appear to work for me. Just out of curiosity, I toggled `RedirectStandardOutput` to be false in the Execute action, which then launched the tool in the currently executing dotnet-monitor.exe window, but that isn't the real way to do it.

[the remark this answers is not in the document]
wiktork: I haven't actually tried this scenario but it looks like dotnet-dump analyze is an interactive shell and even if it wasn't our current results don't capture standard out. If I can think of a better example, I will comment here.

documentation/collectionruleexamples.md
```diff
@@ -0,0 +1,260 @@
...
+
+```json
+{
+  "AssemblyLoadTraceOnStartup": {
+    "Trigger": {
+      "Type": "Startup"
+    },
+    "Actions": [
+      {
+        "Type": "CollectTrace",
```
wiktork, line 17: You'll want something like:
`
"Actions": [
    {
        "Type": "CollectTrace",
        "Settings": {
            "Duration": "00:00:15",
            "Egress": "file",
            "Providers": [
                {
                    "Name": "Microsoft-Windows-DotNETRuntime",
                    "Keywords": "0x8",
                    "EventLevel": "Informational"
                }],

documentation/collectionruleexamples.md
```diff
@@ -0,0 +1,260 @@
...
+
+```json
+{
+  "LargeGCHeapSize": {
+    "Trigger": {
+      "Type": "EventCounter",
+      "Settings": {
+        "ProviderName": "System.Runtime",
+        "CounterName": "gc-heap-size",
+        "GreaterThan": 1000000
```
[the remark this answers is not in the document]
kkeirstead: Is that documented anywhere? When I looked at the page (https://docs.microsoft.com/en-us/dotnet/core/diagnostics/available-counters), the description said, "The number of _bytes_ thought to be allocated based on GC.GetTotalMemory(Boolean)," which is why I assumed bytes. Just figured we should double check this.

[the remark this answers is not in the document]
wiktork: The best source of truth is probably https://github.com/dotnet/runtime/blob/235915ea5e15604c80889c2ad7da9e999bc36e57/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs#L80

I noticed this when writing a trigger experimentally. 

Also if you are using /metrics to find good values for triggers, we normalize to bytes in that endpoint. /livemetrics will be more accurate.

documentation/collectionruleexamples.md
```diff
@@ -0,0 +1,261 @@
...
+        }
+      }
+    ]
+  }
+}
+```
+
+### Explanation
+
+This rule, named "AssemblyLoadTraceOnStartup", will trigger on a process's startup. When the rule is triggered, a trace will be collected for 15 seconds and egressed to the specified `Egress` provider (in this case, `artifacts` has been configured to save the trace to the local filesystem). The trace will capture events from an event provider named `Microsoft-Windows-DotNETRuntime`, and will collect events at or above the `Informational` level using the keyword `0x8` (`LoaderKeyword`). The trace will request rundown by default, and the `BufferSizeInMB` has the default value of 256 MB.
```
wiktork, line 35: Consider adding a link to https://docs.microsoft.com/en-us/dotnet/core/diagnostics/well-known-event-providers.

documentation/collectionruleexamples.md
```diff
@@ -0,0 +1,261 @@
...
+
+```json
+{
+  "LargeGCHeapSize": {
+    "Trigger": {
+      "Type": "EventCounter",
+      "Settings": {
+        "ProviderName": "System.Runtime",
+        "CounterName": "gc-heap-size",
+        "GreaterThan": 1
```
wiktork, line 49: nit: won't 1 trigger right away? Maybe something a little larger?

documentation/collectionruleexamples.md
```diff
@@ -0,0 +1,261 @@
...
+        }
+      }
+    ]
+  }
+}
+```
+
+### Explanation
+
+This rule, named "LargeGCHeapSize", will trigger when the GC Heap Size exceeds 1 MB within the default sliding window duration (1 minute). If the rule is triggered, a GCDump will be collected and egressed to the specified `Egress` provider (in this case, `artifacts` has been configured to save the GCDump to the local filesystem). There is a default `ActionCount` limit stating that this rule may only be triggered 5 times.
```
wiktork, line 66: Not something that needs to be added here but https://github.com/dotnet/dotnet-monitor/issues/1126 does affect gcdumps right now

wiktork (approved):

documentation/collectionruleexamples.md
kkeirstead: I bumped it up to 10 MB after a little playing around, though I still don't have much intuition for this. Let me know if you want me to change this before I check it in (and if not, we can always update the example later if people prefer another value).

PR #1111, opened 2021-11-05
Adds internal/release/ to publish branches

kelltrick:
Makes the prepare release stage execute when building against `internal/release/`

### DIFF
--- a/eng/PrepareRelease.yml
+++ b/eng/PrepareRelease.yml
@@ -2,7 +2,7 @@ stages:
 - stage: PrepareReleaseStage
   displayName: Release Preparation
   dependsOn:
-  - ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest'), startsWith(variables['Build.SourceBranch'], 'refs/heads/release/')) }}:
+  - ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest'), or(startsWith(variables['Build.SourceBranch'], 'refs/heads/release/'), startsWith(variables['Build.SourceBranch'], 'refs/heads/internal/release/'))) }}:
     - publish_using_darc
   jobs:
   - job: PrepareReleaseJob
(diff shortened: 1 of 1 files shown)


eng/PrepareRelease.yml
```diff
@@ -66,8 +124,17 @@ stages:
...
+          and(
+            ne(variables['System.TeamProject'], 'public'), 
+            notin(variables['Build.Reason'], 'PullRequest'), 
+            or(
+              startsWith(variables['Build.SourceBranch'], 'refs/heads/release/'), 
+              startsWith(variables['Build.SourceBranch'], 'refs/heads/internal/release/')
+            )
+          ) }}:
+        - task: tagBuildOrRelease@0
+          displayName: 'Tag Build with Monitor Release'
```
wiktork, line 137: Is test/release for testing the release build? I'm assuming that's why you don't want to tag it?

wiktork (approved):

eng/PrepareRelease.yml
```diff
@@ -2,15 +2,43 @@ stages:
...
+            ne(variables['System.TeamProject'], 'public'), 
+            notin(variables['Build.Reason'], 'PullRequest'), 
+            or(
+              startsWith(variables['Build.SourceBranch'], 'refs/heads/release/'), 
+              startsWith(variables['Build.SourceBranch'], 'refs/heads/internal/release/')
+            )
+          ),
+          and(
+            ne(variables['System.TeamProject'], 'public'), 
+            startsWith(variables['Build.SourceBranch'], 'refs/heads/test/release/')
```
jander-msft, line 39: Can you simplify this by moving this branch check to be inline with the other ones? I don't think we need to allow internal PR builds to attempt these steps; just invoke the build directly on the branch if need be.

jander-msft: Do the same for all of the other conditionals you've updated.

eng/PrepareRelease.yml
kelltrick: I took that part out, but yet, that was a specific branch we made to test the release pipeline, it needed to stay constant because we had to create a DARC channel for it.

eng/PrepareRelease.yml
kelltrick: If I understand your request correctly, I can't do that. The reason this is separated out is because we need to allow build on PR (since that's the whole purpose of the branch).

PR #1143, opened 2021-11-15
Config Show: Array properties without indices

### LINKED ISSUE #1054: `config show` command writes array properties with indexes
The `config show` command will write out the values of each element in an array or list property with indexes. For example:

```json
  "CollectionRules": {
    "CollectAssemblyLoads": {
      "Actions": {
        "0": {
          "Name": "CollectAssemblyLoads",
          "Settings": {
            "Duration": "00:00:10",
            "Egress": "desktop",
            "Providers": {
              "0": {
                "Keywords": "0xC",
                "Name": "Microsoft-Windows-DotNETRuntime"
              }
            },
            "RequestRundown": "False"
          },
          "Type": "CollectTrace"
        },
        "1": {
          "Settings": {
            "Arguments": "$(Actions.CollectAssemblyLoads.EgressPath)",
            "Path": "C:\\Program Files\\Microsoft Visual Studio\\2022\\Preview\\Common7\\IDE\\devenv.exe"
          },
          "Type": "Execute"
        }
      },
      "Trigger": {
        "Type": "Startup"
      }
    }
  }
```

It is probably preferable to write array and list properties as arrays in the json output. Using the same example would look like:

```json
  "CollectionRules": {
    "CollectAssemblyLoads": {
      "Actions": [{
        "Name": "CollectAssemblyLoads",
        "Settings": {
          "Duration": "00:00:10",
          "Egress": "desktop",
          "Providers": [{
            "Keywords": "0xC",
            "Name": "Microsoft-Windows-DotNETRuntime"
          }],
          "RequestRundown": "False"
        }
        "Type": "CollectTrace"
      },{
        "Settings": {
          "Arguments": "$(Actions.CollectAssemblyLoads.EgressPath)",
          "Path": "C:\\Program Files\\Microsoft Visual Studio\\2022\\Preview\\Common7\\IDE\\devenv.exe"
        },
        "Type": "Execute"
      }],
      "Trigger": {
        "Type": "Startup"
      }
    }
  }
```

kkeirstead:
By default, `Config Show` lists the indices of elements in an array. This change counteracts that behavior by printing `Config Show` using the expected format (no indices).

I've currently only wired up three scenarios to take advantage of this code (CollectionRules, Metrics, and DefaultProcess), because they appeared to be the ones that were affected; however, this could easily be extended with the addition of a parameter to any of the other configuration elements.

The least elegant part of this solution is determining the type of actions/triggers in order to get their available properties; I'll leave a few comments in the review where people can weigh in if they have any thoughts on whether it's fine as-is or if there's a better way.

I tested this against the scenarios in my Collection Rule Examples PR as well as a few custom scenarios and it appears to work consistently, but it's possible there's edge cases I've overlooked in the testing, so let me know if anything looks amiss.

Closes #1054

### DIFF
--- a/src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
+++ b/src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
@@ -197,14 +197,14 @@ namespace Microsoft.Diagnostics.Tools.Monitor
         {
             _writer.WritePropertyName(section.Key);
 
-            bool parentIsCR = section.Key.Equals(ConfigurationKeys.CollectionRules);
-
             IEnumerable<IConfigurationSection> children = section.GetChildren();
 
             //If we do not traverse the child sections, the caller is responsible for creating the value
             if (includeChildSections && children.Any())
             {
                 bool isSequentialIndices = CheckForSequentialIndices(children);
+
+                bool parentIsCR = section.Key.Equals(ConfigurationKeys.CollectionRules);
 
                 if (isSequentialIndices && !parentIsCR)
                 {
--- a/src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
+++ b/src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
@@ -222,7 +222,14 @@ namespace Microsoft.Diagnostics.Tools.Monitor
                     }
                     else
                     {
-                        WriteValue(section.Value, redact);
+                        _writer.WriteStartArray();
+
+                        foreach (IConfigurationSection child in children)
+                        {
+                            WriteValue(child.Value, redact);
+                        }
+
+                        _writer.WriteEndArray();
                     }
                 }
                 else
--- a/src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
+++ b/src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
@@ -197,16 +197,67 @@ namespace Microsoft.Diagnostics.Tools.Monitor
         {
             _writer.WritePropertyName(section.Key);
 
+            bool parentIsCR = section.Key.Equals(ConfigurationKeys.CollectionRules);
+
             IEnumerable<IConfigurationSection> children = section.GetChildren();
 
             //If we do not traverse the child sections, the caller is responsible for creating the value
             if (includeChildSections && children.Any())
             {
-                using (new JsonObjectContext(_writer))
+                bool isSequentialIndices = true;
+
+                int indexValue = 0;
+
+                foreach (IConfigurationSection child in children)
                 {
-                    foreach (IConfigurationSection child in children)
+                    if (!child.Key.Equals(indexValue.ToString()))
                     {
-                        ProcessSection(child, includeChildSections, redact);
+                        isSequentialIndices = false;
+                        break;
+                    }
+
+                    indexValue++;
+                }
+
+                if (isSequentialIndices && !parentIsCR)
+                {
+                    if (children.First().GetChildren().Any())
+                    {
+                        _writer.WriteStartArray();
+
+                        foreach (IConfigurationSection child in children)
+                        {
+                            using (new JsonObjectContext(_writer))
+                            {
+                                foreach (IConfigurationSection grandChild in child.GetChildren())
+                                {
+                                    ProcessSection(grandChild, includeChildSections, redact);
+                                }
+                            }
+                        }
+
+                        _writer.WriteEndArray();
+                    }
+                    else
+                    {
+                        if (redact)
+                        {
+                            _writer.WriteStringValue(Strings.Placeholder_Redacted);
+                        }
+                        else
+                        {
+                            _writer.WriteStringValue(section.Value);
+                        }
+                    }
+                }
+                else
+                {
+                    using (new JsonObjectContext(_writer))
+                    {
+                        foreach (IConfigurationSection child in children)
+                        {
+                            ProcessSection(child, includeChildSections, redact);
+                        }
                     }
                 }
             }
--- a/src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
+++ b/src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
@@ -8,6 +8,7 @@ using Microsoft.Diagnostics.Tools.Monitor.Egress.FileSystem;
 using Microsoft.Extensions.Configuration;
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Text.Json;
@@ -259,7 +260,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor
 
             foreach (IConfigurationSection child in children)
             {
-                if (!child.Key.Equals(indexValue.ToString()))
+                if (!child.Key.Equals(indexValue.ToString(CultureInfo.InvariantCulture)))
                 {
                     return false;
                 }
@@ -291,5 +292,4 @@ namespace Microsoft.Diagnostics.Tools.Monitor
             }
         }
     }
-
 }
--- a/src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
+++ b/src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
@@ -209,28 +209,21 @@ namespace Microsoft.Diagnostics.Tools.Monitor
 
                 if (isSequentialIndices && !parentIsCR)
                 {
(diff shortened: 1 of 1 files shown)


src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -20,6 +22,9 @@ namespace Microsoft.Diagnostics.Tools.Monitor
     internal sealed class ConfigurationJsonWriter : IDisposable
     {
         private readonly Utf8JsonWriter _writer;
+        private readonly string _dictionaryName = "IDictionary`2";
```
kkeirstead, line 25: Currently checking via name equality for dictionary/list; do we have a preferred way of doing this?

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -202,28 +207,135 @@ private void ProcessSection(IConfigurationSection section, bool includeChildSect
...
+                        if (section.Key.Equals(nameof(CollectionRuleOptions.Trigger)))
+                        {
+                            optionsType = "Triggers";
+                        }
+                        else if (section.Key.Equals(nameof(CollectionRuleOptions.Actions)))
+                        {
+                            optionsType = "Actions";
+                        }
+
+                        Type t = Type.GetType("Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options" + "." + optionsType + "." + child.Value + "Options");
```
kkeirstead, line 328: This feels fragile to me (especially in the event that we refactor and that path gets changed); is there a better way of constructing this (or something else entirely)?

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -202,28 +207,135 @@ private void ProcessSection(IConfigurationSection section, bool includeChildSect
...
+            {
+                foreach (var child in children)
+                {
+                    // Special case due to not knowing the types of actions/triggers; takes advantage of the fact that we know that a Type variable must be set.
+                    if (child.Key.Equals("Type"))
                     {
-                        _writer.WriteStringValue(section.Value);
+                        string optionsType = "";
+
+                        if (section.Key.Equals(nameof(CollectionRuleOptions.Trigger)))
```
kkeirstead, line 319: I don't love having to special case the trigger and actions, but seeing as the design of collection rules is relatively stable I don't know that it'll really be an issue.

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
wiktork: typeof(IDictionary<>).Name

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -63,12 +68,12 @@ public void Write(IConfiguration configuration, bool full, bool skipNotPresent)
 
                 //No sensitive information
                 ProcessChildSection(configuration, ConfigurationKeys.GlobalCounter, skipNotPresent, includeChildSections: true);
-                ProcessChildSection(configuration, ConfigurationKeys.CollectionRules, skipNotPresent, includeChildSections: true);
+                ProcessChildSection(configuration, ConfigurationKeys.CollectionRules, skipNotPresent, includeChildSections: true, sectionType: typeof(RootOptions).GetProperty(ConfigurationKeys.CollectionRules).PropertyType);
```
wiktork, line 71: Rather than attempting to understand the configuration with types, it might be easier to rely on the keys. If the children are 0,1,2,3,4 etc we show them as an array. Since we control the names, we don't have to worry that someone will create a config key with an actual name of 0 or something like that.

kkeirstead: That was my initial thought too, but is there anything keeping someone from naming their collection rule, "0"? I was concerned about clobbering things actually named as indices by guessing they were arrays

wiktork: I think we can detect that. I.e. if someone names their rules "0", "b", "5" we are not an array. If all the children are 0-n, we are an array. There is one edge cases we couldn't detect, and that is someone naming their rules '0', '1', '2', '3' etc. I honestly think it's ok if we show the wrong thing in that case rather than have to change this code every time we have add an array in configuration.

kkeirstead: That was the case I was worried about (or, in a more basic scenario, someone with only one rule that they just named "0" since they couldn't be bothered to choose a name). On the one hand, we would knowingly output illegal configuration, but that is the much simpler solution. I'll probably start playing around with this in another branch, but will wait until @jander-msft gets back from vacation and has a chance to weigh in before committing to one direction or another.

kkeirstead: This discussion also made me think of something else, which is that the `ConfigurationJsonWriter` uses colons to separate the different keys in the hierarchy (e.g. CollectionRules:abc:Actions:0). Technically, we allow colons in our collection rule names, which means someone can make the output illegally formatted if they put a colon in their collection rule name. That's probably a relatively minor bug, but I also don't see any harm in banning the inclusion of colons in keys unless someone is opposed.

wiktork: The configuration section will properly handle this case. You'll just see a Key with a ':' inside. This shouldn't be an issue.

kkeirstead: Are you sure? This is the behavior I see in `dotnet monitor` (the actual release from dotnet tools):

ColonSeparationBug_ConfigShow
ColonSeparationBug_CollectionRulesStart

wiktork: You're right. I tried a basic top level "a:b": "value" configuration and it worked. But when I changed it to "Nested":{ "a:b": "value"}, "a" and "b" became distinct elements. This looks like a bug in https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Configuration.Json/src/JsonConfigurationFileParser.cs

kkeirstead:
After talking with Wiktor, we decided to take a different approach that should require less maintenance and special-casing. We now treat all sequences of `0,1,2...` as arrays, with the exception of a special case in the case that Collection Rules are named as `0,1,2...`

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -202,28 +202,72 @@ private void ProcessSection(IConfigurationSection section, bool includeChildSect
...
+        }
+
+        private bool CheckForSequentialIndices(IEnumerable<IConfigurationSection> children)
+        {
+            int indexValue = 0;
+
+            foreach (IConfigurationSection child in children)
             {
-                if (!children.Any())
+                if (!child.Key.Equals(indexValue.ToString()))
```
wiktork, line 262: nit: indexValue.ToString(CultureInfo.Invariant)

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -202,28 +202,72 @@ private void ProcessSection(IConfigurationSection section, bool includeChildSect
...
             {
-                using (new JsonObjectContext(_writer))
+                bool isSequentialIndices = CheckForSequentialIndices(children);
+
+                bool parentIsCR = section.Key.Equals(ConfigurationKeys.CollectionRules);
+
+                if (isSequentialIndices && !parentIsCR)
                 {
-                    foreach (IConfigurationSection child in children)
+                    if (children.First().GetChildren().Any())
```
wiktork, line 211: I don't understand this if check? isn't children.First the first element of the array?

kkeirstead: Yep; this check is actually to see if there are "grandchildren". To get rid of the array indices, we jump over them to their children and use those as new keys. However, if the children are values (leaf nodes), then the `else` condition has us write out the value. It's the same fundamental idea as the check on Line 203 to see if there are any children, except we do this specifically for array indices.

kkeirstead: In terms of only checking the first child, given that all elements in the array should be the same type if one has children they should all in theory have children...I'm trying to think if there's an edge case where this would be a problem

wiktork (approved):

PR #1178, opened 2021-12-02
Add support for diagsession packaging

wiktork:
Add the ability to package dumps into diagsession files. For Linux, add the ability to include dac/dbi as part of the diagsession.

TODO:
- dac/dbi packaging should ideally only happen on non-sidecar scenarios.


src/Microsoft.Diagnostics.Monitoring.WebApi/DacLocator.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            ProcessStartInfo processStartInfo = new ProcessStartInfo()
+            {
+                CreateNoWindow = true,
+                Arguments = "--list-runtimes",
+                UseShellExecute = false,
+                FileName = "dotnet",
+                RedirectStandardOutput = true,
+            };
+            //CONSIDER We could get the location of the runtime dll for a specific process by collecting rundown instead.
```
wiktork, line 38: Note this doesn't work across sidecars at all and will produce the wrong result. Consider guarding against that scenario.

src/Microsoft.Diagnostics.Monitoring.WebApi/DacLocator.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        public static async Task<string> LocateRuntime(CancellationToken token)
+        {
+            ProcessStartInfo processStartInfo = new ProcessStartInfo()
+            {
+                CreateNoWindow = true,
+                Arguments = "--list-runtimes",
+                UseShellExecute = false,
+                FileName = "dotnet",
```
jander-msft, line 35: This assumes a shared installation exists and will not work for self-contained applications.

src/Microsoft.Diagnostics.Monitoring.WebApi/DacLocator.cs
```diff
@@ -0,0 +1,88 @@
...
+                RedirectStandardOutput = true,
+            };
+            //CONSIDER We could get the location of the runtime dll for a specific process by collecting rundown instead.
+
+            var process = new Process() { StartInfo = processStartInfo };
```
jander-msft, line 40: This assumes that dotnet-monitor is running from the same file system as the target application.

src/Microsoft.Diagnostics.Monitoring.WebApi/DacLocator.cs
```diff
@@ -0,0 +1,88 @@
...
+                throw new FormatException();
+            }
+
+            string name = line.Substring(0, fxVersionIndex);
+            int pathStart = endFxVersionIndex + 2;
+
+            string version = line.Substring(fxVersionIndex + 1, endFxVersionIndex - fxVersionIndex - 1);
+            string path = line.Substring(pathStart, line.Length - pathStart - 1);
+
+            return new Runtime { Name = name, Version = new Version(version), Path = path };
```
jander-msft, line 85: This assumes the version is a stable version and doesn't account for prerelease versions. You probably want a remove the prerelease label part before parsing into a `Version` and separately store the full version as a separate property.

src/Microsoft.Diagnostics.Monitoring.WebApi/DacLocator.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            //Microsoft.NETCore.App 6.0.0 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
+
+            const char Space = ' ';
+            int fxVersionIndex = line.IndexOf(Space);
+            int endFxVersionIndex = line.IndexOf(Space, fxVersionIndex + 1);
+
+            if (fxVersionIndex == -1 || endFxVersionIndex == -1)
+            {
+                throw new FormatException();
```
jander-msft, line 76: Rather than throwing, refactor the method to be `bool TryParseRuntime(string, out Runtime)` and return false.

src/Microsoft.Diagnostics.Monitoring.WebApi/DacLocator.cs
wiktork: Good catch, surprisingly I have no prerelease versions on my box right now.

src/Microsoft.Diagnostics.Monitoring.WebApi/DacLocator.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        private static Runtime ParseRuntime(string line)
+        {
+            //Microsoft.NETCore.App 6.0.0 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
+
+            const char Space = ' ';
+            int fxVersionIndex = line.IndexOf(Space);
+            int endFxVersionIndex = line.IndexOf(Space, fxVersionIndex + 1);
+
+            if (fxVersionIndex == -1 || endFxVersionIndex == -1)
```
kkeirstead, line 74: Would it also be a good idea to verify the number of spaces? This checks that there's at least two spaces, but doesn't rule out there being additional spaces (which could skew the versioning if the wrong space starting index was used). I guess I'm not aware of whether we have any formatting guarantees that would render this unnecessary, but figured I'd throw it out there.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -203,17 +204,19 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
             [FromQuery]
             Models.DumpType type = Models.DumpType.WithHeap,
             [FromQuery]
-            string egressProvider = null)
+            string egressProvider = null,
+            [FromQuery]
+            Models.PackageMode mode = Models.PackageMode.None)
```
kkeirstead, line 209: nit: We use a variable called `mode` here in reference to a `PackageMode` and `type` for a `DumpType`; however, in the `DumpAsync` method we use `mode` to refer to a `DumpType` and use `packageMode` for a `PackageMode`. Just for consistency and readability, we should probably use the same terms in both places if they mean the same thing.

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/PackageMode.cs
```diff
@@ -0,0 +1,20 @@
...
+using System;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi.Models
+{
+    [Flags]
+    public enum PackageMode
+    {
+        None = 0x0,
+        DiagSession =   0x00000001, //Packages the dump into a diagsession file with no compression
+        IncludeDacDbi = 0x00010000, //Packages
```
kkeirstead, line 14: nit: Is this missing part of its comment? Just based off the line above.

src/Microsoft.Diagnostics.Monitoring.WebApi/Packaging.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        public static async Task<string> CreateDiagSession(string dumpFilePath, string dac, string dbi, CancellationToken token)
+        {
+            string packagePath = Path.ChangeExtension(dumpFilePath, ".diagsession");
+
+            //Note it is critical to have Create/Write here. A Read/Write mode will cause
+            //the entire dump to be placed into MemoryStream
+            using Package package = Package.Open(packagePath, FileMode.Create, FileAccess.Write);
+
+            Guid tool = new Guid("013914d7-d262-4da2-8fa4-ba8b4e328a1e");
```
kkeirstead, line 32: Just to confirm because I haven't seen this before: did we generate a GUID so we would have a unique identifier, but then we manually specify that GUID for consistency instead of always producing new GUIDs?

src/Microsoft.Diagnostics.Monitoring.WebApi/Packaging.cs
```diff
@@ -0,0 +1,93 @@
...
+                new XElement(Ns + "Tools",
+                    new XElement(Ns + "Tool", new XAttribute("Id", FormattableString.Invariant($"{tool:b}")),
+                        new XElement(Ns + "AcquireLink", new XAttribute("Type", "none")))));
+
+            var content = new XElement(Ns + "Content");
+            root.Add(content);
+
+            XElement dumpResource = await CreateResource(package, dumpFilePath, token);
+            content.Add(dumpResource);
+            if (dac != null)
```
kkeirstead, line 44: Should we be using a null or empty check (as well as for the dbi check below)?

src/Microsoft.Diagnostics.Monitoring.WebApi/Packaging.cs
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        private static async Task<XElement> CreateResource(Package package, string path, CancellationToken token)
+        {
+            Guid id = Guid.NewGuid();
+            string name = Path.GetFileName(path);
+
+            string prefix = FormattableString.Invariant($"{id:d}");
+            long timeAdded = DateTime.UtcNow.ToFileTimeUtc();
+            string type = @"DiagnosticsHub.Resource.File";
```
kkeirstead, line 73: Is there a reason this isn't a constant?

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/DumpUtilities.cs
```diff
@@ -9,11 +9,11 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
 {
     internal static class DumpUtilities
     {
-        public static string GenerateDumpFileName()
+        public static string GenerateDumpFileName(Models.PackageMode mode = Models.PackageMode.None)
```
kkeirstead, line 12: nit: See earlier comment about `packageMode` vs `mode`

src/Microsoft.Diagnostics.Monitoring.WebApi/Packaging.cs
wiktork: When VS will consume this file, it will attempt to use the specific tool. The guid of the tool is fixed.

src/Microsoft.Diagnostics.Monitoring.WebApi/DacLocator.cs
```diff
@@ -37,52 +51,102 @@ public static async Task<string> LocateRuntime(CancellationToken token)
...
+            {
+                return null;
+            }
 
             var runtimes = new List<Runtime>();
             string line = null;
             while ( (line = process.StandardOutput.ReadLine()) != null)
             {
-                runtimes.Add(ParseRuntime(line));
+                if (ParseRuntime(line, out Runtime runtime))
```
wiktork, line 79: nit: TryParseRuntime

src/Microsoft.Diagnostics.Monitoring.WebApi/DumpService.cs
```diff
@@ -78,15 +79,26 @@ public async Task<Stream> DumpAsync(IEndpointInfo endpointInfo, Models.DumpType
...
+                {
+                    dac = Path.Combine(runtimeDir, "libmscordaccore.so");
+                    if (!File.Exists(dac)) 
+                    {
+                        dac = null;
+                    }
+                    dbi = Path.Combine(runtimeDir, "libmscordbi.so");
+                    if (!File.Exists(dbi))
+                    {
+                        dbi = null; ;
```
wiktork, line 97: nit: Remove ;

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/DumpTests.cs
```diff
@@ -74,5 +79,75 @@ public Task DumpTest(DiagnosticPortConnectionMode mode, DumpType type)
...
+            return ScenarioRunner.SingleTarget(
+                _outputHelper,
+                _httpClientFactory,
+                mode,
+                TestAppScenarios.AsyncWait.Name,
+                appValidate: async (runner, client) =>
+                {
+                    int processId = await runner.ProcessIdTask;
+
+                    PackageMode mode = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? PackageMode.DiagSession : PackageMode.IncludeDacDbi;
```
kkeirstead, line 96: We have a variable named `mode` in the encompassing context that has a different meaning (`DiagnosticPortConnectionMode`); one of these should probably be changed.

src/Microsoft.Diagnostics.Monitoring.WebApi/DacLocator.cs
```diff
@@ -14,15 +15,28 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
 {
     internal static class DacLocator
     {
+        private const string NetCoreApp = "Microsoft.NETCore.App";
+        private const char VersionSeperator = '-';
```
kkeirstead, line 19: nit: VersionSeparator instead of VersionSeperator

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/HttpApi/ApiClient.cs
```diff
@@ -204,20 +204,29 @@ public async Task<DotnetMonitorInfo> GetInfoAsync(CancellationToken token)
...
-        private async Task<ResponseStreamHolder> CaptureDumpAsync(string processQuery, DumpType dumpType, CancellationToken token)
+        /// <summary>
+        /// Get /dump?pid={pid}&type={dumpType}&mode={mode}
+        /// </summary>
+        public Task<ResponseStreamHolder> CaptureDumpAsync(int pid, DumpType dumpType, PackageMode mode, CancellationToken token)
+        {
+            return CaptureDumpAsync(GetProcessQuery(pid: pid), dumpType, mode, token);
+        }
+
+
```
kkeirstead, line 226: nit: extra newline

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/DumpTests.cs
```diff
@@ -74,5 +79,75 @@ public Task DumpTest(DiagnosticPortConnectionMode mode, DumpType type)
...
+                    using ResponseStreamHolder holder = await client.CaptureDumpAsync(processId, type, mode);
+                    Assert.NotNull(holder);
+
+                    // The dump operation may still be in progress but the process should still be discoverable.
+                    // If this check fails, then the dump operation is causing dotnet-monitor to not be able
+                    // to observe the process any more.
+                    ProcessInfo processInfo = await client.GetProcessAsync(processId);
+                    Assert.NotNull(processInfo);
+
+                    using var fileStream = new FileStream(Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString()), FileMode.Create, FileAccess.ReadWrite, FileShare.None, 4096, FileOptions.DeleteOnClose);
```
kkeirstead, line 107: nit: for readability, consider putting the path name generation onto its own line

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/DumpTests.cs
```diff
@@ -74,5 +79,75 @@ public Task DumpTest(DiagnosticPortConnectionMode mode, DumpType type)
...
+
+                    PackagePart metadata = parts.FirstOrDefault(p => p.Uri.ToString().Contains("metadata.xml"));
+                    Assert.NotNull(metadata);
+                    PackagePart dump = parts.FirstOrDefault(p => p.Uri.ToString().EndsWith(".dmp"));
+                    Assert.NotNull(dump);
+
+                    if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                    {
+                        PackagePart dac = parts.FirstOrDefault(p => p.Uri.ToString().Contains("mscordaccore"));
+                        Assert.NotNull(dump);
```
kkeirstead, line 134: Should be `Assert.NotNull(dac);`

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/DumpTests.cs
```diff
@@ -74,5 +79,75 @@ public Task DumpTest(DiagnosticPortConnectionMode mode, DumpType type)
...
+                        PackagePart dbi = parts.FirstOrDefault(p => p.Uri.ToString().Contains("mscordbi"));
+                        Assert.NotNull(dump);
```
kkeirstead, line 136: Should be `Assert.NotNull(dbi);`

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/DumpTests.cs
```diff
@@ -74,5 +79,75 @@ public Task DumpTest(DiagnosticPortConnectionMode mode, DumpType type)
                     }
                 });
         }
+
+        [ConditionalTheory(typeof(TestConditions), nameof(TestConditions.IsDumpSupported))]
+        [InlineData(DiagnosticPortConnectionMode.Connect, DumpType.Mini)]
+        public Task DiagSessionTest(DiagnosticPortConnectionMode mode, DumpType type)
```
kkeirstead, line 85: Given that we're only testing a single mode and a single type of dump, should this test just be a Fact? Or do you intend to open this up later with additional modes/types?

wiktork: I am thinking that I would like to do this with Full dumps as well, since testing large artifacts may be worthwhile.

src/Microsoft.Diagnostics.Monitoring.WebApi/DacLocator.cs
wiktork: I added some additional parse checks for the version.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectDumpAction.cs
```diff
@@ -70,7 +70,7 @@ public CollectDumpAction(IServiceProvider serviceProvider, IEndpointInfo endpoin
                     EgressOperation egressOperation = new EgressOperation(
                         token => {
                             startCompletionSource.TrySetResult(null);
-                            return _dumpService.DumpAsync(EndpointInfo, dumpType, token);
+                            return _dumpService.DumpAsync(EndpointInfo, dumpType, PackageMode.None, token);
```
jander-msft, line 73: Is this something we're going to want to expose as an option on the CollectDumpOptions at some point? Don't have to do it now.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/DumpTests.cs
```diff
@@ -97,5 +101,77 @@ public Task DumpTest(DiagnosticPortConnectionMode mode, DumpType type)
...
+
+                    int expectedParts = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? 2 : 4;
+
+                    PackagePartCollection parts = package.GetParts();
+                    Assert.Equal(expectedParts, parts.Count());
+
+                    PackagePart metadata = parts.FirstOrDefault(p => p.Uri.ToString().Contains("metadata.xml"));
+                    Assert.NotNull(metadata);
+                    PackagePart dump = parts.FirstOrDefault(p => p.Uri.ToString().EndsWith(".dmp"));
+                    Assert.NotNull(dump);
```
jander-msft, line 153: Won't this check fail on non-Windows systems since it produces a core dump without an extension?

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/PackageMode.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi.Models
+{
+    [Flags]
+    public enum PackageMode
```
jander-msft, line 10: I'm guessing the intent is that the upper WORD is the "what to include" and the lower WORD is the package format and that we want to maintain this separation so that we can easily add more of each type in the future? If so, can you add comments to this effect?

src/Microsoft.Diagnostics.Monitoring.WebApi/DumpService.cs
```diff
@@ -82,7 +88,26 @@ public async Task<Stream> DumpAsync(IEndpointInfo endpointInfo, Models.DumpType
...
 
-            return new AutoDeleteFileStream(dumpFilePath);
+            string resultPath = dumpFilePath;
+            string dac = null;
+            string dbi = null;
+            if (mode.HasFlag(Models.PackageMode.IncludeDacDbi))
+            {
+                try
+                {
+                    string runtimeDir = DacLocator.LocateRuntimeComponents(endpointInfo, out dac, out dbi);
```
jander-msft, line 98: Looks like the return value is not needed. If it's not used anywhere else, please remove it from the method.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -203,17 +205,24 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
...
```
jander-msft, line 210: This needs model validation such that IncludeDacDbi must be used in conjunction with DiagSession.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/DumpTests.cs
wiktork: We scrape the intermediate file, and this change forces all intermediate files to have a dmp extension.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectDumpAction.cs
wiktork: I think we should long term, but I was not going to make this a part of this change.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
wiktork: IncludeDacDbi implies DiagSession.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -203,17 +205,24 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
...
         {
+            if (mode.HasFlag(Models.PackageMode.IncludeDacDbi) && RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                throw new PlatformNotSupportedException();
```
wiktork, line 214: Convert this to a BadResult/Problem details. Add verbose error message.

wiktork:
Closing for now, until the nuget package for diagsession is ready.

PR #1196, opened 2021-12-13
Update sliding window tests and duration tests to use simulated time.

jander-msft:
This change updates the CollectionRulePipeline and related unit test facades to use the ISystemClock interface for acquiring current time information. This allows the unit tests to mock the current time and increment it as necessary in order to remove the inherent instability of relying on actual system timing. This should result in tests that are more stable and can run faster compared to the prior changes. Additional changes were made to unify the pipeline callbacks into one system within the unit tests, which should make it easier to add more callbacks in the future.


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRules/Triggers/ManualTrigger.cs
```diff
@@ -60,11 +62,18 @@ private void NotifyHandler(object sender, EventArgs args)
...
     internal sealed class ManualTriggerService
     {
-        public event EventHandler Notify;
+        public event EventHandler NotifyStarted;
+
+        public event EventHandler NotifyTrigger;
+
+        public void NotifyStartedSubscribers()
+        {
+            NotifyStarted?.Invoke(this, EventArgs.Empty);
```
kkeirstead, line 71: I don't really have experience with `EventHandler`, so I may be overlooking something, but what exactly are we invoking here? I see we add `NotifyHandler` for `NotifyTrigger`, but nothing for `NotifyStarted`.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/MockSystemClock.cs
```diff
@@ -0,0 +1,21 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.AspNetCore.Authentication;
+using System;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    internal class MockSystemClock : ISystemClock
+    {
+        private DateTimeOffset _utcNow = DateTime.UtcNow;
```
kkeirstead, line 12: What's the reasoning behind using `DateTime.UtcNow` here and `DateTimeOffset.UtcNow` for `RealSystemClock`? I see they're both being assigned to `DateTimeOffset` types, but I don't have a good understanding of why we want `DateTime` for the mocked clock and `DateTimeOffset` for the real clock.

jander-msft: I think they are equivalent when you dig down into them. I'll change the mock to use DateTimeOffset to be consistent.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRules/Triggers/ManualTrigger.cs
jander-msft: The `NotifyStarted` event is subscribed and unsubscribed in the new `ManualTriggerAsync` method in the tests. The `NotifyStartedSubscribers` method is invoked by the `ManualTrigger.StartAsync` method.

kkeirstead: Oops, I just realized that when I searched for it the `CollectionRulePipelineTests` diff hadn't loaded automatically, so it didn't show up - I see it now. I'll close this thread

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRulePipelineTests.cs
```diff
@@ -325,11 +367,93 @@ private void VerifyExecutionCount(CallbackActionService service, int expectedCou
...
+        /// that the actions are invoked for the number of expected iterations (<paramref name="expectedCount"/>) and
+        /// are throttled for the remaining number of iterations.
+        /// </summary>
+        private async Task ManualTriggerAsync(
+            ManualTriggerService triggerService,
+            CallbackActionService callbackService,
+            PipelineCallbacks callbacks,
+            int iterationCount,
+            int expectedCount,
+            MockSystemClock clock,
```
wiktork, line 381: Not necessary but would it make sense to register the clock using depdendency injection? Then in the test context register a mock clock?

wiktork (approved):

Issue #1198, opened 2021-12-14
Add Native Build Support


kelltrick:
## Native Build Support
We should add support to the dotnet-monitor repo to build native code specifically to enable compiling test and produciton cor profilers in `dotnet-monitor`.

Ideally this repo should be able to build the test profilers provided in the runtime repo here: [dotnet/runtime: src/tests/profiler/native](https://github.com/dotnet/runtime/tree/main/src/tests/profiler/native)

We will also want to be able to build our own profiler to support features such as trigger on exception.

PR #1199, opened 2021-12-14
Add LoadProfiler action

### LINKED ISSUE #781: Profiler load support
# Loading a Cor Profiler

The goal of this feature is to support loading a cor profiler to a target process via configuration in `dotnet-monitor`. A requirement for this feature is also to allow configuration of the profiler via environment variables, see **Insert GH Issue**. 

### Usage Requirements
This describes the limitations of this design and states the requirements that will be imposed on the target process and profiler loaded into said target process.
- The target dotnet process must be running on `dotnet 6.0` or higher. This is required because the API `SetStartupProfilerAsync` was added in version `6.0` of the runtime.
- The target process must connect to `dotnet-monitor` using `listen` mode, this is where the target process needs the environment variable, `DOTNET_DiagnosticPorts` to be set to the same pipe that `dotnet-monitor` is using for process communication. This is required because we must be present before the runtime starts user code to load a startup profiler. Using `connect` mode will not allow `dotnet-monitor` to be present in the target process early enough.
- The profiler must be able to be loaded as a startup profiler. Future support is being considered for attach profilers, see **Insert GH Issue**.

## Design

We will add a new `ConditionalRuleAction` called `LoadProfiler` to the available actions. This will use the existing action processing pipeline to execute a profiler load. We will call `SetStartupProfilerAsync` on the diagnostic pipe RPC connection with the provided parameters during startup. If called in any state other than the target process startup, `LoadProfiler` will fail; currently we choose to only support startup profilers, see *Optional Enhancements* for future support. This also means that `dotnet-monitor` must be started in `listen` mode because operating in `connect` mode will cause the connection to happen after the process startup sequence. An example configuration will be:

```json
{
    "CollectionRules": {
        "LoadMyProfiler": {
            "Trigger": {
                "Type": "Startup"
            },
            "Actions": [
                {
                    "Type": "LoadProfiler",
                    "Settings": {
                        "Path": {
                            // Format: "[RID]": "[PathToDll]"
                            "win-x64": "C:\\Path\\To\\x86\\MyCorProfiler.dll",
                            "win-x86": "C:\\Path\\To\\x64\\MyCorProfiler.dll",
                            "alpine-x64": "/opt/AwesomeMonitoringAgent/x64/SomeAgentProfiler.dll",
                            "debian-arm64": "/opt/AwesomeMonitoringAgent/arm64/SomeAgentProfiler.dll",
                        },
                        // Path can be an object or string that applies to 'any' RID
                        "Path": "C:\\Path\\To\\MyCorProfiler.dll",
                        "Guid": "{d452068a-a952-4955-9ee5-41a5192144a2}"
                    }
                }
            ]
        }
    }
}
```

## Settings
- **Path** (a string): This setting will be a string that represents a system path to the dll to load as a cor profiler. It is the responsibility of the user to ensure that the path is accessible to the target process (`dotnet-monitor` does not need access).
-  **Path** (an object): An object containing [Runtime Identifiers (RIDs)](https://docs.microsoft.com/en-us/dotnet/core/rid-catalog) as properties and paths as values. The executing dotnet-monitor will attempt to match the closest RID by traversing the runtime graph down to find the most specific match relative to what the target process is running version. We will need to directly import [`src/libraries/Microsoft.NETCore.Platforms/src/runtime.json`](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.NETCore.Platforms/src/runtime.json) via [Microsoft.NETCore.Platforms](https://www.nuget.org/packages/Microsoft.NETCore.Platforms/) and perform the same platform matching that Nuget performs during package restore.
- **Guid**: A string representing the GUID of the cor profiler. The guid must be printed with the [`B` format specifier](https://docs.microsoft.com/en-us/dotnet/api/system.guid.tostring?view=net-6.0). This will look something like `{d452068a-a952-4955-9ee5-41a5192144a2}`, both upper- and lower-case hex characters are allowed.

## Optional Enhancements
- Attaching a cor profiler after startup. See **Insert GH Issue**

kelltrick:
Adds the `LoadProfiler` action to the list of available CollectionRules to support loading a cor profiler during startup.

This PR is the minimal "unit of code" for checkin and the feature is not complete with this PR. Known limitations include:

- Full RID resolving to support all discrimination on platform. Currently we just load whatever is defined as `any`.
- Add end-to-end functional test loading a profiler (this requires adding native build support to the repo and is a much larger work item, see #1198 )

resolves #781


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    public sealed class LoadProfilerActionTests
+    {
+        private readonly ITestOutputHelper _outputHelper;
+        private static readonly TimeSpan TimeoutMs = TimeSpan.FromSeconds(1);
```
kkeirstead, line 28: My gut reaction is that 1 second is probably a bit short for a testing timeout, but Justin can weigh in on this if he has any thoughts.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
```diff
@@ -0,0 +1,92 @@
...
+
+                IOptionsMonitor<CollectionRuleOptions> collectionRuleOptsMonitor = host.Services.GetRequiredService<IOptionsMonitor<CollectionRuleOptions>>();
+                CollectionRuleOptions ruleOptions = collectionRuleOptsMonitor.Get(DefaultRuleName);
+                ILogger<CollectionRuleService> logger = host.Services.GetRequiredService<ILogger<CollectionRuleService>>();
+
+                CollectionRuleContext context = new(DefaultRuleName, ruleOptions, null, logger);
+
+                LoadProfilerOptions configuredOptions = context.Options.Actions.First().Settings as LoadProfilerOptions;
+                Assert.NotNull(configuredOptions);
+                Assert.True(configuredOptions.PathMap.ContainsKey("any"));
```
kkeirstead, line 65: It looks like we have a bunch of references to the string literal `"any"`; maybe toss this into a constant string like you did in `RidResolver`?

src/Tools/dotnet-monitor/CollectionRules/Actions/ICollectionRuleAction.cs
```diff
@@ -13,6 +13,14 @@ internal interface ICollectionRuleAction
         Task StartAsync(CancellationToken token);
 
         Task<CollectionRuleActionResult> WaitForCompletionAsync(CancellationToken token);
+
+        CollectionRuleActionExecutionType ExecutionType { get; }
+    }
+
+    public enum CollectionRuleActionExecutionType
```
jander-msft, line 20: This concept should not be necessary. The way you wrote the `LoadProfilerAction.ExecuteCoreAsync` method causes the execution system to wait for the profiler to be applied before considering the action startup to have been completed. So the target application will not be resumed, nor will any subsequent action start, before setting profiler has occurred.

src/Tools/dotnet-monitor/CollectionRules/Actions/LoadProfilerAction.cs
```diff
@@ -0,0 +1,115 @@
...
+                catch (Exception ex)
+                {
+                    throw new CollectionRuleActionException(ex);
+                }
+            }
+
+            internal enum ProfilerLoadType
+            {
+                Startup,
+                Attach,
```
jander-msft, line 111: I don't see how we'll ever be able to deterministically support attach. There is no way to ask the runtime "are you in a state that allows profiler attach" without just doing it (same actually applies for startup, but we have better heuristic because it is extremely likely that the runtime is suspended when `dotnet monitor` is configured for listen mode). We would have to add some ability to poll a diagnostic command asking whether attach could ever occur for the given instance.

src/Tools/dotnet-monitor/CollectionRules/Actions/LoadProfilerAction.cs
```diff
@@ -0,0 +1,115 @@
...
+                    await client.SetStartupProfilerAsync(_options.ProfilerGuid, _options.PathMap[choosenRid], token);
+
+                    if (!startCompletionSource.TrySetResult(null))
+                    {
+                        throw new InvalidOperationException();
+                    }
+
+                    return new CollectionRuleActionResult()
+                    {
+                        OutputValues = new Dictionary<string, string>(StringComparer.Ordinal)
```
jander-msft, line 95: I'm not certain either of these pieces of information are relevant outside of this action. Maybe for logging purposes, but then the action should just import an ILogger and log that information rather than surfacing it to the action execution system.

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/LoadProfilerOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// Options for the <see cref="CollectionRules.Actions.LoadProfilerAction"/> action.
+    /// </summary>
+    [DebuggerDisplay("LoadProfiler")]
+    internal sealed partial class LoadProfilerOptions
+    {
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_LoadProfilerOptions_Path))]
+        [Required]
+        public object Path { get; set; } = new Dictionary<string, string>();
```
kkeirstead, line 25: Should we be defining this as an object? My understanding is we're doing it for flexibility to be able to assign it to either a string or a dictionary, and then `PathMap` takes the `Path` (if it's a string) and pushes it into that dictionary format, but couldn't we just make `Path` a dictionary and force all assignments to be in the form of a dictionary (even if there's only a single entry)? I could be overlooking a good reason for taking this approach though.

jander-msft, line 25: This adds additional complexity when I don't think we have a scenario where it is needed yet. Let's make this a `string` property and then add support for platform and architecture differentiation in the future if needed (and design it all up for `dotnet monitor` rather than on a case-by-case basis).

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
```diff
@@ -0,0 +1,92 @@
...
+            await TestHostHelper.CreateCollectionRulesHost(_outputHelper, rootOptions =>
+            {
+                CollectionRuleOptions options = rootOptions.CreateCollectionRule(DefaultRuleName)
+                    .AddAction(KnownCollectionRuleActions.LoadProfiler, settings, out CollectionRuleActionOptions actionOptions)
+                    .SetStartupTrigger();
+
+                settings = (LoadProfilerOptions)options.Actions.Last().Settings;
+            }, host =>
+            {
+                using CancellationTokenSource cancellationTokenSource = new CancellationTokenSource(TimeoutMs);
```
jander-msft, line 55: Not used

src/Tools/dotnet-monitor/CollectionRules/Options/ConfigurationHelper.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            Type configureOptionsType = typeof(IConfigureOptions<>).MakeGenericType(optionsType);
+            MethodInfo configureOptionsConfigMethod = configureOptionsType.GetMethod(
+                name: nameof(IConfigureOptions<object>.Configure), 
+                bindingAttr: BindingFlags.Instance | BindingFlags.Public, 
+                binder: null,
+                types: new Type[] { optionsType }, 
+                modifiers: null);
+
+            bool foundConfigSvcs = false;
```
kkeirstead, line 28: I don't technically see anything wrong with doing it this way from a functional perspective, but it appears that `foundConfigSvcs` is really just a check to see whether `configOptsImpls` is empty . Couldn't we just return that check (e.g. `return configOptsImpls != null && configOptsImpls.Any();`) instead of redundantly setting `foundConfigSvcs` to true for each object?

src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
```diff
@@ -551,4 +551,12 @@
   <data name="DisplayAttributeDescription_GlobalCounterOptions_IntervalSeconds" xml:space="preserve">
     <value>Determines the metrics interval for all dotnet-monitor scenarios. This includes prometheus metrics, live metrics, triggers, and traces.</value>
   </data>
+  <data name="DisplayAttributeDescription_LoadProfilerOptions_Path" xml:space="preserve">
+    <value>Defines the path to the dll for the profiler to be loaded. This is typically the same value that would be passed in the environment variable CORECLR_PROFILER_PATH. This may be a single string or a json object containing properties whose name is a runtime identifier (RID) and whose value is the  corresponding path for that RID.</value>
```
kkeirstead, line 555: nit: Extra space between `the` and `corresponding` near the end of the string.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
```diff
@@ -0,0 +1,92 @@
...
+
+                CollectionRuleContext context = new(DefaultRuleName, ruleOptions, null, logger);
+
+                LoadProfilerOptions configuredOptions = context.Options.Actions.First().Settings as LoadProfilerOptions;
+                Assert.NotNull(configuredOptions);
+                Assert.True(configuredOptions.PathMap.ContainsKey("any"));
+                Assert.Equal(TargetPath, configuredOptions.PathMap["any"]);
+            }, serviceCollection =>
+            {
+                serviceCollection.RegisterCollectionRuleAction<LoadProfilerActionFactory, LoadProfilerOptions>(nameof(KnownCollectionRuleActions.LoadProfiler));
```
jander-msft, line 69: This should not be necessary. It's already part of your change [here](https://github.com/dotnet/dotnet-monitor/blob/25d34fe735ef397c6aa0d7712049f04be0944ecb/src/Tools/dotnet-monitor/ServiceCollectionExtensions.cs) and the `CreateCollectionRulesHost` already calls the `ConfigureCollectionRules` method.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
```diff
@@ -0,0 +1,92 @@
...
+            }, loggingBuilder =>
+            {
+                loggingBuilder.AddProvider(new TestLoggerProvider(record));
```
jander-msft, line 72: Is this needed?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
```diff
@@ -0,0 +1,92 @@
...
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    public sealed class LoadProfilerActionTests
+    {
+        private readonly ITestOutputHelper _outputHelper;
+        private static readonly TimeSpan TimeoutMs = TimeSpan.FromSeconds(1);
+        private const string DefaultRuleName = nameof(ActionDependencyAnalyzerTests);
```
kkeirstead, line 29: I'm assuming this just got copy-pasted from `ActionDependencyAnalyzerTests` and wasn't updated for the current file.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            _outputHelper = outputHelper;
+        }
+
+        [Fact]
+        public async Task TranslateSingleStringToAny()
+        {
+            const string TargetPath = @"C:\My\Test\Path";
+            LogRecord record = new LogRecord();
+            LoadProfilerOptions settings = new LoadProfilerOptions()
```
jander-msft, line 41: This local doesn't really seem to be used. I would inline it with the options configuration callback (where you call `AddAction`).

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
```diff
@@ -0,0 +1,92 @@
...
+                    .SetStartupTrigger();
+
+                settings = (LoadProfilerOptions)options.Actions.Last().Settings;
+            }, host =>
+            {
+                using CancellationTokenSource cancellationTokenSource = new CancellationTokenSource(TimeoutMs);
+
+                IOptionsMonitor<CollectionRuleOptions> collectionRuleOptsMonitor = host.Services.GetRequiredService<IOptionsMonitor<CollectionRuleOptions>>();
+                CollectionRuleOptions ruleOptions = collectionRuleOptsMonitor.Get(DefaultRuleName);
+                ILogger<CollectionRuleService> logger = host.Services.GetRequiredService<ILogger<CollectionRuleService>>();
```
jander-msft, line 59: The `logger` and `context` are not really used/necessary.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
```diff
@@ -0,0 +1,92 @@
...
+                Path = TargetPath,
+                ProfilerGuid = Guid.NewGuid(),                    
+            };
+            await TestHostHelper.CreateCollectionRulesHost(_outputHelper, rootOptions =>
+            {
+                CollectionRuleOptions options = rootOptions.CreateCollectionRule(DefaultRuleName)
+                    .AddAction(KnownCollectionRuleActions.LoadProfiler, settings, out CollectionRuleActionOptions actionOptions)
+                    .SetStartupTrigger();
+
+                settings = (LoadProfilerOptions)options.Actions.Last().Settings;
```
kkeirstead, line 52: Is there a reason for this assignment?

jander-msft (changes requested):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
kelltrick: That was there for debugging but doesn't need to be there for real.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
kelltrick: This isn't actually a full up test, just a unit test. it just loads all the services and exercises them, there isn't any dotnet-monitor

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
kelltrick: nope, this was for debugging.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
kelltrick: but I don't actually need it.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
```diff
@@ -0,0 +1,92 @@
...
+            {
+                serviceCollection.RegisterCollectionRuleAction<LoadProfilerActionFactory, LoadProfilerOptions>(nameof(KnownCollectionRuleActions.LoadProfiler));
+            }, loggingBuilder =>
+            {
+                loggingBuilder.AddProvider(new TestLoggerProvider(record));
+            });
+        }
+
+        [Theory]
+        [InlineData("Windows", "x64", "any")]
```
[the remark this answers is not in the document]
kelltrick: yep, this is for when this implementation is filled in.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
kelltrick: nope

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
kelltrick: Nope, didn't realize `ruleOptions` was a post-configuration value.

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/LoadProfilerOptions.cs
kelltrick: removed.

src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
```diff
@@ -551,4 +551,12 @@
   <data name="DisplayAttributeDescription_GlobalCounterOptions_IntervalSeconds" xml:space="preserve">
     <value>Determines the metrics interval for all dotnet-monitor scenarios. This includes prometheus metrics, live metrics, triggers, and traces.</value>
   </data>
+  <data name="DisplayAttributeDescription_LoadProfilerOptions_Path" xml:space="preserve">
+    <value>Defines the path to the dll for the profiler to be loaded. This is typically the same value that would be passed in the environment variable CORECLR_PROFILER_PATH.</value>
```
jander-msft, line 555:
```suggestion
    <value>The path of the profiler library to be loaded. This is typically the same value that would be set as the CORECLR_PROFILER_PATH environment variable.</value>
```

Let's avoid saying "dll" because it's typically a `.dll` on Windows, `.so` on Linux, and `.dylib` on MacOS.

src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
```diff
@@ -551,4 +551,12 @@
...
+    <comment>The description provided for the Path parameter on LoadProfilerOptions.</comment>
+  </data>
+  <data name="DisplayAttributeDescription_LoadProfilerOptions_ProfilerGuid" xml:space="preserve">
+    <value>The Guid of the cor profiler, this is typically the same value that would be passed in the environment variable CORECLR_PROFILER.</value>
```
jander-msft, line 559:
```suggestion
    <value>The class identifier (or CLSID, typically a GUID) of the ICorProfilerCallback implementation. This is typically the same value that would be set as the CORECLR_PROFILER environment variable.</value>
```

src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
```diff
@@ -551,4 +551,12 @@
   <data name="DisplayAttributeDescription_GlobalCounterOptions_IntervalSeconds" xml:space="preserve">
     <value>Determines the metrics interval for all dotnet-monitor scenarios. This includes prometheus metrics, live metrics, triggers, and traces.</value>
   </data>
+  <data name="DisplayAttributeDescription_LoadProfilerOptions_Path" xml:space="preserve">
+    <value>Defines the path to the dll for the profiler to be loaded. This is typically the same value that would be passed in the environment variable CORECLR_PROFILER_PATH.</value>
+    <comment>The description provided for the Path parameter on LoadProfilerOptions.</comment>
+  </data>
+  <data name="DisplayAttributeDescription_LoadProfilerOptions_ProfilerGuid" xml:space="preserve">
```
jander-msft, line 558:
```suggestion
  <data name="DisplayAttributeDescription_LoadProfilerOptions_Clsid" xml:space="preserve">
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Options/CollectionRuleOptionsExtensions.cs
```diff
@@ -156,6 +156,18 @@ public static CollectionRuleOptions AddExecuteActionAppAction(this CollectionRul
...
         }
 
+        public static CollectionRuleOptions AddLoadProfilerAction(this CollectionRuleOptions options, Action<LoadProfilerOptions> configureOptions)
+        {
+           return options.AddAction(
+                KnownCollectionRuleActions.LoadProfiler,
+                callback: actionOptions =>
+                {
+                    LoadProfilerOptions loadProfilerOptions = new();
+                    configureOptions(loadProfilerOptions);
```
jander-msft, line 166:
```suggestion
                    configureOptions?.Invoke(loadProfilerOptions);
```

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectionRuleActionConstants.cs
```diff
@@ -7,5 +7,6 @@ namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Actions
     internal static class CollectionRuleActionConstants
     {
         public const string EgressPathOutputValueName = "EgressPath";
+        public const string ProfilerLoadTypeValueName = "ProfilerLoadType";
```
jander-msft, line 10: Not used anymore

src/Tools/dotnet-monitor/CollectionRules/Actions/LoadProfilerAction.cs
```diff
@@ -0,0 +1,92 @@
...
+            private readonly LoadProfilerOptions _options;
+
+            public LoadProfilerAction(ILogger logger, IEndpointInfo endpointInfo, LoadProfilerOptions options)
+                : base(endpointInfo, options)
+            {
+                _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+                _endpointInfo = endpointInfo ?? throw new ArgumentNullException(nameof(endpointInfo));
+                _options = options ?? throw new ArgumentNullException(nameof(options));
+
+                if (string.IsNullOrEmpty(_options.Path))
```
jander-msft, line 56: These IsNullOrEmpty and Guid.Empty checks are not necessary since the only caller is the outer class which already validates the options. Let's keep options validation consistent and in few places as possible.

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/LoadProfilerOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_LoadProfilerOptions_Path))]
+        [Required]
+        public string Path { get; set; }
+
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_LoadProfilerOptions_ProfilerGuid))]
+        [Required]
+        public Guid ProfilerGuid { get; set; } = Guid.Empty;
```
jander-msft, line 31:
```suggestion
        public Guid Clsid { get; set; } = Guid.Empty;
```

or `ClassId` or `ClassIdentifier`.

src/Tools/dotnet-monitor/LoggingExtensions.cs
```diff
@@ -511,5 +517,10 @@ public static void EndpointTimeout(this ILogger logger, string processId)
         {
             _endpointTimeout(logger, processId, null);
         }
+
+        public static void LoadingProfiler(this ILogger logger, Guid profilerGuid, string path, int processId)
+        {
+            _loadingProfiler(logger, profilerGuid.ToString(), path, processId.ToString(), null);
```
jander-msft, line 523: Remove the `ToString` calls and let the format string handle whatever format it wants to represent the arguments (by adding format specifiers to each hole). Otherwise, you should be passing the format specifiers to these `ToString` calls as well as a `CultureInfo` implementation (probably `CultureInfo.InvariantCulture`).

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -529,6 +529,14 @@
   <data name="LogFormatString_InvalidToken" xml:space="preserve">
     <value>Action '{actionName}' with setting '{setting}' has invalid tokens.</value>
   </data>
+  <data name="LogFormatString_LoadingProfiler" xml:space="preserve">
+    <value>Loading the startup profiler with id {profilerGuid} from {path} into process with id {processId}.</value>
```
jander-msft, line 533:
```suggestion
    <value>Loading the profiler with CLSID {clsid} from {path} into process {processId}.</value>
```

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/LoadProfilerOptions.cs
jander-msft, line 31: Add a `CustomValidation` attribute to this property that validates that the value is not Guid.Empty.

jander-msft, line 31: The assignment of `Guid.Empty` as the default value is not necessary.

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/LoadProfilerOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.Diagnostics;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Actions
+{
+    /// <summary>
+    /// Options for the <see cref="CollectionRules.Actions.LoadProfilerAction"/> action.
+    /// </summary>
+    [DebuggerDisplay("LoadProfiler")]
+    internal sealed partial class LoadProfilerOptions
```
jander-msft, line 19: The `partial` keywrod is not necessary since there is no other partial class.

jander-msft, line 19: Please add additional tests that verify the validation of the properties. See https://github.com/dotnet/dotnet-monitor/blob/main/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs for examples of verifying the minimally required properties that need to be set to pass validation (*_MinimumOptions), round tripping testing (*_RoundTrip), and full validation failure validation (*_PropertyValidation).

jander-msft (changes requested):

jander-msft (approved):

jander-msft:
Please add documentation for this new action under https://github.com/dotnet/dotnet-monitor/blob/main/documentation/configuration.md#actions

jander-msft:
> Please add documentation for this new action under https://github.com/dotnet/dotnet-monitor/blob/main/documentation/configuration.md#actions

And update the releaseNotes.md

PR #1205, opened 2021-12-14
Config Show Testing

### LINKED ISSUE #687: Add UnitTests for `show config`
## Background and Motivation

Currently, we have no automated tests for `dotnet-monitor show config`

## Proposed Feature

Use the new [`MonitorRunner`](https://github.com/dotnet/dotnet-monitor/blob/main/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunner.cs) base class that allows a different root command to be run with `dotnet-monitor` to test `dotnet-monitor show config --level full` and `dotnet-monitor show config --level redacted`

## Usage Examples

Verify output of `dotnet-monitor show config --level redacted --temp-apikey` is (or similar):
```json
{
  "urls": "https://localhost:52323",
  "Kestrel": ":NOT PRESENT:",
  "CorsConfiguration": ":NOT PRESENT:",
  "DiagnosticPort": {
    "ConnectionMode": "Connect",
    "EndpointName": null
  },
  "Metrics": {
    "Enabled": "True",
    "Endpoints": "http://localhost:52325",
    "IncludeDefaultProviders": "True",
    "MetricCount": "3",
    "UpdateIntervalSeconds": "10"
  },
  "Storage": {
    "DumpTempFolder": "C:\\Users\\pafenelo\\AppData\\Local\\Temp\\"
  },
  "DefaultProcess": ":NOT PRESENT:",
  "ApiAuthentication": {
    "ApiKeyHash": ":REDACTED:",
    "ApiKeyHashType": "SHA256"
  },
  "Egress": ":NOT PRESENT:"
}
````

### PR DESCRIPTION
**Pull request by kkeirstead**
Tests the Config Show command for the individual sections in configuration (e.g. Storage, Egress, Logging, etc), for both `Full` and `Redacted` using text comparison. Currently, the testing doesn't include any checks for `Kestrel` or `CorsConfiguration` since there's no documentation on how to configure them; however, adding them in the future should be a relatively straightforward exercise. To keep the tests and configuration from becoming overly large, the tests are not designed to check all possible edge cases or configuration possibilities, but in its current state this should provide a good baseline check that `Config Show` is behaving properly.

Closes #687

### CODE REVIEW

File: src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs [hunk 1]
```diff
@@ -0,0 +1,268 @@
...
+using Xunit.Abstractions;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    public class ConfigurationTests
+    {
+        private readonly ITestOutputHelper _outputHelper;
+
+        // This needs to be updated and kept in order for any future configuration sections
+        private readonly List<string> orderedConfigurationKeys = new()
```
kelltrick, line 21: You might want to pull this list out of `RootOptions` with reflection and avoid needing to keep this up to date.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExpectedConfigurations/CollectionRules.json
```diff
@@ -0,0 +1,30 @@
...
+      }
+    ],
+    "Limits": {
+      "ActionCount": "2",
+      "ActionCountSlidingWindowDuration": "1:00:00"
+    },
+    "Trigger": {
+      "Settings": {
+        "CounterName": "gc-heap-size",
+        "GreaterThan": "10",
```
wiktork, line 24: Won't this trigger immediately, since a webapp is typically > 10Mi?

wiktork: Nvm, you are not actually running these.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -0,0 +1,268 @@
...
+        {
+            string expectedOutput = "{";
+
+            foreach (var key in orderedConfigurationKeys)
+            {
+                expectedOutput += "\"" + key + "\"" + ":";
+
+                if (categoryMapping.ContainsKey(key))
+                {
+                    string expectedPath = Path.Combine(Directory.GetCurrentDirectory(), "ExpectedConfigurations", categoryMapping[key]);
```
wiktork, line 188: I'm not sure current directory advisable. Is this value the same in the context of command line, visual studio, and ci tests? It's usually safer to use test settings (assuming xunit has them) or Assembly.GetExecutingAssembly().Location or something simlilar.

src/Tools/dotnet-monitor/HostBuilderHelper.cs
```diff
@@ -65,37 +67,51 @@ public static IHostBuilder CreateHostBuilder(string[] urls, string[] metricUrls,
...
+                        ConfigureGlobalMetrics(builder);
+                        builder.ConfigureStorageDefaults();
 
-                    builder.AddCommandLine(new[] { "--urls", ConfigurationHelper.JoinValue(urls) });
+                        builder.AddCommandLine(new[] { "--urls", ConfigurationHelper.JoinValue(urls) });
+                    }
                 })
                 .ConfigureAppConfiguration((IConfigurationBuilder builder) =>
                 {
+                    if (TestingMode != ConfigurationTestingMode.None)
```
kkeirstead, line 83: For context, this is the area I've been needing @jander-msft to take a look at; we had previously discussed a way to get rid of some of these checks regarding testing, but in the meantime they're still here and I'm not sure if we resolved what we were going to do about them. If we can get rid of them, then we shouldn't need to pass around `TestingMode`

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
kkeirstead: I like the sentiment, although it's not the same list as what's in `RootOptions` (and the ordering, at least in the current design, does need to be the same). That being said, if there's a way to avoid needing to keep this up to date that would reduce upkeep

src/Tools/dotnet-monitor/HostBuilderHelper.cs
```diff
@@ -52,7 +52,9 @@ internal static class HostBuilderHelper
                     Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), "." + ProductFolderName) :
                     Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), ProductFolderName));
 
-        private static readonly string UserSettingsPath = Path.Combine(UserConfigDirectoryPath, SettingsFileName);
+        public static string UserSettingsPath = Path.Combine(UserConfigDirectoryPath, SettingsFileName);
+
+        public static ConfigurationTestingMode TestingMode = ConfigurationTestingMode.None;
```
wiktork, line 57: I don't believe this gets used anywhere other than CreateHostBuilder. Consider making it a parameter.

src/Tools/dotnet-monitor/Commands/ConfigShowCommandHandler.cs
```diff
@@ -20,8 +20,14 @@ public static void Invoke(string[] urls, string[] metricUrls, bool metrics, stri
             Write(Console.OpenStandardOutput(), urls, metricUrls, metrics, diagnosticPort, noAuth, tempApiKey, level);
         }
 
-        public static void Write(Stream stream, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, ConfigDisplayLevel level)
+        public static void Write(Stream stream, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, ConfigDisplayLevel level, string userSettingsFilePath = null, ConfigurationTestingMode testingMode = ConfigurationTestingMode.None)
         {
+            if (!string.IsNullOrEmpty(userSettingsFilePath))
+            {
+                HostBuilderHelper.UserSettingsPath = userSettingsFilePath;
```
jander-msft, line 27: Statics will lead to problems, especially when two or more tests run at the same time. Or if one test sets them, but another test expects them to not be set or have their default values. When any of these occur, its really difficult to attempt to reproduce the problem.

kkeirstead: I was previously trying to use environment variables, but it looked like when I ran the tests only one would pass (because all the config show tests were sharing a single value for `userSettingsFilePath` instead of each using their own).

jander-msft: The #1372 PR refactors host building so that all of the configuration directories can be specified via a settings. This allows dotnet-monitor itself to use environment variables and fixed paths for its configuration directories and allows tests to redirect those directories as they need. The PR allow adds some extension methods to override the configuration sources that provide configuration via environment variables to be replaced with in-memory sources for testing.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
kkeirstead: Currently leaving as unresolved - I didn't look too much into this yet, but swapping in Assembly.GetExecutingAssembly().Location didn't work.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -147,6 +165,149 @@ public void ConfigurationOrderingTest(ConfigurationLevel level)
...
+        private void CompareOutput(string actual, string expected)
+        {
+            Assert.Equal(CleanWhitespace(expected), CleanWhitespace(actual));
+        }
+
+        private string ConstructExpectedOutput(bool redact)
+        {
+            Dictionary<string, string> categoryMapping = GetConfigurationFileNames(redact);
+
+            string expectedOutput = "{";
```
jander-msft, line 265: Please use Utf8JsonWriter instead of manually constructing the output.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/SampleConfigurations/Metrics.json
```diff
@@ -0,0 +1,20 @@
+{
+  "GlobalCounter": {
```
jander-msft, line 2: To follow the pattern as the rest of the files, this should be in its own file.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -147,6 +165,149 @@ public void ConfigurationOrderingTest(ConfigurationLevel level)
...
+            }
+
+            expectedOutput += "}";
+
+            return expectedOutput;
+        }
+
+        private string CleanWhitespace(string rawText)
+        {
+            return string.Concat(rawText.Where(c => !char.IsWhiteSpace(c)));
```
jander-msft, line 291: FYI, this will remove spaces from within the strings.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -147,6 +165,149 @@ public void ConfigurationOrderingTest(ConfigurationLevel level)
...
+
+                expectedOutput += (!key.Equals(orderedConfigurationKeys.Last())) ? "," : "";
+            }
+
+            expectedOutput += "}";
+
+            return expectedOutput;
+        }
+
+        private string CleanWhitespace(string rawText)
```
jander-msft, line 289: This may not be necessary if you are using System.Text.Json to serialize and write. They have options for setting the indent to empty string and typically minify their outputs by default: https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-how-to?pivots=dotnet-6-0#how-to-write-net-objects-as-json-serialize

jander-msft (changes requested):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -147,6 +165,149 @@ public void ConfigurationOrderingTest(ConfigurationLevel level)
...
+            {
+                string configString = streamReader.ReadToEnd();
+
+                _outputHelper.WriteLine(configString);
+
+                CompareOutput(configString, expectedConfiguration);
+            }
+        }
+
+        private string ConstructUserSettingsJson()
```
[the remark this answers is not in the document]
kkeirstead: Not sure why I didn't just do it this way in the first place - I guess my instinct is still just to manually do it. This was way easier!

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
kkeirstead: Yep, I just figured it didn't matter since it's all internal comparison. I wouldn't use that on anything that would ever be seen/used.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
kkeirstead: Update: using `Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)` did work. The location was giving me the DLL, which naturally wasn't working for referencing relative files

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -147,6 +167,149 @@ public void ConfigurationOrderingTest(ConfigurationLevel level)
...
+        /// <summary>
+        /// Instead of having to explicitly define every expected value, this reuses the individual categories to ensure they
+        /// assemble properly when combined.
+        /// </summary>
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void FullConfigurationTest(bool redact)
+        {
+            using TemporaryDirectory userSettingsDirectory = new(_outputHelper);
```
jander-msft, line 179: Rather than doing this, could you just pass the content that you want written as the settings.json file directly tho the ConfigShowOutputTest method? I don't see any purpose for writing it to disk first; maybe this is a relic of earlier iterations.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -147,6 +167,149 @@ public void ConfigurationOrderingTest(ConfigurationLevel level)
...
+            string userSettingsFilePath = Path.Combine(userSettingsDirectory.FullName, "settings.json");
+            File.WriteAllText(userSettingsFilePath, ConstructUserSettingsJson());
+
+            ConfigShowOutputTest(redact, userSettingsFilePath, ConstructExpectedOutput(redact));
+        }
+
+        private void ConfigShowOutputTest(bool redact, string userSettingsFilePath, string expectedConfiguration)
```
jander-msft, line 186: Please inline this with FullConfigurationTest since it is the only caller.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -147,6 +167,149 @@ public void ConfigurationOrderingTest(ConfigurationLevel level)
...
+        {
+            string[] fileNames = Directory.GetFiles(Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "SampleConfigurations"));
+
+            IDictionary<string, JsonElement> combinedFiles = new Dictionary<string, JsonElement>();
+
+            foreach (var fileName in fileNames)
+            {
+                IDictionary<string, JsonElement> deserializedFile = JsonSerializer.Deserialize<IDictionary<string, JsonElement>>(File.ReadAllText(fileName));
+
+                combinedFiles = combinedFiles.Union(deserializedFile).ToDictionary(pair => pair.Key, pair => pair.Value);
```
jander-msft, line 247: While this is one line and convenient, I'd imagine it's very inefficient (I know, were only processing ~10 files, so overall it probably isn't too slow). This looks to be O(n^2) processing time because the earlier files are readded to new dictionaries ~n times where n is the number of files.

Try taking the `deserializedFile`, iterate it, and insert its key-value pairs into `combinedFiles` using `Dictionary<TKey, TValue>.Add(TKey, TValue)`. You can use the deconstructor syntax to more easily iterate `deserializedFile`, e.g.
```cs
foreach ((string key, JsonElement element) in deserializedFile)
{
    combinedFiles.Add(key, element);
}
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -147,6 +167,149 @@ public void ConfigurationOrderingTest(ConfigurationLevel level)
...
+            using var stream = new MemoryStream();
+            using var writer = new Utf8JsonWriter(stream);
+
+            writer.WriteStartObject();
+
+            foreach (var key in OrderedConfigurationKeys)
+            {
+                writer.WritePropertyName(key);
+
+                if (categoryMapping.ContainsKey(key))
```
jander-msft, line 273:
```suggestion
                if (categoryMapping.TryGetValue(key, out string fileName))
```
which then avoids the need to index into the dictionary again.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -147,6 +167,149 @@ public void ConfigurationOrderingTest(ConfigurationLevel level)
...
+                }
+            }
+
+            writer.WriteEndObject();
+            writer.Flush();
+
+            return Encoding.UTF8.GetString(stream.ToArray());
+        }
+
+        private string CleanWhitespace(string rawText)
```
jander-msft, line 291: This should be inlined as a static local method of the CompareOutput method.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -147,6 +167,149 @@ public void ConfigurationOrderingTest(ConfigurationLevel level)
...
+
+                combinedFiles = combinedFiles.Union(deserializedFile).ToDictionary(pair => pair.Key, pair => pair.Value);
+            }
+
+            string generatedUserSettings = JsonSerializer.Serialize(combinedFiles);
+
+            return generatedUserSettings;
+        }
+
+        private void CompareOutput(string actual, string expected)
```
jander-msft, line 255: This can be inlined with ConfigShowOutputTest; I don't see much value in keeping it separate.

jander-msft (changes requested):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
kkeirstead: I had the same thought but didn't bother because of the small scale - I'll get this fixed in the next commit.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
kkeirstead: Yep, good catch; I did a fair amount of trimming and reorganizing yesterday to improve maintainability and didn't even think twice about the duplication.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
kkeirstead: I've never used these before - does this provide any specific advantage, or is just for readability/organization?

jander-msft: It's good for encapsulation; since the CleanWhitespace method is an implementation detail of the comparison and won't be used elsewhere, the other methods don't need to have it available. A classic example is a recursive helper method that is only really called by one other method (and itself, hence recursive), so you make it a local function. Local functions can also capture the locals and parameters of their outer function without having to pass them via parameters; the static modifier prevents this behavior.

jander-msft (approved):

PR #1212, opened 2021-12-16
Add documentation for Docker images and Kubernetes examples.

### LINKED ISSUE #1200: Kubernetes deployment example
### Documentation Request
I'd like to see specific examples of the resource definitions that should be used to deploy this as a sidecar in Kubernetes

jander-msft:
This adds documentation for the Docker images (the source of the Dockerfiles and the name of the repositories) and a Kubernetes deployment example.

closes #1200


documentation/kubernetes.md
```diff
@@ -1,18 +1,123 @@
...
+* __Dumps Temporary Folder__: The `dotnet monitor` tool is configured to instruct the application runtime to produce dump files at the path specified by the `DOTNETMONITOR_Storage__DumpTempFolder` environment variable. Without specifying this variable, the application runtime will create dump files in the `/tmp` directory of its container, which is not accessible by the `dotnet monitor` container in this example.
+* __Disabled Authentication__: By default, the `dotnet monitor` tool requires authentication for any of the URLs specified by the `--urls` command line parameter or configured via the `ASPNETCORE_Urls`/`DOTNETMONITOR_Urls` environment variables (see [Authenticatoin](./authentication.md) for further details). For the purposes of this example, authentication has been disabled by using the `--no-auth` command line switch. __Use of this command line switch is NOT recommended for production scenarios__ as it would allow anything with access to the URLs to be able to capture dumps, traces, etc (which may contain sensative information, such as keys) of any process that `dotnet monitor` is able to monitor.
```
kkeirstead, line 103:
```suggestion
* __Disabled Authentication__: By default, the `dotnet monitor` tool requires authentication for any of the URLs specified by the `--urls` command line parameter or configured via the `ASPNETCORE_Urls`/`DOTNETMONITOR_Urls` environment variables (see [Authentication](./authentication.md) for further details). For the purposes of this example, authentication has been disabled by using the `--no-auth` command line switch. __Use of this command line switch is NOT recommended for production scenarios__ as it would allow anything with access to the URLs to be able to capture dumps, traces, etc (which may contain sensitive information, such as keys) of any process that `dotnet monitor` is able to monitor.
```

documentation/docker.md
```diff
@@ -0,0 +1,28 @@
+# Running in Docker
+
+In addition to its availability as a .NET CLI tool, the `dotnet monitor` tool is available as a prebuilt Docker image that can be run in container runtimes and orchestrators.
+
+## Dockerfile Source
+
+### Shippping Dockerfile
```
kkeirstead, line 7:
```suggestion
### Shipping Dockerfile
```

documentation/kubernetes.md
```diff
@@ -1,18 +1,123 @@
...
+        args: [ "--no-auth" ]
+        imagePullPolicy: Always
+        env:
+        - name: DOTNETMONITOR_DiagnosticPort__ConnectionMode
+          value: Listen
+        - name: DOTNETMONITOR_DiagnosticPort__EndpointName
+          value: /diag/port
+        - name: DOTNETMONITOR_Storage__DumpTempFolder
+          value: /dumps
+        # DO NOT remove HTTPS from the URL for deployments in production; this is done for demonstration
```
kelltrick, line 75:
```suggestion
        # ALWAYS use the HTTPS form of the URL for deployments in production; this is done for demonstration
```

documentation/kubernetes.md
```diff
@@ -1,18 +1,123 @@
...
+          requests:
+            cpu: 50m
+            memory: 32Mi
+          limits:
+            cpu: 250m
+            memory: 256Mi
+      volumes:
+      - name: diagvol
+        emptyDir: {}
+      - name: dumpsvol
```
kelltrick, line 94: This might be a dumb question since I've always done it this way and that may be because it doesn't work otherwise but...
Why don't we suggest having a `/diag` volume and then write dumps to `/diag/dumps` and use the diag port at `/diag/port` then the user only needs to define a single mount point for everything.

documentation/kubernetes.md
```diff
@@ -1,18 +1,123 @@
...
+        emptyDir: {}
+```
+
+## Example Details
+
+* __Listen Mode__: The `dotnet monitor` tool is configured to run in `listen` mode. The tool establishes a diagnostic communication channel at the specified Unix Domain Socket path by the `DOTNETMONITOR_DiagnosticPort__EndpointName` environment variable. The application container has a `DOTNET_DiagnosticPorts` environment variable specified so that the application's runtime will communicate with the `dotnet monitor` instance at the specified Unix Domain Socket path. The application runtime will be suspended (e.g. no managed code execution) until it establishes communication with `dotnet monitor` (depends on how long it takes for the `dotnet monitor` container to run, but this should be quick).
```
kelltrick, line 100:
```suggestion
* __Listen Mode__: The `dotnet monitor` tool is configured to run in `listen` mode. The tool establishes a diagnostic communication channel at the specified Unix Domain Socket path by the `DOTNETMONITOR_DiagnosticPort__EndpointName` environment variable. The application container has a `DOTNET_DiagnosticPorts` environment variable specified so that the application's runtime will communicate with the `dotnet monitor` instance at the specified Unix Domain Socket path. The application runtime will be suspended (e.g. no managed code execution) until it establishes communication with `dotnet monitor`. Application startup time will depend on how long it takes for the `dotnet monitor` container to run, but this should be quick.
```

kelltrick (approved):

documentation/kubernetes.md
jander-msft: Not a dumb question. I think that should be fine. Since these are all virtual volumes then it probably doesn't matter how they are split up regarding throughput. I'll give it a quick check and update the example as you suggested.

PR #1215, opened 2021-12-16
Ensure dumps folder exist before issuing dump command.

### LINKED ISSUE #1216: HTTP 400 from `/dump` route and `WriteDumpAsync failed - HRESULT: 0x00000000` if dump temp folder doesn't exist.
Issuing a call to the `/dump` route will return an HTTP 400 with error details stating that `WriteDumpAsync failed - HRESULT: 0x00000000`. It is not obvious what the issue is but the problem is that the dumps temp folder doesn't exist when the runtime tries to write out the dump file.

jander-msft:
If some folder segment in the dumps directory path doesn't exist, the `/dump` endpoint will fail, returning an HTTP 400 with an error message of `WriteDumpAsync failed - HRESULT: 0x00000000`. The logs in `dotnet monitor` show this error too but there's no indication what the issue is.

This is fixed by ensuring that the directory exists before issuing the dump command.

closes #1216

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/DumpService.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/DumpService.cs
@@ -38,7 +38,15 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
                 throw new ArgumentNullException(nameof(endpointInfo));
             }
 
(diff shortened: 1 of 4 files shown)


kkeirstead (approved):
LGTM; one quick question -> is there any scenario where `Directory.CreateDirectory(dumpTempFolder)` might throw due to permissions (or any other reason), and should we be handling it?

PR #1218, opened 2021-12-17
Refactor configuration setup and command handling into separate files

jander-msft:
This PR takes the previously existing DiagnosticsMonitorCommandHandler class and splits the command handlers into different files as well as moving the common host building concepts into a sharable helper class.

Each `dotnet monitor` command has been moved to its own source file under the Commands folder; each of the source files contains the implementation details of the associated command. For example, the `collect` command handler was moved to CollectCommandHandler.cs, which now contains the handler code for building the host (shared among multiple callers) and adding the discovery/http/diagnostic services as a private implementation detail.

The `CreateHostBuilder` method from the previously existing `DiagnosticsMonitorCommandHandler` class has been split into two to separate the concerns of loading configuration and provisioning services. The configuration loading aspects were moved to the `HostBuilderHelper` class whereas the service provisioning aspects were moved to the `CollectCommandHandler` class as a private implementation detail.

This refactor will also further enable more integration tests to be converted to unit tests (with some additional small changes) which will help reduce test time, reduce test complexity, improve test stability, and allow more debugability.


src/Tools/dotnet-monitor/Commands/ConfigShowCommandHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using System;
+using System.IO;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Commands
+{
+    internal sealed class ConfigShowCommandHandler
+    {
+        public static void Invoke(string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, bool noHttpEgress, ConfigDisplayLevel level)
```
kkeirstead, line 15: Do we need `noHttpEgress` as a parameter? It's no longer being passed to the `Write` method

jander-msft: We do if we are telling people "take the parameters from the `collect` command and use them on the `config show` command" to get a replica of the configuration. I can add a comment to that effect.

kkeirstead (approved):

PR #1225, opened 2021-12-21
Exclude dump action testing on MacOS for .NET Core 3.1

### LINKED ISSUE #1206: CollectDumpAction* tests should not be run on MacOS .NET Core 3.1
Dump collection on MacOS is first supported in .NET 5, thus the tests should not be run for .NET Core 3.1 (refer to https://docs.microsoft.com/en-us/dotnet/core/diagnostics/dotnet-dump). This is causing tests to sporadically fail: https://runfo.azurewebsites.net/search/tests/?q=started%3A%7E7+definition%3Adotnet-monitor-ci+name%3ACollectDumpAction

jander-msft:
closes #1206


jander-msft:
I see that the tests are included in the MacOS .NET Core 3.1 test runs but the parameters are not really for .NET Core 3.1; will close and reopen with a more correct solution.

PR #1226, opened 2021-12-21
Update dependency flowed runtimes to 7.0.0 versions.

jander-msft:
This change installs the .NET 7 versions of runtime and aspnet frameworks. All other packages have been kept back at currently shipping versions. This change does not impact the functionality of the tool nor change the target framework support matrix at this time.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -5,17 +5,10 @@
   </solution>
   <packageSources>
     <clear />
-    <!--Begin: Package sources managed by Dependency Flow automation. Do not edit the sources below.-->
-    <!--  Begin: Package sources from dotnet-aspnetcore -->
-    <add key="darc-pub-dotnet-aspnetcore-2d82d2c" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-aspnetcore-2d82d2cc/nuget/v3/index.json" />
-    <!--  End: Package sources from dotnet-aspnetcore -->
-    <!--  Begin: Package sources from dotnet-runtime -->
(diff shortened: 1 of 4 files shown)


NuGet.config
```diff
@@ -5,17 +5,10 @@
...
-    <!--Begin: Package sources managed by Dependency Flow automation. Do not edit the sources below.-->
-    <!--  Begin: Package sources from dotnet-aspnetcore -->
-    <add key="darc-pub-dotnet-aspnetcore-2d82d2c" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-aspnetcore-2d82d2cc/nuget/v3/index.json" />
-    <!--  End: Package sources from dotnet-aspnetcore -->
-    <!--  Begin: Package sources from dotnet-runtime -->
-    <add key="darc-pub-dotnet-runtime-cfb9e77" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-runtime-cfb9e770/nuget/v3/index.json" />
-    <!--  End: Package sources from dotnet-runtime -->
-    <!--End: Package sources managed by Dependency Flow automation. Do not edit the sources above.-->
     <!-- Feeds used in Maestro/Arcade publishing -->
     <add key="dotnet5" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5/nuget/v3/index.json" />
```
kelltrick, line 9: do we get to remove the dotnet5 instance yet?

kelltrick (approved):

NuGet.config
jander-msft: I think it's safe to remove now, but I'm thinking this can be removed when we remove .NET 5 all up from this repo.

jander-msft:
Looks like the Alpine image is too old to support running .NET 7 host on it. Requires #1229 to be completed first.

PR #1230, opened 2021-12-22
Add Get+Set Environment Variable Actions

### LINKED ISSUE #1176: Set + Get a target process Environment Variable
# Setting + Getting an Environment Variable

The goal of this feature is to allow users to set an environment in a target process via a conditional trigger. The most common scenario for this feature would be using this to configure a profiler that will be loaded in the target process. As such, this is a companion feature to [#781: Profiler Load Support](https://github.com/dotnet/dotnet-monitor/issues/781).

### Usage Requirements
This describes the limitations of this design and states the requirements that will be imposed on the target process and `dotnet-monitor` setup.
- The target process must connect to `dotnet-monitor` using `listen` mode, this is where the target process needs the environment variable, `DOTNET_DiagnosticPorts` to be set to the same pipe that `dotnet-monitor` is using for process communication. This is required because setting an environment variable after this stage will yield an unknown state for querying the variables in process; while we can set the variable, it is possible that the old value is already cached and will not be updated.

## Design

We will add a new `ConditionalRuleAction` called `SetEnvironmentVariable` to the available actions. This will use the existing action processing pipeline to set a single environment variable. We will call `SetEnvironmentVariableAsync` on the diagnostic pipe RPC connection with the provided parameters .

```json
{
    "CollectionRules": {
        "LoadMyProfiler": {
            "Trigger": {
                "Type": "Startup"
            },
            "Actions": [
                {
                    "Type": "SetEnvironmentVariable",
                    "Settings": {
                        "Name": "MYPROFILER_ENABLEAWESOMENESS",
                        "Value": "1"
                    },
                    "WaitForComplete": true
                }
            ]
        }
    }
}
```

## Settings
- **Name** (a string): This setting will be the string name of the environment variable. It is up to the user to ensure the string in valid on the platform.
-  **Value** (a string): This setting will be the string value of the environment variable. It is up to the user to ensure the string in valid on the platform .

## Optional Enhancements
- We may make this action process a blob of name/value pairs that appear in a `Values` property in `Settings` or directly in `Settings. This would allow the user to more naturally configure multiple values. This will be considered in future versions.
```json
{
    "CollectionRules": {
        "LoadMyProfiler": {
            "Trigger": {
                "Type": "Startup"
            },
            "Actions": [
                {
                    "Type": "SetEnvironmentVariable",
                    "Settings": {
                        "Values": {
                            "MYPROFILER_ENABLEAWESOMENESS": "1",
                            "MYPROFILER_OtherConfig": "Verbosity=Full",
                            "MYPROFILER_DebugLog": "/var/MyProfiler/svc.log",
                        }
                    },
                    "WaitForComplete": true
                }
            ]
        }
    }
}
```

kelltrick:
- Adds the `GetEnvironmentAction` to available collection rule actions. Get allow users to get an environment variable value and use it in a future action. Set allows a user to specify an environment variable in a process connecting to dotnet-monitor. Set is primarily intended for use with the `LoadProfiler` action to configure the profiler being loaded
- Adds the `SetEnvironmentVariable` to available collection rule actions. Set is primarily intended for use with the `LoadProfiler` action to configure the profiler being loaded. However, this may be used generically to set any environment variable in the process. This should be used from a "startup" trigger to ensure the variable is present as soon as possible in the target process.
- Adds tests for the above action. This required adding a "Command Complete" signal sent from our test app to signal when all information (environment variables) have been sent back over the pipe.
- Adds documentation for the above new actions.

resolves #1176

### DIFF
--- a/documentation/configuration.md
+++ b/documentation/configuration.md
@@ -802,11 +802,11 @@ No outputs
 
 ##### Example
 
-Usage that sets a parameter to the profiler you loaded. In this case, your profiler might be looking for an account key defined in `MyProfiler_ApiKey` which is used to communicate to some outside system.
+Usage that sets a parameter to the profiler you loaded. In this case, your profiler might be looking for an account key defined in `MyProfiler_AccountId` which is used to communicate to some outside system.
 
 ```json
 {
-  "Name": "MyProfiler_ApiKey",
+  "Name": "MyProfiler_AccountId",
   "Value": "8fb138d2c44e4aea8545cc2df541ed4c"
 }
 ```
--- a/documentation/configuration.md
+++ b/documentation/configuration.md
@@ -759,6 +759,104 @@ Usage that executes a .NET executable named "myapp.dll" using `dotnet`.
 }
 ```
 
+#### `LoadProfiler` Action
+
+An action that loads an ICorProfilerCallback implementation into a target process as a startup profiler. This action must be used in a collection rule with a `Startup` trigger.
+
+##### Properties
+
+| Name | Type | Required | Description | Default Value |
+|---|---|---|---|---|
+| `Path` | string | true | The path of the profiler library to be loaded. This is typically the same value that would be set as the CORECLR_PROFILER_PATH environment variable. | |
+| `Clsid` | Guid | true | The class identifier (or CLSID, typically a GUID) of the ICorProfilerCallback implementation. This is typically the same value that would be set as the CORECLR_PROFILER environment variable. | |
+
+##### Outputs
+
+No outputs
+
+##### Example
+
+Usage that loads one of the sample profilers from [`dotnet/runtime`: src/tests/profiler/native/gcallocateprofiler/gcallocateprofiler.cpp](https://github.com/dotnet/runtime/blob/9ddd58a58d14a7bec5ed6eb777c6703c48aca15d/src/tests/profiler/native/gcallocateprofiler/gcallocateprofiler.cpp).
+
+```json
+{
+  "Path": "Profilers\\Profiler.dll",
+  "Clsid": "55b9554d-6115-45a2-be1e-c80f7fa35369"
+}
+```
+
+#### `SetEnvironmentVariable` Action
+
+An action that sets an environment variable value in the target process. This action should be used in a collection rule with a `Startup` trigger.
+
+##### Properties
+
+| Name | Type | Required | Description | Default Value |
+|---|---|---|---|---|
+| `Name` | string | true | The name of the environment variable to set. | |
+| `Value` | string | false | The value of the environment variable to set. | `null` |
+
+##### Outputs
+
+No outputs
+
+##### Example
+
+Usage that sets a parameter to the profiler you loaded. In this case, your profiler might be looking for an account key defined in `MyProfiler_ApiKey` which is used to communicate to some outside system.
+
+```json
+{
+  "Name": "MyProfiler_ApiKey",
+  "Value": "8fb138d2c44e4aea8545cc2df541ed4c"
+}
+```
+
+#### `GetEnvironmentVariable` Action
+
+An action that gets an environment varaible from the target process. Its value is set as the `Value` action output.
+
+##### Properties
+
+| Name | Type | Required | Description | Default Value |
+|---|---|---|---|---|
+| `Name` | string | true | The name of the environment variable to get. | |
+
+##### Outputs
+
+| Name | Description |
+|---|---|
+| `Value` | The value of the environment variable in the target process. |
+
+##### Example
+
+Usage that gets a token your app has access to and uses it to send a trace.
+
+***Note:*** the example below is of an entire action list to provide context, only the second json entry represents the `GetEnvironmentVariable` Action.
+
+```json
+[{
+    "Name": "A",
+    "Type": "CollectTrace",
+    "Settings": {
+        "Profile": "Cpu",
+        "Egress": "AzureBlob"
+    }
+},{
+    "Name": "GetEnvAction",
+    "Type": "GetEnvironmentVariable",
+    "Settings": {
+       "Name": "Azure_SASToken",
+    }
+},{
+    "Name": "B",
+    "Type": "Execute",
+    "Settings": {
+        "Path": "azcopy",
+        "Arguments": "$(Actions.A.EgressPath) https://Contoso.blob.core.windows.net/MyTraces/AwesomeAppTrace.nettrace?$(Actions.GetEnvAction.Value)"
+    }
+}]
+```
+
 ### Limits
 
 Collection rules have limits that constrain the lifetime of the rule and how often its actions can be run before being throttled.
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/AppRunner.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/AppRunner.cs
@@ -34,6 +34,8 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Runners
 
         private TaskCompletionSource<object> _currentCommandSource;
 
+        private Dictionary<string, TaskCompletionSource<string>> _waitingForEnvironmentVariables;
+
         private bool _isDiposed;
 
         /// <summary>
@@ -75,6 +77,8 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Runners
                 tfm);
 
             _runner.TargetFramework = tfm;
+
+            _waitingForEnvironmentVariables = new Dictionary<string, TaskCompletionSource<string>>();
 
             _adapter = new LoggingRunnerAdapter(_outputHelper, _runner);
             _adapter.ReceivedStandardOutputLine += StandardOutputCallback;
@@ -168,10 +172,10 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Runners
         {
             Assert.Null(_currentCommandSource);
             _currentCommandSource = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
+
             _runner.StandardInput.WriteLine(command);
 
             await _currentCommandSource.WithCancellation(token).ConfigureAwait(false);
-
             _currentCommandSource = null;
         }
 
@@ -182,9 +186,9 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Runners
 
         private void HandleProgramEvent(ConsoleLogEvent logEvent)
         {
-            switch (logEvent.EventId)
+            switch ((TestAppLogEventIds)logEvent.EventId)
             {
-                case 1: // ScenarioState
+                case TestAppLogEventIds.ScenarioState:
                     Assert.True(logEvent.State.TryGetValue("state", out TestAppScenarios.SenarioState state));
                     switch (state)
                     {
@@ -193,9 +197,10 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Runners
                             break;
                     }
                     break;
-                case 2: // ReceivedCommand
+                case TestAppLogEventIds.ReceivedCommand:
                     Assert.NotNull(_currentCommandSource);
                     Assert.True(logEvent.State.TryGetValue("expected", out bool expected));
+                    Assert.True(logEvent.State.TryGetValue("command", out _));
                     if (expected)
                     {
                         Assert.True(_currentCommandSource.TrySetResult(null));
@@ -205,7 +210,29 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Runners
                         Assert.True(_currentCommandSource.TrySetException(new InvalidOperationException(logEvent.Message)));
                     }
                     break;
+                case TestAppLogEventIds.EnvironmentVariable:
+                    Assert.True(logEvent.State.TryGetValue("name", out string name));
+                    Assert.True(logEvent.State.TryGetValue("value", out string value));
+                    if (_waitingForEnvironmentVariables.TryGetValue(name, out TaskCompletionSource<string> completedGettingVal))
+                    {
+                        _outputHelper.WriteLine($"Processing callback for envVar: {name}");
+                        Assert.True(completedGettingVal.TrySetResult(value));
+                        _waitingForEnvironmentVariables.Remove(name);
+                    }
+                    break;
             }
+        }
+
+        public async Task<string> WaitForEnvironmentVariable(string name, CancellationToken token)
+        {
+            if (_waitingForEnvironmentVariables.ContainsKey(name))
+            {
+                throw new InvalidOperationException();
+            }
+            _waitingForEnvironmentVariables.Add(name, new TaskCompletionSource<string>());
+            TaskCompletionSource<string> waiter = _waitingForEnvironmentVariables[name];
+            string result = await waiter.WithCancellation(token).ConfigureAwait(false);
+            return result;
         }
     }
 }
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/AppRunner.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/AppRunner.cs
@@ -213,26 +213,33 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Runners
                 case TestAppLogEventIds.EnvironmentVariable:
                     Assert.True(logEvent.State.TryGetValue("name", out string name));
                     Assert.True(logEvent.State.TryGetValue("value", out string value));
-                    if (_waitingForEnvironmentVariables.TryGetValue(name, out TaskCompletionSource<string> completedGettingVal))
+                    lock (_waitingForEnvironmentVariables)
                     {
-                        _outputHelper.WriteLine($"Processing callback for envVar: {name}");
-                        Assert.True(completedGettingVal.TrySetResult(value));
-                        _waitingForEnvironmentVariables.Remove(name);
+                        if (_waitingForEnvironmentVariables.TryGetValue(name, out TaskCompletionSource<string> completedGettingVal))
+                        {
+                            _outputHelper.WriteLine($"Processing callback for envVar: {name}");
+                            Assert.True(completedGettingVal.TrySetResult(value));
+                            _waitingForEnvironmentVariables.Remove(name);
+                        }
                     }
                     break;
             }
         }
 
-        public async Task<string> WaitForEnvironmentVariable(string name, CancellationToken token)
+        public Task<string> WaitForEnvironmentVariable(string name, CancellationToken token)
         {
-            if (_waitingForEnvironmentVariables.ContainsKey(name))
+            TaskCompletionSource<string> waiter;
+            lock (_waitingForEnvironmentVariables)
             {
-                throw new InvalidOperationException();
+                if (_waitingForEnvironmentVariables.ContainsKey(name))
+                {
+                    throw new InvalidOperationException();
+                }
+                _waitingForEnvironmentVariables.Add(name, new TaskCompletionSource<string>());
+                waiter = _waitingForEnvironmentVariables[name];
             }
-            _waitingForEnvironmentVariables.Add(name, new TaskCompletionSource<string>());
-            TaskCompletionSource<string> waiter = _waitingForEnvironmentVariables[name];
-            string result = await waiter.WithCancellation(token).ConfigureAwait(false);
-            return result;
+
+            return waiter.WithCancellation(token);
         }
     }
 }
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/AppRunnerExtensions.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/AppRunnerExtensions.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -122,6 +123,15 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Runners
         {
             using CancellationTokenSource cancellation = new(timeout);
             await runner.StartAsync(cancellation.Token).ConfigureAwait(false);
+        }
+
+        public static async Task<string> GetEnvironmentVariable(this AppRunner runner, string name, TimeSpan timeout)
+        {
+            using CancellationTokenSource cancellation = new(timeout);
+            ConfiguredTaskAwaitable<string> getVarTask = runner.WaitForEnvironmentVariable(name, cancellation.Token).ConfigureAwait(false);
+            await runner.SendCommandAsync(TestAppScenarios.Commands.PrintEnvironmentVariables).ConfigureAwait(false);
+            string result = await getVarTask;
+            return result;
         }
 
         public static Task<int> WaitForExitAsync(this AppRunner runner)
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/TestAppLogEventIds.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/TestAppLogEventIds.cs
@@ -1,0 +1,30 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Extensions.Logging;
+using System;
+
+namespace Microsoft.Diagnostics.Monitoring.TestCommon
+{
+    public enum TestAppLogEventIds
+    {
+        ScenarioState = 1,
+        ReceivedCommand = 2,
+        EnvironmentVariable = 3,
+    }
+    public static class TestAppLogEventIdExtensions
+    {
+        public static EventId EventId(this TestAppLogEventIds enumVal)
+        {
+            string name = Enum.GetName(typeof(TestAppLogEventIds), enumVal);
+            int id = enumVal.Id();
+            return new EventId(id, name);
+        }
+        public static int Id(this TestAppLogEventIds enumVal)
+        {
+            int id = (int)enumVal;
+            return id;
+        }
+    }
+}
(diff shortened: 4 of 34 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
```diff
@@ -829,6 +829,193 @@ public Task CollectionRuleOptions_LoadProfilerAction_ClsidPropertyValidation()
...
+        [Fact]
+        public async Task CollectionRuleOptions_GetEnvironmentVariable_NamePropertyValidation()
+        {
+            await ValidateFailure(
+                rootOptions =>
+                {
+                    rootOptions.CreateCollectionRule(DefaultRuleName)
+                        .SetStartupTrigger()
+                        .AddGetEnvironmentVariableAction(
+                        configureOptions: opts =>
```
jander-msft, line 970: nit: With `configureOptions` having a `null` default, you don't need to specify this anonymous function.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
```diff
@@ -829,6 +829,193 @@ public Task CollectionRuleOptions_LoadProfilerAction_ClsidPropertyValidation()
...
+        [Fact]
+        public async Task CollectionRuleOptions_SetEnvironmentVariable_NamePropertyValidation()
+        {
+            await ValidateFailure(
+                rootOptions =>
+                {
+                    rootOptions.CreateCollectionRule(DefaultRuleName)
+                        .SetStartupTrigger()
+                        .AddSetEnvironmentVariableAction(
+                        configureOptions: opts =>
```
jander-msft, line 889: nit: With `configureOptions` having a `null` default, you don't need to specify this anonymous function.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Options/CollectionRuleOptionsExtensions.cs
```diff
@@ -168,6 +168,30 @@ public static CollectionRuleOptions AddLoadProfilerAction(this CollectionRuleOpt
                 });
         }
 
+        public static CollectionRuleOptions AddSetEnvironmentVariableAction(this CollectionRuleOptions options, Action<SetEnvironmentVariableOptions> configureOptions)
```
jander-msft, line 171: Rather than using a fluent-style API here, just have `string name, string value` parameters and set it within the method.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Options/CollectionRuleOptionsExtensions.cs
```diff
@@ -168,6 +168,30 @@ public static CollectionRuleOptions AddLoadProfilerAction(this CollectionRuleOpt
...
+                 KnownCollectionRuleActions.SetEnvironmentVariable,
+                 callback: actionOptions =>
+                 {
+                     SetEnvironmentVariableOptions setEnvOpts = new();
+                     configureOptions?.Invoke(setEnvOpts);
+                     actionOptions.Settings = setEnvOpts;
+                 });
+        }
+
+        public static CollectionRuleOptions AddGetEnvironmentVariableAction(this CollectionRuleOptions options, Action<GetEnvironmentVariableOptions> configureOptions)
```
jander-msft, line 183: Rather than using a fluent-style API here, just have `string name` parameter and set it within the method.

jander-msft (changes requested):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
kelltrick: These are no longer fluent-style

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
kelltrick: These are no longer fluent-style

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EnvironmentVariableActionTests.cs
```diff
@@ -0,0 +1,269 @@
...
+                            await setAction.StartAsync(cancellationTokenSource.Token);
+                            await setAction.WaitForCompletionAsync(cancellationTokenSource.Token);
+                        }
+                        finally
+                        {
+                            await Tools.Monitor.DisposableHelper.DisposeAsync(setAction);
+                        }
+
+                        await runner.SendCommandAsync(TestAppScenarios.EnvironmentVariables.Commands.PrintVars);
+                        string actualVal = runner.GetEnvironmentVariable(DefaultVarName);
```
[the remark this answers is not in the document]
kelltrick: changed this as discuessed.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/EnvironmentVariablesScenario.cs
```diff
@@ -0,0 +1,73 @@
...
+                            TestAppScenarios.Commands.EndScenario
+                        },
+                        logger,
+                        handleCommand: (string command) =>
+                        {
+                            switch (command)
+                            {
+                                case TestAppScenarios.EnvironmentVariables.Commands.PrintVars:
+                                    IDictionary vars = Environment.GetEnvironmentVariables();
+                                    foreach (string envVarKey in vars.Keys)
```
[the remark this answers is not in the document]
kelltrick: This syntax sugar doesn't work here because the non-generic `IDictionary` here doesn't have a deconstructor.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EnvironmentVariableActionTests.cs
```diff
@@ -0,0 +1,269 @@
...
+                        if (actualVal != DefaultVarValue)
+                        {
+                            _outputHelper.WriteLine("Mismatch vars");
+                        }
+
+                        Assert.Equal(DefaultVarValue, actualVal);
+
+                        await runner.SendCommandAsync(TestAppScenarios.Commands.EndScenario);
```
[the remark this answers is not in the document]
kelltrick: changed this as discuessed.

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/AppRunner.cs
```diff
@@ -205,7 +210,28 @@ private void HandleProgramEvent(ConsoleLogEvent logEvent)
...
+                    {
+                        _outputHelper.WriteLine($"Processing callback for envVar: {name}");
+                        Assert.True(completedGettingVal.TrySetResult(value));
+                        _waitingForEnvironmentVariables.Remove(name);
+                    }
+                    break;
+            }
+        }
+
+        public async Task<string> WaitForEnvironmentVariable(string name, CancellationToken token)
```
kelltrick, line 226: @jander-msft this is what I changed this scenario to use. The test app still prints environment variables one at a time.

If you want to read on or more values, you call this and store the Task (`Task<string> getSetVar = runner.WaitForEnvironmentVariable(DefaultVarName, waitTimeout.Token);`) on one or more values, then call `await runner.SendCommandAsync(TestAppScenarios.EnvironmentVariables.Commands.PrintVars);`
then you can await the tasks above to get the values.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/EnvironmentVariablesScenario.cs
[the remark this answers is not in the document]
jander-msft: Sorry, I totally missed that it is the non-generic interface, let alone it's not the actual type.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/LoggingExtensions.cs
```diff
@@ -12,16 +12,28 @@ internal static class LoggingExtensions
...
                 formatString: "State: {state}");
 
         private static readonly Action<ILogger, string, bool, Exception> _receivedCommand =
             LoggerMessage.Define<string, bool>(
-                eventId: new EventId(2, "ReceivedCommand"),
+                eventId: TestAppLogEventIds.ReceivedCommand.EventId(),
                 logLevel: LogLevel.Debug,
                 formatString: "Received command: {command}; Expected: {expected}");
 
+        private static readonly Action<ILogger, string, Exception> _completedCommand =
```
jander-msft, line 25: This is no longer used.

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/TestAppLogEventIds.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.Extensions.Logging;
+using System;
+
+namespace Microsoft.Diagnostics.Monitoring.TestCommon
+{
+    public enum TestAppLogEventIds
+    {
+        ScenarioState = 1,
+        ReceivedCommand = 2,
+        CompletedCommand = 3,
```
jander-msft, line 14: This is no longer used.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/EnvironmentVariablesScenario.cs
```diff
@@ -0,0 +1,66 @@
...
+            return ScenarioHelpers.RunScenarioAsync(async logger =>
+            {
+                while (true)
+                {
+                    string command = await ScenarioHelpers.WaitForCommandAsync(acceptableCommands, logger);
+
+                    switch (command)
+                    {
+                        case TestAppScenarios.EnvironmentVariables.Commands.PrintVars:
+                            IDictionary vars = Environment.GetEnvironmentVariables();
```
jander-msft, line 44: Because the facility to process these log messages is part of the AppRunner, I think this should move to something that is callable by all tests, such as the ScenarioHelpers class; something like `LogAllEnvironmentVariables()`.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EnvironmentVariableActionTests.cs
```diff
@@ -0,0 +1,219 @@
...
+
+                    Task<IEndpointInfo> newEndpointInfoTask = endpointInfoCallback.WaitAddedEndpointInfoAsync(runner, CommonTestTimeouts.StartProcess);
+
+                    await runner.ExecuteAsync(async () =>
+                    {
+                        IEndpointInfo endpointInfo = await newEndpointInfoTask;
+
+                        ICollectionRuleAction setAction = setFactory.Create(endpointInfo, setOpts);
+
+                        using CancellationTokenSource cancellationTokenSource = new CancellationTokenSource(CommonTestTimeouts.EnvVarsTimeout);
```
jander-msft, line 83: These `cancellationTokenSource` variables in all of the new tests are unused.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EnvironmentVariableActionTests.cs
```diff
@@ -0,0 +1,219 @@
...
+
+                        await ActionTestsHelper.ExecuteAndDisposeAsync(setAction, CommonTestTimeouts.EnvVarsTimeout);
+
+                        using CancellationTokenSource waitTimeout = new CancellationTokenSource(CommonTestTimeouts.SendCommand);
+                        Task<string> getSetVar = runner.WaitForEnvironmentVariable(DefaultVarName, waitTimeout.Token);
```
jander-msft, line 88: Add an extension method to AppRunnerExtensions that takes the environment variable name and uses the default timeout so that the test can be simplified a little.

Consider wrapping all of the WaitForEnvironmentVariable, SendCommandAsync, and awaited value into a new extension method that does all of that since this seems to be repeated:

```cs
public static async Task<string> GetEnvironmentVariable(this AppRunner runner, string name)
{
    Task<string> getVarTask = runner.WaitForEnvironmentVariable(name);
    await runner.SendCommandAsync(TestAppScenarios.EnvironmentVariables.Commands.PrintVars);
    return getVarTask;
}
```

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/EnvironmentVariablesScenario.cs
kelltrick: Moving this to the `ScenarioHelper` would entail inserting this into the `WaitForCommandAsync` handler and simply allowing all tests to call `PrintVars` whenever they want let it be an app level command.

jander-msft: That would be one step further than I was suggesting, which I think I would be okay with, although that means you would be forced to wait for some other command in the test app if all you wanted to do was send the environment variables.

kelltrick: Alright, I did that. That's basically what the `EnvironmentVariable` scenario already does. It has it's own `ShutdownScenario` command to turn it off.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
```diff
@@ -829,6 +829,156 @@ public Task CollectionRuleOptions_LoadProfilerAction_ClsidPropertyValidation()
...
+                {
+                    Assert.Single(ruleOptions.Actions);
+                    ruleOptions.VerifySetEnvironmentVariableAction(0, VariableName, VariableValue);
+                });
+        }
+
+        [Fact]
+        public async Task CollectionRuleOptions_SetEnvironmentVariable_NamePropertyValidation()
+        {
+            await ValidateFailure(
```
kkeirstead, line 874: Since this looks like the exact same validation three times in a row (just with a different illegal environment variable name), should we consider converting this to theory and just passing the illegal name as a parameter?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
```diff
@@ -829,6 +829,156 @@ public Task CollectionRuleOptions_LoadProfilerAction_ClsidPropertyValidation()
...
+                {
+                    Assert.Single(ruleOptions.Actions);
+                    ruleOptions.VerifyGetEnvironmentVariableAction(0, VariableName);
+                });
+        }
+
+        [Fact]
+        public async Task CollectionRuleOptions_GetEnvironmentVariable_NamePropertyValidation()
+        {
+            await ValidateFailure(
```
kkeirstead, line 939: Same comment as above -> consider converting to theory with the environment variable name passed in as a parameter.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Options/CollectionRuleOptionsExtensions.cs
```diff
@@ -168,6 +168,35 @@ public static CollectionRuleOptions AddLoadProfilerAction(this CollectionRuleOpt
...
+
+        public static CollectionRuleOptions AddGetEnvironmentVariableAction(this CollectionRuleOptions options, string name)
+        {
+            return options.AddAction(
+                 KnownCollectionRuleActions.GetEnvironmentVariable,
+                 callback: actionOptions =>
+                 {
+                     GetEnvironmentVariableOptions getEnvOpts = new()
+                     {
+                         Name = name,
```
kkeirstead, line 194: nit: trailing comma

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Options/CollectionRuleOptionsExtensions.cs
```diff
@@ -168,6 +168,35 @@ public static CollectionRuleOptions AddLoadProfilerAction(this CollectionRuleOpt
...
+        public static CollectionRuleOptions AddSetEnvironmentVariableAction(this CollectionRuleOptions options, string name, string value = null)
+        {
+            return options.AddAction(
+                 KnownCollectionRuleActions.SetEnvironmentVariable,
+                 callback: actionOptions =>
+                 {
+                     SetEnvironmentVariableOptions setEnvOpts = new()
+                     {
+                         Name = name,
+                         Value = value,
```
kkeirstead, line 180: nit: trailing comma

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/ScenarioHelpers.cs
```diff
@@ -57,18 +59,39 @@ public static async Task<int> RunScenarioAsync(Func<ILogger, Task<int>> func, Ca
...
 
             while (!receivedExpected && null != (line = await Console.In.ReadLineAsync()))
             {
-                receivedExpected = string.Equals(expectedCommand, line, StringComparison.Ordinal);
+                if (await AppCommands.TryProcessAppCommand(line, logger))
+                {
+                    // We successfully received an app-level command and processed it.
+                    // Restart the loop and wait for the command the scenario is expecting.
+                    receivedExpected = false;
+                    // Still acknoledge this command and say it was expected
```
kkeirstead, line 80:
```suggestion
                    // Still acknowledge this command and say it was expected
```

src/Tools/dotnet-monitor/CollectionRules/Actions/SetEnvironmentVariableAction.cs
```diff
@@ -0,0 +1,82 @@
...
+    internal sealed partial class SetEnvironmentVariableActionFactory :
+        ICollectionRuleActionFactory<SetEnvironmentVariableOptions>
+    {
+        private readonly IServiceProvider _serviceProvider;
+        private readonly ILogger<SetEnvironmentVariableActionFactory> _logger;
+
+        public SetEnvironmentVariableActionFactory(IServiceProvider serviceProvider, ILogger<SetEnvironmentVariableActionFactory> logger)
+        {
+            _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
+            _logger = logger ?? throw new ArgumentNullException(nameof(serviceProvider));
```
kkeirstead, line 26:
```suggestion
            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
kelltrick: I had it does this way to begin with but since the action within `ValidateFailure(...)` is so fast, I didn't want to add the extra overhead of extra test method launches, since we pay a penalty for each test case.

kelltrick:
@wiktork, I added you to this PR because I believed Justin would be out of the office.

src/Tools/dotnet-monitor/CollectionRules/Actions/GetEnvironmentVariableAction.cs
```diff
@@ -0,0 +1,90 @@
...
+            protected override async Task<CollectionRuleActionResult> ExecuteCoreAsync(
+                TaskCompletionSource<object> startCompletionSource,
+                CancellationToken token)
+            {
+                try
+                {
+                    DiagnosticsClient client = new DiagnosticsClient(_endpointInfo.Endpoint);
+
+                    _logger.GettingEnvironmentVariable(_options.Name, _endpointInfo.ProcessId);
+                    Dictionary<string, string> envBlock = await client.GetProcessEnvironmentAsync(token);
```
jander-msft, line 67: Something to think about... This call could be expensive because of the number of environment variable entries or the size of the values for each entry (think about how long PATH is on Windows). Is there a way we could cache this and use it on subsequent `GetEnvironmentVariable` action invocations within the same action list? I don't have a good suggestion here nor do I expect this to be addressed here, but just wanted to make you aware.

src/Tools/dotnet-monitor/CollectionRules/Actions/GetEnvironmentVariableAction.cs
```diff
@@ -0,0 +1,90 @@
...
+                    envBlock.TryGetValue(Options.Name, out string value);
```
jander-msft, line 68: If this fails, it probably should throw an exception with some human readable text e.g. `The environment block does not contain the '{0}' variable.`. In the future, we could add an optional setting that allows silent failure and to report an empty value.

jander-msft (approved):

src/Tools/dotnet-monitor/CollectionRules/Actions/SetEnvironmentVariableAction.cs
```diff
@@ -0,0 +1,82 @@
...
+                try
+                {
+                    DiagnosticsClient client = new DiagnosticsClient(_endpointInfo.Endpoint);
+
+                    _logger.SettingEnvironmentVariable(_options.Name, _endpointInfo.ProcessId);
+                    await client.SetEnvironmentVariableAsync(_options.Name, _options.Value, token);
+
+                    if (!startCompletionSource.TrySetResult(null))
+                    {
+                        throw new InvalidOperationException();
```
wiktork, line 70: Add message?

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/GetEnvironmentVariableOptions.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// Options for the <see cref="CollectionRules.Actions.GetEnvironmentVariableAction"/> action.
+    /// </summary>
+    [DebuggerDisplay("GetEnvironmentVariable")]
+    internal class GetEnvironmentVariableOptions
+    {
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_GetEnvironmentVariableOptions_Name))]
+        [Required]
+        public string Name { get; set; }
```
wiktork, line 21: Did you want to add some validation over the name? I don't believe every character is allowed for environment names.

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/AppRunnerExtensions.cs
```diff
@@ -124,6 +125,15 @@ public static async Task StartAsync(this AppRunner runner, TimeSpan timeout)
             await runner.StartAsync(cancellation.Token).ConfigureAwait(false);
         }
 
+        public static async Task<string> GetEnvironmentVariable(this AppRunner runner, string name, TimeSpan timeout)
+        {
+            using CancellationTokenSource cancellation = new(timeout);
+            ConfiguredTaskAwaitable<string> getVarTask = runner.WaitForEnvironmentVariable(name, cancellation.Token).ConfigureAwait(false);
+            await runner.SendCommandAsync(TestAppScenarios.Commands.PrintEnvironmentVariables).ConfigureAwait(false);
+            string result = await getVarTask;
```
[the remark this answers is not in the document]
kelltrick: There is a second `await` in this method, so it can't be a straight conversion. I can convert the `SetCommandAsync` to a wait and return the `getVarTask` but this seems to be better as-is.

documentation/configuration.md
```diff
@@ -759,6 +759,104 @@ Usage that executes a .NET executable named "myapp.dll" using `dotnet`.
...
+
+#### `SetEnvironmentVariable` Action
+
+An action that sets an environment variable value in the target process. This action should be used in a collection rule with a `Startup` trigger.
+
+##### Properties
+
+| Name | Type | Required | Description | Default Value |
+|---|---|---|---|---|
+| `Name` | string | true | The name of the environment variable to set. | |
```
wiktork, line 796: Have we considered the ability to set multiple environment variables with one action? For example, if i wanted to add clrie here i would have _a lot_ of actions

src/Tools/dotnet-monitor/CollectionRules/Actions/GetEnvironmentVariableAction.cs
kelltrick: This was something I thought about. Caching in a single action list does seem like a reasonable solution but it seemed like overkill for now. My guess is that this is never going to be used, if we get complaints we can fix it.

documentation/configuration.md
kelltrick: The answer is to use many actions. I was thinking it'll be overkill for most people that need to set 2 variables.

wiktork (approved):

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/GetEnvironmentVariableOptions.cs
kelltrick: I looked into it and didn't want to over-validate because Windows and Linux have different characters that are allowed in names.

documentation/configuration.md
jander-msft: This is something we can add in the future and dynamically determine during options binding if we really need it.

src/Tools/dotnet-monitor/CollectionRules/Actions/SetEnvironmentVariableAction.cs
kelltrick: This isn't a customer diagnosable failure. It should never actually fail. I just have the exception there for completeness.

documentation/configuration.md
kelltrick: Exactly, reducing the config size would be the only benefit, which would be meaningful, but we can do that later if needed.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Options/CollectionRuleOptionsExtensions.cs
kelltrick: In the absence of a style guide, I usually do this because when we add more options, it only causes a single line to be changed which can help git do merges easier

PR #1234, opened 2021-12-22
Convert LoggingEventIds to enum

kelltrick:
Let's convert the huge list of constants in [LoggingEventIds.cs](https://github.com/dotnet/dotnet-monitor/blob/main/src/Tools/dotnet-monitor/LoggingEventIds.cs) to an enum to these 2 advantages:
- Compiler will enforce that we don't accidently introduce duplicate IDs, this is very common because we always append to this list and just use the next number
- Avoid having an extra point where we define the ID and a string representing the event. We can just ask the runtime for the enum name and so the code will always match the event names produced at runtime.

The motivation for this is to avoid having to worry about and verify that we don't introduce conflicting IDs in 2 PRs that add an event to this list.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorCollectRunner.CollectionRule.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorCollectRunner.CollectionRule.cs
@@ -20,27 +20,27 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests.Runners
 
         public Task WaitForCollectionRuleActionsCompletedAsync(string ruleName, CancellationToken token)
         {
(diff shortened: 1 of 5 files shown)


jander-msft (approved):
Too bad C# doesn't allow cast operators as extension methods, then you wouldn't need to explicitly cast or call Id().

kelltrick:
> Too bad C# doesn't allow cast operators as extension methods, then you wouldn't need to explicitly cast or call Id().

I'd also settle for calling an extension method to define a `case` in a `switch` statement.

jander-msft:
> > Too bad C# doesn't allow cast operators as extension methods, then you wouldn't need to explicitly cast or call Id().
> 
> I'd also settle for calling an extension method to define a `case` in a `switch` statement.

I suppose an available solution is to switch on `(logEvent.EventId)logEvent.EventId` so that the multiple `int` casts are not necessary.

kelltrick:
> I suppose an available solution is to switch on (LoggingEventIds)logEvent.EventId so that the multiple int casts are not necessary.

I like that suggesting, I'm going to do it.

PR #1235, opened 2021-12-23
Allow tests for each TFM of multi-target test projects to run.

### LINKED ISSUE #1185: Test failure in a multi-targeted test assembly prevents tests in higher TFMs from running.
For example, the Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests assembly is a multi-targeted assembly for the netcoreapp3.1, net5.0, and net6.0 TFMs. When running the tests on the project, if a test fails in the netcoreapp3.1 version, it will not run the tests in the net5.0 and net6.0 versions. This prevents us from evaluating the full extent of a test failure to see if the issue cuts across all TFMs or a subset of TFMs.

jander-msft:
closes #1185

### DIFF
--- a/src/Tests/CrossTargeting.targets
+++ b/src/Tests/CrossTargeting.targets
@@ -1,11 +1,10 @@
 <Project>
-
   <!--
     Arcade uses target batching to run each test assembly TFM for a single test project.
     This causes subsequent items in the batch to not be run if a prior items fail. For example,
     if a test project has the 3.1, 5.0, and 6.0 TFMs and its tests are run in that order and
     the 3.1 TFM has test failures, the 5.0 and 6.0 TFMs will not be run. Override Arcade's
-    RunTests target to use MSBuild to call the Test for each TFM.
+    RunTests target to use MSBuild to call the Test target for each TFM.
    -->
   <Target Name="RunTests">
     <ItemGroup>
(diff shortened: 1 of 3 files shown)


src/Tests/CrossTargeting.targets
```diff
@@ -0,0 +1,22 @@
+<Project>
+  <!--
+    Arcade uses target batching to run each test assembly TFM for a single test project.
+    This causes subsequent items in the batch to not be run if a prior items fail. For example,
+    if a test project has the 3.1, 5.0, and 6.0 TFMs and its tests are run in that order and
+    the 3.1 TFM has test failures, the 5.0 and 6.0 TFMs will not be run. Override Arcade's
+    RunTests target to use MSBuild to call the Test target for each TFM.
```
wiktork, line 7: Does this target change frequently in arcade itself? My concern here is that if Arcade changes this target, we will have difficult to diagnose test run issues.

wiktork (approved):

src/Tests/CrossTargeting.targets
jander-msft: No, it hasn't been changed in years. It will also be very clear that it changed when it does because all legs of the build will fail when it cannot find the RunTests target that we expect in our redefinition of the Test target.

PR #1257, opened 2022-01-03
Update setup and documentation pages to encourage using latest versions.

jander-msft:
Remove the fixed versions of the releases to encourage using the latest versions. Customers can look up exact versions from the sources cited in the existing documentation on the docker.md page as well as in the new section at the bottom of the setup.md page.

### DIFF
--- a/documentation/authentication.md
+++ b/documentation/authentication.md
@@ -96,7 +96,7 @@ spec:
       secretName: apikey
   containers:
   - name: dotnetmonitoragent
-    image: mcr.microsoft.com/dotnet/dotnet-monitor
+    image: mcr.microsoft.com/dotnet/monitor:6.0
     volumeMounts:
       - name: apikey
         mountPath: /etc/dotnet-monitor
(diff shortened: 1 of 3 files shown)


documentation/authentication.md
```diff
@@ -96,7 +96,7 @@ spec:
       secretName: apikey
   containers:
   - name: dotnetmonitoragent
-    image: mcr.microsoft.com/dotnet/dotnet-monitor:6.0.0-preview.8
+    image: mcr.microsoft.com/dotnet/dotnet-monitor
```
wiktork, line 99: A lot of AKS docs discourage using latest tags in production. IIRC we also decided that breaking changes are ok for major version boundaries so we probably would want customers to be explicit about their major version tag.

jander-msft: For AKS, I'll add back the tag, but only as `6.0` to discourage pinning to patch releases.

wiktork (approved):

PR #1273, opened 2022-01-05
Write logger entries to ITestOutputHelper.

jander-msft:
These changes take the log entries and write them to ITestOutputHelper so that the entries are in the test output. This will help diagnose test failures around components that log their state.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/TestOutputLogger.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/TestOutputLogger.cs
@@ -1,0 +1,59 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Extensions.Logging;
+using System;
+using System.Collections.Concurrent;
+using Xunit.Abstractions;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    internal sealed class TestOutputLoggerProvider : ILoggerProvider
+    {
+        private readonly ConcurrentDictionary<string, TestOutputLogger> _loggers = new(StringComparer.OrdinalIgnoreCase);
+        private readonly ITestOutputHelper _outputHelper;
+
+        public TestOutputLoggerProvider(ITestOutputHelper outputHelper)
+        {
+            _outputHelper = outputHelper;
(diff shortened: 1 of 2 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/TestOutputLogger.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        public TestOutputLogger(ITestOutputHelper outputHelper, string categoryName)
+        {
+            _categoryName = categoryName;
+            _outputHelper = outputHelper;
+        }
+
+        public IDisposable BeginScope<TState>(TState state)
+        {
+            return null;
```
wiktork, line 46: Are all callers going to handle null properly here?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/TestOutputLogger.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        public bool IsEnabled(LogLevel logLevel)
+        {
+            return true;
+        }
+
+        public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+        {
+            _outputHelper.WriteLine($"[Logger:{_categoryName}][Id:{eventId.Id}] {formatter(state, exception)}");
```
wiktork, line 56: nit: Maybe just [{_categoryName}][{eventId.Id}]?

wiktork (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/TestOutputLogger.cs
jander-msft: There's a lot of other test output entries that use [] for clarify where they come from. Putting "Logger" in front makes it easier to scan through and find the relevant lines. I can tack the event Id onto the end of the category name to simplify it a little bit.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/TestOutputLogger.cs
jander-msft: It appears to handle it correctly. This might change in the future if we want to log the scopes too, but I don't think that is necessary at this time.

PR #1311, opened 2022-01-13
Add BuildKicker to push PRs through check-run gates

kelltrick:
Well... here is the awful idea. You call it by commenting `/pushrope` with an optional number of retries to allow (default is 3 retries).

### DIFF
--- a/.github/workflows/buildkicker.yml
+++ b/.github/workflows/buildkicker.yml
@@ -55,7 +55,7 @@ jobs:
       uses: actions/checkout@v2
       with:
         fetch-depth: 0
-        ref: 'dev/kelltrick/ropePusher'
+        ref: 'dev/kelltrick/buildkicker'
     - name: Run BuildKicker
       uses: ./eng/actions/buildkicker
       with:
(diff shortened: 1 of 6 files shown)


kelltrick:
This is an example from testing in my own fork:

### Rope Pusher

Started pushing on rope, **Rope Pusher** may re-request each check-run at most 3 time(s).<details>
<summary>Details</summary>

**Rope Pusher** is a github action defined at `.github/workflows/ropepusher.yml` used to automatically retry failed check-runs.

This instance was invoked by `@kelltrick` by adding a comment with the text `/pushrope`. 

> This is called _Rope Pusher_ because I think retrying tests until they succeed is about as useful as pushing on rope, which is to say, not that useful.

[GitHub Action Details](https://github.com/kelltrick/dotnet-monitor/actions/runs/1690924063)</details>

- `2022-01-13T03:33:02.129Z` Found 0 runs and all are in the `completed` + `success` state.

`2022-01-13T03:33:03.277Z` @kelltrick, **Rope Pusher** run complete!

kelltrick:
FYI @jander-msft, you should probably be aware of this

PR #1316, opened 2022-01-14
Add buildkicker

kelltrick:


kelltrick:
Current plan is to check-in this PR for testing, then based on the outcome either:
- If we still get rerequestRun 401 errors, revert the change and abandon the project
- If we get a successful rerequestRun, add any PR feedback changes to a new PR and push that in,

eng/actions/buildkicker/buildkicker.js
```diff
@@ -0,0 +1,336 @@
...
+                            owner: repo_owner,
+                            repo: repo_name,
+                            check_run_id: run.id,
+                          };
+                        console.log("Executing rerequestRun on octokit: " + JSON.stringify(reqParams));
+                        await octokit.rest.checks.rerequestRun(reqParams);
+                    }
+                    else {
+                        // If we are out of retries, abort
+                        throw new BuildKickerException(`Error: out of retries for \`${run.name}\`.`);
```
wiktork, line 271: So what happens when this throws or errors out? Does it interact with the build state in any way?

wiktork (approved):

eng/actions/buildkicker/buildkicker.js
kelltrick: Built state, no. If we throw an exception here, it is caught by the outer try/catch and attempts to append the error message to the issue_comment on the PR.

It looks like this:

### Build Kicker

Started kicking build, **Build Kicker** may re-request each check-run at most 3 time(s).<details>
<summary>Details</summary>

**Build Kicker** is a github action defined at `.github/workflows/buildkicker.yml` used to automatically retry failed check-runs.

This instance was invoked by `@kelltrick` by adding a comment with the text `/kickbuild`. 

> Retrying tests until they suceed isn't a great engineering practice; so, use this command with caution.

[GitHub Action Details](https://github.com/kelltrick/dotnet-monitor/actions/runs/1695213077)</details>


`2022-01-13T22:52:36.387Z` @kelltrick, **Build Kicker** run failed!

```
Cannot read property '0' of undefined
```

PR #1358, opened 2022-01-20
Fix DiagnosticsReleaseTool to allow duplicate files

kelltrick:
Allows duplicate publish paths when output contents SHA512 match.


eng/release/DiagnosticsReleaseTool/Core/Release.cs
```diff
@@ -239,21 +240,30 @@ private async Task<int> LayoutFilesAsync(CancellationToken ct)
...
-                            if (relativePublishPathsUsed.Contains(dstPath))
+                            if (relativePublishPathsToHash.ContainsKey(dstPath))
                             {
-                                _logger.LogError("File {srcPath} is getting published to relative path {dstPath} which is already in use.", srcPath, dstPath);
-                                return -1;
+                                if (!string.IsNullOrEmpty(fileMetadata.Sha512) &&
+                                    !string.IsNullOrEmpty(relativePublishPathsToHash[dstPath]) &&
+                                    relativePublishPathsToHash[dstPath] == fileMetadata.Sha512)
+                                {
+                                    _logger.LogInformation("[{buildFilePath}, {worker}, {layoutInd}: {srcPath} -> {dstPath}, {fileMetadata}] File already published to {dstPath} with same hash: {hash}. This is being allowed.", file.FullName, worker.GetType().FullName, i, srcPath, dstPath, fileMetadata, dstPath, fileMetadata.Sha512);
```
jander-msft, line 249: While this allows the build to succeed, is this what the release team will expect? What's going to happen is that we take a nupkg that was published as to a nuget feed and to a blob feed and only claim that it was published as a NuGetAsset. Do they need to know that it is a BlobAsset as well given that's how it really was published as part of build?

kelltrick: AFIK, the release team doesn't care about the format of our publish feed. The only ask I heard was to keep everything in the single container they assigned to us.

kelltrick: As a side note, I did look through the AzureBlobPublisher and we will publish the file twice, and verify the contents twice; which is fine. I didn't want to duplicate this logic again, so I left it as-is.

eng/release/DiagnosticsReleaseTool/Core/Release.cs
```diff
@@ -240,12 +240,22 @@ private async Task<int> LayoutFilesAsync(CancellationToken ct)
                             (FileMapping fileMap, FileMetadata fileMetadata) = layoutResultArray[i];
                             string srcPath = fileMap.LocalSourcePath;
                             string dstPath = fileMap.RelativeOutputPath;
-                            if (relativePublishPathsUsed.Contains(dstPath))
+                            if (relativePublishPathsToHash.ContainsKey(dstPath))
                             {
-                                _logger.LogError("[{buildFilePath}, {worker}, {layoutInd}: {srcPath} -> {dstPath}, {fileMetadata}] Destination path {dstPath2} already in use.", file.FullName, worker.GetType().FullName, i, srcPath, dstPath, fileMetadata, dstPath);
-                                return -1;
+                                if (!string.IsNullOrEmpty(fileMetadata.Sha512) &&
```
jander-msft, line 245: This has the interesting side effect that if the first layout worker reports a file without a hash value and a subsequent worker reports the same file with a hash value, then this check will fail and publish will fail. We might want to consider hardening the hash assumptions to say that all reported files must have hash values, thus we don't have to think about whether they have hashes or not later.

I'm not asking for this to be corrected now, just something to think about that might simplify this in the future.

jander-msft (approved):

eng/release/DiagnosticsReleaseTool/Core/Release.cs
kelltrick: This is mostly there as a safety check. All our implementations route `FileMetadata` creation through `Helpers.GetDefaultFileMetadata(file, FileClass.XXX)` which will always fill in the SHA512. 

I don't think that assumption is true for all the workers in the Diag repo, so we may need to harden their repo too.

PR #1364, opened 2022-01-21
Port DiagnosticsReleaseTool fix to main

kelltrick:
Ports the following PRs (made to `release/6.x`) into main:
- #1350
- #1351
- #1358
- #1359

This fixes a bug in the release tool that caused it to error when duplicate files exist in the DARC drop.

@hoyosjs let me know if we should also port this over to the Diagnositcs repo. It should be safe for you but it does change one of the core assumptions (that there won't be duplicate files in the DARC drop).


kelltrick:
`/kickbuild`

-- testing new build kicker here

jander-msft (approved):

Issue #277, opened 2021-05-06
dotnet-monitor config show should list all used configuration sources


shirhatti:
`dotnet-monitor config show` should all the configuration providers. I'm currently debugging an issue with my dotnet monitor deployment and it'd be great to know if my config file was read.

PR #1371, opened 2022-01-24
Simplified Diagnostic Port Configuration

### LINKED ISSUE #1292: Proposal: Simplified Diagnostic Port Configuration
The current diagnostic port settings in configuration is currently a child section with two properties. For example:

```json
{
  "DiagnosticPort": {
    "ConnectionMode": "Listen",
    "EndpointName": "/diag/port"
  }
}
```

When `ConnectionMode` is not specified, it is implicitly understood to have a value of `Connect`. Additionally, the `EndpointName` property is only used when `ConnectionMode` is set to `Listen`.

## Proposal

Update the diagnostic port configuration binding to allow the `DiagnosticPort` section to be described with a single string value. When specified in this manner, it is implicitly understood that the port is in `Listen` mode with the specified address. For example:

```json
{
  "DiagnosticPort": "/diag/port"
}
```

This example is equivalent to the first example.

The existing behavior should be preserved as well to maintain backward compatibility.

## Optional

Update the schema for the complex configuration section to suggest using the simplified syntax.

kkeirstead:
Allows users to use a condensed configuration format that omits the need to specify the `ConnectionMode` as `Listen`. This should not regress existing functionality and accepts either format. There were some challenges associated with merged configuration (e.g. if JSON configuration used the standard format and environment variables used the new format, unintended behavior would result); the only complexity in this PR stems from that issue, as seen in `ResolveDiagPortConfigurationCollisions`. The format of `config show` also does not change - using the shortened format will still result in the standard format being used for `config show`.

Closes #1292

### DIFF
--- a/src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
+++ b/src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
@@ -200,8 +200,10 @@ namespace Microsoft.Diagnostics.Tools.Monitor
 
             IEnumerable<IConfigurationSection> children = section.GetChildren();
 
+            bool hasValidChildren = SectionHasValidChildren(section, children);
+
             //If we do not traverse the child sections, the caller is responsible for creating the value
-            if (includeChildSections && children.Any())
+            if (includeChildSections && hasValidChildren)
             {
                 bool isSequentialIndices = CheckForSequentialIndices(children);
 
@@ -230,10 +232,20 @@ namespace Microsoft.Diagnostics.Tools.Monitor
                     ProcessChildren(section, includeChildSections, redact);
                 }
             }
-            else if (!children.Any())
+            else if (!hasValidChildren)
             {
                 WriteValue(section.Value, redact);
             }
+        }
+
+        private bool SectionHasValidChildren(IConfigurationSection section, IEnumerable<IConfigurationSection> children)
+        {
+            if (section.Key.Equals(nameof(RootOptions.DiagnosticPort)))
+            {
+                return string.IsNullOrEmpty(section.Value);
+            }
+
+            return children.Any();
         }
 
         private void WriteValue(string value, bool redact)
--- a/src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
+++ b/src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
@@ -200,10 +200,10 @@ namespace Microsoft.Diagnostics.Tools.Monitor
 
             IEnumerable<IConfigurationSection> children = section.GetChildren();
 
-            bool hasValidChildren = SectionHasValidChildren(section, children);
+            bool canWriteChildren = CanWriteChildren(section, children);
 
             //If we do not traverse the child sections, the caller is responsible for creating the value
-            if (includeChildSections && hasValidChildren)
+            if (includeChildSections && canWriteChildren)
             {
                 bool isSequentialIndices = CheckForSequentialIndices(children);
 
@@ -232,15 +232,15 @@ namespace Microsoft.Diagnostics.Tools.Monitor
                     ProcessChildren(section, includeChildSections, redact);
                 }
             }
-            else if (!hasValidChildren)
+            else if (!canWriteChildren)
             {
                 WriteValue(section.Value, redact);
             }
         }
 
-        private bool SectionHasValidChildren(IConfigurationSection section, IEnumerable<IConfigurationSection> children)
+        private bool CanWriteChildren(IConfigurationSection section, IEnumerable<IConfigurationSection> children)
         {
-            if (section.Key.Equals(nameof(RootOptions.DiagnosticPort)))
+            if (section.Path.Equals(nameof(RootOptions.DiagnosticPort)))
             {
                 return string.IsNullOrEmpty(section.Value);
             }
--- a/src/Tools/dotnet-monitor/DiagnosticPortPostConfigureOptions.cs
+++ b/src/Tools/dotnet-monitor/DiagnosticPortPostConfigureOptions.cs
@@ -8,12 +8,12 @@ using Microsoft.Extensions.Options;
 
 namespace Microsoft.Diagnostics.Tools.Monitor
 {
-    internal sealed class DiagnosticPortNamedOptions :
+    internal sealed class DiagnosticPortPostConfigureOptions :
         IPostConfigureOptions<DiagnosticPortOptions>
     {
         private readonly IConfiguration  _configuration;
 
-        public DiagnosticPortNamedOptions(
+        public DiagnosticPortPostConfigureOptions(
             IConfiguration configuration)
         {
             _configuration = configuration;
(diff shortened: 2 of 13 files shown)


jander-msft:
If I understand correctly, this takes the approach that the highest precedence configuration with the `DiagnosticPort` setting (or subsetting) will win (most of the time). This runs contrary to the normal understanding that configuration is composed using a precedence of sources and interpreted as the whole composition. It makes it easier to explain that "if `DiagnosticPort` is specified, it wins over `DiagnosticPort:EndpointName` and `DiagnosticPort:ConnectionMode`" or vice versa rather than "the highest precedence configuration source with any `DiagnosticPort` will win, kind of" because that requires users to finely understand the order of configuration sources.

The current implementation, if I've ready it correctly, makes that assumption that the `DiagnosticPort:ConnectionMode` and `DiagnosticPort:EndpointName` are configured in the same configuration source. If `DiagnosticPort:ConnectionMode` is configured in a lower source, `DiagnosticPort` in a middle source, and `DiagnosticPort:EndpointName` at a higher source, then `DiagnosticPort` is going to win.

kkeirstead:
Are you suggesting that we should just make a new precedent where we disregard the ordering of configuration sources for this specific case and (in the event of a conflict) prioritize the diagnostic port value over its children (`EndpointName` and `ConnectionMode`)? That would reduce some complexity, although it technically is contradictory to what we do for the rest of configuration. I should be able to make that change relatively easily though if you think it'll be a cleaner standard for users.

Also, the scenario you mentioned was actually an oversight on my part (I can fix it if you'd rather it work this way) - in that situation, `DiagnosticPort` should actually lose because `DiagnosticPort:EndpointName` was specified as higher priority. I just would need to adjust the conditional checks to look not only for `connectionMode` having a value, but also `endpointName`. Then again, I guess you could end up with a weird scenario where `DiagnosticPort:EndpointName` is set, `DiagnosticPort` is set as a lower priority, and then `DiagnosticPort:ConnectionMode` isn't set at all...in which case I guess we would just have to decide how to handle that kind of special scenario.

jander-msft:
> Are you suggesting that we should just make a new precedent where we disregard the ordering of configuration sources for this specific case and (in the event of a conflict) prioritize the diagnostic port value over its children (`EndpointName` and `ConnectionMode`)? That would reduce some complexity, although it technically is contradictory to what we do for the rest of configuration. I should be able to make that change relatively easily though if you think it'll be a cleaner standard for users.

The consumption of the configuration does not understand the ordering of configuration sources. Given that the tool does not yet have any logic of picking one piece of information over another from configuration when both of those represent the same exact aspect, there is no contradiction. The consumption of the information still needs to be reconciled when both are specified, and it's easier to say that "A wins over B" rather than "A or B could win depending on ordering" (although this is true for configuration sources that provide values for the same configuration paths, but that's natively understood by the configuration system itself).

I would suggest looking at implementing a IPostConfigureOptions<DiagnosticPortOptions> that will reconcile the winner and update the options instance so that it is normalized. This way the downstream consumers do not have to think about the reconciliation. The actual implementation of the reconciliation could be in a helper method that is shared among this IPostConfigureOptions implementation and the `config show` implementation. Doing it this leaves the configuration itself unchanged and allows us to more accurately represent it if we decided to implement #277, which reports the source of the configuration value.

kkeirstead:
Ah, in my initial version of this PR I was instead using `IConfigureOptions` until I ran into the ordering issue and switched approaches, so I'm guessing that's more in-line with what you're thinking. I'll work on that for this afternoon.

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -230,12 +232,22 @@ private void ProcessSection(IConfigurationSection section, bool includeChildSect
...
                 }
             }
-            else if (!children.Any())
+            else if (!hasValidChildren)
             {
                 WriteValue(section.Value, redact);
             }
         }
 
+        private bool SectionHasValidChildren(IConfigurationSection section, IEnumerable<IConfigurationSection> children)
```
jander-msft, line 241:
```suggestion
        private bool CanWriteChildren(IConfigurationSection section, IEnumerable<IConfigurationSection> children)
```

Even when the "DiagnosticPort", "DiagnosticPort:ConnectionMode", and "DiagnosticPort:EnpointName" are specified, they are all technically valid (we are not causing configuration to fail nor are the options validation going to fail it). What it's really asking is if it can write the child entries.

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -200,8 +200,10 @@ private void ProcessSection(IConfigurationSection section, bool includeChildSect
 
             IEnumerable<IConfigurationSection> children = section.GetChildren();
 
+            bool hasValidChildren = SectionHasValidChildren(section, children);
```
jander-msft, line 203:
```suggestion
            bool canWriteChildren = CanWriteChildren(section, children);
```

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -230,12 +232,22 @@ private void ProcessSection(IConfigurationSection section, bool includeChildSect
...
-            else if (!children.Any())
+            else if (!hasValidChildren)
             {
                 WriteValue(section.Value, redact);
             }
         }
 
+        private bool SectionHasValidChildren(IConfigurationSection section, IEnumerable<IConfigurationSection> children)
+        {
+            if (section.Key.Equals(nameof(RootOptions.DiagnosticPort)))
```
jander-msft, line 243:
```suggestion
            if (section.Path.Equals(nameof(RootOptions.DiagnosticPort)))
```

To make sure it is only filtering on based on the rooted path named "DiagnosticPort" rather than accidentally considering some other configuration section that happens to have a key of "DiagnosticPort".

src/Tools/dotnet-monitor/DiagnosticPortPostConfigureOptions.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        private readonly IConfiguration  _configuration;
+
+        public DiagnosticPortPostConfigureOptions(
+            IConfiguration configuration)
+        {
+            _configuration = configuration;
+        }
+
+        public void PostConfigure(string name, DiagnosticPortOptions options)
```
jander-msft, line 22: You'll want to write a unit test to validate this once #1372 is merged: setup configuration, bind DiagnosticPortOptions, validate values for difference combinations that exercise this code.

jander-msft (approved):

src/Tools/dotnet-monitor/DiagnosticPortPostConfigureOptions.cs
kkeirstead: Sounds good; do you want me to hold this PR for that, or check this in and then do a separate PR for the test code?

jander-msft: You can do a separate PR for it, but I would say get it done soon so we don't add technical debt long term.

PR #1372, opened 2022-01-25
Add configuration ordering tests.

jander-msft:
Refactor host building to better support unit testing.
Add extension methods for replacing environment configuration sources.

cc @kkeirstead

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.OpenApiGen/Program.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.OpenApiGen/Program.cs
@@ -38,17 +38,18 @@ namespace Microsoft.Diagnostics.Monitoring.OpenApiGen
             // Create directory if it does not exist
             Directory.CreateDirectory(Path.GetDirectoryName(outputPath));
 
+            HostBuilderSettings settings = HostBuilderSettings.CreateMonitor(
+                urls: null,
+                metricUrls: null,
+                metrics: false,
+                diagnosticPort: null,
+                authConfiguration: HostBuilderHelper.CreateAuthConfiguration(noAuth: false, tempApiKey: false));
+
             // Create all of the same services as dotnet-monitor and add
             // OpenAPI generation in order to have it inspect the ASP.NET Core
             // registrations and descriptions.
             IHost host = HostBuilderHelper
-                .CreateHostBuilder(
-                    urls: Array.Empty<string>(),
-                    metricUrls: Array.Empty<string>(),
-                    metrics: true,
-                    diagnosticPort: null,
-                    noAuth: false,
-                    tempApiKey: false)
+                .CreateHostBuilder(settings)
                 .ConfigureServices(services =>
                 {
                     services.AddSwaggerGen(options =>
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
@@ -31,7 +31,7 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
 
         private static readonly Dictionary<string, string> DotnetEnvironmentVariables = new(StringComparer.Ordinal)
         {
-            { WebHostDefaults.ServerUrlsKey, nameof(ConfigurationLevel.DotnetEnviroment) }
+            { WebHostDefaults.ServerUrlsKey, nameof(ConfigurationLevel.DotnetEnvironment) }
         };
 
         private static readonly Dictionary<string, string> MonitorEnvironmentVariables = new(StringComparer.Ordinal)
@@ -39,7 +39,7 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
             { WebHostDefaults.ServerUrlsKey, nameof(ConfigurationLevel.MonitorEnvironment) }
         };
 
-        private static readonly string[] MonitorUrls = new[] { nameof(ConfigurationLevel.MonitorUrls) };
+        private static readonly string[] MonitorUrls = new[] { nameof(ConfigurationLevel.HostBuilderSettingsUrl) };
 
         private static readonly Dictionary<string, string> SharedSettingsContent = new(StringComparer.Ordinal)
         {
@@ -65,8 +65,8 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
         /// </summary>
         [Theory]
         [InlineData(ConfigurationLevel.None)]
-        [InlineData(ConfigurationLevel.DotnetEnviroment)]
-        [InlineData(ConfigurationLevel.MonitorUrls)]
+        [InlineData(ConfigurationLevel.DotnetEnvironment)]
+        [InlineData(ConfigurationLevel.HostBuilderSettingsUrl)]
         [InlineData(ConfigurationLevel.AspnetEnvironment)]
         [InlineData(ConfigurationLevel.AppSettings)]
         [InlineData(ConfigurationLevel.UserSettings)]
@@ -87,12 +87,12 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
                 SharedConfigDirectory = sharedConfigDir.FullName,
                 UserConfigDirectory = userConfigDir.FullName
             };
-            if (level >= ConfigurationLevel.MonitorUrls)
+            if (level >= ConfigurationLevel.HostBuilderSettingsUrl)
             {
                 settings.Urls = MonitorUrls;
             }
 
-            // Write all all of the test files
+            // Write all of the test files
             if (level >= ConfigurationLevel.AppSettings)
             {
                 // This is the appsettings.json file that is normally next to the entrypoint assembly.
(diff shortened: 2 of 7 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -0,0 +1,173 @@
...
+                ContentRootDirectory = contentRootDirectory.FullName,
+                SharedConfigDirectory = sharedConfigDir.FullName,
+                UserConfigDirectory = userConfigDir.FullName
+            };
+            if (level >= ConfigurationLevel.MonitorUrls)
+            {
+                settings.Urls = MonitorUrls;
+            }
+
+            // Write all all of the test files
```
jander-msft, line 95:
```suggestion
            // Write all of the test files
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -0,0 +1,173 @@
...
+                Assert.Equal(Enum.GetName(level), configuredUrls);
+            }
+        }
+
+        /// This is the order of configuration sources where a name with a lower
+        /// enum value has a lower precedence in configuration.
+        public enum ConfigurationLevel
+        {
+            None,
+            DotnetEnviroment,
```
kkeirstead, line 163: nit: `DotnetEnvironment` instead of `DotnetEnviroment`

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -0,0 +1,173 @@
...
+                File.WriteAllText(Path.Combine(sharedConfigDir.FullName, WebHostDefaults.ServerUrlsKey), nameof(ConfigurationLevel.SharedKeyPerFile));
+            }
+
+            // Create the initial host builder.
+            IHostBuilder builder = HostBuilderHelper.CreateHostBuilder(settings);
+
+            // Override the environment configurations to use predefined values so that the test host
+            // doesn't inadvertently provide unexpected values. Passing null replaces with an empty
+            // in-memory collection source.
+            builder.ReplaceAspnetEnvironment(level >= ConfigurationLevel.AspnetEnvironment ? AspnetEnvironmentVariables : null); ;
```
jander-msft, line 129: nit: extra semi-colon

src/Tests/Microsoft.Diagnostics.Monitoring.OpenApiGen/Program.cs
```diff
@@ -38,17 +38,18 @@ public static void Main(string[] args)
             // Create directory if it does not exist
             Directory.CreateDirectory(Path.GetDirectoryName(outputPath));
 
+            HostBuilderSettings settings = HostBuilderSettings.CreateMonitor(
+                urls: null,
+                metricUrls: null,
+                metrics: false,
```
kkeirstead, line 44: Just double checking that this change from `true` to `false` is intentional

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -0,0 +1,173 @@
...
+            }
+            else
+            {
+                Assert.Equal(Enum.GetName(level), configuredUrls);
+            }
+        }
+
+        /// This is the order of configuration sources where a name with a lower
+        /// enum value has a lower precedence in configuration.
+        public enum ConfigurationLevel
```
kkeirstead, line 160: One point of confusion here - is `MonitorUrls` supposed to be equivalent to command line parameters? I expected a command line enum value to be roughly where `MonitorUrls` is, but I don't fully know what that means.

src/Tests/Microsoft.Diagnostics.Monitoring.OpenApiGen/Program.cs
jander-msft: The setting here doesn't impact the functionality of this executable (to generate the OpenAPI document).

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HostBuilderExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            static bool IsEnvironmentConfigurationProvider(IConfigurationProvider provider, string prefix)
+            {
+                if (!(provider is EnvironmentVariablesConfigurationProvider envProvider))
+                {
+                    return false;
+                }
+
+                FieldInfo prefixField = typeof(EnvironmentVariablesConfigurationProvider)
+                    .GetField("_prefix", BindingFlags.Instance | BindingFlags.NonPublic);
```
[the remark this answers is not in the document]
jander-msft: That's the field within the EnvironmentVariablesConfigurationProvider class that contains the environment variable prefix information. It's value will be `ASPNETCORE_` in this context. There's no public way of interrogating what the prefix value is for the EnvironmentVariablesConfigurationProvider.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
jander-msft: In terms of thinking of this as the host builder for dotnet-monitor's `collect` and `config show` commands, yes. However, the is testing at the configuration building level, so it's saying that the Urls parameter of the settings is being set. Calling it `CommandLine` would be conflating it with the default configuration settings where command line arguments are inserted (which our host builder doesn't use).

jander-msft: I've renamed it to `HostBuilderSettingsUrl` to make it clear what is being set.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -0,0 +1,173 @@
...
+
+            // Test that the value of the Urls property is the same as it was set
+            // for the level of configuration of the test.
+            if (level == ConfigurationLevel.None)
+            {
+                Assert.Null(configuredUrls);
+            }
+            else if (level == ConfigurationLevel.DotnetEnvironment)
+            {
+                // Not sure this was the intent that setting DOTNET_Urls via environment yields no
```
jander-msft, line 146: @wiktork, not sure if you had thoughts on this. It feels like setting DOTNET_Urls should override whatever the `--urls` argument is. It's kind of in a weird state right now where DOTNET_Urls has lower precedence but ASPNETCORE_Urls has higher precedence.

wiktork: I was under the impression DOTNET_Urls already supersedes the command line parameter. At least the intent was that the command line parameter is secondary to environment variables since we wanted docker configurability through environment variables.

jander-msft: I can make a separate change to fix that.

wiktork: Sounds good. I probably just tested ASPNET_Urls in the initial implementation.

jander-msft: Fix in #1382

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HostBuilderExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+        public static IHostBuilder ReplaceDotnetEnvironment(this IHostBuilder builder, IDictionary<string, string> values = null)
+        {
+            return builder.ConfigureHostConfiguration(builder =>
+            {
+                ReplaceEnvironment(builder.Sources, prefix: "DOTNET_", values);
+            });
+        }
+
+        /// <summary>
+        /// Replaces the DOTNETMONITOR_* environment configuration source with an in-memory configuration source.
```
kkeirstead, line 67: I'm pretty sure the capitalization doesn't matter, but pointing out for consistency: I noticed that we say DOTNETMONITOR here, as well as DOTNET in the prior method, but `HostBuilderHelper.ConfigPrefix` is actually set to DotnetMonitor.

src/Tools/dotnet-monitor/HostBuilderHelper.cs
```diff
@@ -18,76 +18,39 @@ internal static class HostBuilderHelper
...
             return Host.CreateDefaultBuilder()
-                .UseContentRoot(AppContext.BaseDirectory) // Use the application root instead of the current directory
+                .UseContentRoot(settings.ContentRootDirectory)
                 .ConfigureHostConfiguration((IConfigurationBuilder builder) =>
                 {
                     //Note these are in precedence order.
-                    ConfigureEndpointInfoSource(builder, diagnosticPort);
-                    ConfigureMetricsEndpoint(builder, metrics, metricUrls);
+                    ConfigureEndpointInfoSource(builder, settings.DiagnosticPort);
+                    ConfigureMetricsEndpoint(builder, settings.EnableMetrics, settings.MetricsUrls ?? Array.Empty<string>());
```
kkeirstead, line 30: Is there a reason to do the `?? Array.Empty<string>()` here as opposed to in `CreateMonitor` (this also applies to urls)? I don't have any real issue with this - I was mostly just curious why we didn't deal with the issue at the source on creation.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/HostBuilderExtensions.cs
jander-msft: Correct, capitalization doesn't matter for the prefix. See https://github.com/dotnet/runtime/blob/f85ea976f81945ea18cd5dc71959cccecdc93cd2/src/libraries/Microsoft.Extensions.Configuration.EnvironmentVariables/src/EnvironmentVariablesConfigurationProvider.cs#L48

kkeirstead (approved):

src/Tools/dotnet-monitor/HostBuilderHelper.cs
jander-msft: It simplifies the `HostBuilderSettings` class so that it doesn't have to validate that the property is not null. Couple of options are:
1. Leave as written.
2. Add validation to `HostBuilderSettings` to make sure it is not null.
3. Change `Urls` and `MetricsUrls` to be readonly properties with an empty `List<string>` as its default value. Then call `ToArray` or update downstream to handle `List<string>` instead.

I personally like options 1 or 3, neither requires extra code to validate something that is very unlikely to happen in real usage.

jander-msft: Of the two, option 1 is better at this time because we get the urls as `string[]` and can pass them directly through to the settings. Option 3 would require iterating/copying the array (which isn't going to be too expensive since we're only dealing with probably less than three items 99% of the time).

PR #1382, opened 2022-01-26
Push command line url parameters to lowest precedence.

jander-msft:
This change pushes the command line url parameters to the lowest possible precedence so that DOTNET_* environment variables (e.g. DOTNET_Urls) can override them.

cc @wiktork

### DIFF
--- a/src/Tools/dotnet-monitor/HostBuilderHelper.cs
+++ b/src/Tools/dotnet-monitor/HostBuilderHelper.cs
@@ -61,8 +61,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor
 
         public static IHostBuilder CreateHostBuilder(string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, AuthConfiguration authenticationOptions)
         {
-            return Host.CreateDefaultBuilder()
-                .UseContentRoot(AppContext.BaseDirectory) // Use the application root instead of the current directory
+            return new HostBuilder()
                 .ConfigureHostConfiguration((IConfigurationBuilder builder) =>
                 {
                     //Note these are in precedence order.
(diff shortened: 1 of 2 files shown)


src/Tools/dotnet-monitor/HostBuilderHelper.cs
```diff
@@ -33,6 +32,8 @@ public static IHostBuilder CreateHostBuilder(HostBuilderSettings settings)
 
                     builder.AddCommandLine(new[] { "--urls", ConfigurationHelper.JoinValue(settings.Urls ?? Array.Empty<string>()) });
                 })
+                .ConfigureDefaults(args: null)
```
wiktork, line 35: So you want all the values specified in the HostBuilderSettings to be replaced by the defaults, (such as those from appsettings.json and enviornment variables). Then we process our own custom json files + key value pairs, followed by our environment variables. Consider updating the comments in ConfigureKestrel to reflect this behavior.

wiktork (approved):

src/Tools/dotnet-monitor/HostBuilderHelper.cs
jander-msft: > So you want all the values specified in the HostBuilderSettings to be replaced by the defaults, (such as those from appsettings.json and enviornment variables). Then we process our own custom json files + key value pairs, followed by our environment variables. Consider updating the comments in ConfigureKestrel to reflect this behavior.

I think it would be inconsistent to make the urls argument overridable by everything, but not the metrics arguments and the diagnostic port arguments. In general, I went with the approach that all of the values from the command line (notable exception is the temp API key generation) are considered to be at the same configuration ordering.

PR #1415, opened 2022-01-31
Simplified Diagnostic Port Configuration Testing

kkeirstead:
Provides unit testing for #1371. Checks that `config show` outputs the diagnostic port correctly and that `DiagnosticPortOptions` are set correctly.


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/DiagnosticPortConfigurations/Connect.txt
```diff
@@ -0,0 +1,4 @@
+"DiagnosticPort": {
```
kkeirstead, line 1: Note that these are txt files. This was done because valid JSON would require the opening and closing braces, which would then need to be removed to do the `contains` check. I left it this way because they're small snippets, but this can be changed if we'd rather not use this approach.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/DiagnosticPortTestsHostHelper.cs
```diff
@@ -0,0 +1,66 @@
...
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Options;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Xunit.Abstractions;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    internal static class DiagnosticPortTestsHostHelper
```
kkeirstead, line 17: This is closely modeled after `TestHostHelper`; I originally planned to integrate with that, but decided it carried a lot of things that aren't relevant for this use-case and made a separate version for this set of tests.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/DiagnosticPortTestsHelper.cs
```diff
@@ -0,0 +1,51 @@
...
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    internal static class DiagnosticPortTestsHelper
+    {
+        public static IEnumerable<object[]> GetFileNamesAndEnvironmentVariables()
+        {
+            yield return new object[] { "SimplifiedListen.txt", DiagnosticPortTestsConstants.SimplifiedListen_EnvironmentVariables };
+            yield return new object[] { "FullListen.txt", DiagnosticPortTestsConstants.FullListen_EnvironmentVariables };
+            yield return new object[] { "Connect.txt", DiagnosticPortTestsConstants.Connect_EnvironmentVariables };
+            yield return new object[] { "SimplifiedListen.txt", DiagnosticPortTestsConstants.AllListen_EnvironmentVariables };
```
kkeirstead, line 20: Using `SimplifiedListen.txt` here is intentional (since that was the behavior we settled on when there's a conflict); however, if we're concerned about this looking like a bug in the code, I can add a comment stating that explicitly or an extra txt file that's identical to `SimplifiedListen.txt`

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/DiagnosticPortTestsHostHelper.cs
jander-msft: It's okay that `TestHostHelper` adds other services and configuration so long as it doesn't impact the diagnostic port testing; I'd rather reuse the common class rather than copying it and slimming it down for each individual use case.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/DiagnosticPortTestsHelper.cs
```diff
@@ -0,0 +1,23 @@
...
+using Microsoft.Diagnostics.Tools.Monitor;
+using Microsoft.Extensions.Hosting;
+using System.Collections.Generic;
+using Xunit.Abstractions;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    internal static class DiagnosticPortTestsHelper
+    {
+        public static IEnumerable<object[]> GetFileNamesAndEnvironmentVariables()
```
jander-msft, line 15: Move this to be a method on the ConfigurationTests class. I don't think any other test is going to use this.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -202,14 +202,62 @@ public void FullConfigurationTest(bool redact)
...
+            string generatedConfig = WriteAndRetrieveConfiguration(rootConfiguration, redact);
+
+            Assert.Equal(CleanWhitespace(generatedConfig), CleanWhitespace(ConstructExpectedOutput(redact)));
+        }
+
+        /// <summary>
+        /// Tests that the connection mode is set correctly for various configurations of the diagnostic port
+        /// </summary>
+        [Theory]
+        [MemberData(nameof(DiagnosticPortTestsHelper.GetFileNamesAndEnvironmentVariables), MemberType = typeof(DiagnosticPortTestsHelper))]
```
jander-msft, line 218: If the GetFileNamesAndEnvironmentVariables method is moved to this class, then you can just refer to it as `[MemberData(nameof(DiagnosticPortTestsHelper.GetFileNamesAndEnvironmentVariables))]` without the member type qualification.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -202,14 +202,62 @@ public void FullConfigurationTest(bool redact)
...
+        public void TestConnectionMode(string fileName, IDictionary<string, string> diagnosticPortEnvironmentVariables)
```
jander-msft, line 219: nit: ConnectionModeTest

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/DiagnosticPortTestsHelper.cs
jander-msft: Maybe rename to `GetConnectionModeTestArguments`

jander-msft (approved):

PR #1431, opened 2022-02-02
Add first entry in troubleshooting guide

### LINKED ISSUE #1326: High CPU usage for process 'dotnet-moni+' when running in Azure Web App (Linux)
We are experiencing high CPU usage from the process 'dotnet-moni+' when running in Azure Web App (Linux) affecting the performance of our API in Production. The API we have in Production has about 900,000 requests a day so does have a decent amount of traffic. When we do see this high CPU usage pop up we restart the web app and it goes back to normal for a few weeks but after a few weeks later we get the same issue of high CPU from that same process again so it seems cumulative. Has anyone experienced this issue?

See the screenshot below taken from the Azure diagnostic tools. We run the API on the plan P1v2 with 5 instances. You can see all 5 instances of our API has high CPU usage of that same process.
HighCpuUsage

kelltrick:
Adds the first entry in our troubleshooting guide to address the issue found in #1326, which ended up being high log volume.

Closes #1326

### DIFF
--- a/documentation/troubleshooting.md
+++ b/documentation/troubleshooting.md
@@ -1,0 +1,15 @@
+# Troubleshooting Guide
+
+Here is our guide for diagnosing specific issues with `Dotnet-Monitor`. This is an ongoing effort to add issues and solutions to this guide and is never complete but you can help us by letting us know in the issues sections if we should add something here
+
+## Problem: In Azure App Service on Linux, dotnet-monitor is consuming lots of CPU time
+
+If you find that the `dotnet-monitor` is consuming lots of CPU time, this typically means that there is a high load of logging statements being emitted from the process. `dotnet-monitor` is used to capture `AppServiceAppLogs` and `AllMetrics` from your process before sending them onto other systems; these settings are configured under Monitoring > Diagnostic Settings in the Azure Portal. `dotnet-monitor` is also always enabled to collect a baseline level of metrics from the process to display in the Azure Portal, however this activity is fast enough that it should not show up as an impact on CPU time.
+
+> Solution #1 (temporary fix):
+>
+> Turn off `AppServiceAppLogs` under  Monitoring > Diagnostic Settings
+
+> Solution #2:
+>
+> Adjust the logging being emitted in your app to be in-line with CPU performance.


documentation/troubleshooting.md
```diff
@@ -0,0 +1,15 @@
+# Troubleshooting Guide
+
+Here is our guide for diagnosing specific issues with `Dotnet-Monitor`. This is an ongoing effort to add issues and solutions to this guide and is never complete but you can help us by letting us know in the issues sections if we should add something here
```
wiktork, line 3: nit: 'here.'

documentation/troubleshooting.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Problem: In Azure App Service on Linux, dotnet-monitor is consuming lots of CPU time
+
+If you find that the `dotnet-monitor` is consuming lots of CPU time, this typically means that there is a high load of logging statements being emitted from the process. `dotnet-monitor` is used to capture `AppServiceAppLogs` and `AllMetrics` from your process before sending them onto other systems; these settings are configured under Monitoring > Diagnostic Settings in the Azure Portal. `dotnet-monitor` is also always enabled to collect a baseline level of metrics from the process to display in the Azure Portal, however this activity is fast enough that it should not show up as an impact on CPU time.
```
wiktork, line 7: While not applicable directly to this issue, consider adding a link to our container limits doc.

documentation/troubleshooting.md
```diff
@@ -0,0 +1,15 @@
...
+
+> Solution #1 (temporary fix):
+>
+> Turn off `AppServiceAppLogs` under  Monitoring > Diagnostic Settings
+
+> Solution #2:
+>
+> Adjust the logging being emitted in your app to be in-line with CPU performance.
```
wiktork, line 15: Consider a link to https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-6.0#configure-logging

documentation/troubleshooting.md
wiktork, line 7:
```suggestion
If you find that the `dotnet-monitor` is consuming lots of CPU time, this typically means that there is a high load of logging statements being emitted from the process. `dotnet-monitor` is used to capture `AppServiceAppLogs` and `AllMetrics` from your process before sending them onto other systems; these settings are configured under Monitoring > Diagnostic Settings in the Azure Portal. `dotnet-monitor` is also always enabled to collect a baseline level of metrics from the process to display in the Azure Portal. However this activity is fast enough that it should not show up as an impact on CPU time.
```

wiktork (approved):

PR #1438, opened 2022-02-03
Config Show - List configuration sources

### LINKED ISSUE #277: dotnet-monitor config show should list all used configuration sources

kkeirstead:
With the addition of the `--show-sources` flag, users can now view the configuration sources that `dotnet monitor` is using to generate the configuration for debugging purposes. NOTE: this does not show which parts of the configuration are associated with each provider - if we'd rather invest in trying to do that, we can close this PR and I can work on a different solution.

Closes #277

### DIFF
--- a/src/Tools/dotnet-monitor/Commands/ConfigShowCommandHandler.cs
+++ b/src/Tools/dotnet-monitor/Commands/ConfigShowCommandHandler.cs
@@ -7,6 +7,7 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using System;
 using System.IO;
+using System.Linq;
 
 namespace Microsoft.Diagnostics.Tools.Monitor.Commands
 {
@@ -15,12 +16,12 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Commands
         // Although the "noHttpEgress" parameter is unused, it keeps the entire command parameter set a superset
         // of the "collect" command so that users can take the same arguments from "collect" and use it on "config show"
         // to get the same configuration without telling them to drop specific command line arguments.
-        public static void Invoke(string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, bool noHttpEgress, ConfigDisplayLevel level)
+        public static void Invoke(string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, bool noHttpEgress, ConfigDisplayLevel level, bool showSources)
         {
-            Write(Console.OpenStandardOutput(), urls, metricUrls, metrics, diagnosticPort, noAuth, tempApiKey, level);
+            Write(Console.OpenStandardOutput(), urls, metricUrls, metrics, diagnosticPort, noAuth, tempApiKey, level, showSources);
         }
 
-        public static void Write(Stream stream, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, ConfigDisplayLevel level)
+        public static void Write(Stream stream, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, ConfigDisplayLevel level, bool showSources)
         {
             IAuthConfiguration authConfiguration = HostBuilderHelper.CreateAuthConfiguration(noAuth, tempApiKey);
             HostBuilderSettings settings = HostBuilderSettings.CreateMonitor(urls, metricUrls, metrics, diagnosticPort, authConfiguration);
@@ -28,6 +29,30 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Commands
             IConfiguration configuration = host.Services.GetRequiredService<IConfiguration>();
             using ConfigurationJsonWriter jsonWriter = new ConfigurationJsonWriter(stream);
             jsonWriter.Write(configuration, full: level == ConfigDisplayLevel.Full, skipNotPresent: false);
+
+            if (showSources)
+            {
+                WriteConfigurationProviders(stream, configuration);
+            }
+        }
+
+        private static void WriteConfigurationProviders(Stream stream, IConfiguration configuration)
+        {
+            var configurationProviders = ((IConfigurationRoot)configuration).Providers.Reverse();
+
+            StreamWriter writer = new StreamWriter(stream);
+
+            writer.WriteLine("ConfigurationProviders (High to Low Priority):");
+
+
+            foreach (var provider in configurationProviders)
+            {
+                writer.WriteLine(" - " + provider.ToString());
+            }
+
+            writer.WriteLine();
+
+            writer.Flush();
         }
     }
 }
--- a/src/Tools/dotnet-monitor/Commands/ConfigShowCommandHandler.cs
+++ b/src/Tools/dotnet-monitor/Commands/ConfigShowCommandHandler.cs
@@ -7,7 +7,6 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using System;
 using System.IO;
-using System.Linq;
 
 namespace Microsoft.Diagnostics.Tools.Monitor.Commands
 {
@@ -32,26 +31,8 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Commands
 
             if (showSources)
             {
-                WriteConfigurationProviders(stream, configuration);
+                jsonWriter.WriteProviders(configuration);
             }
-        }
-
-        private static void WriteConfigurationProviders(Stream stream, IConfiguration configuration)
-        {
-            var configurationProviders = ((IConfigurationRoot)configuration).Providers.Reverse();
-
-            StreamWriter writer = new StreamWriter(stream);
-
-            writer.WriteLine("Configuration Providers (High to Low Priority):");
-
-            foreach (var provider in configurationProviders)
-            {
-                writer.WriteLine(" - " + provider.ToString());
-            }
-
-            writer.WriteLine();
-
-            writer.Flush();
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/Tools/dotnet-monitor/Commands/ConfigShowCommandHandler.cs
```diff
@@ -15,19 +16,42 @@ internal sealed class ConfigShowCommandHandler
...
             jsonWriter.Write(configuration, full: level == ConfigDisplayLevel.Full, skipNotPresent: false);
+
+            if (showSources)
+            {
+                WriteConfigurationProviders(stream, configuration);
+            }
+        }
+
+        private static void WriteConfigurationProviders(Stream stream, IConfiguration configuration)
+        {
```
kkeirstead, line 40: I wasn't sure what an appropriate place would be for this method, so I left it in `ConfigShowCommandHandler` for now. I considered relocating it to be with the rest of the configuration writer, but that's specifically for writing JSON (and this is just text). Let me know if we want this to be somewhere else.

src/Tools/dotnet-monitor/Commands/ConfigShowCommandHandler.cs
```diff
@@ -15,19 +16,42 @@ internal sealed class ConfigShowCommandHandler
...
+            var configurationProviders = ((IConfigurationRoot)configuration).Providers.Reverse();
+
+            StreamWriter writer = new StreamWriter(stream);
```
wiktork, line 43: using StreamWriter =
You will probably need the overload that has a leaveOpen:true option so that you don't dispose the underlying console stream.

src/Tools/dotnet-monitor/Commands/ConfigShowCommandHandler.cs
wiktork: I'm ok with this although I think it's a bit odd to mix json and non-json results as command output. It may be cumbersome to create tests around that for example.

src/Tools/dotnet-monitor/Commands/ConfigShowCommandHandler.cs
kkeirstead: Ah; I had tried adding `using` but then it closed the stream - I didn't know about that overload.

src/Tools/dotnet-monitor/Commands/ConfigShowCommandHandler.cs
```diff
@@ -15,19 +16,42 @@ internal sealed class ConfigShowCommandHandler
...
+
+            writer.WriteLine("Configuration Providers (High to Low Priority):");
+
+            foreach (var provider in configurationProviders)
```
wiktork, line 47: While I think this is going to be very useful, how difficult would it be to do it on a per value basis? Consider something like IConfigurationRoot.GetDebugView() https://github.com/dotnet/runtime/blob/50e6f078bcf4aa5bca1e7e110b9a6cb846390b82/src/libraries/Microsoft.Extensions.Configuration.Abstractions/src/ConfigurationRootExtensions.cs#L36

jander-msft: I agree with this sentiment. I'm hoping we can do something more like [GetDebugView](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.configurationrootextensions.getdebugview?view=dotnet-plat-ext-3.1). Here's example output of calling this: https://stackoverflow.com/questions/37688027/how-can-i-list-all-of-the-configuration-sources-or-properties-in-asp-net-core

I'm thinking that we could add the source as a comment above the JSON element using [Utf8JsonWriter.WriteCommentValue](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonwriter.writecommentvalue?view=net-6.0) so that it is still valid JSON.

We can also do better than what GetDebugView writes out by writing useful information for each provider type:
- EnvironmentVariablesConfigurationProvider would write out the prefix
- JsonConfigurationProvider would write out the file path or name
- KeyPerFileConfigurationProvider would write out the directory

etc

wiktork: +1 on the Json comments. This also removes the need for an additional non-json text block at the end of the output.

jander-msft: Maybe just start with writing out the provider type and then do a follow up PR to write more specific information.

jander-msft: To further clarify, I would like to see a comment added to every value as to where that value came from in configuration, not just running through the sources and printing them at the top (because everyone is going to have the same list), e.g.

```
{
  // Source: Environment variable with prefix 'DOTNET_'
  "urls": "http://localhost:8080"
  // Source: JSON file 'settings.json'
  "DiagnosticPort": "/diag/port"
}
```

kkeirstead: Ah, that makes sense - I'll work on that

kkeirstead: So, I've gotten a version of this working...my only concern is that readability isn't great. Do people have thoughts about having the sources listed at the top/bottom, but numbered - so every value just has a comment with the number pointing to the source, instead of writing the whole thing out every time? Or something else along those lines?

jander-msft: Can you paste in a sample of the output as you have it?

kkeirstead: ListConfigurationSources_Raw
ListConfigurationSources_References

I included screenshots of the two approaches to maintain formatting; this isn't the whole output, but should give you an idea. Currently, the source is appearing after the key/value instead of before it just due to the way we write the JSON; this can probably be worked around, but I expect that would be a bit more intrusive than the current solution.

kkeirstead: Also, ignore the bug in the first image under "Subject" (I fixed it when I was making the second version); forgot a break statement after finding the provider.

kkeirstead: Pinging - wanted to make sure people didn't forget about this thread

wiktork: I'm ok with the first one. I don't think the verbosity is really an issue, since you're asking for the information and can use tooling to help analyze the file.

jander-msft: Also, please put the comment above the property/value instead under it.

kkeirstead:
In doing some investigating, I discovered that I can't place comments before Key/Value pairs; the writer only places a comma right before the next key, resulting in the configuration not printing correctly. See https://github.com/dotnet/runtime/issues/52408#:~:text=Currently%2C%20you%20are%20unable%20to%20write%20comments%20after,after%20commas%20that%20are%20delimiting%20properties%20in%20Json under Detailed Description for context. In the meantime, the current version of the code is putting the provider directly **after** the key/value pair. For now, I'm holding off on adding any test code until we decide if we want to make any changes to this format, but I am planning on adding tests later for this feature.

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -21,14 +22,16 @@ namespace Microsoft.Diagnostics.Tools.Monitor
...
         private readonly Utf8JsonWriter _writer;
+        private IConfiguration _configuration;
         public ConfigurationJsonWriter(Stream outputStream)
         {
             JsonWriterOptions options = new() { Indented = true };
             _writer = new Utf8JsonWriter(outputStream, options);
         }
 
-        public void Write(IConfiguration configuration, bool full, bool skipNotPresent)
+        public void Write(IConfiguration configuration, bool full, bool skipNotPresent, bool showSources=false)
```
wiktork, line 32: incredible nit:
```suggestion
        public void Write(IConfiguration configuration, bool full, bool skipNotPresent, bool showSources = false)
```

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -217,25 +220,64 @@ private void ProcessSection(IConfigurationSection section, bool includeChildSect
...
+
+                StringBuilder comment = new StringBuilder();
+
+                foreach (var provider in configurationProviders)
+                {
+                    provider.TryGet(section.Path, out string value);
+
+                    if (!string.IsNullOrEmpty(value))
+                    {
+                        comment.Append(provider.ToString());
```
wiktork, line 272: Since we are not actually merging multiple providers here, can we just use a string instead of StringBuilder?

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -217,25 +220,64 @@ private void ProcessSection(IConfigurationSection section, bool includeChildSect
...
+                        break;
+                    }
+                }
+
+                return comment.ToString();
             }
+
+            return "";
```
wiktork, line 280: string.Empty. Alternatively consider a bool TryGetConfigurationProvider(..., out string comment) instead.

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -217,25 +220,64 @@ private void ProcessSection(IConfigurationSection section, bool includeChildSect
...
-                            ProcessChildren(child, includeChildSections, redact);
+                            ProcessChildren(child, includeChildSections, redact,  showSources: showSources);
                         }
                         else
                         {
                             WriteValue(child.Value, redact);
+
+                            string comment = GetConfigurationProvider(child, showSources);
+
+                            if (comment.Length > 0)
```
wiktork, line 231: Can you put a TODO regarding the comment ordering, explaining why it's after the value instead of before?

wiktork (approved):

PR #1462, opened 2022-02-06
Add common release scripts

jander-msft:
This change adds the Get BAR ID and Get Release Version inline scripts from the release definition into the repo with several improvements:
- Caller can decide if it wants verbose information (e.g. release data and versions) by using the -Verbose flag.
- Caller can decide whether to push final output value into a task variable.
- Use standard error stream for reporting errors and failing the script immediately.
- Allows calling script via file path so that script parameters are printed into task output.
- Allows reuse among multiple build tasks.

cc @kelltrick, @hoyosjs

### DIFF
--- a/eng/release/Scripts/GetBarId.ps1
+++ b/eng/release/Scripts/GetBarId.ps1
@@ -15,7 +15,7 @@ $tagsUri = "${env:SYSTEM_TEAMFOUNDATIONCOLLECTIONURI}${env:SYSTEM_TEAMPROJECT}/_
 $buildData = Invoke-RestMethod `
     -Uri $tagsUri `
     -Method 'GET' `
-    -Headers @{ 'accept' = 'application/json'; 'Authorization' = 'Bearer ' + $env:System_AccessToken }
+    -Headers @{ 'accept' = 'application/json'; 'Authorization' = "Bearer ${env:System_AccessToken}" }
 
 Write-Verbose 'BuildData:'
 $buildDataJson = $buildData | ConvertTo-Json
--- a/eng/release/Scripts/GetBarId.ps1
+++ b/eng/release/Scripts/GetBarId.ps1
@@ -1,0 +1,46 @@
+[CmdletBinding()]
+Param(
+    [Parameter(Mandatory=$true)][string] $BuildId,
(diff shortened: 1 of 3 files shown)


eng/release/Scripts/GetReleaseVersion.ps1
```diff
@@ -0,0 +1,39 @@
...
+    [Parameter(Mandatory=$false)][string] $MaestroApiEndPoint = 'https://maestro-prod.westus2.cloudapp.azure.com',
+    [Parameter(Mandatory=$false)][string] $TaskVariableName = $null,
+    [Parameter(Mandatory=$false)][switch] $IncludeV
+)
+
+$ErrorActionPreference = 'Stop'
+Set-StrictMode -Version 2.0
+
+$releaseData = Invoke-RestMethod `
+    -Uri "$MaestroApiEndPoint/api/assets?buildId=$BarId&name=dotnet-monitor&api-version=2020-02-20" `
```
hoyosjs, line 14: nit: parametrize the API version. The path for this one doesn't change much.

eng/release/Scripts/GetBarId.ps1
```diff
@@ -0,0 +1,46 @@
+[CmdletBinding()]
```
hoyosjs, line 1: For my learning: Are we using BAR ID for anything other that correlation purposes anymore in dotnet-monitor?

hoyosjs (approved):

eng/release/Scripts/GetBarId.ps1
jander-msft: We use it for calling `darc gather-drop` in the AcquireBuild script and for getting the release version number. In the near future, I'll use it to get the build version number as well.

eng/release/Scripts/GetBarId.ps1
```diff
@@ -43,4 +43,4 @@ if ($TaskVariableName) {
         -Value $barId
 }
 
-Write-Ouptut $barId
\ No newline at end of file
+Write-Output $barId
```
hoyosjs, line 46: GarfieldAnnoyedGIF

That's a good catch

Issue #445, opened 2021-06-15
Azure egress should support more restrictive SAS tokens


wiktork:
Currently if an egress container is pre-created and a container level sas token (such as one created with Storage Explorer) is used, dotnet-monitor fails because it does not have enough permissions to issue the CreateContainerIfExists call. This is not an issue with account level SAS tokens (such as those created by the Storage Portal blade). We should consider supporting highly restricted tokens.

Also consider other auth options, such as MSI.

DrewRomanyk:
@wiktork It would be great to rely on AAD auth in order to access rather than having a secret to have to manage especially with AKS & pod identity.

From my brief and naive look it seems like [this function would just need another egress option to specify to use DefaultAzureCredential](https://github.com/dotnet/dotnet-monitor/blob/main/src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs#L242) rather than using a SAS token or an Account Key credentials, [similar to how this guide describes](https://docs.microsoft.com/en-us/azure/storage/blobs/authorize-managed-identity). If that is true, I'm happy to try to spin up a PR that has that change.

PR #1464, opened 2022-02-07
Push a Message to a Queue When Writing to Azure Egress

### LINKED ISSUE #163: Blob Egress should also have a queue option
When egressing to blob storage, we should also allow egress to a queue for better integration with Azure based workflows such as web jobs or azure functions. Detection of blobs alone is very slow.

kkeirstead:
Optionally allows users to push a message to a queue when writing a blob to Azure egress. I noticed this was an older feature suggestion, but it didn't look like anyone else had taken it.

Closes #163


eng/Versions.props
```diff
@@ -67,6 +67,7 @@
   </PropertyGroup>
   <PropertyGroup Label="Manual">
     <AzureStorageBlobsVersion>12.6.0</AzureStorageBlobsVersion>
+    <AzureStorageQueuesVersion>12.8.0</AzureStorageQueuesVersion>
```
kkeirstead, line 70: I believe this is the latest stable version, but I don't know if we have a process for how we choose which version we want to use.

wiktork: For these particular apis, latest stable should work.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -129,6 +142,74 @@ private Uri GetAccountUri(AzureBlobEgressProviderOptions options, out string acc
...
+
+        public async Task EgressMessageToQueue(string artifactName, AzureBlobEgressProviderOptions options, CancellationToken token)
+        {
+            try
+            {
+                QueueClient queueClient = await GetQueueClientAsync(options, token);
+
+                if (queueClient.Exists())
+                {
+                    queueClient.SendMessage(string.Format(CultureInfo.CurrentCulture, Strings.Message_AzureQueueEgress, artifactName));
```
wiktork, line 165: Note that this should be a machine parsable message. Consider a small json message with a reference to the url of the uploaded blob.

wiktork: e.g.
``` json
{
    "artifact": "name",
    "uri": "bloburi"
}

File: src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs [hunk 3]
```diff
@@ -129,6 +142,74 @@ private Uri GetAccountUri(AzureBlobEgressProviderOptions options, out string acc
             return blobUriBuilder.ToUri();
         }
 
+        private Uri GetQueueAccountUri(AzureBlobEgressProviderOptions options, out string accountName)
+        {
+            var queueUriBuilder = new QueueUriBuilder(options.QueueAccountUri);
```
kelltrick, line 147: This can be null here. We probably need to check both `QueueName` and `QueueAccountUri` before entering this. If one is null with the other being set we should probably print a warning and not send anything to a queue.

kelltrick (approved):
My 2 strong recommendations for this is:

1.  Check both options parameters (`QueueName ` and `QueueAccountUri`) before proceeding to sending a message to the queue.
2. Wiktors comment on sending a json blob instead of a message string.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -117,7 +132,20 @@ public AzureBlobEgressProvider(ILogger<AzureBlobEgressProvider> logger)
             }
         }
 
-        private Uri GetAccountUri(AzureBlobEgressProviderOptions options, out string accountName)
+        private bool CheckQueueEgressOptions(AzureBlobEgressProviderOptions options)
+        {
+            bool queueNameSet = !string.IsNullOrEmpty(options.QueueName);
+            bool queueAccountUriSet = null != options.QueueAccountUri;
+
+            if ((queueNameSet && !queueAccountUriSet) || (!queueNameSet && queueAccountUriSet))
```
jander-msft, line 140:
```suggestion
            if (queueNameSet ^ queueAccountUriSet)
```

XOR - one of them is true but not both

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -129,6 +157,91 @@ private Uri GetAccountUri(AzureBlobEgressProviderOptions options, out string acc
...
+
+            queueUriBuilder.Query = null;
+            queueUriBuilder.QueueName = null;
+
+            accountName = queueUriBuilder.AccountName;
+
+            return queueUriBuilder.ToUri();
+        }
+
+        public async Task EgressMessageToQueue(string artifactName, AzureBlobEgressProviderOptions options, CancellationToken token)
```
jander-msft, line 172:
```suggestion
        private async Task EgressMessageToQueue(string artifactName, AzureBlobEgressProviderOptions options, CancellationToken token)
```

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -129,6 +157,91 @@ private Uri GetAccountUri(AzureBlobEgressProviderOptions options, out string acc
...
+        {
+            try
+            {
+                QueueClient queueClient = await GetQueueClientAsync(options, token);
+
+                if (queueClient.Exists())
```
jander-msft, line 178: This code path silently fails to write to the queue because the queue doesn't exist. Because we can't upload the blob storage and push to the queue as an atomic operation and there's potential benefit to allowing the egress operation to complete even if it can't write to the queue, I think this is okay.

However, if the queue doesn't exist, we should log a warning to that effect so that the user or operator know why the blob appeared but the message was not queued. In that same vein, we shouldn't fail the whole operation if there is a problem in writing to the queue either; we should handle the exception and log it appropriately, probably as a warning.

To sum up, either all types of failures should fail the operation or all types of failures should be handled and logged; I prefer the latter.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -129,6 +157,91 @@ private Uri GetAccountUri(AzureBlobEgressProviderOptions options, out string acc
...
+                serviceClient = new QueueServiceClient(accountUri, credential);
+            }
+            else
+            {
+                throw CreateException(Strings.ErrorMessage_EgressMissingSasOrKey);
+            }
+
+            QueueClient queueClient = serviceClient.GetQueueClient(options.QueueName);
+
+            await queueClient.CreateIfNotExistsAsync(cancellationToken: token);
```
jander-msft, line 240: The SAS token may allow for putting items into the queue but not allow for creation of the queue itself. Please explore if there is a way to check that the queue exists with a limited SAS token and then fallback to trying to call CreateIfNotExistsAsync. We have the same issue with the blob storage portion already but let's not pile more onto it: #445

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -129,6 +157,91 @@ private Uri GetAccountUri(AzureBlobEgressProviderOptions options, out string acc
...
+            {
+                throw CreateException(innerException);
+            }
+            catch (RequestFailedException ex)
+            {
+                throw CreateException(ex);
+            }
+        }
+
+        private string ConstructQueueMessage(string artifactName, AzureBlobEgressProviderOptions options)
```
jander-msft, line 194: The format of the message needs to be documented.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -129,6 +157,91 @@ private Uri GetAccountUri(AzureBlobEgressProviderOptions options, out string acc
...
+        {
+            using var stream = new MemoryStream();
+            using var writer = new Utf8JsonWriter(stream);
+
+            writer.WriteStartObject();
+
+            writer.WriteString("artifact", artifactName);
+            writer.WriteString("uri", options.AccountUri.ToString());
```
jander-msft, line 202: May need to strip out the query portion before writing the account uri.

jander-msft (changes requested):

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
kkeirstead: Off the top of your head, any reason why we couldn't just use `queueClient.Exists()` here? Looking at documentation, it says in the Remarks section for `CreateIfNotExistsAsync`, "This API requires Create or Write permissions," but there's no disclaimer for `Exists()`. I figure we could do an `Exists` check, and then if it doesn't exist we fall back to `CreateIfNotExistsAsync`?

jander-msft: That would probably work, but use `ExistsAsync` (I think that's a method). Add a comment as to why we do "if !ExistsAsync() { CreateIfNotExistsAsync() }" because it won't be obvious why we shouldn't just call `CreateIfNotExistsAsync`.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -129,6 +157,100 @@ private Uri GetAccountUri(AzureBlobEgressProviderOptions options, out string acc
...
+                    queueClient.SendMessage(queueMessage);
+                }
+                else
+                {
+                    Logger.LogWarning(string.Format(CultureInfo.CurrentCulture, Strings.Message_QueueDoesNotExist, options.QueueName));
+                }
+            }
+            catch (AggregateException ex) when (ex.InnerException is RequestFailedException innerException)
+            {
+                Logger.LogWarning(string.Format(CultureInfo.CurrentCulture, Strings.Message_WritingMessageToQueueFailed, options.QueueName));
```
wiktork, line 190: Consider propagating the actual message into the log as well. You can use AggregateException.GetBaseException to get to the root failure.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -129,6 +157,100 @@ private Uri GetAccountUri(AzureBlobEgressProviderOptions options, out string acc
...
+
+        private string ConstructQueueMessage(string artifactName, AzureBlobEgressProviderOptions options)
+        {
+            using var stream = new MemoryStream();
+            using var writer = new Utf8JsonWriter(stream);
+
+            writer.WriteStartObject();
+
+            writer.WriteString("artifact", artifactName);
+            writer.WriteString("uri", options.AccountUri.GetLeftPart(UriPartial.Path));
```
wiktork, line 206: The Uri should be the full path to the artifact in blob storage.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
wiktork: Thinking about this some more, we should consider enabling something like https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-storage-blob-input?tabs=csharp#example. In this particular case, the payload wouldn't be json, but just the name of the file. 

This would greatly limit the amount of information about the blob that we would be sending via the queue, but it's likely that the person doing the egress would have a good understanding of the relationship between the queue and the blob already.

kkeirstead: I switched to just using the artifact name; if we change our minds I can always revert back to the previous format (or something else entirely).

jander-msft: I think is is _okay_... I guess we could always add another option to have it write out a JSON object with more detailed information.

FYI, The [QueueTrigger] example that @wiktork linked to is not limited to just string values. It can, for the most part, use any deserializable type such as a POCO (which is implicitly understood to be a JSON object).

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -129,6 +157,83 @@ private Uri GetAccountUri(AzureBlobEgressProviderOptions options, out string acc
...
+
+        private async Task EgressMessageToQueue(string artifactName, AzureBlobEgressProviderOptions options, CancellationToken token)
+        {
+            try
+            {
+                QueueClient queueClient = await GetQueueClientAsync(options, token);
+
+                if (queueClient.Exists())
+                {
+                    queueClient.SendMessage(artifactName);
```
jander-msft, line 180: This should write the full URI to the blob item, which is already available in the `blobUriString` variable from the `EgressAsync` method before calling `EgressMessageToQueue`.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -52,6 +55,12 @@ public AzureBlobEgressProvider(ILogger<AzureBlobEgressProvider> logger)
 
                 string blobUriString = GetBlobUri(blobClient);
                 Logger?.EgressProviderSavedStream(EgressProviderTypes.AzureBlobStorage, blobUriString);
+
+                if (CheckQueueEgressOptions(options))
+                {
+                    await EgressMessageToQueue(artifactSettings.Name, options, token);
```
jander-msft, line 61: It cannot just be the artifact name that is passed in. It most likely needs to be the reuslt of the `GetBlobName` call from above because the name of the blob can contain "folders".

For example, for the URI `https://dotnetcli.azureedge.net/dotnet/diagnostics/monitor/6.1.0/dotnet-monitor.6.1.0.nupkg`:
- The account is `dotnetcli`
- The container is `dotnet`
- The blob is `diagnostics/monitor/6.1.0/dotnet-monitor.6.1.0.nupkg`

Setting just `dotnet-monitor.6.1.0.nupkg` as the queue item would not be sufficient for the association within an Azure function to automatically create a stream from that since it is missing `diagnostics/monitor/6.1.0`.

kkeirstead: I changed this in the latest version; in my example I actually was using a blob that was more than just the artifact name (artifacts/artifact_name), but I had just changed the Azure Function so that it knew to look in the artifacts directory so this would work. Removing the need for Azure Function users to do that step should make it easier, however.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -129,6 +161,88 @@ private Uri GetAccountUri(AzureBlobEgressProviderOptions options, out string acc
...
+                if (queueClient.Exists())
+                {
+                    queueClient.SendMessage(blobName);
+                }
+                else
+                {
+                    Logger.LogWarning(string.Format(CultureInfo.CurrentCulture, Strings.Message_QueueDoesNotExist, options.QueueName));
+                }
+            }
+            catch (AggregateException ex) when (ex.InnerException is RequestFailedException innerException)
```
jander-msft, line 191: Why does it matter that the inner exception is a `RequestFailedException`?

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -129,6 +161,88 @@ private Uri GetAccountUri(AzureBlobEgressProviderOptions options, out string acc
...
+            {
+                Logger.LogWarning(string.Format(CultureInfo.CurrentCulture, Strings.Message_WritingMessageToQueueFailed, options.QueueName));
```
jander-msft, line 193: Move to LoggingExtensions

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -129,6 +161,88 @@ private Uri GetAccountUri(AzureBlobEgressProviderOptions options, out string acc
...
+            }
+            catch (RequestFailedException)
+            {
+                Logger.LogWarning(string.Format(CultureInfo.CurrentCulture, Strings.Message_WritingMessageToQueueFailed, options.QueueName));
```
jander-msft, line 197: Move to LoggingExtensions

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
jander-msft, line 193: Please pass the exception into the Log* call as well so that the exception itself will be logged.

eng/Versions.props
jander-msft: Please choose an Azure.Storage.Queues version that matches the dependencies of Azure.Storage.Blobs. There have been binary breaking changes in the past for these packages and I'd rather avoid missing methods and type load failures.

jander-msft (changes requested):

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
kkeirstead: I actually carried the exception handling logic over from the `EgressAsync` methods (i.e. Line 129), since I figured it was best to be consistent with what was already in place for egressing to blob storage. That being said, would you prefer something different?

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -129,6 +161,83 @@ private Uri GetAccountUri(AzureBlobEgressProviderOptions options, out string acc
...
+
+        private async Task EgressMessageToQueue(string blobName, AzureBlobEgressProviderOptions options, CancellationToken token)
+        {
+            try
+            {
+                QueueClient queueClient = await GetQueueClientAsync(options, token);
+
+                if (queueClient.Exists())
+                {
+                    queueClient.SendMessage(Convert.ToBase64String(Encoding.UTF8.GetBytes(blobName)));
```
kkeirstead, line 184: Previously, I was using `QueueClientOptions` to handle the Base64 conversion; however, after switching to an older version of Azure Queues, the message encoding option doesn't appear to be available. This should have the same effect.

jander-msft: Why does it need to be base64 encoded before calling SendMessage? Shouldn't we just be able to give it whatever is our raw data and it will encode in whatever format their protocol requires? This seems like the consumer would have to understand that the data needs to be base64 decoded, which I would not expect.

kkeirstead: Originally, I wasn't doing the encoding, and noticed that all of my messages were going into a queue named `dotnet-monitor-testing-queue-poison` - after looking around online and trying that, it resolved the issue. The error I was getting was "The input is not a valid Base-64 string as it contains a non-base 64 character, more than two padding characters, or an illegal character among the padding characters azure function queue message."

jander-msft: How about: `await queueClient.SendMessageAsync(BinaryData.FromString(blobName), cancellationToken: token);`?

eng/Versions.props
```diff
@@ -67,6 +67,7 @@
   </PropertyGroup>
   <PropertyGroup Label="Manual">
     <AzureStorageBlobsVersion>12.6.0</AzureStorageBlobsVersion>
+    <AzureStorageQueuesVersion>12.4.2</AzureStorageQueuesVersion>
```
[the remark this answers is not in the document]
jander-msft: Try Blobs @ 12.10.0 and Queues @ 12.8.0; they have the same version ranges for overlapping dependencies and both happen to be the latest stable version.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
jander-msft: Mostly redundant, but do please use the async method.

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -590,6 +590,16 @@
 1 Format Parameter:
 1. failure: The failure message from validation</comment>
   </data>
+  <data name="LogFormatString_QueueDoesNotExist" xml:space="preserve">
+    <value>The queue {0} does not exist; ensure that the queueName and queueAccountUri fields are set correctly.</value>
```
jander-msft, line 594: nit: We might want to put in holes for `queueName` and `queueAccoutUri` and then fill them in programmatically with constants (e.g. nameof(...)) so that localization doesn't accidentally localize these.

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -590,6 +590,16 @@
...
+    <comment>Gets the format string that is printed in the 57:QueueDoesNotExist event.
+1 Format Parameter:
+1. queueName: The name of the Azure Queue where messages are egressed</comment>
+  </data>
+  <data name="LogFormatString_QueueOptionsPartiallySet" xml:space="preserve">
+    <value>Queue message egress requires queueName and queueAccountUri to be set</value>
```
jander-msft, line 600: nit: Same about adding holes for the property names.

jander-msft (approved):

src/Tools/dotnet-monitor/LoggingExtensions.cs
```diff
@@ -567,5 +585,20 @@ private static string GetFwLinkWithCurrentLcidUri(long fwlinkId)
                 fwlinkId,
                 CultureInfo.CurrentUICulture.LCID);
         }
+
+        public static void QueueDoesNotExist(this ILogger logger, string queueName, string queueNameField, string queueAccountUriField)
+        {
+            _queueDoesNotExist(logger, queueName, queueNameField, queueAccountUriField, null);
```
jander-msft, line 591: Rather than exposing these as parameters, just use nameof(...) here since these should not be configurable.

PR #1477, opened 2022-02-08
Release tool updates

### LINKED ISSUE #1386: Add --dry-run switch to release tool
This would allow the internal builds to run the release tool without publishing a release manifest in order to validate that the tool can successfully process build information on a regular basis.

jander-msft:
- Change NugetLayoutWorker to only consider files under the packages directory
- Add checksum file to ChecksumAssets in publish manifest
- Add unprocessed files to BlobAssets in publish manifest
- Add --dry-run option to run tool without publishing

The changes will make it easier to publish blob assets to the desired target storage account, checksums to a different storage account, and more easily have NuGet publishing only consider NuGet packages (fixes issue where the nupkg was listed twice under NuGetAssets).

Example dry run publish manifest:
```json
{
  "ReleaseVersion": "20220127.7",
  "RepoUrl": "https://github.com/dotnet/dotnet-monitor",
  "Branch": "refs/heads/release/6.0",
  "Commit": "9774a8ccc1a015d47960ed76de1bc6a5a1227f2c",
  "DateProduced": "2022-01-28T05:44:47+00:00",
  "BuildNumber": "20220127.7",
  "BarBuildId": 124762,
  "BlobAssets": [
    {
      "PublishRelativePath": "BlobAssets/dotnet-monitor.6.0.2.nupkg",
      "PublishedPath": "BlobAssets/dotnet-monitor.6.0.2.nupkg",
      "Sha512": "3203E455FE4FA9E7B251BA957DDF286A3A3FCF1C88663315547D5F4F7DF9EE1C1F164DFCC42EDEEBAF786BF1B893753C5A5339377B259798208D5F07C48F9A10"
    }
  ],
  "NugetAssets": [
    {
      "PublishRelativePath": "NugetAssets/dotnet-monitor.6.0.2.nupkg",
      "PublishedPath": "NugetAssets/dotnet-monitor.6.0.2.nupkg",
      "Sha512": "3203E455FE4FA9E7B251BA957DDF286A3A3FCF1C88663315547D5F4F7DF9EE1C1F164DFCC42EDEEBAF786BF1B893753C5A5339377B259798208D5F07C48F9A10"
    }
  ],
  "SymbolNugetAssets": [
    {
      "PublishRelativePath": "SymbolNugetAssets/dotnet-monitor.6.0.2.symbols.nupkg",
      "PublishedPath": "SymbolNugetAssets/dotnet-monitor.6.0.2.symbols.nupkg",
      "Sha512": "3203E455FE4FA9E7B251BA957DDF286A3A3FCF1C88663315547D5F4F7DF9EE1C1F164DFCC42EDEEBAF786BF1B893753C5A5339377B259798208D5F07C48F9A10"
    }
  ],
  "ChecksumAssets": [
    {
      "PublishRelativePath": "ChecksumAssets/dotnet-monitor.6.0.2.nupkg.sha512",
      "PublishedPath": "ChecksumAssets/dotnet-monitor.6.0.2.nupkg.sha512",
      "Sha512": "9D6045C11F71F380D97D8A3A4CC0B6C07FD866DB21DD19F422F973DB6F00C5F4FC037316443C727DD7E810303B9079EAD50289EE703A50782C306E5059A1F387"
    }
  ],
  "UnknownAssets": []
}
```

closes #1386

### DIFF
--- a/eng/release/DiagnosticsReleaseTool/Common/BlobLayoutWorker.cs
+++ b/eng/release/DiagnosticsReleaseTool/Common/BlobLayoutWorker.cs
@@ -5,7 +5,7 @@ namespace ReleaseTool.Core
     public sealed class BlobLayoutWorker : PassThroughLayoutWorker
     {
         public BlobLayoutWorker(string stagingPath) : base(
-            shouldHandleFileFunc: static file => true,
+            shouldHandleFileFunc: static _ => true,
             getRelativePublishPathFromFileFunc: static file => Helpers.GetDefaultPathForFileCategory(file, FileClass.Blob),
             getMetadataForFileFunc: static file => Helpers.GetDefaultFileMetadata(file, FileClass.Blob),
             stagingPath
--- a/eng/release/DiagnosticsReleaseTool/Common/BlobLayoutWorker.cs
+++ b/eng/release/DiagnosticsReleaseTool/Common/BlobLayoutWorker.cs
@@ -1,0 +1,14 @@
+using System.IO;
+
+namespace ReleaseTool.Core
+{
+    public sealed class BlobLayoutWorker : PassThroughLayoutWorker
+    {
+        public BlobLayoutWorker(string stagingPath) : base(
+            shouldHandleFileFunc: static file => true,
+            getRelativePublishPathFromFileFunc: static file => Helpers.GetDefaultPathForFileCategory(file, FileClass.Blob),
+            getMetadataForFileFunc: static file => Helpers.GetDefaultFileMetadata(file, FileClass.Blob),
+            stagingPath
+        ){}
+    }
+}
--- a/eng/release/DiagnosticsReleaseTool/Common/NugetLayoutWorker.cs
+++ b/eng/release/DiagnosticsReleaseTool/Common/NugetLayoutWorker.cs
@@ -1,11 +1,12 @@
+using System;
 using System.IO;
 
 namespace ReleaseTool.Core
 {
     public sealed class NugetLayoutWorker : PassThroughLayoutWorker
     {
-        public NugetLayoutWorker(string stagingPath) : base(
-            shouldHandleFileFunc: static file => file.Extension == ".nupkg" && file.Directory.Name == "packages",
+        public NugetLayoutWorker(string stagingPath, Func<FileInfo, bool> additionalFileConstraint = null) : base(
+            shouldHandleFileFunc: file => file.Extension == ".nupkg" && (null == additionalFileConstraint || additionalFileConstraint(file)),
             getRelativePublishPathFromFileFunc: static file => Helpers.GetDefaultPathForFileCategory(file, FileClass.Nuget),
             getMetadataForFileFunc: static file => Helpers.GetDefaultFileMetadata(file, FileClass.Nuget),
             stagingPath
--- a/eng/release/DiagnosticsReleaseTool/Common/NugetLayoutWorker.cs
+++ b/eng/release/DiagnosticsReleaseTool/Common/NugetLayoutWorker.cs
@@ -5,7 +5,7 @@ namespace ReleaseTool.Core
     public sealed class NugetLayoutWorker : PassThroughLayoutWorker
     {
         public NugetLayoutWorker(string stagingPath) : base(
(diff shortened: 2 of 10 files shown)


eng/release/DiagnosticsReleaseTool/Common/BlobLayoutWorker.cs
```diff
@@ -0,0 +1,14 @@
+using System.IO;
+
+namespace ReleaseTool.Core
+{
+    public sealed class BlobLayoutWorker : PassThroughLayoutWorker
+    {
+        public BlobLayoutWorker(string stagingPath) : base(
+            shouldHandleFileFunc: static file => true,
```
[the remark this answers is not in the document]
hoyosjs: Does this mean that this always handles all files? I see that you relaxed the assumption that only one handler should handle things - instead you create cascading ones. I am OK with that, but is there a way to assert this is the last one?

eng/release/DiagnosticsReleaseTool/Common/NugetLayoutWorker.cs
```diff
@@ -5,7 +5,7 @@ namespace ReleaseTool.Core
     public sealed class NugetLayoutWorker : PassThroughLayoutWorker
     {
         public NugetLayoutWorker(string stagingPath) : base(
-            shouldHandleFileFunc: static file => file.Extension == ".nupkg" && !file.Name.EndsWith(".symbols.nupkg"),
+            shouldHandleFileFunc: static file => file.Extension == ".nupkg" && file.Directory.Name == "packages",
```
hoyosjs, line 8: Two things:
- Can you make the packages folder a constant in `DarcHelpers.cs`? I try to keep all the things where we depend on darc's layout there just for when they change it.
- Also, given that this is the common shared one (I don't have an overly strong stance here) - I think it's better to have the constructor in the runner pass down the override. This is tying it closely to darc, which changes things around in layouts. Something like 
```c#
new NugetLayoutWorker(
  stagingPath: releaseConfig.StagingDirectory.FullName,
  shouldHandleFileFunc: static file => file.Extension == ".nupkg" && file.Directory.Name == DarcHelpers.PackageDirName
),
```

hoyosjs (approved):

eng/release/DiagnosticsReleaseTool/Common/BlobLayoutWorker.cs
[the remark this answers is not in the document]
jander-msft: > Does this mean that this always handles all files?

Yes, because there isn't a good way to say what is a blob file and what isn't since it's just given a file path to figure everything out.

> I see that you relaxed the assumption that only one handler should handle things - instead you create cascading ones.

That assumption was not relaxed; it still remains. What I did change was allowing additional handlers to attempt to handle an already handled file, which was odd because then there was a validation check to make sure that a file wasn't handled more than once. So rather than allowing the bad behavior to occur and then logging an error when the bad behavior happens, I changed it so only one can handle it and the remaining handlers are ignored.

> I am OK with that, but is there a way to assert this is the last one?

Not sure, maybe remove it from the list and then explicitly add it as the fallback in the `Release` ctor.

eng/release/DiagnosticsReleaseTool/Core/Release.cs
```diff
@@ -224,14 +224,6 @@ private async Task<int> LayoutFilesAsync(CancellationToken ct)
...
                     {
-                        if (isProcessed)
-                        {
-                            // TODO: Might be worth to relax this limitation. It just needs to turn the
-                            //      source -> fileData relationship to something like source -> List<FileData>).
-                            _logger.LogError("[{buildFilePath}, {worker}] File {file} is getting handled by several workers.", file.FullName, worker.GetType().FullName, file);
-                            return -1;
-                        }
-
                         isProcessed = true;
```
kkeirstead, line 227: Do we need `isProcessed` anymore? It doesn't appear that it's currently being evaluated anywhere

jander-msft: It's still used below to make sure that the file was actually processed and logs a warning if it was not.

kkeirstead: Ah, I see - another case of GitHub not expanding to show enough lines 👍

PR #1491, opened 2022-02-10
Scripts and pipeline for release to blob storage

jander-msft:
These changes add scripts and a pipeline for publishing the `dotnet monitor` files from the staging location to the expected production location. They have options for performing a dry-run (exercises all of the publish steps but not actually push the files to the destination accounts) and for testing with test accounts (exercises all of the publish steps but publishes to test accounts rather than `dotnetcli` and `dotnetclichecksums`). The test and dry-run modes are the default to prevent accidental promotion. When not running a test run, the promotion will use an environment that requires an explicit approval, whereas the test run uses an environment that has automatic approval.

Example runs:
- [6.1.0 candidate test (dry-run)](https://dev.azure.com/dnceng/internal/_build/results?buildId=1605372&view=results)
- [6.1.0 candidate test](https://dev.azure.com/dnceng/internal/_build/results?buildId=1605407&view=results)
- [7.0.0 P1 candidate test](https://dev.azure.com/dnceng/internal/_build/results?buildId=1605409&view=results)

Over time, I think the other existing release process (push to NuGet.org, create GitHub release) can be folded into this pipeline.

### DIFF
--- a/eng/release.yml
+++ b/eng/release.yml
@@ -1,0 +1,183 @@
+trigger: none
+pr: none
+
+parameters:
+- name: IsTest
+  type: boolean
+  default: true
+- name: IsDryRun
+  type: boolean
+  default: true
+
+variables:
+- name: _TeamName
+  value: DotNetCore
+- group: Release-Pipeline
+
+resources:
+  pipelines:
+  - pipeline: Build
+    source: dotnet-dotnet-monitor
+
+stages:
+- stage: Validation
+
+  pool:
+    name: NetCore1ESPool-Internal
+    demands: ImageOverride -equals Build.Windows.10.Amd64.VS2019
+
+  jobs:
+  - job: Validate
+
+    variables:
+    # Allow for differentiation of runs of this pipeline
+    # when running it with the same build repeatedly.
+    - name: RunRevision
+      value: $[counter(variables['resources.pipeline.Build.runID'],1)]
+
+    workspace:
+      clean: all
+
+    steps:
+    - download: none
+
+    - task: PowerShell@2
+      displayName: Get BAR ID
+      inputs:
+        filePath: $(Build.SourcesDirectory)/eng/release/Scripts/GetBarId.ps1
+        arguments: >-
+          -BuildId $(resources.pipeline.Build.runID)
+          -TaskVariableName 'BarId'
+      env:
+        SYSTEM_ACCESSTOKEN: $(System.AccessToken)
+
+    - task: PowerShell@2
+      displayName: Get Release Version
+      inputs:
+        filePath: $(Build.SourcesDirectory)/eng/release/Scripts/GetReleaseVersion.ps1
+        arguments: >-
+          -BarId $(BarId)
+          -MaestroToken $(MaestroAccessToken)
+          -TaskVariableName 'ReleaseVersion'
+    
+    - task: PowerShell@2
+      displayName: Get Build Version
+      inputs:
+        filePath: $(Build.SourcesDirectory)/eng/release/Scripts/GetBuildVersion.ps1
+        arguments: >-
+          -BarId $(BarId)
+          -MaestroToken $(MaestroAccessToken)
+          -TaskVariableName 'BuildVersion'
+
+    - powershell: Write-Host "##vso[build.updatebuildnumber]$(ReleaseVersion) ($(BuildVersion)) (R$(RunRevision))"
+      displayName: Set Name
+
+- stage: Publish
+  dependsOn:
+  - Validation
+
+  pool:
+    name: NetCore1ESPool-Internal
+    demands: ImageOverride -equals Build.Windows.10.Amd64.VS2019
+
+  jobs:
+  - deployment: PublishToStorageAccounts
+    displayName: Publish to Storage Accounts
+
+    ${{ if eq(parameters.IsTest, 'true') }}:
+      environment: Diagnostics-Monitor-Storage-Test
+    ${{ else }}:
+      environment: Diagnostics-Monitor-Storage-DotNetCli
+
+    variables:
+    - ${{ if eq(parameters.IsTest, 'true') }}:
+      - group: DotNet-Diagnostics-Storage-Test
+    - group: DotNet-DotNetStage-Storage
+
+    workspace:
+      clean: all
+
+    strategy:
+      runOnce:
+        deploy:
+          steps:
+          - checkout: self
+          - download: none
+
+          - task: PowerShell@2
+            displayName: Install AzCopy
+            inputs:
+              filePath: $(Build.SourcesDirectory)/eng/release/Scripts/InstallAzCopy.ps1
+              arguments: >-
+                -ToolsDirectory $(Agent.ToolsDirectory)
+                -TaskVariableName 'AzCopyPath'
+
+          - task: PowerShell@2
+            displayName: Get BAR ID
+            inputs:
+              filePath: $(Build.SourcesDirectory)/eng/release/Scripts/GetBarId.ps1
+              arguments: >-
+                -BuildId $(resources.pipeline.Build.runID)
+                -TaskVariableName 'BarId'
+            env:
+              SYSTEM_ACCESSTOKEN: $(System.AccessToken)
+
+          - task: PowerShell@2
+            displayName: Get Release Version
+            inputs:
+              filePath: $(Build.SourcesDirectory)/eng/release/Scripts/GetReleaseVersion.ps1
+              arguments: >-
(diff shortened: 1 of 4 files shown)


eng/release/Scripts/PublishToBlobAccounts.ps1
```diff
@@ -0,0 +1,121 @@
...
+    Write-Verbose "Copy $From -> $To"
+
+    if ($From -eq $to) {
+        Write-Verbose 'Skipping copy because source and destination are the same.'
+    } else {
+        [array]$azCopyArgs = "$from$fromToken"
+        $azCopyArgs += "$to$toToken"
+        $azCopyArgs += "--s2s-preserve-properties"
+        $azCopyArgs += "--s2s-preserve-access-tier=false"
+        if ($WhatIfPreference) {
```
kkeirstead, line 69: What's the context of `WhatIf` and `WhatIfPreference`? Mostly just confused about the naming convention, so figured I'd ask

jander-msft: WhatIf/$WhatIfPreference is the dry run concept but built into PowerShell.

If you specify the `-WhatIf` switch on a cmdlet that supports it (by adding `SupportsShouldProcess` to the `CmdletBinding` attribute), then the cmdlet should do all of the work it normally does EXCEPT for making a material impact. The cmdlet can inspect if `-WhatIf` was specified by checking the `$WhatIfPreference` variable which will be `$true` if it was specified and otherwise `$false`. 

You can read up on all of the details here: https://docs.microsoft.com/en-us/powershell/scripting/learn/deep-dives/everything-about-shouldprocess?view=powershell-7.2

eng/release.yml
```diff
@@ -0,0 +1,200 @@
+trigger: none
+pr: none
+
+parameters:
+- name: IsTestRun
```
kelltrick, line 5: What is `IsTestRun` for? I'm somewhat confused that there is a `IsDryRun` too.

kelltrick (approved):

eng/release/Scripts/InstallAzCopy.ps1
```diff
@@ -0,0 +1,48 @@
...
+    Rename-Item -Path $unpackDirPath -NewName 'azcopy'
+
+    # Delete zip
+    Remove-Item -Path $zipPath
+
+    Write-Verbose 'Finished'
+}
+
+if ($TaskVariableName) {
+    & $PSScriptRoot\SetTaskVariable.ps1 `
```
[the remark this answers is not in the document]
jander-msft: Yes

eng/release.yml
jander-msft: IsTestRun determines if the destination accounts are dotnetcli/dotnetclichecksums vs some test accounts that I set up. IsDryRun determines if the files are actually pushed to those accounts or not.

eng/release.yml
```diff
@@ -0,0 +1,200 @@
...
+
+    ${{ if eq(parameters.IsTestRun, 'true') }}:
+      environment: Diagnostics-Monitor-Storage-Test
+    ${{ else }}:
+      environment: Diagnostics-Monitor-Storage-DotNetCli
+
+    variables:
+    - ${{ if eq(parameters.IsTestRun, 'true') }}:
+      - group: DotNet-Diagnostics-Storage-Test
+    - group: DotNet-DotNetStage-Storage
```
wiktork, line 112: This is meant to be additive right?

eng/release/Scripts/GetBuildVersion.ps1
```diff
@@ -0,0 +1,38 @@
+[CmdletBinding()]
+Param(
+    [Parameter(Mandatory=$true)][string] $BarId,
+    [Parameter(Mandatory=$true)][string] $MaestroToken,
+    [Parameter(Mandatory=$false)][string] $MaestroApiEndPoint = 'https://maestro-prod.westus2.cloudapp.azure.com',
```
wiktork, line 5: Is this always stable or should we be getting it from an environment variable or variable group?

eng/release/Scripts/InstallAzCopy.ps1
```diff
@@ -0,0 +1,48 @@
...
+Param(
+    [Parameter(Mandatory=$true)][string] $ToolsDirectory,
+    [Parameter(Mandatory=$false)][string] $TaskVariableName = $null
+)
+
+$ErrorActionPreference = 'Stop'
+Set-StrictMode -Version 2.0
+
+$url = 'https://aka.ms/downloadazcopy-v10-windows'
+$basePath = Join-Path $ToolsDirectory 'azcopy'
```
wiktork, line 11: The old release definitions had a task to install azcopy. I'm a little surprised we have to download it ourselves.

eng/release/Scripts/InstallAzCopy.ps1
```diff
@@ -0,0 +1,48 @@
...
+
+if (Test-Path $azCopyPath) {
+    Write-Verbose 'Already installed'
+} else {
+    if (!(Test-Path $basePath)) {
+        New-Item -ItemType 'Directory' -Path $basePath | Out-Null
+    }
+
+    Write-Verbose 'Fetching...'
+    Invoke-WebRequest -Uri $url -OutFile $zipPath
```
wiktork, line 25: I'm not sure this has been fixed, but Invoke-WebRequest is known to be pretty slow. There may be alternatives that are a bit faster. Not sure how big the file is.

eng/release/Scripts/InstallAzCopy.ps1
```diff
@@ -0,0 +1,48 @@
...
+Set-StrictMode -Version 2.0
+
+$url = 'https://aka.ms/downloadazcopy-v10-windows'
+$basePath = Join-Path $ToolsDirectory 'azcopy'
+
+$zipPath = Join-Path $basePath 'azcopy.zip'
+$toolDirPath = Join-Path $basePath 'azcopy'
+$azCopyPath = Join-Path $toolDirPath 'azcopy.exe'
+
+if (Test-Path $azCopyPath) {
```
wiktork, line 17: Is the file persistent from build to build? Will this ever pick up updated versions of azcopy?

eng/release/Scripts/PublishToBlobAccounts.ps1
```diff
@@ -0,0 +1,121 @@
...
+    [CmdletBinding()]
+    Param(
+        [Parameter(Mandatory=$true)][string]$StorageAccountName,
+        [Parameter(Mandatory=$true)][string]$ContainerName,
+        [Parameter(Mandatory=$true)][string]$AccountKey,
+        [Parameter(Mandatory=$true)][string]$Permissions
+    )
+
+    $context = New-AzStorageContext -StorageAccountName $StorageAccountName -StorageAccountKey $AccountKey;
+    return New-AzStorageContainerSASToken -Container $ContainerName -Context $context -Permission $Permissions -ExpiryTime (Get-Date).AddHours(1.0);
```
wiktork, line 48: Note that it's possible that if StartTime defaults to the current time and the transfer operation is attempted very quickly, the server may reject it due to clock skew. Consider setting an explicit StartTime.

wiktork (approved):

eng/release.yml
jander-msft: Correct. It will always require this group since it will always copy from dotnetstage account.

eng/release/Scripts/GetBuildVersion.ps1
jander-msft: This is fairly stable. It's parameterized for easier update if it starts failing all of a sudden.

eng/release/Scripts/InstallAzCopy.ps1
jander-msft: I couldn't find the task when mocking some of this up using the editor.

jander-msft: Also, not sure if I would be able to manually figure out what the path to the azcopy executable would be if using that task.

eng/release/Scripts/InstallAzCopy.ps1
jander-msft: It's somewhat slow, takes a little less than 30 seconds for the whole task to execute.

eng/release/Scripts/InstallAzCopy.ps1
jander-msft: It won't persist from build to build because of how the agents are acquired. This is more of a "if someone calls this again in the same job, don't do extra work." type of thing.

PR #1500, opened 2022-02-15
List Configuration Sources - Testing

kkeirstead:
Testing for listing out configuration sources. This follows the format `"Key": "Value" /*Comment*/`, which means that if we ever decide to change to listing comments first, this test will break. This test reuses components from the original `config show` tests.

Note: This doesn't include KeyPerFile, because it appears that the comment would change depending on the user: `/*KeyPerFileConfigurationProvider for files in 'C:\Users\kkeirstead\AppData\Local\Temp\9fcc0e1b0ed1401b81f5bb40cc7e427e\' (Optional)*/`,


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -298,10 +356,12 @@ private string ConstructUserSettingsJson()
             return generatedUserSettings;
         }
 
-        private string ConstructExpectedOutput(bool redact)
+        private string ConstructExpectedOutput(bool redact, bool showSources = false)
```
wiktork, line 359: Maybe have a string parameter and make "Expected*" constants

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -277,19 +332,22 @@ private string CleanWhitespace(string rawText)
             return string.Concat(rawText.Where(c => !char.IsWhiteSpace(c)));
         }
 
-        private string ConstructUserSettingsJson()
+        private string ConstructSettingsJson(List<string> permittedFileNames = null)
```
wiktork, line 335: nit: params string[] may be more convenient

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -211,6 +211,61 @@ public void FullConfigurationTest(bool redact)
...
+            // to all users on the machine e.g. /etc/dotnet-monitor on Unix systems.
+            File.WriteAllText(Path.Combine(sharedConfigDir.FullName, "settings.json"), ConstructSettingsJson(new() { "Logging.json", "Metrics.json" }));
+
+            // Create the initial host builder.
+            IHostBuilder builder = HostBuilderHelper.CreateHostBuilder(settings);
+
+            // Override the environment configurations to use predefined values so that the test host
+            // doesn't inadvertently provide unexpected values. Passing null replaces with an empty
+            // in-memory collection source.
+            builder.ReplaceAspnetEnvironment(ShowSourcesTestsConstants.DefaultProcess_EnvironmentVariables);
```
wiktork, line 256: Are we picking up any sources from environment variables? I didn't notice any in the baselines.

wiktork (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
kkeirstead: Just to clarify what you mean here - are you suggesting we pass in `directoryNameLocation` from the caller instead of determining it here using `showSources` (and that we save the two directory names as constants in `ConfigurationTests`)?

kkeirstead: I pushed an update that does what I think you're getting at - if you want it to be something different, just let me know before I check it in

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
jander-msft: The `params` keyword is still missing; if you add that, it will alleviate the need to check for null on this parameter and the need to manually create a new array at each callsite

kkeirstead: Oh, my mistake - I wasn't aware of the `params` keyword and just read it as Wiktor suggesting I use string[] as my parameter type. I'll fix that.

jander-msft (approved):

PR #1540, opened 2022-02-23
Simplified Process Filter Schema

### LINKED ISSUE #636: Proposal: Allow simplified process filter schema
The current process filter schema allows for setting a key and a value and an optional match type, for example:

```json
{
  "Key": "ProcessName",
  "Value": "dotnet"
}
```

This can be simplified to:

```json
{
  "ProcessName": "dotnet"
}
```

in order to help minimize the verbosity.

## Proposed Change

Add `ProcessName`, `ProcessId`, and `CommandLine` properties to the process filter descriptor in order to help minimize the verbosity of the configuration. The underlying transform can then check if the values are pulled from these new properties or from the `Key`/`Value` properties.

kkeirstead:
Allows users to use a shorthand means of specifying the process filter instead of listing the key and value separately. This is not a breaking change and should not interfere with the existing functionality. Note: In the testing, I also noticed that the `Key` field could be omitted for a value, and it defaults to checking for a `ProcessId`; since altering this behavior would be a breaking change, it's now listed explicitly in documentation.

Closes #636

### DIFF
--- a/documentation/configuration.md
+++ b/documentation/configuration.md
@@ -249,13 +249,22 @@ Unlike the other diagnostic artifacts (for example, traces), memory dumps aren't
 
 ## Default Process Configuration
 
-Default process configuration is used to determine which process is used for metrics and in situations where the process is not specified in the query to retrieve an artifact. A process must match all the specified filters.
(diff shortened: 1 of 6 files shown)


wiktork:
> it defaults to checking for a `ProcessId`; since altering this behavior would be a breaking change, it's now listed explicitly in documentation.

That may have been accidental on my part. I think your approach here is correct though.

wiktork (approved):

PR #1453, opened 2022-02-04
Add native build support

jander-msft:
This change adds native build support to the repository. It is broken down into three separate changes:

1) Copy (without changing) PAL, CMake support files, and native support scripts from dotnet/runtime @ 59e643b9932122a12e71ceb360a9ae3bb9420dec
1) Copy build scripts from dotnet/diagnostics @ b6b2bb644102806864da991760b6d2a599c312d4
1) Make adjustments to files to successfully build PAL in repository.

This separation should make it easier to review such that you only really need to look at https://github.com/dotnet/dotnet-monitor/commit/1489bd3f7de3a9158a80c9e09b7679ecc51acec1

closes https://github.com/dotnet/dotnet-monitor/issues/1198

PR #1553, opened 2022-03-02
Add native build support

### LINKED ISSUE #1198: Add Native Build Support

jander-msft:
This change adds native build support to the repository. It is broken down into four separate changes:
1. Copy (without changing) PAL, CMake support files, and native support scripts from dotnet/runtime @ 6dd808ff7ae62512330d2f111eb1f60f1ae40125 (a recent commit in the release/6.0 branch)
2. Copy build scripts from dotnet/diagnostics @ b6b2bb644102806864da991760b6d2a599c312d4
3. Make build adjustments specific to dotnet-monitor repository (https://github.com/dotnet/dotnet-monitor/commit/4fc6c1ff43b37121f336ba378abf0f12f7f03ab4)
4. Add test profiler project that successfully builds and produces a viable profiler (https://github.com/dotnet/dotnet-monitor/commit/6ddb8784dd849656a1ddc9bb0661f9ae8a066e96)

CMake is used for all platforms (including Windows) to build native assemblies.

These changes minimize the PAL inclusion significantly (~30 files vs ~1500 files in #1453).

I would recommend reviewing the last two commits as the first two are direct copies from the runtime and diagnostic repos.

Changes not in this PR that will be done subsequently:
- Integration test that exercises the test profiler.
- SAL annotation and PreFAST validation
- Profiler included in product
- Refactor packaging to include native assemblies from all platforms and architectures
- Better integration with CMake Tools for Visual Studio

close #1198


CMakeLists.txt
```diff
@@ -1,6 +1,34 @@
...
+include(${CLR_ENG_NATIVE_DIR}/configurecompiler.cmake)
+
+include_directories(
+    src/external/runtime/src/coreclr/pal/prebuilt/inc
+    src/inc
+    )
+
+if(CLR_CMAKE_HOST_UNIX)
+    include_directories(
+        src/external/runtime/src/coreclr/pal/inc/rt
```
kkeirstead, line 15: Is there a reason why we explicitly include a directory that is part of the directory on the next line?

jander-msft: Subdirectories are not probed automatically. If this isn't done, then source code would have to include headers using something like `#include "rt/palrt.h"` rather than just doing `#include "palrt.h"`. I want to avoid updating the PAL file if absolutely avoidable.

eng/empty.csproj
```diff
@@ -0,0 +1,17 @@
...
+
+  <!-- This project exists purely to work around Arcade
+       behavior. Arcade knows how to restore tasks needed for
+       publishing to blob feeds, but it requires a project file that
+       imports nuget restore targets to be specified for this to
+       work. See details in
+       https://github.com/dotnet/arcade/commit/f657be5cb7cd4920334dd9162173b131211a1e17#r31728598.
+  -->
+
+  <PropertyGroup>
```
wiktork, line 11: Is this needed?

src/Tests/TestProfiler/ClassFactory.cpp
```diff
@@ -0,0 +1,50 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#include "ClassFactory.h"
+#include "EnvProfiler/EnvProfiler.h"
+#include "ProfilerBase.h"
+
+ClassFactory::ClassFactory(REFCLSID guidClsid) :
```
wiktork, line 9: Do you expect the test profiler to be very different from the real one? Perhaps we can make the real one as well (or first).

src/Tests/TestProfiler/EnvProfiler/EnvProfiler.cpp
```diff
@@ -0,0 +1,65 @@
...
+    return { 0x6A494330, 0x5848, 0x4A23,{ 0x9D, 0x87, 0x0E, 0x57, 0xBB, 0xF6, 0xDE, 0x79 } };
+}
+
+STDMETHODIMP EnvProfiler::Initialize(IUnknown *pICorProfilerInfoUnk)
+{
+    HRESULT hr = S_OK;
+
+    if (FAILED(hr = ProfilerBase::Initialize(pICorProfilerInfoUnk)))
+    {
+        std::cerr << "Failed to initialize profiler." << std::endl;
```
wiktork, line 26: Consider a level of indirection for logging. I would not be surprised if we changed the sink at some point.

src/inc/com.h
```diff
@@ -0,0 +1,104 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#pragma once
```
wiktork, line 5: Where did this implementation come from?

src/Tests/TestProfiler/EnvProfiler/EnvProfiler.cpp
jander-msft: Absolutely. But I don't want to invest in a logging infrastructure yet in this PR.

src/Tests/TestProfiler/ClassFactory.cpp
jander-msft: I think we can reuse the `ProfilerBase` "base" class in product too. Maybe the `ClassFactory` as well if we want to `#ifdef` between test and product or come up with some registration mechanism.

eng/empty.csproj
jander-msft: This is here to satisfy the expectation of the eng/native/build-commons.sh script that the project exists. It can be excluded if everyone remembers to always pass `-skipconfigure`. Maybe I can change the higher level scripts to always pass this and then eliminate this project and the targets file. Alternatively, we keep it and try to incorporate the version information file it generates (I would do that post merge).

src/inc/com.h
jander-msft: It's an amalgamation of:
https://github.com/dotnet/runtime/blob/release/6.0/src/coreclr/interop/comwrappers.hpp#L229
https://github.com/dotnet/runtime/blob/release/6.0/src/tests/profiler/native/holder.h#L7
if I recall correctly

jander-msft: And https://github.com/microsoft/CLRInstrumentationEngine/blob/main/src/unix/inc/atl/atlcomcli.h#L638 for the macros.

wiktork (approved):

PR #1555, opened 2022-03-03
Vaildate numerical values

wiktork:


src/Tools/dotnet-monitor/DiagnosticPortValidateOptions.cs
```diff
@@ -12,10 +14,29 @@ internal class DiagnosticPortValidateOptions :
...
             if (options.ConnectionMode == DiagnosticPortConnectionMode.Listen
                 && string.IsNullOrEmpty(options.EndpointName))
             {
-                return ValidateOptionsResult.Fail(Strings.ErrorMessage_DiagnosticPortMissingInListenMode);
+                failures.Add(Strings.ErrorMessage_DiagnosticPortMissingInListenMode);
+            }
+
+            // On Windows, the server is implemented using Named Pipes.
+            // -1 means use as many as resources allow.
+            // Otherwise, the allowed instances are between 1 and 254.
```
kelltrick, line 27:
```suggestion
            // Otherwise, the allowed instances are between 1 and 254, inclusive.
```

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -232,6 +232,9 @@
 1. value: The value of the string that could not be parsed.
 2. errorMessage: Additional error message text.</comment>
   </data>
+  <data name="ErrorMessage_MaxConnections" xml:space="preserve">
+    <value>Value must be between 1 and 254, or -1 (to obtain the maximum number allowed by system resources).</value>
```
kelltrick, line 236:
```suggestion
    <value>Value must be between 1 and 254, inclusive, or -1 (to obtain the maximum number allowed by system resources).</value>
```

kelltrick (approved):

src/Tools/dotnet-monitor/DiagnosticPortValidateOptions.cs
```diff
@@ -12,10 +14,29 @@ internal class DiagnosticPortValidateOptions :
...
+
+            // For Linux domain sockets, all values are valid with 0 representing 1 connection
+            // and other invalid values being normalized to a system defined maximum.
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) &&
+               ((options.MaxConnections < -1) || (options.MaxConnections == 0) || (options.MaxConnections > 254)))
```
jander-msft, line 32: There should be a MaxConnections property on the server endpoint info source that you can use, which is an OS agnostic way of declaring to use the maximum number of connections allowed by the OS.

Also, you should need to check if it's Windows.

src/Tools/dotnet-monitor/ServiceCollectionExtensions.cs
```diff
@@ -37,7 +37,9 @@ public static IServiceCollection ConfigureGlobalCounter(this IServiceCollection
 
         public static IServiceCollection ConfigureMetrics(this IServiceCollection services, IConfiguration configuration)
         {
-            return ConfigureOptions<MetricsOptions>(services, configuration, ConfigurationKeys.Metrics);
+            return ConfigureOptions<MetricsOptions>(services, configuration, ConfigurationKeys.Metrics)
+                .AddSingleton<IValidateOptions<MetricsOptions>, DataAnnotationValidateOptions<MetricsOptions>>();
+
```
kkeirstead, line 42: nit: Looks like there's an extra newline here

src/Tools/dotnet-monitor/DiagnosticPortValidateOptions.cs
wiktork: I'm going to keep this as is, since I copied this message from the NamedPipeServerStream exception.

src/Tools/dotnet-monitor/CollectionRules/Options/Triggers/AspNetRequestDurationOptions.cs
```diff
@@ -19,6 +19,7 @@ internal sealed class AspNetRequestDurationOptions :
             ResourceType = typeof(OptionsDisplayStrings),
             Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_AspNetRequestDurationOptions_RequestCount))]
         [Required]
+        [Range(1, int.MaxValue)]
         public int RequestCount { get; set; }
 
         [Display(
```
kkeirstead, line 25: I just noticed that we don't specify the range for `RequestDuration`; should we be? Just wanted to double check on this before the PR goes in since in `AspNetRequestDurationTriggerSettings` in `EventPipe` we specify a min of `00:00:00` and a max of `01:00:00`

kkeirstead, line 25: I just noticed that we don't specify the range for `RequestDuration`; should we be? Just wanted to double check on this before the PR goes in since in `AspNetRequestDurationTriggerSettings` in `EventPipe` we specify a min of `00:00:00` and a max of `01:00:00`

wiktork: I added this. I think the original idea was that the EventPipe types would validate, but it's a much clearer experience when we validate all the config upfront in dotnet-monitor.

PR #1573, opened 2022-03-09
Add Unit Testing for AspNet* Triggers

kkeirstead:
Adds unit testing for `AspNet*` triggers. Noticed that `RequestCount` and `ResponseCount` were not causing errors when omitted in the testing; this was because their values were not nullable (they defaulted to 0). Users will now see the error message when these properties are omitted due to switching to the `int?` type.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
@@ -314,7 +314,7 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
                     string[] failures = ex.Failures.ToArray();
 
                     Assert.Single(failures);
-                    VerifyRequiredMessage(failures, 0, nameof(AspNetRequestCountOptions.RequestCount));
+                    VerifyRangeMessage<int>(failures, 0, nameof(AspNetRequestCountOptions.RequestCount), "1", int.MaxValue.ToString()); // Since non-nullable, defaults to 0
                 });
         }
 
@@ -410,7 +410,7 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
                     string[] failures = ex.Failures.ToArray();
 
                     Assert.Single(failures);
-                    VerifyRequiredMessage(failures, 0, nameof(AspNetRequestDurationOptions.RequestCount));
+                    VerifyRangeMessage<int>(failures, 0, nameof(AspNetRequestDurationOptions.RequestCount), "1", int.MaxValue.ToString()); // Since non-nullable, defaults to 0
                 });
         }
 
@@ -513,7 +513,7 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
 
                     Assert.Equal(2, failures.Length);
                     VerifyRequiredMessage(failures, 0, nameof(AspNetResponseStatusOptions.StatusCodes));
-                    VerifyRequiredMessage(failures, 1, nameof(AspNetResponseStatusOptions.ResponseCount));
+                    VerifyRangeMessage<int>(failures, 1, nameof(AspNetResponseStatusOptions.ResponseCount), "1", int.MaxValue.ToString()); // Since non-nullable, defaults to 0
                 });
         }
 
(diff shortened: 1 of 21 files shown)


src/Tools/dotnet-monitor/CollectionRules/Options/Triggers/AspNetResponseStatusOptions.cs
```diff
@@ -31,7 +31,7 @@ internal sealed class AspNetResponseStatusOptions :
             Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_AspNetResponseStatusOptions_ResponseCount))]
         [Required]
         [Range(1, int.MaxValue)]
-        public int ResponseCount { get; set; }
+        public int? ResponseCount { get; set; }
```
wiktork, line 34: Since there is now validation on these properties, can they stay non-nullable? Nullable and [Required] are a bit confusing.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
```diff
@@ -249,6 +249,301 @@ public Task CollectionRuleOptions_EventCounterTrigger_GreaterThanLargerThanLessT
...
+                    Assert.Equal(ExpectedRequestCount, requestCountOptions.RequestCount);
+                });
+        }
+
+        [Fact]
+        public Task CollectionRuleOptions_AspNetRequestCountTrigger_RoundTrip()
+        {
+            const int ExpectedRequestCount = 10;
+            TimeSpan ExpectedSlidingWindowDuration = TimeSpan.FromSeconds(45);
+            string[] ExpectedIncludePaths = { "IncludePath1", "IncludePath2" };
```
wiktork, line 278: Not something we need to worry about for this PR, but I am wondering if we should also bring in the custom validator for these properties from the diag repo.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
```diff
@@ -249,6 +249,301 @@ public Task CollectionRuleOptions_EventCounterTrigger_GreaterThanLargerThanLessT
...
+        }
+
+        [Fact]
+        public Task CollectionRuleOptions_AspNetResponseStatusTrigger_RoundTrip()
+        {
+            const int ExpectedResponseCount = 10;
+            TimeSpan ExpectedSlidingWindowDuration = TimeSpan.FromSeconds(45);
+            string[] ExpectedIncludePaths = { "IncludePath1", "IncludePath2" };
+            string[] ExpectedExcludePaths = { "ExcludePath1", "ExcludePath2" };
+            string[] ExpectedStatusCodes = { "400", "500" };
```
wiktork, line 475: Not sure we have tests for this already, but there is a regex that forces to be valid status code ranges.

wiktork (approved):

src/Tools/dotnet-monitor/CollectionRules/Options/Triggers/AspNetResponseStatusOptions.cs
kkeirstead: So, if I revert to making them non-nullable, the error message you'll get for omitting them is "The field `RequestCount` must be between 1 and 2147483647". I guess the question is, do we want an error message that says the field is required, or just that the field must be in that range? I can imagine this being a slight source of confusion if I see that error message and think that doesn't make sense since I didn't set it at all, because I assumed there was a default (as opposed to the "default" just being that of an int, which is 0). I'll defer to what you all think makes more sense though.

jander-msft: I'm guessing that the schema.json (which you'll need to regenerate for this change; otherwise it will fail the json schema tests) will say that type of this property will be `[ "integer", "null" ]` but the parent object will have it be required. This means that:
```json
"Trigger": {
    "Type": "AspNetRequestCount",
    "Settings": {
        "RequestCount": null
    }
},
```
is allowed by the schema, which isn't what we want.

If the property remains non-nullable and I got a message stating "The field RequestCount must be between 1 and 2147483647", while I think it's mildly strange that I wasn't told it was required, it's easy to understand what the mitigation is.

wiktork: I would be ok with non-nullable even if the error message is a bit indirect.

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -267,6 +267,13 @@
 1. parameterValue: The value in the parameter that was rejected
 2. minBytes: The minimum number of bytes acceptable
 3. maxBytes: The maximum number of bytes acceptable</comment>
+  </data>
+  <data name="ErrorMessage_RegularExpressionDoesNotMatch" xml:space="preserve">
+    <value>The field {0} must match the regular expression '{1}'.</value>
```
wiktork, line 272: nit: While I think this is ok, it might actually be clearer to give a couple of examples. E.g.:
"The field {0} must be a status code or status code range between 100 and 500. E.g. 200, 400-500

PR #1581, opened 2022-03-10
Rename test profiler to be product profiler

jander-msft:
This change largely moves and renames (from `(lib)TestProfiler` to `(lib)MonitorProfiler`) the test profiler to be the primary product profiler. None of the functionality has been changed at this point. Additional changes were made to use macros for checking pointers before using them.


src/MonitorProfiler/ClassFactory.cpp
```diff
@@ -16,25 +18,20 @@ STDMETHODIMP ClassFactory::CreateInstance(
     REFIID riid,
     void **ppvObject)
 {
-    if (nullptr == ppvObject)
-    {
-        return E_POINTER;
-    }
+    IfNullPointer(ppvObject);
```
kkeirstead, line 21: I figured out what they do, but I think the naming for these macros (`IfNullPointer` and `IfNullRet`) seemed unintuitive. That may just be due to my lack of context, but I also thought it was worth pointing out that even though they both return an error, only one has `Ret` in the name from a consistency standpoint, and I also don't know what's being returned in either case.

src/MonitorProfiler/DllMain.cpp
```diff
@@ -16,18 +18,14 @@ STDMETHODIMP_(BOOL) DLLEXPORT DllMain(HMODULE hModule, DWORD ul_reason_for_call,
...
-    ComPtr<ClassFactory> pfactory(new ClassFactory(rclsid));
-    if (nullptr == pfactory)
-    {
-        return E_OUTOFMEMORY;
-    }
-
-    return pfactory->QueryInterface(riid, ppv);
+    IfNullPointer(ppv);
+
+    *ppv = nullptr;
```
kkeirstead, line 23: Just a sanity check that we intend to set this to `nullptr` here - in the code being replaced we don't do this right now, so I wanted to make sure that this change was intentional.

src/MonitorProfiler/ClassFactory.cpp
jander-msft: `IfNullRet` is a fairly standard Windows macro (it's in corhlpr.h) that means "if the value is null, return `E_OUTOFMEMORY`". So it's good for checking if something could not be allocated and reporting such an occurance.

`IfNullPointer` is implemented in this PR to mimic `IfNullRet` except that it returns E_POINTER when the value is null. This is typically used to validate double pointers are not null (if they were null, you couldn't assign a value to the outer point becuase `*ptr` would AV). I can rename this one to something like `ExpectedPtr` which seems to be commonly used elsewhere for this.

src/MonitorProfiler/DllMain.cpp
jander-msft: These were intentional. I should have added these in the initial merge but missed it.

We should be initializing double pointers to a reasonable value (typically nullptr) before doing anything so that if we return from the method early then the points are already initialized with appropriate values.

kkeirstead (approved):

PR #1586, opened 2022-03-11
Pack and sign from combined build artifacts

jander-msft:
These changes allow individual build jobs to contribute their native assemblies to a single pipeline artifact so that all platform variations of each native assembly can be incorporated into the same package. Note that the dotnet-monitor package does not pack any native assemblies at this time, but these changes allow that to happen more easily in a subsequent change.

Summary of Changes:
- Remove pack, sign, and publish from the individual build jobs
- Build jobs publish only the binaries the contribute to the NuGet packages
  - Windows x64 Release contributes managed and native assemblies
  - All other release builds contribute native assemblies only
- Add single job for pack, sign, and publish
- Add pipeline parameter to allow skipping test runs; by default, tests are run

### DIFF
--- a/dotnet-monitor.yml
+++ b/dotnet-monitor.yml
@@ -9,6 +9,12 @@ pr:
     - internal/release/*
     - feature/*
 
+parameters:
+- name: RunTests
+  displayName: 'Run Tests'
+  type: boolean
+  default: true
+
 variables:
 - name: _TeamName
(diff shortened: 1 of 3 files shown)


jander-msft:
@hoyosjs, if you wouldn't mind, I'd like to get your feedback on this PR

hoyosjs (approved):

PR #1594, opened 2022-03-14
ARM64 cross build support

jander-msft:
Add support for building native libraries for arm64 Windows, Linux, Linux Musl, and MacOS without testing them. Package projects have not been updated to include these binaries because the package projects do not include native libraries yet.

~~Internal build (skipped tests): https://dev.azure.com/dnceng/internal/_build/results?buildId=1660023&view=results~~
Internal build (skipped tests): https://dev.azure.com/dnceng/internal/_build/results?buildId=1663048&view=results

### DIFF
--- a/eng/build.yml
+++ b/eng/build.yml
@@ -70,11 +70,17 @@ jobs:
           demands: ImageOverride -equals Build.Windows.10.Amd64.VS2019
 
     ${{ if eq(parameters.osGroup, 'Linux') }}:
-      container: mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7-20210714125435-9b5bbc2
+      ${{ if eq(parameters.architecture, 'arm64') }}:
+        container: mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-16.04-cross-arm64-20220223205945-8a8d3be
+      ${{ else }}:
+        container: mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7-20220107135107-9b5bbc2
 
-    # CMake + Clang is broken on Alpine 3.14 prereqs image
     ${{ if eq(parameters.osGroup, 'Linux_Musl') }}:
-      container: mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.13-WithNode-20210910135845-c401c85
+      ${{ if eq(parameters.architecture, 'arm64') }}:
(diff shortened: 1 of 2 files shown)


eng/build.yml
```diff
@@ -70,11 +70,17 @@ jobs:
...
+      ${{ if eq(parameters.architecture, 'arm64') }}:
+        container: mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-16.04-cross-arm64-20220223205945-8a8d3be
+      ${{ else }}:
+        container: mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7-20220107135107-9b5bbc2
 
-    # CMake + Clang is broken on Alpine 3.14 prereqs image
     ${{ if eq(parameters.osGroup, 'Linux_Musl') }}:
-      container: mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.13-WithNode-20210910135845-c401c85
+      ${{ if eq(parameters.architecture, 'arm64') }}:
+        container: mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-16.04-cross-arm64-alpine-20220223205945-538077f
```
wiktork, line 80: Is there an image that's later than 16.04?

wiktork (approved):

eng/build.yml
```diff
@@ -141,14 +152,21 @@ jobs:
           env:
             Token: $(dn-bot-dnceng-artifact-feeds-rw)
 
+    - ${{ if and(eq(parameters.architecture, 'arm64'), eq(parameters.osGroup, 'MacOS')) }}:
+      - script: /bin/bash -c "sudo xcode-select -s /Applications/Xcode_12.4.app/Contents/Developer"
```
kkeirstead, line 156: Is there a reason why we're specifically using Xcode 12.4 here?

eng/build.yml
jander-msft: Missed that there are 18.04 images. I'll change to that and then evaluate if we need to build for 16.04 for older glibc versions.

hoyosjs: This is a crossbuild - it doesn't use the system's glibc.

eng/build.yml
hoyosjs: In diagnostics it's mostly needed for LLDB testing.

eng/build.yml
jander-msft: Sure but it's not easy to discern what glibc version it is using though. For now, I'm just going to build it and check what it linked.

Issue #1337, opened 2022-01-18
Enable configuration templating via shortcuts


willpe:
* Could work with config import concept

PR #1595, opened 2022-03-14
Collection Rule Defaults

kkeirstead:
This is the first part of #1337; this PR provides 7 built-in collection rule defaults that users can set to simplify their configuration. More defaults can be added as part of this PR if people have ones they would like to see, and they can also be added in future releases if customers ask for them. Tests have been included to check for correct behavior when setting (or omitting) defaults, as well as for configuration. The documentation also includes an example of how to use collection rule defaults.


src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleDefaultsInterfaces/EgressProviders.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.CollectionRuleDefaultsInterfaces
+{
+    internal interface EgressProviders
```
wiktork, line 7: Consider something like IEgressProviderProperties or IEgressProviderSettings. Similiar to the other interfaces.

wiktork, line 7: Consider something like IEgressProviderProperties or IEgressProviderSettings. Similiar to the other interfaces.

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleDefaultsOptions.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.ComponentModel.DataAnnotations;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options
+{
+    internal class CollectionRuleDefaultsOptions
+    {
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_CollectionRuleDefaultOptions_Egress))]
+        public string Egress { get; set; }
```
wiktork, line 17: Something to consider:
- We want this to be relatively simple but over time this single tier of properties could potentially get large. 
- While this system doesn't have to worry about name collisions, we are bringing together unrelated properties. I'm wondering if we should at least split up the defaults into multiple tiers such as Trigger defaults, Action Defaults, etc.

wiktork, line 17: Something to consider:
- We want this to be relatively simple but over time this single tier of properties could potentially get large. 
- While this system doesn't have to worry about name collisions, we are bringing together unrelated properties. I'm wondering if we should at least split up the defaults into multiple tiers such as Trigger defaults, Action Defaults, etc.

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleDefaultsOptions.cs
```diff
@@ -0,0 +1,55 @@
...
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_CollectionRuleDefaultOptions_RuleDuration))]
+        [Range(typeof(TimeSpan), CollectionRuleOptionsConstants.RuleDuration_MinValue, CollectionRuleOptionsConstants.RuleDuration_MaxValue)]
+        public TimeSpan? RuleDuration { get; set; }
+
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_CollectionRuleDefaultOptions_RequestCount))]
+        [Range(1, int.MaxValue)]
+        public int RequestCount { get; set; }
```
wiktork, line 47: Consider making all non-reference types nullable since we don't have to have a default set here.

wiktork, line 47: Consider making all non-reference types nullable since we don't have to have a default set here.

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleDefaultsOptions.cs
kkeirstead: I had thought about this too (and didn't necessarily reach a resolution); was your thinking just to have these separated internally (i.e. having a `CollectionRuleActionDefaultsOptions` and `CollectionRuleTriggerDefaultsOptions`), or to also have this be user-facing? Ideally I'd like to keep the outward facing configuration surface as simple as possible, but I could see the merit of a hierarchy along the lines of `CollectionRuleDefaults -> Actions -> Egress` or `CollectionRuleDefaults -> Limits -> ActionCount`.

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
```diff
@@ -161,6 +161,18 @@ private static void AddCollectionRuleActionSchema<TOptions>(GenerationContext co
...
             settingsProperty.Reference = context.AddTypeIfNotExist<TOptions>();
 
+            // Don't require properties that have a corresponding collection rule default
+            foreach (var propName in typeof(CollectionRuleDefaultsOptions).GetProperties().Select(x => x.Name))
+            {
+                settingsProperty.Reference.RequiredProperties.Remove(propName);
+            }
+
+            // Have to special case for Traces since Profile/Providers aren't listed as required, but one of them must be included
+            if (settingsProperty.Reference.RequiredProperties.Count > 0 || typeof(TOptions) == typeof(CollectTraceOptions))
```
jander-msft, line 171: Let's not special case this particular set of options. I don't want us to have to add more special cases for future actions. The algorithm should say that "if a property is required, does not have a default value, and has no corresponding default property, then the settings are required".

jander-msft, line 171: Let's not special case this particular set of options. I don't want us to have to add more special cases for future actions. The algorithm should say that "if a property is required, does not have a default value, and has no corresponding default property, then the settings are required".

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
```diff
@@ -191,6 +202,16 @@ private static void AddCollectionRuleTriggerSchema<TOptions>(GenerationContext c
 
             settingsProperty.Reference = context.AddTypeIfNotExist<TOptions>();
 
+            foreach (var propName in typeof(CollectionRuleDefaultsOptions).GetProperties().Select(x => x.Name))
+            {
+                settingsProperty.Reference.RequiredProperties.Remove(propName);
+            }
+
+            if (settingsProperty.Reference.RequiredProperties.Count > 0)
```
jander-msft, line 210: This is almost correct; you have to check that the required properties don't have default values and there is no corresponding default property.

jander-msft, line 210: This is almost correct; you have to check that the required properties don't have default values and there is no corresponding default property.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Options/CollectionRuleDefaultsOptionsExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using Microsoft.Diagnostics.Tools.Monitor;
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options;
+using System;
+
+namespace Microsoft.Diagnostics.Monitoring.TestCommon.Options
+{
+    internal static partial class CollectionRuleDefaultsOptionsExtensions
+    {
+        public static CollectionRuleDefaultsOptions CreateCollectionRuleDefaults(this RootOptions rootOptions, Action<CollectionRuleDefaultsOptions> callback = null)
```
jander-msft, line 13:
```suggestion
        public static RootOptions AddCollectionRuleDefaults(this RootOptions rootOptions, Action<CollectionRuleDefaultsOptions> callback = null)
```

None of the callers currently use the returned value; instead, return `RootOptions` to allow chaining of method calls.

jander-msft, line 13:
```suggestion
        public static RootOptions AddCollectionRuleDefaults(this RootOptions rootOptions, Action<CollectionRuleDefaultsOptions> callback = null)
```

None of the callers currently use the returned value; instead, return `RootOptions` to allow chaining of method calls.

jander-msft: I think there are other RootOptions extensions, so this should be moved there instead of its own file.

jander-msft: I think there are other RootOptions extensions, so this should be moved there instead of its own file.

src/Tools/dotnet-monitor/Commands/ConfigShowCommandHandler.cs
```diff
@@ -24,10 +24,20 @@ public static void Write(Stream stream, string[] urls, string[] metricUrls, bool
...
-            IHost host = HostBuilderHelper.CreateHostBuilder(settings).Build();
+            IHost host = HostBuilderHelper.CreateHostBuilder(settings).ConfigureCollectionRuleServices().Build();
             IConfiguration configuration = host.Services.GetRequiredService<IConfiguration>();
+
             using ConfigurationJsonWriter jsonWriter = new ConfigurationJsonWriter(stream);
-            jsonWriter.Write(configuration, full: level == ConfigDisplayLevel.Full, skipNotPresent: false, showSources: showSources);
+            jsonWriter.Write(configuration, full: level == ConfigDisplayLevel.Full, skipNotPresent: false, showSources: showSources, host.Services);
+        }
+
+        private static IHostBuilder ConfigureCollectionRuleServices(this IHostBuilder builder)
```
jander-msft, line 34: Why is this needed?

jander-msft, line 34: Why is this needed?

src/Tools/dotnet-monitor/Commands/CollectCommandHandler.cs
```diff
@@ -148,6 +148,8 @@ private static IHostBuilder ConfigureServices(this IHostBuilder builder, AuthCon
 
                 services.ConfigureGlobalCounter(context.Configuration);
 
+                services.ConfigureCollectionRuleDefaults(context.Configuration);
```
jander-msft, line 151: This shouldn't be necessary since ConfigureCollectionRules already calls this.

jander-msft, line 151: This shouldn't be necessary since ConfigureCollectionRules already calls this.

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleDefaultsPostConfigure/SlidingWindowDurationPostConfigure.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            if (null != options.Trigger)
+            {
+                _triggerOperations.TryCreateOptions(options.Trigger.Type, out object triggerSettings);
+
+                if (null != triggerSettings && typeof(SlidingWindowDurations).IsAssignableFrom(triggerSettings.GetType()))
+                {
+                    if (null == ((SlidingWindowDurations)options.Trigger.Settings).SlidingWindowDuration)
+                    {
+                        ((SlidingWindowDurations)options.Trigger.Settings).SlidingWindowDuration = _defaultOptions.CurrentValue.SlidingWindowDuration ?? TimeSpan.Parse(TriggerOptionsConstants.SlidingWindowDuration_Default);
```
jander-msft, line 37: Put `TimeSpan.Parse(TriggerOptionsConstants.SlidingWindowDuration_Default)` in a static readonly field so we don't have to recompute it every time. Another option is if SlidingWindowDuration_Default is not used by anything else anymore, just change it to be a TimeSpan.

jander-msft, line 37: Put `TimeSpan.Parse(TriggerOptionsConstants.SlidingWindowDuration_Default)` in a static readonly field so we don't have to recompute it every time. Another option is if SlidingWindowDuration_Default is not used by anything else anymore, just change it to be a TimeSpan.

src/Tools/dotnet-monitor/Commands/CollectCommandHandler.cs
kkeirstead: I actually (unintentionally) overloaded `ConfigureCollectionRuleDefaults`; the call here corresponds to a `ConfigureOptions` call, and the one in `ConfigureCollectionRules` is more of a logical grouping for the `IPostConfigureOptions`. I can in-line the latter since it doesn't really serve any practical purpose (and to get rid of the confusion around the names).

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Options/CollectionRuleDefaultsOptionsExtensions.cs
kkeirstead: I'll change this - but do you want me to actually chain this call with the other set-up, or is this just so we _could_ in the future? I followed the model we use in some of the other tests where we set up `CollectionRules` and `FileSystemEgress` separately, but if you'd prefer I can chain them all together - whatever you'd consider more readable.

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
kkeirstead: I'm not sure I'm interpreting this correctly, so just to clarify what you mean:
1) If there exists a `[Required]` property that doesn't have a default value (as in `[DefaultValue(40)]`) and isn't one of the `CollectionRuleDefaults` properties, then mark `Settings` as `Required` - my confusion here is that the preceding loop removes any `[Required]` property that is also in `CollectionRuleDefaults`, and I'm not sure why a `[Required]` property would have a `[DefaultValue]` on it; or
2) If there exists a property that's `[Required]` but the `CollectionRuleDefaults` property for it hasn't been set, `Settings` is considered `Required`, but then is no longer considered `Required` if that `CollectionRuleDefaults` property is set - I'm not actually sure if this is feasible, but would be a more nuanced way of only showing settings as `Required` when it truly should be.

Hopefully that makes sense - I realized that `Required` is being used in two different contexts between the annotation and the schema, but I tried to use brackets to help with that.

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleDefaultsOptions.cs
wiktork: I was thinking user-facing. My concern is that having a flat list, especially if it grows, will be confusing since it brings together unrelated properties.

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -222,43 +239,75 @@ private void ProcessSection(IConfigurationSection section, bool includeChildSect
                     {
                         if (child.GetChildren().Any())
                         {
-                            ProcessChildren(child, includeChildSections, redact,  showSources: showSources);
+                            if (loadCRDefaults && section.Key.Equals("Actions"))
```
wiktork, line 242: nit: nameof(CollectionRuleOptions.Actions)

documentation/schema.json
```diff
@@ -1125,6 +1118,124 @@
...
+            }
+          ]
+        }
+      }
+    },
+    "CollectionRuleTriggerDefaultsOptions": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "RequestCount": {
```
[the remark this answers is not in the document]
kkeirstead: That's an interesting idea...I'm not opposed to it, but I think we also want to make sure we consider what level of granularity we want to provide users here given that one of the key objectives with this feature is to minimize configuration and make it simpler. If people have four or five different `SlidingDurationWindow` defaults, does that create more clutter and confusion? Especially when a generic default is set in one place and a specific one overrides it in a different configuration source - I could see that being confusing/annoying. I do think it's worth discussing how many knobs we want to give customers here though.

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
```diff
@@ -150,7 +151,6 @@ private static void AddCollectionRuleSchemas(GenerationContext context)
         private static void AddCollectionRuleActionSchema<TOptions>(GenerationContext context, JsonSchema actionTypeSchema, string actionType)
         {
             JsonSchema subSchema = new JsonSchema();
-            subSchema.RequiredProperties.Add(nameof(CollectionRuleActionOptions.Settings));
```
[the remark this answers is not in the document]
kkeirstead: I'm not completely sure where this comment fits in given some of the prior discussions on this thread (https://github.com/dotnet/dotnet-monitor/pull/1595#discussion_r829377774 and https://github.com/dotnet/dotnet-monitor/pull/1595#discussion_r829376140); other than the special case with Traces, the code later in this function should ensure that `Settings` is required if the defaults don't cover all the required properties.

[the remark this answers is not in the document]
jander-msft: I think GH goofed when submitting the last feedback, so you can ignore this.

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleDefaultsPostConfigure/EgressPostConfigure.cs
```diff
@@ -0,0 +1,41 @@
...
+            _actionOperations = actionOperations;
+        }
+
+        public void PostConfigure(string name, CollectionRuleOptions options)
+        {
+            foreach (var action in options.Actions)
+            {
+                _actionOperations.TryCreateOptions(action.Type, out object actionSettings);
+
+                if (null != actionSettings && typeof(IEgressProviderProperties).IsAssignableFrom(actionSettings.GetType()))
```
jander-msft, line 31:
```suggestion
                if (null != actionSettings && actionSettings is IEgressProviderProperties egressProviderProperties)
```

Then you can use `egressProviderProperties` later without casting.

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleDefaultsPostConfigure/RequestCountsPostConfigure.cs
```diff
@@ -0,0 +1,41 @@
...
+            _triggerOperations = triggerOperations;
+        }
+
+        public void PostConfigure(string name, CollectionRuleOptions options)
+        {
+            if (null != options.Trigger)
+            {
+                _triggerOperations.TryCreateOptions(options.Trigger.Type, out object triggerSettings);
+
+                if (null != triggerSettings && typeof(IRequestCountProperties).IsAssignableFrom(triggerSettings.GetType()))
```
jander-msft, line 31:
```suggestion
                if (null != triggerSettings && triggerSettings is IRequestCountProperties requestCountProperties)
```

Then you can use `requestCountProperties` later without casting.

jander-msft: Do this for the other post configure classes.

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleDefaultsPostConfigure/EgressPostConfigure.cs
kkeirstead: Without really investigating this, my tests failed after making this change. My first guess would be that it's an issue related to getting a copy and thus the assignment isn't reflected in the actual value - but I can look more into this since I don't have experience doing assignment with the `is` operator.

kkeirstead: Oh, nevermind - it was actually that we're only using `actionSettings` for its type and that it's not connected to `action.Settings`. If I set `actionSettings = action.Settings` before we do this check, then it works since then things are properly connected.

documentation/schema.json
```diff
@@ -1125,6 +1118,124 @@
...
+          "type": [
+            "integer",
+            "null"
+          ],
+          "description": "The default threshold of the number of requests that start within the sliding window of time; to be used when no other request count is specified.",
```
jander-msft, line 1169: Do we really need to say that "; to be used when no other request count is specified." for all of these options? Isn't that understood when its already said that these are the default values?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LimitsTestsHelper.cs
```diff
@@ -0,0 +1,22 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    internal static class LimitsTestsHelper
```
jander-msft, line 12: Not sure this is used anymore.

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleDefaultsOptions.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.ComponentModel.DataAnnotations;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options
+{
+    internal class CollectionRuleDefaultsOptions
+    {
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_CollectionRuleDefaultsOptions_Triggers))]
+        public CollectionRuleTriggerDefaultsOptions Triggers { get; set; } = new();
```
jander-msft, line 15: I don't think the `= new();` on these properties is necessary. It might make it more difficult to determine if these properties were actually set by the user rather than happen to be filled by deserialization.

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleDefaultsPostConfigure/EgressPostConfigure.cs
```diff
@@ -0,0 +1,36 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Actions;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.CollectionRuleDefaultsInterfaces
+{
+    internal sealed class EgressPostConfigure :
+        IPostConfigureOptions<CollectionRuleOptions>
```
jander-msft, line 11: Please combined all of these new `IPostConfigureOptions<CollectionRuleOptions>` implementations into one class since they are all doing the same type of work on the same options type with the same default information. Maybe call it `DefaultCollectionRulePostConfigureOptions`. Trying to avoid calling it `CollectionRuleDefaultsPostConfigureOptions` since that implies it is an `IPostConfigureOptions<CollectionRuleDefaultsOptions>`.

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleDefaultsPostConfigure/LimitsPostConfigure.cs
```diff
@@ -0,0 +1,49 @@
...
+        public LimitsPostConfigure(IOptionsMonitor<CollectionRuleDefaultsOptions> defaultOptions)
+        {
+            _defaultOptions = defaultOptions;
+        }
+
+        public void PostConfigure(string name, CollectionRuleOptions options)
+        {
+            if (null == options.Limits)
+            {
+                if (!_defaultOptions.CurrentValue.Limits.ActionCount.HasValue
```
jander-msft, line 23: Rather than constantly accessing `_defaultOptions.CurrentValue.Limits`, cache it in a variable.

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleDefaultsPostConfigure/EgressPostConfigure.cs
```diff
@@ -0,0 +1,36 @@
...
+            )
+        {
+            _defaultOptions = defaultOptions;
+        }
+
+        public void PostConfigure(string name, CollectionRuleOptions options)
+        {
+            foreach (var action in options.Actions)
+            {
+                if (null != action.Settings && action.Settings is IEgressProviderProperties egressProviderProperties)
```
jander-msft, line 26: The null check should not be necessary since `action.Settings is IEgressProviderProperties egressProviderProperties` should return `false` if `action.Settings` is `null`.

jander-msft: Same for the other type testing.

src/Tools/dotnet-monitor/ConfigurationJsonWriter.cs
```diff
@@ -352,4 +352,4 @@ public void Dispose()
             }
         }
     }
-}
+}
```
jander-msft, line 355: Please add back the newline

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LimitsTestsHelper.cs
kkeirstead: I believe it's being used in `CollectionRuleDefaultsTests`

src/Tools/dotnet-monitor/CollectionRules/Options/DefaultCollectionRulePostConfigureOptions.cs
```diff
@@ -0,0 +1,155 @@
...
+
+            if (null == options.Limits.RuleDuration)
+            {
+                options.Limits.RuleDuration = limitsDefaults.RuleDuration;
+            }
+        }
+
+        public void ConfigureRequestCounts(CollectionRuleOptions options)
+        {
+            var triggerDefaults = _defaultOptions.CurrentValue.Triggers;
```
wiktork, line 94: nit: It might be clearer if you expand some of these vars to their actual type.

documentation/configuration.md
```diff
@@ -1401,3 +1401,95 @@ The following example shows the `Limits` portion of a collection rule that has t
...
+    }
+  }
+  ```
+</details>
+
+<details>
+  <summary>Environment Variables</summary>
+  
+  ```bash
+  export DotnetMonitor_Egress__AzureBlobStorage__monitorBlob__accountUri="https://exampleaccount.blob.core.windows.net"
```
wiktork, line 1481: Would this be more useful as yaml?

src/Tests/Microsoft.Diagnostics.Monitoring.ConfigurationSchema/SchemaGenerator.cs
```diff
@@ -191,9 +203,34 @@ private static void AddCollectionRuleTriggerSchema<TOptions>(GenerationContext c
 
             settingsProperty.Reference = context.AddTypeIfNotExist<TOptions>();
 
+            var propertyNames = GetCollectionRuleDefaultsPropertyNames();
```
wiktork, line 206: nit: Consider using full types to make the code more readable.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/TriggerTestsHelper.cs
```diff
@@ -0,0 +1,22 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    internal static class TriggerTestsHelper
```
wiktork, line 12: Consider unifying LimitsTestsHelper and TriggerTestsHelper into a more generic helper that gets Options.

wiktork (approved):

documentation/configuration.md
kkeirstead: I guess I'm not sure if you've seen it, but for QFW I added a lot of sections like this throughout configuration: https://github.com/dotnet/dotnet-monitor/pull/1686

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/TriggerTestsHelper.cs
kkeirstead: Presumably then we'd also pull `GetActionOptions` out of ActionTestsHelper.cs - I could put all three of these helpers together in a new class, but if we want to add more helper methods for limits/triggers in the future, then we'll end up with everything categorized by actions/limits/triggers, with the exception of getting options, which would be in its own bubble. I don't have an issue either way - just more of a question of how we want to keep test helpers organized.

jander-msft (approved):

PR #1597, opened 2022-03-15
Pack native files and symbols in tool package

jander-msft:
These changes pack the monitor profiler libraries under the `profiler/<rid>` folder for each target framework and each native platform. This allows dotnet-monitor to carry the native libraries for all supported platforms in a single package.

Internal build (no tests): https://dev.azure.com/dnceng/internal/_build/results?buildId=1665317&view=results


src/Tools/Directory.Build.props
```diff
@@ -10,4 +10,28 @@
...
+    <NativeArtifactDirectories Include="$(ArtifactsBinDir)Windows_NT.arm64.$(Configuration)" TargetRid="win-arm64" LibraryPrefix="" LibraryExtension=".dll" SymbolsExtension=".pdb" />
+    <NativeArtifactDirectories Include="$(ArtifactsBinDir)Windows_NT.x64.$(Configuration)" TargetRid="win-x64" LibraryPrefix="" LibraryExtension=".dll" SymbolsExtension=".pdb" />
+    <NativeArtifactDirectories Include="$(ArtifactsBinDir)Windows_NT.x86.$(Configuration)" TargetRid="win-x86" LibraryPrefix="" LibraryExtension=".dll" SymbolsExtension=".pdb" />
+  </ItemGroup>
+  <PropertyGroup>
+    <GenerateNuspecDependsOn>$(GenerateNuspecDependsOn);GatherCustomPackageFiles</GenerateNuspecDependsOn>
+    <GatherCustomPackageFilesDependsOn>$(GatherCustomPackageFilesDependsOn)</GatherCustomPackageFilesDependsOn>
+  </PropertyGroup>
+  <!--
+    These items are added into the nupkg at the path sepcified by the PackagePath item metadata.
```
kkeirstead, line 29:
```suggestion
    These items are added into the nupkg at the path specified by the PackagePath item metadata.
```

src/Tools/Directory.Build.targets
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <Import Project="$(MSBuildThisFileDirectory)..\Directory.Build.targets"/>
+  <!--
+    Target that adds CustomPackageFile files to the nupkg at the path specified by the PackagePath item metadata.
+  -->
+  <Target Name="GatherCustomPackageFiles" DependsOnTargets="$(GatherCustomPackageFilesDependsOn)">
```
hoyosjs, line 7: Maybe a way to not depends on the internal is something like: https://github.com/dotnet/diagnostics/blob/4c0a75d8c11618c733efd7a87b2744a9ba4e805f/src/dbgshim/pkg/Microsoft.Diagnostics.DbgShim.props#L25-L31

jander-msft: Sure, I'll take a look at that.

jander-msft: Unfortunately, that particular property doesn't work because of:
- https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.targets#L374
- https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.targets#L460

due to IncludeBuildOutput being forced to false due to https://github.com/dotnet/sdk/blob/main/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Common.targets#L46 since this is a tool package whereas the one that you referenced is a library package.

jander-msft: I was able to do it with the TargetsForTfmSpecificContentInPackage property.

src/Tools/Directory.Build.props
jander-msft: Fixed by removal :smile:

src/Tools/dotnet-monitor/dotnet-monitor.csproj
```diff
@@ -57,4 +57,29 @@
...
+    </ItemGroup>
+    <ItemGroup>
+      <!-- Create profiler symbols items for all of the native platforms. -->
+      <MonitorProfilerSymbolsFile Include="@(NativeArtifactDirectories->'%(Identity)\%(LibraryPrefix)MonitorProfiler%(SymbolsExtension)')">
+        <PackagePath>tools/$(TargetFramework)/any/profiler/%(TargetRid)</PackagePath>
+      </MonitorProfilerSymbolsFile>
+    </ItemGroup>
+    <ItemGroup>
+      <!-- Pack the profiler library for each platform if it exists. -->
+      <TfmSpecificPackageFile Include="@(MonitorProfilerLibraryFile->Exists())" />
```
kelltrick, line 79: Isn't this going it yield something like `true;true;true;true;false;true`?

kelltrick (approved):

src/Tools/dotnet-monitor/dotnet-monitor.csproj
jander-msft: It's a filter, not a transform: https://docs.microsoft.com/en-us/visualstudio/msbuild/item-functions?view=vs-2022

wiktork (approved):

PR #1604, opened 2022-03-16
[main] Add experience survey link

jander-msft:
Backport of #1602 to main

### DIFF
--- a/src/Tools/dotnet-monitor/Commands/GenerateApiKeyCommandHandler.cs
+++ b/src/Tools/dotnet-monitor/Commands/GenerateApiKeyCommandHandler.cs
@@ -55,6 +55,8 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Commands
             }
             else
             {
+                outputBldr.AppendLine(ExperienceSurvey.ExperienceSurveyMessage);
+                outputBldr.AppendLine();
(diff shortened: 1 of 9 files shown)


src/Tools/dotnet-monitor/Commands/GenerateApiKeyCommandHandler.cs
```diff
@@ -55,6 +55,8 @@ public static Task<int> Invoke(CancellationToken token, OutputFormat output, ICo
             }
             else
             {
+                outputBldr.AppendLine(ExperienceSurvey.ExperienceSurveyMessage);
```
wiktork, line 58: Are we sure we want to add this for GenerateKey? For example, this breaks my setup scripts since I haven't yet switched to machinejson.

wiktork (approved):

src/Tools/dotnet-monitor/Commands/GenerateApiKeyCommandHandler.cs
jander-msft: All of the other non`machinejson` output formats have a lot of human  text already so I don't see this as changing much. I assume your script does some kind of beeline splitting and assuming an index if you haven't moved to `machinejson` yet.

PR #1613, opened 2022-03-19
Skip netcoreapp3.1 and net5.0 tests in PR and CI builds

jander-msft:
Allow selecting subset of tests to run.
Schedule once-a-week build that runs all tests.
Test selection uses xUnit trait filtering.

### DIFF
--- a/dotnet-monitor.yml
+++ b/dotnet-monitor.yml
@@ -9,11 +9,25 @@ pr:
     - internal/release/*
     - feature/*
 
+schedules:
+- cron: 0 12 * * 6
+  displayName: Weekly Full Test Run
+  branches:
+    include:
+    - main
+  always: true
+
 parameters:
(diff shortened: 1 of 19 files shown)


dotnet-monitor.yml
```diff
@@ -9,11 +9,25 @@ pr:
     - internal/release/*
     - feature/*
 
+schedules:
+- cron: 0 12 * * 6
```
kelltrick, line 13:
```suggestion
- cron: 15 11 * * 6
```

I've had issues with everyone kicking off scheduled jobs at regular hours like "noon Saturday" or "midnight Sunday", this would cause build pools would get overwhelmed. Don't really know if that would happen here but might be worth it to pin the startup time to something weirder.

kelltrick (approved):

PR #1621, opened 2022-03-22
Update README.md

noahfalk:
People wanting to use dotnet-monitor are likely to come here, but the README didn't forward them to the docs.
They might guess that the documentation folder is a good place to go on their own, but I am trying to make it a pit of
success.


jander-msft (approved):

jander-msft:
@noahfalk, if you have no objections, I'm going to override the status check and merge it myself. The build completed successfully but the webhook seems to not have reported the state change.

noahfalk:
I have no objections at all : )

PR #1639, opened 2022-03-28
Include generated version sources in compilation

jander-msft:
Reuse Unix version generation in Windows version generation.

Prior to this change, product and version information were missing from the native libraries.

Example of the product and version information with this change:
dlldetails


CMakeLists.txt
```diff
@@ -31,4 +31,17 @@ set(PROFILER_SOURCES
     ${CMAKE_CURRENT_LIST_DIR}/src/external/runtime/src/coreclr/pal/prebuilt/idl/corprof_i.cpp
     )
 
+# Add version files
+if(CLR_CMAKE_HOST_WIN32)
+    set(SOURCES
+        ${SOURCES}
+        ${PROJECT_BINARY_DIR}/NativeVersion.rc
+        )
+else(CLR_CMAKE_HOST_WIN32)
```
kkeirstead, line 40: Without having much familiarity with this format, is there a reason why we have a condition for an `else`?

jander-msft: It's the same condition in the above `if` but it's not really a condition. I believe older versions of CMake required this pairing and in some newer versions it issues a warning if it doesn't have the repeated condition.

jander-msft: See https://cmake.org/cmake/help/latest/command/if.html

eng/empty.csproj
```diff
@@ -10,6 +10,7 @@
 
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
+    <AssemblyName>.NET Monitor</AssemblyName>
```
wiktork, line 13: So does this value become the original file name and the file description?

jander-msft: Correct

jander-msft: This generation could be done on a per-file basis so that the description and original name are more tied to the file, but that would take a lot more work than this initial cleanup and inclusion.

wiktork: Did you want to use dotnet-monitor or perhaps dotnet monitor?

wiktork (approved):

eng/empty.csproj
jander-msft: Now using `dotnet-monitor`

PR #1659, opened 2022-03-31
Add LoadProfilerAction test

jander-msft:
Adds a basic success test for the LoadProfilerAction collection rule action. This also tests that the MonitorProfiler library can be successfully loaded into processes of the testable architectures (at this time, only amd64).


src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/NativeLibraryHelper.cs
```diff
@@ -0,0 +1,66 @@
...
+#endif
+
+        public static readonly Guid MonitorProfilerClsid = new Guid("6A494330-5848-4A23-9D87-0E57BBF6DE79");
+
+        public static string GetMonitorProfilerPath(string architecture) =>
+            GetSharedLibraryPath(architecture, "MonitorProfiler");
+
+        private static string GetSharedLibraryPath(string architecture, string rootName)
+        {
+            string artifactsBinPath = Path.GetFullPath(Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "..", "..", ".."));
```
kkeirstead, line 28: I'm guessing there isn't a cleaner way to get this path? My only concern is that dropping back three levels feels somewhat fragile (as in if we move code around in the future this could break), but I don't actually have a suggestion for how to resolve this since you can't use an absolute path.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionTestsHelper.cs
```diff
@@ -90,6 +90,29 @@ public static IEnumerable<object[]> GetTfmsAndLogFormat()
...
+                {
+                    yield return new object[] { tfm, profilerPath };
+                }
+            }
+            else
+            {
+                // If the x64 library could not be found, likely built the x86 library (it shouldn't
+                // be arm64 since tests are not run on arm64). Check that x86 was built and pass the
+                // test since the actual test cannot be run without the x86 runtimes.
+                Assert.True(File.Exists(NativeLibraryHelper.GetMonitorProfilerPath("x86")));
```
kkeirstead, line 112: Would it be appropriate to log something stating that we didn't run the actual test?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
```diff
@@ -0,0 +1,106 @@
...
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    public sealed class LoadProfilerActionTests
+    {
+        private const string DefaultRuleName = "ProfilerTestRule";
+
+        private const string ProductVersionEnvVarName = "DOTNETMONITOR_ProductVersion";
```
kkeirstead, line 27: Given that it has `EnvVarName` in its name this probably goes without saying; should we have a comment saying not to change this without also changing what's in `MainVariable.cpp`? It's probably not really an issue - I just realized these are two distinct copies of the exact same value that are co-dependent.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
```diff
@@ -0,0 +1,106 @@
...
+            {
+                LoadProfilerCallback callback = new(_outputHelper, host);
+                await using ServerSourceHolder sourceHolder = await _endpointUtilities.StartServerAsync(callback);
+
+                AppRunner runner = _endpointUtilities.CreateAppRunner(sourceHolder.TransportName, tfm);
+                runner.ScenarioName = TestAppScenarios.AsyncWait.Name;
+
+                await runner.ExecuteAsync(async () =>
+                {
+                    // At this point, the profiler has alreay been initialized and managed code is already running.
```
kkeirstead, line 66:
```suggestion
                    // At this point, the profiler has already been initialized and managed code is already running.
```

eng/empty.csproj
```diff
@@ -13,6 +13,37 @@
     <AssemblyName>dotnet-monitor</AssemblyName>
   </PropertyGroup>
 
-  <Import Project="versioning.targets" />
+  <!-- This target is invoked by:
+       - eng/native/build-common.sh on Linux and MacOS
+       - eng/Build-Native.cmd on Windows
+
+       It is used to generate native file versioning resources. Here, it is also
+       used to provide produce version information in a header file.
```
kkeirstead, line 21:
```suggestion
       used to provide product version information in a header file.
```

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/NativeLibraryHelper.cs
jander-msft: Correct. but the relativeness between this assembly and the profiler assembly isn't likely going to change unless someone is moving the output folders around.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionTestsHelper.cs
jander-msft: From here, we cannot log. Interestingly, when the member data is populated with an empty array, xUnit will fail the test stating "System.InvalidOperationException : No data found for Microsoft.Diagnostics.Monitoring.Tool.UnitTests.LoadProfilerActionTests.LoadProfilerAsStartupProfilerTest". I'll probably change this to return the profiler path regardless and then let the test opt out if it needs to do so.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/LoadProfilerActionTests.cs
jander-msft: I'll add a comment to specify with what it needs to be kept in sync.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionTestsHelper.cs
```diff
@@ -90,6 +91,31 @@ public static IEnumerable<object[]> GetTfmsAndLogFormat()
             }
         }
 
+        public static IEnumerable<object[]> GetTfmsAchitectureProfilerPath()
```
kkeirstead, line 94:
```suggestion
        public static IEnumerable<object[]> GetTfmsArchitectureProfilerPath()
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionTestsHelper.cs
```diff
@@ -90,6 +91,31 @@ public static IEnumerable<object[]> GetTfmsAndLogFormat()
...
+        {
+            // There isn't a good way to check if the test should be using the x86 or x64 profiler
```
kkeirstead, line 96: nit: outdated comment (doesn't reference `Arm64`)

kkeirstead (approved):

PR #1686, opened 2022-04-06
Add Environment Variable examples in Configuration

kkeirstead:
JSON samples now have corresponding environment variables.

### DIFF
--- a/documentation/collectionruleexamples.md
+++ b/documentation/collectionruleexamples.md
@@ -5,30 +5,49 @@ The following examples provide sample scenarios for using a collection rule. The
 ## Collect Trace - Startup (`Startup` Trigger)
 
 ### JSON
+<details>
+  <summary>Expand for code sample</summary>
 
-```json
-{
-  "AssemblyLoadTraceOnStartup": {
-    "Trigger": {
-      "Type": "Startup"
-    },
-    "Actions": [
-      {
-        "Type": "CollectTrace",
-        "Settings": {
-          "Providers": [{
-              "Name": "Microsoft-Windows-DotNETRuntime",
-              "EventLevel": "Informational",
-              "Keywords": "0x8"
-          }],
-          "Duration": "00:00:15",
-          "Egress": "artifacts"
+  ```json
+  {
+    "AssemblyLoadTraceOnStartup": {
+      "Trigger": {
+        "Type": "Startup"
+      },
+      "Actions": [
+        {
+          "Type": "CollectTrace",
+          "Settings": {
+            "Providers": [{
+                "Name": "Microsoft-Windows-DotNETRuntime",
+                "EventLevel": "Informational",
+                "Keywords": "0x8"
+            }],
+            "Duration": "00:00:15",
+            "Egress": "artifacts"
+          }
         }
-      }
-    ]
+      ]
+    }
   }
-}
+  ```
+</details>
+
+### Environment Variables
+
+<details>
+  <summary>Expand for code sample</summary>
+  
+```bash
+export DotnetMonitor_CollectionRules__AssemblyLoadTraceOnStartup__Trigger__Type="Startup"
+export DotnetMonitor_CollectionRules__AssemblyLoadTraceOnStartup__Actions__0__Settings__Type="CollectTrace"
+export DotnetMonitor_CollectionRules__AssemblyLoadTraceOnStartup__Actions__0__Settings__Providers__0__Name="Microsoft-Windows-DotNETRuntime"
+export DotnetMonitor_CollectionRules__AssemblyLoadTraceOnStartup__Actions__0__Settings__Providers__0__EventLevel="Informational"
+export DotnetMonitor_CollectionRules__AssemblyLoadTraceOnStartup__Actions__0__Settings__Providers__0__Keywords="0x8"
+export DotnetMonitor_CollectionRules__AssemblyLoadTraceOnStartup__Actions__0__Settings__Duration="00:00:15"
+export DotnetMonitor_CollectionRules__AssemblyLoadTraceOnStartup__Actions__0__Settings__Egress="artifacts"
 ```
+</details>
 
 ### Explanation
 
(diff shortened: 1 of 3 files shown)


documentation/collectionruleexamples.md
```diff
@@ -4,39 +4,56 @@ The following examples provide sample scenarios for using a collection rule. The
...
-              "Name": "Microsoft-Windows-DotNETRuntime",
-              "EventLevel": "Informational",
-              "Keywords": "0x8"
-          }],
-          "Duration": "00:00:15",
-          "Egress": "artifacts"
+<details>
+  <summary>JSON</summary>
+
+  ```json
```
jander-msft, line 10: Some of these are indented to line up with the `<sumarry>` tag whereas other examples are no indented. Let's keep it consistent.

documentation/configuration.md
```diff
@@ -327,69 +434,119 @@ Match pid 1 (Shorthand)
...
 
-Metrics urls can be configured via the command line:
+<details>
+  <summary>Command Line</summary>
 
-```cmd
-dotnet monitor collect --metricUrls http://*:52325/
-```
+  ```cmd
+  dotnet monitor collect --metricUrls http://*:52325/
```
jander-msft, line 465: Let's remove the trailing `/` so its consistent with the json and env var examples.

documentation/configuration.md
```diff
@@ -327,69 +434,119 @@ Match pid 1 (Shorthand)
...
+  <summary>JSON</summary>
 
-```json
-{
-  "Metrics": {
-    "Endpoints": "http://localhost:52325"
+  ```json
+  {
+    "Metrics": {
+      "Endpoints": "http://localhost:52325"
```
jander-msft, line 475:
```suggestion
      "Endpoints": "http://*:52325"
```

documentation/configuration.md
```diff
@@ -327,69 +434,119 @@ Match pid 1 (Shorthand)
...
-}
-```
+  ```
+</details>
+
+<details>
+  <summary>Environment Variables</summary>
+  
+  ```bash
+  export DotnetMonitor_Metrics__Endpoints="http://localhost:52325"
```
jander-msft, line 485:
```suggestion
  export DotnetMonitor_Metrics__Endpoints="http://*:52325"
```

documentation/configuration.md
```diff
@@ -577,13 +834,28 @@ The `IncludePaths` and `ExcludePaths` support [wildcards and globbing](#aspnet-r
...
+    "IncludePaths": [ "/api/**/*" ]
+  }
+  ```
+</details>
+
+<details>
+  <summary>Environment Variables</summary>
+  
+  ```bash
+  export DotnetMonitor_CollectionRules__RuleName__Trigger__Type="AspNetRequestCount"
```
jander-msft, line 853: The json example doesn't have this since it is only demonstrating the settings of the trigger.

jander-msft: Same applies to all of the other trigger and action settings examples.

documentation/collectionruleexamples.md
```diff
@@ -255,6 +368,27 @@ This rule, named "LongRequestDuration", will trigger when 5 requests each take g
...
+  export DotnetMonitor_CollectionRules__CollectDumpAndExecute__Trigger__Type="AspNetResponseStatus"
+  export DotnetMonitor_CollectionRules__CollectDumpAndExecute__Trigger__Settings__ResponseCount="3"
+  export DotnetMonitor_CollectionRules__CollectDumpAndExecute__Trigger__Settings__StatusCodes__0="400"
+  export DotnetMonitor_CollectionRules__CollectDumpAndExecute__Actions__0__Name="MyDump"
+  export DotnetMonitor_CollectionRules__CollectDumpAndExecute__Actions__0__Type="CollectDump"
+  export DotnetMonitor_CollectionRules__CollectDumpAndExecute__Actions__0__Settings__Egress="artifacts"
+  export DotnetMonitor_CollectionRules__CollectDumpAndExecute__Actions__0__Settings__Type="Mini"
+  export DotnetMonitor_CollectionRules__CollectDumpAndExecute__Actions__0__WaitForCompletion="true"
+  export DotnetMonitor_CollectionRules__CollectDumpAndExecute__Actions__1__Type="Execute"
+  export DotnetMonitor_CollectionRules__CollectDumpAndExecute__Actions__1__Settings__Path="C:\\Program Files\\Microsoft Visual Studio\\2022\\Preview\\Common7\\IDE\\devenv.exe"
```
jander-msft, line 387: Given that this is bash, using a Windows path example is a bit out of place.

documentation/configuration.md
kkeirstead: Yep, I figured this would come up - the reason I added it was for copy-pastability, so people wouldn't just add the settings lines and assume it should work, even though we don't have it in the json. That being said, if you'd rather not have it for the sake of consistency, I'll yank them all out

documentation/collectionruleexamples.md
kkeirstead: Would you want to switch it to something else? I wanted to keep it consistent with the JSON example, but we could also change that one too if you had a better idea for them

documentation/configuration.md
jander-msft: People already can't just take the json and copy/paste it as-is. Let's remove them to make them consistent.

documentation/collectionruleexamples.md
jander-msft: I'm torn about this; it's not that big of a deal the way it is, but at the same time Linux and MacOS users won't be able to use the example as-is. I can't really think of a trivial alternative that would work across platforms except for maybe using [dotnet-dump analyze](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/dotnet-dump#dotnet-dump-analyze), but that requires user interaction.

jander-msft (approved):

PR #1731, opened 2022-04-11
[release/6.x] Add CMakeSettings.json to support CMake Tools for Visual Studio

jander-msft:


CMakeSettings.json
```diff
@@ -0,0 +1,162 @@
...
+          "value": "${projectDir}\\artifacts\\bin",
+          "type": "FILEPATH"
+        }
+      ]
+    },
+    {
+      "name": "Windows_NT.x64.Debug",
+      "generator": "Ninja",
+      "configurationType": "Debug",
+      "inheritEnvironments": [ "msvc_x64" ],
```
wiktork, line 63: should we be using msvc_x64_x64 instead?

wiktork (approved):

CMakeSettings.json
jander-msft: I looked at the description for these and the difference is that the current one "Target x64 Windows with the MSVC compiler" whereas the one with the extra _x64 "Target x64 Windows with the 64-bit MSVC compiler". I don't think we need to be that specific about the compiler bitness.

PR #1751, opened 2022-04-14
Built In Shortcuts

kkeirstead:
This PR adds the first three built-in trigger shortcuts: `CPUUsage`, `GCHeapSize`, and `ThreadpoolQueueLength`. This should provide a template that can be easily extended for any future `EventCounter` shortcuts we add.


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/BuiltInTriggerTests.cs
```diff
@@ -0,0 +1,172 @@
...
+                    .AddCollectDumpAction(ActionTestsConstants.ExpectedEgressProvider)
+                    .SetIEventCounterTrigger(typeof(T), triggerName, callback);
+            }, host =>
+            {
+                T options = TriggerTestsHelper.GetTriggerOptions<T>(host, DefaultRuleName);
+
+                ICollectionRuleTriggerFactoryProxy factory;
+                Assert.True(host.Services.GetService<ICollectionRuleTriggerOperations>().TryCreateFactory(triggerName, out factory));
+
+                EventPipeTrigger<EventCounterTriggerSettings> trigger = (EventPipeTrigger<EventCounterTriggerSettings>)factory.Create(new EndpointInfo(), null, options);
```
kkeirstead, line 159: This is the reason for the build failures - I forgot that I had needed to change the visibility of some things in the diagnostics repo to do this, which meant the tests build and succeed locally. Since I'm not sure if those are changes we should be making in that repo, I'll hold off on pushing anything there until I get feedback on this - or maybe there's just a better way to test this.

jander-msft: I don't think you should be testing that the settings are correctly set at the pipeline settings level. You should be able to allow the configuration system to process the configuration, pull out the options instance (using something like `IOptions<EventCounterOptions>.Value`), and check its values. Testing the translation from `dotnet monitor` options to pipeline options should not be the responsibility of this test.

documentation/collectionrules/TriggerShortcuts.md
```diff
@@ -0,0 +1,201 @@
+# Trigger Shortcuts
```
wiktork, line 1: Nit: **t**rigger**s**hortcuts.md

documentation/collectionrules/TriggerShortcuts.md
```diff
@@ -0,0 +1,201 @@
...
+      }
+    }
+  }
+  ```
+</details>
+
+<details>
+  <summary>Environment Variables</summary>
+  
+  ```bash
```
wiktork, line 39: Let's switch to yaml/configmap format as with your other PR?

documentation/collectionrules/TriggerShortcuts.md
```diff
@@ -0,0 +1,201 @@
...
+
+### `GCHeapSize` Trigger Shortcut
+
+Execute a trigger when the target application's GC Heap Size is continuously greater than or less than a specified value for a set duration of time.
+
+#### Properties
+
+| Name | Type | Required | Description | Default Value | Min Value | Max Value |
+|---|---|---|---|---|---|---|
+| `GreaterThan` | double? | false | The threshold level in MBs that the counter must maintain (or higher) for the specified duration. | `10` | `0` | |
```
wiktork, line 79: Did we want a larger default here? I understand this is not prescriptive guidance, but 10 feels like something we will simply always trigger on. I would expect a default asp.net core application to **not** trigger the default.

documentation/collectionrules/TriggerShortcuts.md
```diff
@@ -0,0 +1,201 @@
...
+
+### `ThreadpoolQueueLength` Trigger Shortcut
+
+Execute a trigger when the target application's threadpool queue length is continuously greater than or less than a specified value for a set duration of time.
+
+#### Properties
+
+| Name | Type | Required | Description | Default Value | Min Value | Max Value |
+|---|---|---|---|---|---|---|
+| `GreaterThan` | double? | false | The threshold level the counter must maintain (or higher) for the specified duration. | `200` | `0` | |
```
wiktork, line 145: @poppastring  Has done more research here, but should we be supplying a default, given how workload dependent this value is?

documentation/collectionrules/TriggerShortcuts.md
```diff
@@ -0,0 +1,201 @@
...
+  <summary>Environment Variables</summary>
+  
+  ```bash
+  export DotnetMonitor_CollectionRules__RuleName__Trigger__Type="GCHeapSize"
+  ```
+</details>
+
+### `ThreadpoolQueueLength` Trigger Shortcut
+
+Execute a trigger when the target application's threadpool queue length is continuously greater than or less than a specified value for a set duration of time.
```
wiktork, line 139: Doesn't have to be part of this review, but we should provide additional guidance here on how this can be used.

wiktork (approved):

documentation/collectionrules/TriggerShortcuts.md
kkeirstead: I'm open to setting these defaults to anything since I don't have a good intuition for what they should be, but I actually look at it the opposite way - if I'm using a shortcut for the first time and the default is too aggressive, I might be wondering why it didn't work (or if it did work). If it is a pretty easy mark to hit, then I'll know it worked, in which case my next question will be - how can I make this trigger _less_, in which case they can add their own `GreaterThan` value.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/BuiltInTriggerTests.cs
kkeirstead: I think I'm a bit confused on what you're suggesting - is the implication that we could call `optionsMonitor.Get(DefaultRuleName).Trigger.Settings` and check the `EventCounterOptions`? My understanding (and what I've been seeing) is that this would return the original `CPUUsageOptions`, because the translation is only used for the pipeline and not changing the options that we pass into the factory. Hence the current strategy of inspecting the pipeline itself. I realized we could also do testing like `CollectionRulePipeline_EventCounterTriggerTest` but for the built-in shortcuts.

jander-msft: If you are getting the rule options through an IOptions*<CollectionRuleOptions> instance, that means that `IConfigureOptions<CollectionRuleOptions>`, `IPostConfigureOptions<CollectionRuleOptions>`, and `IValidateOptions<CollectionRuleOptions>` have already run. I would expect that the `CollectionRuleOptions` instance you get should be completely filled at that point and not require further mutation of anything on it. That means the `CPUUsageOptions` should have already been translated to `EventCounterOptions` and placed into the `ruleOptions.Trigger.Settings` property. You should be able to use that property path to validate that the options were set correctly from the shortcut options.

kkeirstead: Ah - in which case, I think the change needs to be made in the implementation, and not just the testing. Instead of doing the translation in `EventCounterTriggerFactory` on create, moving the translation logic into `IConfigureOptions` or `IPostConfigureOptions` would actually change the `CollectionRuleOptions`, in which case `EventCounterTriggerFactory` doesn't need to handle any of the different `IEventCounterShortcuts`, assuming that's the desired behavior.

jander-msft: That's what I would expect. I haven't looked at implementation, only the test code you pointed out.

documentation/collectionrules/TriggerShortcuts.md
jander-msft: You probably need to delete this file and then readd it again as lowercase. Otherwise, it'll keep the original casing.

kkeirstead:
I think something weird might have happened when rolling back those three commits (particularly the two merges with main)...I'll see if I can get those changes to go away

src/Tools/dotnet-monitor/CollectionRules/Options/Triggers/EventCounterShortcuts/IEventCounterShortcutsConstants.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Triggers.EventCounterShortcuts
+{
+    internal static class IEventCounterShortcutsConstants
```
jander-msft, line 7: nit: KnownEventCounterConstants

src/Tools/dotnet-monitor/CollectionRules/Triggers/EventPipeTriggerFactory.cs
```diff
@@ -34,11 +34,11 @@ internal sealed class EventPipeTriggerFactory
                 callback);
         }
 
-        private sealed class EventPipeTrigger<TSettings> :
+        internal sealed class EventPipeTrigger<TSettings> :
```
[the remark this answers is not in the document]
kkeirstead: That was a remnant of the prior way of testing -> I switched them back to private

documentation/collectionrules/triggershortcuts.md
```diff
@@ -0,0 +1,273 @@
...
+    }
+  }
+  ```
+</details>
+
+<details>
+  <summary>Kubernetes ConfigMap</summary>
+  
+  ```yaml
+  DotnetMonitor_CollectionRules__RuleName__Trigger__Type: "EventCounter"
```
jander-msft, line 40: Remove the `DotnetMonitor_` prefix

documentation/collectionrules/triggershortcuts.md
```diff
@@ -0,0 +1,273 @@
...
+    }
+  }
+  ```
+</details>
+
+<details>
+  <summary>Kubernetes ConfigMap</summary>
+  
+  ```yaml
+  DotnetMonitor_CollectionRules__RuleName__Trigger__Type: "CPUUsage"
```
jander-msft, line 82: Remove the `DotnetMonitor_` prefix

jander-msft (approved):

PR #1766, opened 2022-04-18
Profiler logging infrastructure

jander-msft:
## Logging
These changes add a logging infrastructure that has a similar shape as [ILogger](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.ilogger?view=dotnet-plat-ext-6.0) in managed code. There currently only two logger implementations: a debug output logger (logs to the output window when debugging on Windows) and an aggregate logger (dispatches log calls to a list of loggers). Choosing a similar abstraction with the same log levels allows easier integration with managed code if we decide to serialize log entries back to the dotnet-monitor process. Finally, common macros and implementation specific macros were added to aid in logging HRESULT failures and log entries in general.

## Environment
These changes also refactor the notion of getting and setting environment variables information into separate abstraction (the `IEnvironment` interface) that allows for sourcing the environment information from different sources. The only source available in this change is from the `ICorProfilerInfo*` instance (provided as the `ProfilerEnvironment` class). This will allow testing more easily independent of having a profiler if we choose to add those tests in the future. An additional helper class, named `EnvironmentHelper`, is added to aid in consolidating the logic for getting and setting specifically known environment variables.

### DIFF
--- a/src/MonitorProfiler/Environment/Environment.h
+++ b/src/MonitorProfiler/Environment/Environment.h
@@ -1,0 +1,24 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#pragma once
+
+#include "corhlpr.h"
+#include "tstring.h"
+
+/// <summary>
+/// Interface allowing getting and setting of environment variables.
+/// </summary>
+DECLARE_INTERFACE(IEnvironment)
+{
+    /// <summary>
+    /// Gets the value of the specified environment variable.
+    /// </summary>
+    STDMETHOD(GetEnvironmentVariable)(const tstring name, tstring& value) PURE;
+
+    /// <summary>
+    /// Sets the value of the specified environment variable.
+    /// </summary>
+    STDMETHOD(SetEnvironmentVariable)(const tstring name, const tstring value) PURE;
+};
--- a/src/MonitorProfiler/Environment/ProfilerEnvironment.cpp
+++ b/src/MonitorProfiler/Environment/ProfilerEnvironment.cpp
@@ -1,0 +1,50 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#include "ProfilerEnvironment.h"
+
+using namespace std;
+
+ProfilerEnvironment::ProfilerEnvironment(ICorProfilerInfo11* pCorProfilerInfo) :
+    m_pCorProfilerInfo(pCorProfilerInfo)
+{
+}
+
+STDMETHODIMP ProfilerEnvironment::GetEnvironmentVariable(const tstring name, tstring& value)
+{
+    HRESULT hr = S_OK;
+
+    ULONG cchValue;
+    IfFailRet(m_pCorProfilerInfo->GetEnvironmentVariable(
+        name.c_str(),
+        0,
+        &cchValue,
+        nullptr));
+
+    unique_ptr<WCHAR[]> pwszValue(new (nothrow) WCHAR[cchValue + 1]);
+    IfNullRet(pwszValue);
+
+    IfFailRet(m_pCorProfilerInfo->GetEnvironmentVariable(
+        name.c_str(),
+        cchValue + 1,
+        nullptr,
+        pwszValue.get()
+        ));
+
+    value.assign(pwszValue.get());
+
+    return S_OK;
+}
+
+STDMETHODIMP ProfilerEnvironment::SetEnvironmentVariable(const tstring name, const tstring value)
+{
+    HRESULT hr = S_OK;
+
+    IfFailRet(m_pCorProfilerInfo->SetEnvironmentVariable(
+        name.c_str(),
+        value.c_str()
+        ));
+
+    return S_OK;
+}
--- a/src/MonitorProfiler/Logging/DebugLogger.cpp
+++ b/src/MonitorProfiler/Logging/DebugLogger.cpp
@@ -1,0 +1,29 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#include "DebugLogger.h"
+#include "LogLevelHelper.h"
+
+using namespace std;
+
+STDMETHODIMP DebugLogger::Log(LogLevel level, const string format, va_list args)
+{
+    HRESULT hr = S_OK;
+
+    CHAR szMessage[s_nMaxEntrySize];
+    _vsnprintf_s(szMessage, s_nMaxEntrySize, _TRUNCATE, format.c_str(), args);
+
+    string strLevel;
+    if (FAILED(LogLevelHelper::GetShortName(level, strLevel)))
+    {
+        strLevel.assign("ukwn");
+    }
+
+    CHAR szString[s_nMaxEntrySize];
+    _snprintf_s(szString, s_nMaxEntrySize, _TRUNCATE, "[profiler]%s: %s\r\n", strLevel.c_str(), szMessage);
+
+    OutputDebugStringA(szString);
+
+    return S_OK;
+}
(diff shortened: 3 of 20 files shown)


src/MonitorProfiler/Logging/Logger.h
```diff
@@ -0,0 +1,103 @@
...
+};
+
+// Checks if EXPR is a failed HRESULT
+// If failed, logs the failure and returns the HRESULT
+#define IfFailLogRet_(pLogger, EXPR) \
+    do { \
+        hr = (EXPR); \
+        if(FAILED(hr)) { \
+            if (nullptr != pLogger) { \
+                IfFailRet(pLogger->Log( \
```
wiktork, line 51: I'm not sure the Log failure should be overriding the EXPR failure. That may be misleading to callers.

jander-msft: Maybe we need some cross plat debug assertion notion. Maybe something that writes out to stderr for debug builds.

src/MonitorProfiler/Environment/Environment.h
```diff
@@ -0,0 +1,24 @@
...
+
+#pragma once
+
+#include "corhlpr.h"
+#include "tstring.h"
+
+/// <summary>
+/// Interface allowing getting and setting of environment variables.
+/// </summary>
+DECLARE_INTERFACE(IEnvironment)
```
wiktork, line 13: I'm ok if you keep this, but do we really need an interface here? Consider just folding all the environment classes into one utility class that always uses the profiler.

src/MonitorProfiler/Environment/ProfilerEnvironment.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "ProfilerEnvironment.h"
+
+using namespace std;
+
+ProfilerEnvironment::ProfilerEnvironment(ICorProfilerInfo11* pCorProfilerInfo) :
+    m_pCorProfilerInfo(pCorProfilerInfo)
+{
+}
+
+STDMETHODIMP ProfilerEnvironment::GetEnvironmentVariable(const tstring name, tstring& value)
```
wiktork, line 14: Maybe I missed these, but add comments regarding the env caching and why this is necessary.

src/MonitorProfiler/MainProfiler/MainProfiler.cpp
```diff
@@ -22,21 +24,37 @@ GUID MainProfiler::GetClsid()
...
-    IfFailRet(m_pCorProfilerInfo->SetEnvironmentVariable(
-        _T("DOTNETMONITOR_ProductVersion"),
-        MonitorProductVersion_TSTR
-        ));
+    // Set product version environment variable to allow discovery of if the profiler
+    // as been applied to a target process. Diagnostic tools must use the diagnostic
+    // communication channel's GetProcessEnvironment command to get this value.
+    IfFailLogRet(EnvironmentHelper::SetProductVersion(m_pEnvironment, m_pLogger));
+
+#ifdef WIN32
```
wiktork, line 42: I've been using TARGET_WINDOWS. Is there a preferred set of defines here?

src/MonitorProfiler/MainProfiler/MainProfiler.cpp
```diff
@@ -45,3 +63,31 @@ STDMETHODIMP MainProfiler::LoadAsNotficationOnly(BOOL *pbNotificationOnly)
 
     return S_OK;
 }
+
+HRESULT MainProfiler::InitializeEnvironment()
+{
+    m_pEnvironment.reset(new (nothrow) ProfilerEnvironment(m_pCorProfilerInfo));
```
wiktork, line 69: std::make_shared?

src/MonitorProfiler/MainProfiler/MainProfiler.cpp
```diff
@@ -45,3 +63,31 @@ STDMETHODIMP MainProfiler::LoadAsNotficationOnly(BOOL *pbNotificationOnly)
...
+    IfNullRet(m_pEnvironment);
+
+    return S_OK;
+}
+
+HRESULT MainProfiler::InitializeLogging()
+{
+    // Create an aggregate logger to allow for multiple logging implementations
+    unique_ptr<AggregateLogger> pAggregateLogger(new (nothrow) AggregateLogger());
```
wiktork, line 78: I would like to say std::make_unique but I'm pretty sure we won't be able to cross compile it due to c++ version :(

jander-msft: Correct

src/MonitorProfiler/MainProfiler/MainProfiler.cpp
jander-msft: For some reason, I don't think this was compiling either, but maybe I'm misremembering. I'll try it again.

src/MonitorProfiler/Environment/ProfilerEnvironment.cpp
```diff
@@ -0,0 +1,50 @@
...
+        nullptr,
+        pwszValue.get()
+        ));
+
+    value.assign(pwszValue.get());
+
+    return S_OK;
+}
+
+STDMETHODIMP ProfilerEnvironment::SetEnvironmentVariable(const tstring name, const tstring value)
```
wiktork, line 40: This helper will be very useful for my current changes.

src/MonitorProfiler/Environment/Environment.h
jander-msft: I wanted to keep how the environment is read separate from an inherent dependency on the profiler so that way we have the possibility of unit testing different classes without having to mock an `ICorProfilerInfo*` implementation. With the way it is, tests can implement an `IEnvironment` to pass in default environment values and read out any changes that were made during testing.

src/MonitorProfiler/Logging/DebugLogger.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "DebugLogger.h"
+#include "LogLevelHelper.h"
+
+using namespace std;
+
+STDMETHODIMP DebugLogger::Log(LogLevel level, const string format, va_list args)
+{
+    HRESULT hr = S_OK;
+
+    CHAR szMessage[s_nMaxEntrySize];
```
wiktork, line 14: Should we be using Unicode for logging? A lot of the data we are going to be working with will likely be WCHAR. We probably don't want to be doing conversions constantly when logging.

wiktork (approved):

src/MonitorProfiler/MainProfiler/MainProfiler.cpp
jander-msft: `TARGET_*` sounds good to me

src/MonitorProfiler/Logging/DebugLogger.cpp
jander-msft: I know we chatted about this, but I think I'm going to defer this for now, but try to resolve it soon.

There's a couple of things I need to figure out first, such as:
- While the debug option logger can use `OutputDebugStringW` natively, the formatting functions will not support `WCHAR*` strings. Thus, it would require conversion from `WCHAR*` to `CHAR*` first and then output using `OutputDebugStringA`. Alternatively, the format string could be parsed to find the format specifiers, format each value individually, and recombine before printing. This is what the PAL does, but we haven't taken any PAL implementations yet and I'm hesitant to write this myself.
- In either case (`string` vs `tstring`), mixing string types for the log statements is going to be problematic. For example, the `IfFailLogRet` macros use `__func__` to write the function name out the log; this variable is typically a `char*`. If we were to convert the `Log` methods to use `tstring`, I think we'd expect the variable list to be non-`char` types. So then we get into having to inspect each value in the `va_list`, determining if it is `string`-y or `char`-y, and convert it to the appropriate `string` or `char` type.

A possible solution here is to have a `Log` overload for `string` and for `tstring`; caller chooses which one to use and must use it consistently with the args. Each logger can determine to which size it should be converted before writing it out.

jander-msft: I was able to convert to using tstring fairly easily: https://github.com/dotnet/dotnet-monitor/pull/1775

We'll just have to keep in mind that we cannot pass `CHAR` or `CHAR*` args to the formatter.

PR #1790, opened 2022-04-22
Add command ipc to profiler

wiktork:
Add a communication mechanism between the profiler and dotnet-monitor.
- Currently using Unix Domain Socket on both Windows and Linux
- Current model is to receive small control messages from dotnet-monitor. 
- Requests are handled serially, long/con-current sessions are not really supported with the current model.


src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -507,6 +507,47 @@ public ActionResult<Models.DotnetMonitorInfo> GetInfo()
...
+                using System.Net.Sockets.Socket s =
+                    new System.Net.Sockets.Socket(System.Net.Sockets.AddressFamily.Unix, System.Net.Sockets.SocketType.Stream, System.Net.Sockets.ProtocolType.Unspecified);
+
+#if NET6_0_OR_GREATER
+                await s.ConnectAsync(ep);
+                await s.SendAsync(new ReadOnlyMemory<byte>(new byte[6]), System.Net.Sockets.SocketFlags.None, HttpContext.RequestAborted);
+                return Ok();
+#else
+                await Task.CompletedTask;
+                throw new InvalidOperationException("Unsupported for 3.1");
```
jander-msft, line 545: Is that UDS is not supported in netcoreapp3.1 or that the implementation detail of the communication channel (using a profiler that uses ICorProfilerInfo12, which is only available on .NET 5+) the limiting factor? If the latter, I think we should not fail just because `dotnet monitor` is running as netcoreapp3.1 but should do better detection of the target process and fail if it is not .NET 5+.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -507,6 +507,47 @@ public ActionResult<Models.DotnetMonitorInfo> GetInfo()
...
+        [EgressValidation]
+        public async Task<ActionResult> CaptureStacks(
+            [FromQuery]
+            int? pid = null,
+            [FromQuery]
+            Guid? uid = null,
+            [FromQuery]
+            string name = null,
+            [FromQuery][Range(-1, int.MaxValue)]
+            int durationSeconds = 5,
```
jander-msft, line 525: Duration doesn't make sense for stacks collection

jander-msft (commented):
Some initial feedback; still reviewing the rest.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -507,6 +507,47 @@ public ActionResult<Models.DotnetMonitorInfo> GetInfo()
...
+            [FromQuery]
+            string egressProvider = null)
+        {
+            ProcessKey? processKey = GetProcessKey(pid, uid, name);
+
+            return await InvokeForProcess(async processInfo =>
+            {
+                System.Net.Sockets.UnixDomainSocketEndPoint ep =
```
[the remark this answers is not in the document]
wiktork: I am probably going to take this out entirely before checking in. I am reluctant to add an endpoint that does nothing.

src/Microsoft.Diagnostics.Monitoring.WebApi/RequestThrottling/RequestLimitTracker.cs
```diff
@@ -33,6 +33,7 @@ public RequestLimitTracker(ILogger<RequestLimitTracker> logger)
             _requestLimitTable.Add(Utilities.ArtifactType_Logs, 3);
             _requestLimitTable.Add(Utilities.ArtifactType_Trace, 3);
             _requestLimitTable.Add(Utilities.ArtifactType_Metrics, 3);
+            _requestLimitTable.Add(Utilities.ArtifactType_Stacks, 3);
```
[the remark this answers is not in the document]
wiktork: The current infrastructure could serialize the requests on the profiler side. I'm ok with starting with 1, esp since this is not too time consuming

src/MonitorProfiler/CMakeLists.txt
```diff
@@ -20,11 +22,18 @@ set(SOURCES
...
+    Communication/IpcCommServer.cpp
+    Communication/IpcCommClient.cpp
+    Communication/CommandServer.cpp
     )
 
 # Build library and split symbols
 add_library_clr(MonitorProfiler SHARED ${SOURCES})
 
+if(CLR_CMAKE_HOST_WIN32)
+    target_link_libraries(MonitorProfiler ws2_32)
```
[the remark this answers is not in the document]
wiktork: I'm pretty sure it's always ws2_32, but I will double check.

src/MonitorProfiler/Communication/IpcCommServer.cpp
```diff
@@ -0,0 +1,130 @@
...
+        return E_INVALIDARG;
+    }
+
+    //We don't error check this on purpose
+    std::remove(rootAddress.c_str());
+
+    address.sun_family = AF_UNIX;
+#pragma warning(push)
+#pragma warning(disable:4996)
+    strncpy(address.sun_path, rootAddress.c_str(), sizeof(address.sun_path));
```
jander-msft, line 46: Could you use strncpy_s so that way we don't have to suppress the warning. Then on Unix, we can add a macro that replaces strncpy_s with strncpy?

src/MonitorProfiler/Communication/IpcCommClient.cpp
```diff
@@ -0,0 +1,106 @@
...
+// See the LICENSE file in the project root for more information.
+
+#include "IpcCommClient.h"
+#include <memory>
+
+HRESULT IpcCommClient::Receive(IpcMessage& message)
+{
+    if (_shutdown.load())
+    {
+        return E_UNEXPECTED;
```
jander-msft, line 12: These are good candidates for a macro

src/MonitorProfiler/Communication/IpcCommServer.cpp
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    _rootAddress = rootAddress;
+
+    sockaddr_un address;
+    ZeroMemory(&address, sizeof(address));
+
+    if ((rootAddress.length() == 0) || (rootAddress.length() >= sizeof(address.sun_path)))
+    {
+        return E_INVALIDARG;
```
jander-msft, line 37: This is a good candidate for a macro

src/MonitorProfiler/Communication/IpcCommServer.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+    if (_domainSocket.Valid())
+    {
+        return E_UNEXPECTED;
+    }
+
+    _rootAddress = rootAddress;
+
+    sockaddr_un address;
+    ZeroMemory(&address, sizeof(address));
```
jander-msft, line 33: nit: use `memset` to avoid the Windows-ism

src/MonitorProfiler/Communication/IpcCommServer.cpp
```diff
@@ -0,0 +1,130 @@
...
+        {
+#if TARGET_UNIX
+            if (errno == EINTR)
+            {
+                continue;
+            }
+#endif
+            return SocketWrapper::GetSocketError();
+        }
+        if (result == 0)
```
jander-msft, line 103: This doesn't seem to be necessary if the while is checking `result <= 0`

src/MonitorProfiler/Communication/IpcCommServer.cpp
```diff
@@ -0,0 +1,130 @@
...
+    } while (result <= 0);
+
+    //TODO server socket should be in non-blocking mode
+
+    SocketWrapper clientSocket = SocketWrapper(accept(_domainSocket, nullptr, nullptr));
+    if (!clientSocket.Valid())
+    {
+        return SocketWrapper::GetSocketError();
+    }
+    if (setsockopt(clientSocket, SOL_SOCKET, SO_RCVTIMEO, reinterpret_cast<const char*>(&ReceiveTimeoutMilliseconds), sizeof(ReceiveTimeoutMilliseconds)) != 0)
```
jander-msft, line 117: A pointer to a `timeval` struct should be passed as the option value, not an `int`. And update the size to the size of the structure.

src/MonitorProfiler/Communication/CommandServer.cpp
```diff
@@ -0,0 +1,104 @@
...
+        response.MessageType = SUCCEEDED(hr) ? MessageType::OK : MessageType::Error;
+        response.Parameters = hr;
+
+        hr = client->Send(response);
+        if (FAILED(hr))
+        {
+            _logger->Log(LogLevel::Error, _T("Unexpected error when sending data: 0x%08x"), hr);
+            continue;
+        }
+
```
jander-msft, line 82: Should there be a call to `client.Shutdown()` here?

src/MonitorProfiler/Communication/IpcCommClient.cpp
```diff
@@ -0,0 +1,106 @@
...
+        }
+    } while (sent < 0);
+
+    return S_OK;
+}
+
+void IpcCommClient::Shutdown()
+{
+    _shutdown.store(true);
+    shutdown(_socket,
```
jander-msft, line 95: Should the return value be checked if its not 0 and then return `SocketWrapper::GetSocketError()`? Requires `Shutdown` to be updated to have an `HRESULT` return type.

src/MonitorProfiler/Communication/IpcCommClient.cpp
```diff
@@ -0,0 +1,106 @@
...
+#if TARGET_UNIX
+            if (errno == EINTR)
+            {
+                //Signal can interrupt operations. Try again.
+                continue;
+            }
+#endif
+            return SocketWrapper::GetSocketError();
+        }
+    } while (sent < 0);
```
jander-msft, line 87: I think you need to account for partial writes.

src/MonitorProfiler/Communication/IpcCommClient.cpp
```diff
@@ -0,0 +1,106 @@
...
+    {
+        int read = recv(_socket, &buffer[offset], sizeof(buffer) - offset, 0);
+        if (read == 0)
+        {
+            return E_ABORT;
+        }
+        if (read < 0)
+        {
+#if TARGET_UNIX
+            if (errno == EINTR)
```
kkeirstead, line 34: Is there any risk of an infinite loop here? Do we need to have a counter of how many times we try again before we error out?

src/MonitorProfiler/Communication/IpcCommServer.cpp
```diff
@@ -0,0 +1,130 @@
...
+    {
+        return E_UNEXPECTED;
+    }
+
+    _rootAddress = rootAddress;
+
+    sockaddr_un address;
+    ZeroMemory(&address, sizeof(address));
+
+    if ((rootAddress.length() == 0) || (rootAddress.length() >= sizeof(address.sun_path)))
```
kkeirstead, line 35: Is it guaranteed that `rootAddress` isn't null?

src/MonitorProfiler/Communication/IpcCommClient.cpp
wiktork: I don't think it's a practical concern? The app would need to be continuously interrupted by signals.

src/MonitorProfiler/Communication/IpcCommClient.cpp
wiktork: I can push the error handling to the caller, although I am unlikely to check the error code anyway since my action is to close the socket descriptor either way.

src/MonitorProfiler/Communication/IpcCommServer.cpp
wiktork: Resolving to my other comment.

src/MonitorProfiler/Communication/IpcCommClient.cpp
jander-msft: Sure, but it's good to be able to log an error when a failure does occur. I want us to get logging on bad results as much as possible.

src/MonitorProfiler/Communication/CommandServer.cpp
wiktork: That would probably be cleaner, yes. My original approach involved multiple clients and a fully blocking socket so shutdown was a much stricter requirement.

src/MonitorProfiler/Communication/IpcCommServer.cpp
```diff
@@ -0,0 +1,141 @@
...
+#include "IpcCommClient.h"
+#include "IpcCommServer.h"
+
+IpcCommServer::IpcCommServer() : _shutdown(false)
+{
+}
+
+IpcCommServer::~IpcCommServer()
+{
+    _domainSocket.~SocketWrapper();
```
jander-msft, line 14: Why does the dtor have to be explicitly called? Is it because you explicitly implement the dtor for `IpcCommServer`?

src/MonitorProfiler/MainProfiler/MainProfiler.cpp
```diff
@@ -91,3 +97,34 @@ HRESULT MainProfiler::InitializeLogging()
 
     return S_OK;
 }
+
+HRESULT MainProfiler::InitializeCommandServer()
+{
+    HRESULT hr = S_OK;
+
+    tstring instanceId;
+    IfFailRet(m_pEnvironment->GetEnvironmentVariable(_T("DotnetMonitorProfiler_InstanceId"), instanceId));
```
jander-msft, line 106: Can you move this to EnvironmentHelper? I'd like to try to keep all of the `DotnetMonitorProfiler_` variables in one spot for easy determination of what variables are in use.

src/MonitorProfiler/MainProfiler/MainProfiler.cpp
```diff
@@ -91,3 +97,34 @@ HRESULT MainProfiler::InitializeLogging()
...
+
+#if TARGET_UNIX
```
jander-msft, line 108: Do you think we'll add more file system type code in the future? I can see us doing more work in the `/tmp` dir in future chagnes. Maybe we should put this in a helper class so it doesn't need to be repeated.

src/MonitorProfiler/ProfilerBase.h
```diff
@@ -14,7 +14,7 @@ class ProfilerBase :
     public ICorProfilerCallback11
 {
 protected:
-    ComPtr<ICorProfilerInfo11> m_pCorProfilerInfo;
+    ComPtr<ICorProfilerInfo12> m_pCorProfilerInfo;
```
jander-msft, line 17: I think this can be reverted until its necessary to bump up the requirement.

src/MonitorProfiler/Utilities/BlockingQueue.h
```diff
@@ -0,0 +1,58 @@
...
+
+#include <queue>
+#include <mutex>
+#include <condition_variable>
+
+template<typename T>
+class BlockingQueue final
+{
+public:
+    HRESULT Push(const T& item)
```
jander-msft, line 13: nit: Enqueue

src/MonitorProfiler/Utilities/BlockingQueue.h
```diff
@@ -0,0 +1,58 @@
...
+                return E_UNEXPECTED;
+            }
+            _queue.push(item);
+        }
+        _condition.notify_all();
+
+        return S_OK;
+    }
+
+    HRESULT BlockingDequeue(T& item)
```
jander-msft, line 28: nit: Dequeue

src/Tools/dotnet-monitor/CollectionRules/Actions/LoadProfilerAction.cs
```diff
@@ -62,6 +62,8 @@ public LoadProfilerAction(ILogger logger, IEndpointInfo endpointInfo, LoadProfil
                 {
                     DiagnosticsClient client = new DiagnosticsClient(_endpointInfo.Endpoint);
 
+                    await client.SetEnvironmentVariableAsync("DotnetMonitorProfiler_InstanceId", _endpointInfo.RuntimeInstanceCookie.ToString("D"), token);
```
jander-msft, line 65: I don't think this should be here. This is very specific to the monitor profiler whereas the LoadProfilerAction is mean for general usage.

jander-msft (approved):

src/MonitorProfiler/Communication/IpcCommServer.cpp
wiktork: I want to make sure the inner socket is closed first before I attempt any path cleanup.

src/Tools/dotnet-monitor/CollectionRules/Actions/LoadProfilerAction.cs
wiktork: Will use pid for now. We will likely need to add a token for runtime instance id to add to the set environment action.

src/MonitorProfiler/Environment/EnvironmentHelper.cpp
```diff
@@ -22,3 +22,32 @@ HRESULT EnvironmentHelper::SetProductVersion(
...
+}
+
+HRESULT EnvironmentHelper::GetTempFolder(const std::shared_ptr<IEnvironment>& pEnvironment, const std::shared_ptr<ILogger>& pLogger, tstring& tempFolder)
+{
+    HRESULT hr = S_OK;
+
+    tstring tmpDir;
+    hr = pEnvironment->GetEnvironmentVariable(s_wszTempEnvVar, tmpDir);
+#if TARGET_WINDOWS
+    IfFailLogRet(hr);
```
jander-msft, line 42: This is not useful for logging purposes because it'll just log that the expression "hr" resulted in failure. I would call the same method on both Windows and non-Windows but wrap the Windows one in the `IfFailLogRet` macro.

jander-msft (approved):

src/MonitorProfiler/Environment/EnvironmentHelper.cpp
```diff
@@ -22,3 +22,33 @@ HRESULT EnvironmentHelper::SetProductVersion(
...
+    tstring tmpDir;
+#if TARGET_WINDOWS
+    IfFailLogRet(pEnvironment->GetEnvironmentVariable(s_wszTempEnvVar, tmpDir));
+#else
+    hr = pEnvironment->GetEnvironmentVariable(s_wszTempEnvVar, tmpDir);
+#endif
+
+    if (FAILED(hr))
+    {
+        tmpDir = s_wszDefaultTempFolder;
```
jander-msft, line 48: s_wszDefaultTempFolder is `/tmp` on all platforms as the code is right now. This fallback should probably only be done for non-Windows platforms.

wiktork: I can address this on the next PR

PR #1791, opened 2022-04-23
[Egress Extensibility] Setup configuration system for extensibility

kelltrick:
This is the first PR for egress extensibility, this is a work in progress and should not go into `main` yet. The PR includes other changes that are required to test the configuration system but aren't ready for `main`.
- [Should be reviewed] Adds configuration system, the main workaround and change is that `TOptions` was previously relied on as a unique key in the configuration system. This makes `TOptions` not need to be unique.
- [Not intended to be reviewed yet, files under `src/Extensions/azure-storage/*`] Adds a `AzureStorage` assembly to test upload.
- [Not intended to be reviewed yet, one file `src/Tools/dotnet-monitor/Egress/ExtensionEgress/ExtensionEgressProvider.cs`] Adds an `ExtensionEgressProvider` class to call external extensions, this is a work in progress and not complete.

### DIFF
--- a/eng/common/cross/build-rootfs.sh
+++ b/eng/common/cross/build-rootfs.sh
@@ -4,12 +4,13 @@ set -e
 
 usage()
 {
-    echo "Usage: $0 [BuildArch] [CodeName] [lldbx.y] [--skipunmount] --rootfsdir <directory>]"
+    echo "Usage: $0 [BuildArch] [CodeName] [lldbx.y] [llvmx[.y]] [--skipunmount] --rootfsdir <directory>]"
     echo "BuildArch can be: arm(default), armel, arm64, x86"
     echo "CodeName - optional, Code name for Linux, can be: xenial(default), zesty, bionic, alpine, alpine3.13 or alpine3.14. If BuildArch is armel, LinuxCodeName is jessie(default) or tizen."
     echo "                              for FreeBSD can be: freebsd12, freebsd13"
     echo "                              for illumos can be: illumos."
     echo "lldbx.y - optional, LLDB version, can be: lldb3.9(default), lldb4.0, lldb5.0, lldb6.0 no-lldb. Ignored for alpine and FreeBSD"
+    echo "llvmx[.y] - optional, LLVM version for LLVM related packages."
     echo "--skipunmount - optional, will skip the unmount of rootfs folder."
     echo "--use-mirror - optional, use mirror URL to fetch resources, when available."
     exit 1
@@ -48,6 +49,7 @@ __AlpinePackages+=" gettext-dev"
 __AlpinePackages+=" icu-dev"
 __AlpinePackages+=" libunwind-dev"
 __AlpinePackages+=" lttng-ust-dev"
+__AlpinePackages+=" compiler-rt-static"
 
 # CoreFX dependencies
 __UbuntuPackages+=" libcurl4-openssl-dev"
@@ -164,6 +166,15 @@ while :; do
         no-lldb)
             unset __LLDB_Package
             ;;
+        llvm*)
+            version="$(echo "$lowerI" | tr -d '[:alpha:]-=')"
+            parts=(${version//./ })
+            __LLVM_MajorVersion="${parts[0]}"
+            __LLVM_MinorVersion="${parts[1]}"
+            if [[ -z "$__LLVM_MinorVersion" && "$__LLVM_MajorVersion" -le 6 ]]; then
+                __LLVM_MinorVersion=0;
+            fi
+            ;;
         xenial) # Ubuntu 16.04
             if [ "$__CodeName" != "jessie" ]; then
                 __CodeName=xenial
@@ -260,6 +271,10 @@ if [ "$__BuildArch" == "armel" ]; then
     __LLDB_Package="lldb-3.5-dev"
 fi
 __UbuntuPackages+=" ${__LLDB_Package:-}"
+
+if [ ! -z "$__LLVM_MajorVersion" ]; then
+    __UbuntuPackages+=" libclang-common-${__LLVM_MajorVersion}${__LLVM_MinorVersion:+.$__LLVM_MinorVersion}-dev"
+fi
 
 if [ -z "$__RootfsDir" ] && [ ! -z "$ROOTFS_DIR" ]; then
     __RootfsDir=$ROOTFS_DIR
--- a/src/Extensions/azure-storage/azure-storage/AzureBlobEgressProvider.AutoFlushStream.cs
+++ b/src/Extensions/azure-storage/azure-storage/AzureBlobEgressProvider.AutoFlushStream.cs
@@ -1,0 +1,107 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Egress.AzureBlob
+{
+    partial class AzureBlobEgressProvider
+    {
+        /// <summary>
+        /// Automatically flushes the stream after a certain amount of bytes have been written.
+        /// </summary>
+        private sealed class AutoFlushStream : Stream
+        {
+            private readonly Stream _baseStream;
+            private readonly long _flushSize;
+            private long _written;
+
+            public AutoFlushStream(Stream stream, long flushSize)
+            {
+                _flushSize = flushSize >= 0 ? flushSize : throw new ArgumentOutOfRangeException(nameof(flushSize));
+                _baseStream = stream ?? throw new ArgumentNullException(nameof(stream));
+            }
+
+            public override bool CanRead => _baseStream.CanRead;
+            public override bool CanSeek => _baseStream.CanSeek;
+            public override bool CanWrite => _baseStream.CanWrite;
+            public override long Length => _baseStream.Length;
+            public override bool CanTimeout => _baseStream.CanTimeout;
+            public override int WriteTimeout { get => _baseStream.WriteTimeout; set => _baseStream.WriteTimeout = value; }
+            public override int ReadTimeout { get => _baseStream.ReadTimeout; set => _baseStream.ReadTimeout = value; }
+            public override long Position { get => _baseStream.Position; set => _baseStream.Position = value; }
+            public override int Read(byte[] buffer, int offset, int count) => _baseStream.Read(buffer, offset, count);
+            public override long Seek(long offset, SeekOrigin origin) => _baseStream.Seek(offset, origin);
+            public override void SetLength(long value) => _baseStream.SetLength(value);
+            public override void Close() => _baseStream.Close();
+            public override Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken) =>
+                _baseStream.CopyToAsync(destination, bufferSize, cancellationToken);
+            public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken) =>
+                _baseStream.ReadAsync(buffer, offset, count, cancellationToken);
+            public override int ReadByte() => _baseStream.ReadByte();
+
+            //CONSIDER These are not really used, but should still autoflush.
+            public override IAsyncResult BeginRead(byte[] buffer, int offset, int count, AsyncCallback callback, object state) =>
+                _baseStream.BeginRead(buffer, offset, count, callback, state);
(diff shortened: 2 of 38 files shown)


eng/common/cross/build-rootfs.sh
```diff
@@ -4,12 +4,13 @@ set -e
 
 usage()
 {
-    echo "Usage: $0 [BuildArch] [CodeName] [lldbx.y] [--skipunmount] --rootfsdir <directory>]"
+    echo "Usage: $0 [BuildArch] [CodeName] [lldbx.y] [llvmx[.y]] [--skipunmount] --rootfsdir <directory>]"
```
jander-msft, line 7: For some reason, infra changes are in your PR.

src/Extensions/azure-storage/azure-storage/AzureBlobEgressProvider.AutoFlushStream.cs
```diff
@@ -0,0 +1,107 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Egress.AzureBlob
+{
+    partial class AzureBlobEgressProvider
```
jander-msft, line 14: Please do not copy code wholesale. Either move it to the new location or link it in. If you need to add a temporary ProjectReference to dotnet-monitor, I would be fine with that too.

src/Extensions/azure-storage/azure-storage/Microsoft.Diagnostics.Monitoring.AzureStorage.csproj
```diff
@@ -0,0 +1,26 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
jander-msft, line 1: Remove the tabs. And the indentation should be 2 spaces for csproj files.

src/Extensions/azure-storage/azure-storage/Microsoft.Diagnostics.Monitoring.AzureStorage.csproj
```diff
@@ -0,0 +1,26 @@
...
+
+	<PropertyGroup>
+		<TargetFrameworks>$(ToolTargetFrameworks)</TargetFrameworks>
+		<RuntimeIdentifiers>linux-x64;linux-musl-x64;win-x64</RuntimeIdentifiers>
+		<PackAsToolShimRuntimeIdentifiers>linux-x64;linux-musl-x64;win-x64</PackAsToolShimRuntimeIdentifiers>
+		<OutputType>Exe</OutputType>
+		<TargetFramework>net6.0</TargetFramework>
+		<RootNamespace>Microsoft.Diagnostics.Monitoring.AzureStorage</RootNamespace>
+		<ImplicitUsings>enable</ImplicitUsings>
```
jander-msft, line 10: I would recommend against enabling these. Since you will have to backport to `release/6.x`, I don't think this is going to work for the `netcoreapp3.1` TFM.

src/Extensions/azure-storage/azure-storage/Microsoft.Diagnostics.Monitoring.AzureStorage.csproj
```diff
@@ -0,0 +1,26 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+	<PropertyGroup>
+		<TargetFrameworks>$(ToolTargetFrameworks)</TargetFrameworks>
+		<RuntimeIdentifiers>linux-x64;linux-musl-x64;win-x64</RuntimeIdentifiers>
```
jander-msft, line 5: These two identifier lists are probably not necessary.

src/Extensions/azure-storage/azure-storage/Microsoft.Diagnostics.Monitoring.AzureStorage.csproj
```diff
@@ -0,0 +1,26 @@
...
+		<PackAsToolShimRuntimeIdentifiers>linux-x64;linux-musl-x64;win-x64</PackAsToolShimRuntimeIdentifiers>
+		<OutputType>Exe</OutputType>
+		<TargetFramework>net6.0</TargetFramework>
+		<RootNamespace>Microsoft.Diagnostics.Monitoring.AzureStorage</RootNamespace>
+		<ImplicitUsings>enable</ImplicitUsings>
+		<Nullable>disable</Nullable>
```
jander-msft, line 11: This isn't enabled by default, so you shouldn't have to disable it. If left explicitly disabled, when we go to enable it across the board, this may fall under the radar.

src/Extensions/azure-storage/azure-storage/Program.cs
```diff
@@ -0,0 +1,91 @@
...
+using System.Text.Json;
+
+namespace Microsoft.Diagnostics.Monitoring.AzureStorage
+{
+    internal class Program
+    {
+        private static Stream StdInStream = null;
+        static async Task Main(string[] args)
+        {
+            string jsonConfig = Console.ReadLine();
```
jander-msft, line 16: I think this should be a command line parameter as a file path:
1. Avoids mixed mode content on the stdin stream.
2. Makes it way easier to statically create a file for automated testing and debuggability purposes.

src/Extensions/azure-storage/azure-storage/Program.cs
```diff
@@ -0,0 +1,91 @@
...
+            //Console.TreatControlCAsInput = true;
+            Console.CancelKeyPress += Console_CancelKeyPress;
+
+            string blobUri = await provider.EgressAsync(options, action, config.Settings, CancellationToken.None);
+            Console.Write(blobUri);
+        }
+
+        private static void Console_CancelKeyPress(object sender, ConsoleCancelEventArgs e)
+        {
+            StdInStream.Close();
```
jander-msft, line 69: Why would Ctrl+C close stdin? I would expect Ctrl+C to cancel a CancellationTokenSource, from which a token was gotten and passed into EgressAsync

src/Extensions/azure-storage/azure-storage/Program.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        private static string GetConfig(Dictionary<string, string> configDict, string propKey)
+        {
+            if (configDict.ContainsKey(propKey))
```
jander-msft, line 74: `Dictionary.TryGetValue(string key, out T value)`

src/Extensions/azure-storage/azure-storage/Program.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            if (options.SharedAccessSignature == null && options.SharedAccessSignatureName != null && config.Properties.ContainsKey(options.SharedAccessSignatureName))
+            {
+                options.SharedAccessSignature = config.Properties[options.SharedAccessSignatureName];
+            }
+
+            Func<CancellationToken, Task<Stream>> action = new Func<CancellationToken, Task<Stream>>(
+                (CancellationToken token) =>
+                {
+                    StdInStream = Console.OpenStandardInput();
```
jander-msft, line 48: Couple of things:
- I think using both `Console.ReadX` and `Console.OpenStandardInput` at the same time are typically not advisable.
- You shouldn't assign a property in a callback that is also used asynchronously in an event handler. You should establish the value of the property (probably as a field) before any assumed usage of it.

src/Tools/dotnet-monitor/Egress/ExtensionEgress/ExtensionEgressDiscoverer.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        private bool IsBuiltInSection(IConfigurationSection s)
+        {
+            Type egressType = typeof(EgressOptions);
+            return egressType.GetProperty(s.Key) != null || egressType.GetField(s.Key) != null;
+        }
+    }
+
+    public interface IExternalEgressDiscovererConfigurationProvider : IEnumerable<string>
```
jander-msft, line 63: I don't think this has any purpose.

src/Microsoft.Diagnostics.Monitoring.Options/EgressCategoryOptions.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Egress
+{
+    /// <summary>
+    /// Egress provider options for external egress providers.
+    /// </summary>
+    internal sealed partial class EgressCategoryOptions
```
jander-msft, line 15: Only use `partial` when it has actual usage. This is happening in several locations. Otherwise I look through the code trying to find the other part of the class and don't find it.

src/Microsoft.Diagnostics.Monitoring.Options/EgressCategoryOptions.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public object ConfigurationSection { get; set; }
```
jander-msft, line 17: From what I can tell, this property isn't used (yet?). I would also ask you to not just store the configuration in the options type since that is leaking configuration through options (separation of responsibilities). Just pull out what is needed from configuration and store it on the options instance.

jander-msft (changes requested):

jander-msft:
> * [Not intended to be reviewed yet, files under `src/Extensions/azure-storage/*`] Adds a `AzureStorage` assembly to test upload.
> * [Not intended to be reviewed yet, one file `src/Tools/dotnet-monitor/Egress/ExtensionEgress/ExtensionEgressProvider.cs`] Adds an `ExtensionEgressProvider` class to call external extensions, this is a work in progress and not complete.

I would suggest backing these changes out and keep them in a private branch until they are ready for review. That will greatly simplify this PR. Additionally, I don't think it's a good policy to allow unreviewed code to be merged and then later ask folks to review it for a multitude of reasons, some being that it's difficult to know what actually changed (especially since a bunch of it seems to be wholesale copies with possible minor adjustments) or it's forgotten that it needs to be reviewed.

src/Tools/dotnet-monitor/Egress/EgressService.cs
```diff
@@ -137,26 +159,26 @@ private static EgressArtifactSettings CreateSettings(IEndpointInfo source, strin
...
         private void Reload()
         {
-            _providerTypeMap.Clear();
-            _optionsTypeMap.Clear();
+            _egressProviderMap.Clear();
 
             // Deliberately fill the maps in the reverse order of how they are accessed
             // by the GetProvider method to avoid indeterminate accessing of the option
             // information.
             foreach (IEgressProviderConfigurationProvider provider in _providers)
```
jander-msft, line 167: If we change the `IEgressProviderConfigurationProvider.ProviderType` property to `IEnumerable<string> ProviderTypes { get; }`, then the service configuration doesn't need to iterate the configuration settings. The `EgressService` then just needs to iterate the ProviderTypes in an extra loop and the remainder largely remains unchanged.

Then we can change the `Configuration` property into something like `IConfiguration GetConfiguration(string)` method that takes a provider name (not the provider type) and returns the associated configuration. This will allow the `EgressProviderConfigureNamedOptions<T>` to bind the options to the correct configuration section.

Finally, `EgressProviderConfigurationChangeTokenSource<T>` can be updated to build a [CompositeChangeToken](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.primitives.compositechangetoken?view=dotnet-plat-ext-6.0) from the iterative calls to `IEgressProviderConfigurationProvider.GetConfiguration`.

With those changes, I think everything higher in the stack (other than passing the provider name through IEgressProvider) won't be necessary.

src/Tools/dotnet-monitor/Egress/EgressProviderTypes.cs
```diff
@@ -8,6 +8,8 @@ internal static class EgressProviderTypes
     {
         public const string AzureBlobStorage = nameof(AzureBlobStorage);
 
+        public const string External = nameof(External);
```
jander-msft, line 11: This shouldn't be statically known since extensions will register their custom types via existence in the file system.

kelltrick:
[Replacing PR]

eng/common/cross/build-rootfs.sh
kelltrick: Removed

src/Extensions/azure-storage/azure-storage/Microsoft.Diagnostics.Monitoring.AzureStorage.csproj
kelltrick: Removed

PR #1797, opened 2022-04-25
Custom Shortcuts For Collection Rules

kkeirstead:
This PR allows users to define their own custom shortcuts that can be used to improve readability/reuse in configuration. By associating a name with an action/trigger/limit/filter, the name can be referenced alongside the existing style of configuration when defining a collection rule, reducing the need to copy-paste blocks of frequently used configuration between multiple rules.


documentation/collectionrules/customshortcuts.md
```diff
@@ -0,0 +1,294 @@
...
+    }
+  }
+  ```
+</details>
+
+<details>
+  <summary>Kubernetes ConfigMap</summary>
+
+  ```yaml
+  DotnetMonitor_CustomShortcuts__Actions__NameOfActionShortcut__Type: "CollectTrace"
```
wiktork, line 52: Remove DotnetMonitor_ prefix from all the yaml examples.

documentation/collectionrules/customshortcuts.md
```diff
@@ -0,0 +1,294 @@
...
+          "Settings": {
+            "RequestCount": 10,
+            "SlidingWindowDuration": "00:01:00"
+          }
+        }
+      },
+      "Filters": {
+        "AppPID": {
+          "Key": "ProcessId",
+          "Value": "12345",
```
wiktork, line 130: nit: Consider a process name example. I think it would be more useful as a shortcut.

src/Tools/dotnet-monitor/CollectionRules/Configuration/CustomShortcutsConfigureNamedOptions.cs
```diff
@@ -0,0 +1,224 @@
...
+            _configuration = configuration;
+            _logger = logger;
+        }
+
+        public void PostConfigure(string name, CollectionRuleOptions options)
+        {
+            // Action Shortcuts
+            foreach (var key in _shortcutOptions.CurrentValue.Actions.Keys)
+            {
+                IConfigurationSection section = _configuration.GetSection($"{nameof(RootOptions.CustomShortcuts)}:{nameof(CustomShortcutOptions.Actions)}:{key}");
```
wiktork, line 43: Formattable.Invariant.

wiktork, line 43: You can use ConfigurationPath.Combine to combine key values

src/Tools/dotnet-monitor/CollectionRules/Configuration/CustomShortcutsConfigureNamedOptions.cs
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+        private bool InsertCustomFiltersIntoFilterList(CollectionRuleOptions ruleOptions, string name)
+        {
+            bool boundAllFilters = false;
+            int index = 0;
+
+            while (!boundAllFilters)
+            {                
+                IConfigurationSection section = _configuration.GetSection(ConfigurationPath.Combine(nameof(RootOptions.CollectionRules), name, nameof(CollectionRuleOptions.Filters), index.ToString(CultureInfo.InvariantCulture)));
```
wiktork, line 135: Why not get all the children for the element and iterate through them instead of maintaining and testing an index?

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRuleConfigureNamedOptions.cs
```diff
@@ -57,14 +57,28 @@ private void BindActionSettings(IConfigurationSection ruleSection, CollectionRul
...
                 _actionOperations.TryCreateOptions(actionOptions.Type, out object actionSettings))
             {
-                IConfigurationSection settingsSection = ruleSection.GetSection(ConfigurationPath.Combine(
-                    nameof(CollectionRuleOptions.Actions),
-                    actionIndex.ToString(CultureInfo.InvariantCulture),
-                    nameof(CollectionRuleActionOptions.Settings)));
+                int configurationIndex = 0;
+                int actionsFound = 0;
+
+                while (actionsFound <= actionIndex)
```
wiktork, line 63: I am not really understanding this logic. Why would there be action elements that are empty?

src/Tools/dotnet-monitor/CollectionRules/Configuration/CustomShortcutsConfigureNamedOptions.cs
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+        private bool InsertCustomShortcut<T>(IDictionary<string, T> shortcutsOptions, string shortcutKey, List<T> options, int index) where T : class
+        {
+            if (!ValidateCustomShortcutExists(shortcutsOptions, shortcutKey))
+            {
+                return false;
+            }
+
+            options.Insert(index, shortcutsOptions[shortcutKey]);
```
wiktork, line 178: So does this mean that the bound action settings are adjacent to the original settings, rather than replacements? Would it be easier if we just replaced the originals?

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRuleConfigureNamedOptions.cs
kkeirstead: This logic is checking for whether the current action is a custom shortcut or a normal action. A custom shortcut will have a value set, whereas for a standard action the value will be empty, but it'll have children elements describing the different properties. So, when we check `actionSection.Exists() && string.IsNullOrEmpty(actionSection.Value)`, that's basically saying: is there anything in this section, and is it populated by children, or directly by a value. If it's populated by children, then we know it's a normal action, in which case we do the binding. If it's populated by a value (aka "MyCustomShortcut"), then we leave it alone, and we handle it in `CustomShortcutsConfigureNamedOptions`. I tried to mostly separate the new logic from the old, but maybe it's more confusing having these two separate files for configuring options.

src/Tools/dotnet-monitor/CollectionRules/Configuration/CustomShortcutsConfigureNamedOptions.cs
kkeirstead: IIRC, the `options` in this case only includes the standard actions. So, if I had four actions (normal, custom, normal, custom), the `options` would have a length of 2. I can't just stick the custom actions on the end, or else we'd have ordering problems, so it would do: (normal, normal) -> (normal, custom, normal) ... and ... (normal, custom, normal) -> (normal, custom, normal, custom). I could be mis-remembering though

wiktork: Would it be reasonable to change CollectionRuleOptions.Actions to include all the actions and then substitute accordingly? We can validate that if a user specifies an action with no shortcut, it fails.

kkeirstead: This wasn't an intentional decision (more-so just something I discovered organically, and then worked around) - I'm assuming that since `Actions` expects `List<CollectionRuleActionOptions`, and not a list that includes strings, that we can't get the custom shortcuts as part of that list...unless we could force it to just create empty actions as placeholders.

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRuleConfigureNamedOptions.cs
kkeirstead: I'm about to put up a cleaned-up version of this (and the other file) that uses your idea with the children - you can let me know if that feels any better to you, and then we can go from there.

src/Tools/dotnet-monitor/CollectionRules/Configuration/CustomShortcutsConfigureNamedOptions.cs
```diff
@@ -0,0 +1,214 @@
...
+            ILogger<CustomShortcutsConfigureOptions> logger)
+        {
+            _triggerOperations = triggerOperations;
+            _actionOperations = actionOperations;
+            _shortcutOptions = shortcutOptions;
+            _configuration = configuration;
+            _logger = logger;
+        }
+
+        public void PostConfigure(string name, CollectionRuleOptions options)
```
jander-msft, line 40: I don't think the concerns are separated correctly. I would expect that there is an `IPostConfigureOptions<CustomShortcutOptions>` implementation that effectively does the same thing as https://github.com/dotnet/dotnet-monitor/blob/main/src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRuleConfigureNamedOptions.cs where it binds the settings for each action and trigger. Then this `IPostConfigureOptions<CollectionRuleOptions>` can just assign the filters, trigger, actions, limits wholesale without have to bind anything itself.

src/Tools/dotnet-monitor/ConfigurationKeys.cs
```diff
@@ -29,5 +29,7 @@ internal static class ConfigurationKeys
         public const string GlobalCounter = nameof(RootOptions.GlobalCounter);
 
         public const string CollectionRuleDefaults = nameof(RootOptions.CollectionRuleDefaults);
+
+        public const string CustomShortcuts = nameof(RootOptions.CustomShortcuts);
```
jander-msft, line 33: I know that when we've discussed this feature, we've called it "custom shortcuts", but I'm not sure this is the best name for it. First, it's too generic, because by the name, you have no idea what it applies to and when you think of shortcut in general terms, a shortcut is something that takes you from one place to another via a shorter route. Additionally, outside of *.lnk files used as shortcuts to other files on Windows, I'm not sure I've heard "shortcut" used elsewhere in computing.

Consider an alternate name that is more consistent with its usage and context, such as "CollectionRuleTemplates".

kkeirstead: I would agree, but just for branding consistency - does that mean we should change the name of "Built-In Shortcuts"? That PR is already checked in (but hasn't shipped), so we could do that - unless you only wanted to change the naming of this one.

src/Tools/dotnet-monitor/CollectionRules/Configuration/CustomShortcutsConfigureNamedOptions.cs
kkeirstead: So, I just pushed changes that I _think_ hopefully reflects what you're getting at; I didn't do a lot of testing, but it isn't a huge change and it passes the tests (and a small amount of manual testing). Let me know if I misinterpreted what you had in mind.

src/Tools/dotnet-monitor/CollectionRules/Configuration/CustomShortcutsConfigureNamedOptions.cs
```diff
@@ -0,0 +1,91 @@
...
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Actions;
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options;
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Triggers;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Configuration
+{
+    internal sealed class CustomShortcutsConfigureOptions :
+        IConfigureOptions<CustomShortcutOptions>
```
jander-msft, line 14: This should be an IPostConfigureOptions<T>. You cannot assume in an IConfigureOptions<T> implementation that your instance happens to run after the first instance that populates the options from binding; that is the purpose of IPostConfigureOptions, to change the options after they've been populated from their initial source(s).

src/Tools/dotnet-monitor/CollectionRules/Configuration/CustomShortcutsConfigureNamedOptions.cs
```diff
@@ -0,0 +1,91 @@
...
+        public void Configure(CustomShortcutOptions options)
+        {
+            // Action Shortcuts
+            foreach (var key in options.Actions.Keys)
+            {
+                IConfigurationSection section = _configuration.GetSection(ConfigurationPath.Combine(nameof(RootOptions.CustomShortcuts), nameof(CustomShortcutOptions.Actions), key));
+
+                if (section.Exists())
+                {
+                    section.Bind(options);
```
jander-msft, line 39: `options` is of type `CustomShortcutOptions`. Here, you are binding this options instance to the "CustomShortcuts:Actions:<key>" section, which I expect to be representative of a `CollectionRuleActionOptions` instance. As is, I think this code is not doing any under normal circumstances; under misconfigured circumstances, I think it will lead to unexpected behavior.

src/Tools/dotnet-monitor/CollectionRules/Configuration/CustomShortcutsConfigureNamedOptions.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        private void BindCustomActions(IConfigurationSection ruleSection, CustomShortcutOptions shortcutOptions, string shortcutName)
+        {
+            CollectionRuleActionOptions actionOptions = shortcutOptions.Actions[shortcutName];
+
+            if (null != actionOptions &&
+                _actionOperations.TryCreateOptions(actionOptions.Type, out object actionSettings))
+            {
+                IConfigurationSection settingsSection = ruleSection.GetSection(ConfigurationPath.Combine(
```
jander-msft, line 66: no need to call `ConfigurationPath.Combine` if the path is only one segment.

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRulePostConfigureNamedOptions.cs
```diff
@@ -0,0 +1,147 @@
...
+            IOptionsMonitor<CustomShortcutOptions> shortcutOptions,
+            IConfiguration configuration,
+            ILogger<CollectionRulePostConfigureOptions> logger)
+        {
+            _shortcutOptions = shortcutOptions;
+            _configuration = configuration;
+            _logger = logger;
+        }
+
+        public void PostConfigure(string name, CollectionRuleOptions options)
```
jander-msft, line 31: This is doing some complicated math and relying on configuration binding to not bind if the trigger/action/etc is not fully specified. This can be simplified by doing something like:
- In `CollectionRuleConfigureNamedOptions`, remove the `BindTriggerSettings` and `BindActionSettings` methods/calls and move them to this class. That class will only bind the `CollectionRuleOptions` instance to the collection rule configuration section.
- Instead of calculating where something was fulfilled or not, just clear each of the lists on `CollectionRuleOptions`. For each child section in each list section, either (1) if the section doesn't have a body, then assign the body from corresponding shortcut options, or (2) bind and call `BindXSettings` method as usual.

There's no math involved in determining the order of items or where to insert new items.

src/Tools/dotnet-monitor/CollectionRules/Configuration/CustomShortcutsConfigureNamedOptions.cs
kkeirstead: Related question (which was my original logic behind doing it this way) - can we assume that `CustomShortcutsPostConfigureNamedOptions` would run before `CollectionRulePostConfigureNamedOptions`?

documentation/collectionrules/customshortcuts.md
```diff
@@ -0,0 +1,294 @@
...
+Custom Shortcuts associate a name with a single Filter, Trigger, Action, or Limit; this name can then be used throughout configuration to represent the use of that Filter/Trigger/Action/Limit. This is ideal for scenarios where multiple collection rules re-use the same functionality, allowing the author to write/edit the configuration in a single place.
+
+You can easily translate existing configuration to Custom Shortcuts using the format in the following sample; once defined, the Custom Shortcut is referenced by its name in the collection rule.
+
+<details>
+  <summary>JSON</summary>
+
+  ```json
+  {
+    "CustomShortcuts": {
```
wiktork, line 14: Templates

documentation/collectionrules/customshortcuts.md
```diff
@@ -0,0 +1,294 @@
...
+      "Actions": {
```
wiktork, line 15: CollectionRule*

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRuleConfigureNamedOptions.cs
```diff
@@ -57,13 +57,16 @@ private void BindActionSettings(IConfigurationSection ruleSection, CollectionRul
...
-                IConfigurationSection settingsSection = ruleSection.GetSection(ConfigurationPath.Combine(
-                    nameof(CollectionRuleOptions.Actions),
-                    actionIndex.ToString(CultureInfo.InvariantCulture),
-                    nameof(CollectionRuleActionOptions.Settings)));
+                var section = ruleSection.GetSection(nameof(CollectionRuleOptions.Actions));
 
-                settingsSection.Bind(actionSettings);
+                section.GetChildren().Where(child => string.IsNullOrEmpty(child.Value)); // Doesn't include string values for custom shortcuts
+
+                var actionSections = section.GetChildren().Where(child => string.IsNullOrEmpty(child.Value));
```
wiktork, line 64: nit: var to actual Type

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExpectedShowSourcesConfigurations/CustomShortcuts.json
```diff
@@ -0,0 +1,40 @@
+{
+  "Actions": {
```
wiktork, line 2: Per @kelltrick Use jsonc instead of json.

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRulePostConfigureNamedOptions.cs
kkeirstead: Without preserving its contents, are we sure we could actually clear out the lists on `CollectionRuleOptions`? If we throw away the `CollectionRuleActionOptions`, when we try to `BindActionSettings` it'll try to get `actionOptions.Type` on the current action - which, if we throw them all away and just fill them in as we go, means we wouldn't have a type here. Or am I misunderstanding this part?

jander-msft: You can clear it out and rebind it when necessary. The existing `CollectionRuleConfigureNamedOptions` already knows how to bind the `CollectionRuleOptions` without any help; the `CollectionRulePostConfigureNamedOptions` has the same context, so it can recreate it.

The code would look something like:
```cs
options.Actions.Clear();

IConfigurationSection actionsSection = ruleSection.GetSection(nameof(CollectionRuleOptions.Actions));
if (actionsSection.Exists())
{
    foreach (IConfigurationSection actionSection in actionsSection.GetChildren())
    {
        // actionSection.Key is the index of the action e.g. "0", "1", "2"
        // actionSection.Value is null (if specified inline) or non-empty string if a simple name
        if (string.IsNullOrEmpty(actionSection.Value))
        {
            // Bind action options from inline configuration
            CollectionRuleActionOptions actionOptions = new();

            actionSection.Bind(actionOptions);

            BindActionSettings(actionSection, actionOptions);

            options.Actions.Add(actionOptions);
        }
        else
        {
            // Get action options from template
            // options.Actions.Add(actionOptionsFromTemplate);
        }
    }
}

...

private void BindActionSettings(IConfigurationSection actionSection, CollectionRuleActionOptions actionOptions)
{
    if (null != actionOptions &&
        _actionOperations.TryCreateOptions(actionOptions.Type, out object actionSettings))
    {
        IConfigurationSection settingsSection = actionSection.GetSection(nameof(CollectionRuleActionOptions.Settings));

        settingsSection.Bind(actionSettings);

        actionOptions.Settings = actionSettings;
    }
}

```

kkeirstead: Gotcha, that makes sense - I had been saving a copy of the list as a queue before clearing it, and then just dequeuing instead of creating a new `CollectionRuleActionOptions` and binding it to `actionSection`, but I'll switch to this approach.

documentation/collectionrules/customshortcuts.md
jander-msft: There really isn't any template-y about the "built-in shortcuts". Those are just other triggers that happen to be event counter based. A template is typically something where you take the content as-is and do some token replacement, if any.

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRulePostConfigureNamedOptions.cs
```diff
@@ -0,0 +1,192 @@
...
+
+        private bool ResolveActionList(CollectionRuleOptions ruleOptions, string name)
+        {
+            ruleOptions.Actions.Clear();
+
+            IConfigurationSection actionsSection = _configuration.GetSection(ConfigurationPath.Combine(nameof(RootOptions.CollectionRules), name, nameof(CollectionRuleOptions.Actions)));
+
+            foreach (IConfigurationSection actionSection in actionsSection.GetChildren())
+            {
+                if (!string.IsNullOrEmpty(actionSection.Value))
```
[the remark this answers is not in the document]
jander-msft: Please add this comment to all usage of `string.IsNullOrEmpty(section.Value)`. Better yet, you can add a method that does this determination, name it something like `SectionHasValue(IConfigurationSection section)` and add the comments there in one place.

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRulePostConfigureNamedOptions.cs
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        private bool TryGetTemplate<T>(IDictionary<string, T> templatesOptions, string templateKey, out T templatesValue)
+        {
+            if (templatesOptions.TryGetValue(templateKey, out templatesValue))
+            {
+                return true;
+            }
+
+            _logger.LogError(Strings.ErrorMessage_TemplateNotFound, templateKey);
```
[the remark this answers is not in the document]
jander-msft: I'm fine with adding a validation list at the `CollectionRuleOptions` level. The one downfall is that you lose the natural association between each error and the thing that the error applies to. But I'm not sure that's terribly important and I'm not sure the validation system actually reports that hierarchy anyway. So I think this is a good place to start.

In terms of if/when we have more template types, I think it'll have to be the individual system's responsibility to figure out where to store that information, so long as it fits the pattern that `IConfigureOptions`/`IPostConfigureOptions` is about configurating and `IValidateOptions` is about reporting the errors.

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRulePostConfigureNamedOptions.cs
```diff
@@ -76,117 +64,79 @@ private bool ResolveActionList(CollectionRuleOptions ruleOptions, string name)
...
-            if (null != triggerOptions &&
-                _triggerOperations.TryCreateOptions(triggerOptions.Type, out object triggerSettings))
-            {
-                IConfigurationSection settingsSection = triggerSection.GetSection(nameof(CollectionRuleTriggerOptions.Settings));
-
-                settingsSection.Bind(triggerSettings);
-
-                triggerOptions.Settings = triggerSettings;
-            }
+            // If the section has a value, the value is the name of a template.
```
jander-msft, line 138: Isn't it the opposite though? "If the section does not have a value, then the key is the name of the template"

kkeirstead: Actually, I think the comment is right, but I just have two inversions cancelling each other out. It should be `!string.IsNullOrEmpty(section.Value)`, and then the caller should be `SectionHasValue(section)`

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRuleBindingHelper.cs
```diff
@@ -0,0 +1,40 @@
...
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Actions;
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options;
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Triggers;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Configuration
+{
+    internal class CollectionRuleBindingHelper
+    {
+        internal static void BindActionSettings(IConfigurationSection actionSection, CollectionRuleActionOptions actionOptions, ICollectionRuleActionOperations actionOperations)
```
wiktork, line 14: public static

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleOptions.Validate.cs
```diff
@@ -15,6 +15,13 @@ IEnumerable<ValidationResult> IValidatableObject.Validate(ValidationContext vali
         {
             List<ValidationResult> results = new();
 
+            results.AddRange(ErrorList);
+
+            if (results.Count > 0)
+            {
+                return results; // QUESTION: Do we want to short-circuit here? Invalid template names may cause other error messages that distract from the actual issue (e.g. saying the Type field is required on a trigger when the trigger template wasn't found)
```
wiktork, line 22: I think as long as we show the error that the template is invalid, further validation makes no sense

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleOptions.Validate.cs
```diff
@@ -15,6 +15,13 @@ IEnumerable<ValidationResult> IValidatableObject.Validate(ValidationContext vali
         {
             List<ValidationResult> results = new();
 
+            results.AddRange(ErrorList);
```
wiktork, line 18: Can you add some comments regarding how the ErrorList object gets populated?

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRulePostConfigureNamedOptions.cs
```diff
@@ -0,0 +1,142 @@
...
+                ruleOptions.Limits = limitsOptions;
+            }
+        }
+
+        private void TryGetTemplate<T>(CollectionRuleOptions ruleOptions, IDictionary<string, T> templatesOptions, string templateKey, out T templatesValue) where T : new()
+        {
+            if (!templatesOptions.TryGetValue(templateKey, out templatesValue))
+            {
+                templatesValue = new();
+                ruleOptions.ErrorList.Add(new ValidationResult(string.Format(CultureInfo.CurrentCulture, Strings.ErrorMessage_TemplateNotFound, templateKey)));
```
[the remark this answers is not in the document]
kkeirstead: Previously this was returning a bool, but with the updated error-handling it wasn't being used so I switched it to void. That being said, there's probably no harm in returning a bool even if it isn't being used, so I can change that back.

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRulePostConfigureNamedOptions.cs
```diff
@@ -0,0 +1,142 @@
...
+        {
+            ruleOptions.Actions.Clear();
+
+            IConfigurationSection actionSections = ruleSection.GetSection(nameof(CollectionRuleOptions.Actions));
+
+            foreach (IConfigurationSection actionSection in actionSections.GetChildren())
+            {
+                if (SectionHasValue(actionSection))
+                {
+                    TryGetTemplate(ruleOptions, _templateOptions.CollectionRuleActions, actionSection.Value, out CollectionRuleActionOptions templateActionOptions);
```
wiktork, line 57: Test if TryGetTemplate succeeded and only add the result if it does.

wiktork (dismissed):
I left some feedback, but looks good otherwise.

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRulePostConfigureNamedOptions.cs
```diff
@@ -0,0 +1,142 @@
...
+
+            ResolveActionList(options, ruleSection);
+            ResolveTrigger(options, ruleSection);
+            ResolveFilterList(options, ruleSection);
+            ResolveLimits(options, ruleSection);
+        }
+
+        private void ResolveActionList(CollectionRuleOptions ruleOptions, IConfigurationSection ruleSection)
+        {
+            ruleOptions.Actions.Clear();
```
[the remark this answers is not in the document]
kkeirstead: We're clearing out the action list to avoid having to insert actions at the right spots - Justin wanted to avoid having to figure out the right index for where to put templates to maintain ordering, so we just re-populate the full set of actions from scrach here, regardless of whether or not they're templates.

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleOptions.Validate.cs
kkeirstead: My thinking on this is that it makes more sense to short-circuit than to show error-messages that might be misleading (as the comment suggests). Technically, we could probably add in the logic to recognize that, for example, if the trigger template failed, we don't do the validation on the trigger (but to do the other pieces), but I also don't think we want/need to be adding additional complexity just for this. We do already have a precedent for this (kind of) - in `CollectionRuleOptionsTests`, there's a comment noting that validation short-circuits before checking `GreaterThan` and `LessThan` for `EventCounterTrigger`. Once the user has fixed their templates, they'll still see the errors - fixing config just might require an extra iteration, which doesn't _seem_ unreasonable.

src/Tools/dotnet-monitor/CollectionRules/Configuration/CollectionRulePostConfigureNamedOptions.cs
kkeirstead: I believe the issue with doing this (specifically in the case of the trigger) is that we will then fail `DataAnnotationValidateOptions` with "The Trigger field is required", not the template error. In the current design, since we always set it, it gets put to the `new()` (or default) value, and then fails due to the template issue in the `ErrorList`.

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleOptions.Validate.cs
jander-msft: As a user, it annoys me when VS tells me "you have 3 errors", I fix them, and then it says "you now have 10 errors", fix them, etc, until some indeterminate point in the future when I have no errors left. I would understand that if fixing some error uncovers other related errors, but errors that are independent of each other should be reported at the same time.

This is also something that could be fixed later.

src/Tools/dotnet-monitor/ServiceCollectionExtensions.cs
```diff
@@ -102,6 +107,9 @@ public static IServiceCollection ConfigureCollectionRules(this IServiceCollectio
             services.AddSingleton<ICollectionRuleActionOperations, CollectionRuleActionOperations>();
             services.AddSingleton<ICollectionRuleTriggerOperations, CollectionRuleTriggerOperations>();
 
+            services.AddSingleton<IPostConfigureOptions<TemplateOptions>, TemplatesPostConfigureOptions>();
+            services.AddSingleton<IPostConfigureOptions<CollectionRuleOptions>, CollectionRulePostConfigureOptions>();
```
jander-msft, line 111: nit: I think it makes sense to move this to between lines 113 and 114 so that all of the `CollectionRuleOptions` options configuration implementations are grouped together. You'd also have to move the `DefaultCollectionRulePostConfigureOptions` registration to this group as well to make this true.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfigurationTests.cs
```diff
@@ -339,9 +340,9 @@ private string CleanWhitespace(string rawText)
             return string.Concat(rawText.Where(c => !char.IsWhiteSpace(c)));
         }
 
-        private string ConstructSettingsJson(params string[] permittedFileNames)
+        internal static string ConstructSettingsJson(string sampleConfigurationsDirectory, params string[] permittedFileNames)
```
jander-msft, line 343: The changes related to adding this parameter seem like they can be undone.

kkeirstead: This is also being used in `TemplatesTests`; I originally had just included the template testing directly in this file, but decided it made sense to separate it as its own thing. If we don't like having another test file accessing this method within `ConfigurationTests`, I could break it out into a helper class - I wasn't sure if that was considered necessary, but given that it wasn't obvious that this is being used elsewhere, I suspect that I should.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/TemplatesTests.cs
```diff
@@ -0,0 +1,137 @@
...
+                string[] failures = ex.Failures.ToArray();
+                Assert.Equal(2, failures.Length);
+
+                Assert.Equal(string.Format(TemplateNotFoundErrorMessage, "TriggerTemplateINVALID"), failures[0]);
+                Assert.Equal(string.Format(TemplateNotFoundErrorMessage, "FilterTemplateINVALID"), failures[1]);
+
+            }, builder: (HostBuilder)builder);
+        }
+
+        private IHostBuilder GetHostBuilder(TemporaryDirectory userConfigDir)
```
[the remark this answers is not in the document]
kkeirstead: I guess I'm not sure what the process is for converting the current approach to using `overrideSource`, but what's the downside of the current approach (other than adding another parameter)? I like that it currently follows the pattern we use in `ConfigurationTests`, but it doesn't need to if there's a reason not to.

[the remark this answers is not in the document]
jander-msft: > I guess I'm not sure what the process is for converting the current approach to using `overrideSource`, but what's the downside of the current approach (other than adding another parameter)?

You are trying to add a new parameter that has an already existing solution that shouldn't be painful to use. If there is no actual benefit for the alternative, then the alternatives shouldn't be added.

For how to use `overrideSource`, you can try (1) deserialize the JSON doc, traverse it and flatten it into a dictionary (key is configuration key path, value is value), or (2) refactor `overrideSource` to be an `IConfigurationSource` that is just added rather than calling AddInMemoryCollection, change the existing usage to pass `MemoryConfigurationSource`, and have your new usage pass a `JsonConfigurationSource` or `JsonStreamConfigurationSource` (the latter should let you use a MemoryStream over your JSON data rather than having to write it disk).

jander-msft (approved):

PR #1801, opened 2022-04-25
Merged main into Release/7.x for Preview 4

kkeirstead:


kkeirstead:
I also did not touch the `BlobGroupBuildQuality`; it was set to `release` by default, but it wasn't clear from the docs when it should be `prerelease` vs `release`.

jander-msft:
> I also did not touch the `BlobGroupBuildQuality`; it was set to `release` by default, but it wasn't clear from the docs when it should be `prerelease` vs `release`.

Keeping it at `release` is fine. I'll make a change to remove `prerelease` as an option since it really isn't necessary. We just need something to differentiate daily builds from release candidate builds.

jander-msft:
> For the version of `Microsoft.Diagnostics.Monitoring`, I used the version that I updated to earlier today in #1798 (there had been a merge conflict with main, which used a preview version).

There will always be a merge conflict regarding these packages because `main` gets daily-ish prerelease versions of these packages.

> Additionally, I was confused about the version number itself (v5.0.320703 instead of v6.0.320703)

As of right now, it should be `5.0.X`. We can probably change it to float with the rest of the dotnet/diagnostics repo to avoid confusion.

kkeirstead:
See #1806 - the commit history was squashed on this PR, so I replaced it with another one.

Issue #1635, opened 2022-03-26
Extendable Egress


kelltrick:
We are designing the ability to provide custom egress providers to dotnet-monitor to allow egress of artifacts to other cloud providers or custom endpoints.

jander-msft:
Egress extensibility has been available since .NET Monitor 8. See documentation for details: https://github.com/dotnet/dotnet-monitor/blob/main/documentation/egress.md

PR #1812, opened 2022-04-26
Add configuration section for Egress Extensibility

kelltrick:
- Adds the configuration system for egress extensibility
Replaces PR #1791 
Initial work on #1635


src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressPayload.cs
```diff
@@ -0,0 +1,16 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Egress
+{
+    public class ExtensionEgressPayload
+    {
+        public EgressArtifactSettings Settings { get; set; }
+        public Dictionary<string, string> Properties { get; set; }
```
kelltrick, line 12: `IDictionary`

kelltrick: Done.

src/Tools/dotnet-monitor/Egress/Configuration/EgressProviderConfigurationProvider.cs
```diff
@@ -14,19 +16,45 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Egress.Configuration
...
         {
-            ProviderType = providerType;
-            Configuration = configuration.GetEgressSection().GetSection(providerType);
+            _optionsMapper  = optionsMapper;
+            _egressSection = configuration.GetEgressSection();
         }
 
         /// <inheritdoc/>
-        public IConfiguration Configuration { get; }
+        public Type OptionsType => typeof(TOptions);
```
wiktork, line 29: Since we've unified the Options type and now reference all the categories here, does it even make sense to have this generic parameter anymore?

wiktork: Ah I see these can still be the old azure/filesystem types as well. Can you add comments to clarify this?

src/Tools/dotnet-monitor/Egress/EgressProviderInternal.cs
```diff
@@ -20,7 +21,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Egress
     /// </summary>
     internal class EgressProviderInternal<TOptions> :
         IEgressProviderInternal<TOptions>
-        where TOptions : class
+        where TOptions : new()
```
wiktork, line 24: Why the constraint change?

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            ExtensionEgressPayload payload = new ExtensionEgressPayload()
+            {
+                Settings = artifactSettings,
+                Configuration = options,
+                Properties = props,
+                ProfileName = providerName,
+            };
+
+            string startingPayload = JsonSerializer.Serialize<ExtensionEgressPayload>(payload);
```
wiktork, line 62: Consider JsonSerializer.SerializeAsync directly against the BaseStream, rather than a string representation first.

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            // [TODO] add a new service to dynamically find these extensions
+            const string extensionProcessPath = "MyPathTo\\Extension.exe";
+            ProcessStartInfo pStart = new ProcessStartInfo()
+            {
+                FileName = extensionProcessPath,
```
wiktork, line 68: I would be explicit about parameters such as CreateNoWindow and UseShellExecute even if they have reasonable defaults on net core.

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+            await action(p.StandardInput.BaseStream, token);
+            p.StandardInput.BaseStream.Flush();
+
+            Logger?.LogInformation("Stream done, sending close...");
+
+            p.StandardInput.Close();
+
+            string procContent = await p.StandardOutput.ReadToEndAsync();
+
+            Logger?.LogInformation("Waiting for exit...");
```
wiktork, line 91: Log standard error as well?

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+            p.WaitForExit();
```
wiktork, line 92: I think we should have some guards here and timeout after a certain point.

wiktork (approved):

src/Tools/dotnet-monitor/Egress/Configuration/EgressProviderConfigurationProvider.cs
kelltrick: I need to update all the documentation comments for `IEgressProviderConfigurationProvider`.. let me go do that now.

src/Tools/dotnet-monitor/Egress/EgressProviderInternal.cs
kelltrick: This is required because the manual enumeration of `TOptions` requires the class to make a new instance of `TOptions` to bind to. Normally this was done via reflection by the IConfiguraitonOptions internal goo, but adding this constraint makes it easy to do manually.

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
kelltrick: I've got a new chageset that re-writes all of this, so I'll leave this part to the new revision.

src/Tools/dotnet-monitor/Egress/Configuration/IEgressProviderConfigurationProvider.cs
```diff
@@ -4,33 +4,43 @@
 
 using Microsoft.Extensions.Configuration;
 using System;
+using System.Collections.Generic;
 
 namespace Microsoft.Diagnostics.Tools.Monitor.Egress.Configuration
 {
     /// <summary>
-    /// Provides access to an Egress:{ProviderType} section of the configuration.
+    /// Provides access to all of the egress provider configuraiton blocks that use the given <see cref="OptionsType"/>.
```
wiktork, line 12: configuration

src/Tools/dotnet-monitor/Egress/Configuration/IEgressProviderConfigurationProvider.cs
```diff
@@ -4,33 +4,43 @@
...
+    /// These configuration blocks will be from "Egress:{ProviderType}" where "{ProviderType}" is one of the strings definded in <see cref="ProviderTypes"/>.
     /// </summary>
     internal interface IEgressProviderConfigurationProvider
     {
         /// <summary>
-        /// The configuration section associated with the egress provider.
+        /// The name of the categories defined in configuration of this options type. These keys can be passed to <see cref="GetConfigurationSection(string)"/> to get the specific configuration section.
```
wiktork, line 18: types instead of categories? We should be consistent in the naming throughout the code and comments.

src/Tools/dotnet-monitor/Egress/Configuration/IEgressProviderConfigurationProvider.cs
```diff
@@ -4,33 +4,43 @@
...
-        /// The type of the options class associated with the egress provider.
+        /// Gets the <see cref="IConfigurationSection"/> associated with the given <paramref name="providerType"/>.
         /// </summary>
-        Type OptionsType { get; }
+        /// <param name="providerType">The provider type, this is element under "Egress:" in configuration. You can use <see cref="ProviderTypes"/> to get valid strings to use here.</param>
+        IConfigurationSection GetConfigurationSection(string providerType);
+
+        /// <summary>
+        /// The configuration section that should be monitored for changes. This will return the minimum set of <see cref="IConfigurationSection"/> to monitor for changes.
+        /// Note: a change in this section does not guarentee a meaningful change to configuration.
```
wiktork, line 35: guarantee

src/Tools/dotnet-monitor/Egress/Configuration/IEgressProviderConfigurationProvider.cs
```diff
@@ -4,33 +4,43 @@
...
         /// </summary>
-        string ProviderType { get; }
+        Type OptionsType { get; }
 
         /// <summary>
-        /// The type of the options class associated with the egress provider.
+        /// Gets the <see cref="IConfigurationSection"/> associated with the given <paramref name="providerType"/>.
         /// </summary>
-        Type OptionsType { get; }
+        /// <param name="providerType">The provider type, this is element under "Egress:" in configuration. You can use <see cref="ProviderTypes"/> to get valid strings to use here.</param>
```
wiktork, line 30: the element

src/Tools/dotnet-monitor/Egress/EgressService.cs
```diff
@@ -84,23 +94,34 @@ public async Task<EgressResult> EgressAsync(string providerName, Func<Stream, Ca
...
-            if (!_providerTypeMap.TryGetValue(providerName, out string providerType))
+            if (!_providerNameToTypeMap.TryGetValue(providerName, out string providerType))
             {
                 throw new EgressException(string.Format(CultureInfo.CurrentCulture, Strings.ErrorMessage_EgressProviderDoesNotExist, providerName));
             }
+            return providerType;
+        }
 
-            if (!_optionsTypeMap.TryGetValue(providerType, out Type optionsType))
+        private IEgressProviderConfigurationProvider GetConfigProvider(string providerType)
```
jander-msft, line 106: I think this method and `GetProvider` can be merged into one again since they are not used separately.

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressPayload.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Egress
+{
+    public class ExtensionEgressPayload
+    {
+        public EgressArtifactSettings Settings { get; set; }
+        public IDictionary<string, string> Properties { get; set; }
+        public IDictionary<string, string> Configuration { get; set; }
+        public string ProfileName { get; set; }
```
jander-msft, line 14: `ProviderName` to keep things consistent. In future changes, I would imagine that we'll refer to the fact that egress providers have a name and the word "profile" is not going to appear in any documentation or implementation other than here.

src/Tools/dotnet-monitor/Egress/Configuration/EgressPropertiesProvider.cs
```diff
@@ -19,6 +21,12 @@ public EgressPropertiesProvider(IEgressPropertiesConfigurationProvider provider)
             _provider = provider;
         }
 
+        /// <inheritdoc/>
+        public IEnumerable<string> GetKeys()
```
jander-msft, line 25: Maybe just return a `Dictionary<string, string>` that's already populated all of the values. The current usage in your changes seems that it is getting the keys and then one-by-one getting values via `TryGetPropertyValue`. I think evolution of egress will eliminate the need for `TryGetPropertyValue` because dotnet monitor itself will not need to pull out individual values.

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
```diff
@@ -0,0 +1,107 @@
...
+            Process p = new Process()
+            {
+                StartInfo = pStart,
+            };
+
+            Logger?.LogInformation("Starting process...");
+            p.Start();
+            Logger?.LogInformation("starting stream...");
+
+            await JsonSerializer.SerializeAsync<ExtensionEgressPayload>(p.StandardInput.BaseStream, payload, options: null, token);
```
jander-msft, line 82: Maybe flush just after serialization so that the json payload is immediately available to the egress process in case the artifact isn't written to the stream for a while. That will allow the egress process to examine the settings and set up its implementation before necessarily having to wait for some portion of the artifact to be written.

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
```diff
@@ -0,0 +1,107 @@
...
+
+            await action(p.StandardInput.BaseStream, token);
```
jander-msft, line 84: Shouldn't a newline be written to the stream between the json payload and the artifact payload?

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
```diff
@@ -0,0 +1,107 @@
...
+                RedirectStandardError = true,
+                CreateNoWindow = true,
+                UseShellExecute = false,
+            };
+            Process p = new Process()
+            {
+                StartInfo = pStart,
+            };
+
+            Logger?.LogInformation("Starting process...");
```
jander-msft, line 78: All of these logger statements (for whichever ones you want to keep) need to be moved to LoggingExtensions.

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
```diff
@@ -0,0 +1,107 @@
...
+            Logger?.LogInformation("Stream done, sending close...");
+
+            p.StandardInput.Close();
+
+            Logger?.LogInformation("Waiting for exit...");
+
+            string procStdOutContent = await p.StandardOutput.ReadToEndAsync();
+            string procErrorContent = await p.StandardError.ReadToEndAsync();
+
+            p.WaitForExit();
```
jander-msft, line 96: The tool should asynchronously wait for the process to finish rather than blocking the current thread. See [ExecuteAction](https://github.com/dotnet/dotnet-monitor/blob/main/src/Tools/dotnet-monitor/CollectionRules/Actions/ExecuteAction.cs#L44) for an example of how to do this.

src/Tools/dotnet-monitor/Egress/EgressProviderInternal.cs
jander-msft: I don't see why this is change is necessary here.

jander-msft (changes requested):

src/Tools/dotnet-monitor/Egress/EgressProviderInternal.cs
```diff
@@ -68,7 +75,8 @@ private TOptions GetOptions(string providerName)
         {
             try
             {
-                return _monitor.Get(providerName);
+                TOptions opts = _monitor.Get(providerName);
```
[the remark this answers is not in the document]
kelltrick: This complies to the same code but makes it easier to debug this code, so I'd like to leave it.

src/Tools/dotnet-monitor/Egress/EgressProviderInternal.cs
kelltrick: Ahh, this isn't required now that `IEgressProviderConfigurationProvider` doesn't do the `TOptions` creation anymore.

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
kelltrick: Yep, I changed this from Wiktor's feedback and dropped the newline

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
kelltrick: I'm in the process of re-writing this whole thing, these were just for prototyping. This is all going to be encapsulated in an `IExtension` class.

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
kelltrick: Added to Todos

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
kelltrick: Added to Todos

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
jander-msft: `TextWriter.WriteLineAsync` uses `Environment.NewLine`, which is different on Windows vs non-Windows (see [test](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Runtime.Extensions/tests/System/Environment.NewLine.cs) validating this). Because this is a protocol that you are trying to establish, I don't think the "newline" notion should depend on the operating system. For example, `\r\n` is always used in HTTP newlines.

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+            {
+                StartInfo = pStart,
+            };
+
+            Logger?.LogInformation("Starting process...");
+            p.Start();
+            Logger?.LogInformation("starting stream...");
+
+            await JsonSerializer.SerializeAsync<ExtensionEgressPayload>(p.StandardInput.BaseStream, payload, options: null, token);
+            await p.StandardInput.WriteLineAsync();
```
jander-msft, line 79: Do any of these async writing methods taken cancellation tokens? If so, then please pass `token`.

src/Tools/dotnet-monitor/Egress/Configuration/EgressPropertiesProvider.cs
```diff
@@ -19,6 +21,20 @@ public EgressPropertiesProvider(IEgressPropertiesConfigurationProvider provider)
             _provider = provider;
         }
 
+        /// <inheritdoc/>
+        public IDictionary<string, string> GetAllProperties()
+        {
+            Dictionary<string, string> properties = new Dictionary<string, string>();
+            foreach (IConfigurationSection section in _provider.Configuration.GetChildren())
+            {
+                if (section.Exists())
```
jander-msft, line 30: nit: Do you need to check that it exists if you just got it from `GetChildren` method?

src/Tools/dotnet-monitor/Egress/EgressArtifactSettings.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Microsoft.Diagnostics.Tools.Monitor.Egress
 {
-    internal sealed class EgressArtifactSettings
+    public sealed class EgressArtifactSettings
```
jander-msft, line 6: Does this need to be public?

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            await p.WaitForExitAsync(token);
+            Logger?.LogInformation($"Exited with code: {p.ExitCode}");
+
+            if (!string.IsNullOrWhiteSpace(procErrorContent))
+            {
+                Logger?.LogWarning($"Process Error output: {procErrorContent}");
+            }
+
+            return procStdOutContent;
```
jander-msft, line 102: Please add a TODO comment that this need to be the egressed path. You might have to devise a standards-based protocol for stdout to report the path and a success value, and possibly logging at various levels too.

src/Tools/dotnet-monitor/ServiceCollectionExtensions.cs
```diff
@@ -202,22 +208,23 @@ public static IServiceCollection ConfigureDiagnosticPort(this IServiceCollection
...
+            where TOptions : class, new()
         {
+            // These Singletons are "IEnumerable<T>" services where there are multiple services registered (if TOptions is the same)
             // Add services to provide raw configuration for the options type
-            services.AddSingleton(sp => new EgressProviderConfigurationProvider<TOptions>(sp.GetRequiredService<IConfiguration>(), name));
+            services.AddSingleton<EgressProviderConfigurationProvider<TOptions>>();
             services.AddSingletonForwarder<IEgressProviderConfigurationProvider<TOptions>, EgressProviderConfigurationProvider<TOptions>>();
             services.AddSingletonForwarder<IEgressProviderConfigurationProvider, EgressProviderConfigurationProvider<TOptions>>();
 
+            services.AddSingleton<IOptionsChangeTokenSource<TOptions>, EgressProviderConfigurationChangeTokenSource<TOptions>>();
```
jander-msft, line 221: Can you move this back to where it was so that all of the `IOptionsChangeTokenSource<TOptions>` implementations are grouped together?

jander-msft (approved):

src/Tools/dotnet-monitor/Egress/Configuration/EgressPropertiesProvider.cs
kelltrick: The docs say this can return a `IConfigurationSection` with a `null` parameter, so I just want to omit those if present.

src/Tools/dotnet-monitor/Egress/EgressArtifactSettings.cs
kelltrick: Yes and no, I initially changed this because the extension I wrote used an assembly reference to `dotnet-monitor` but I think I'm going to replace that with a source file reference, so this doesn't need to be public anymore.

jander-msft: Rather than going the route of using source file references, consider making another assembly e.g. Microsoft.Diagnostics.Monitoring.Egress that can hold sharable types, even if they are still marked internal. I know that we have a bunch of collection rule code that uses source file references, but that was done so we could get the feature done in time for 6.0, but I would like to refactor those to be in common assemblies such as the existing Microsoft.Diagnostics.Monitoring.Options assembly.

src/Tools/dotnet-monitor/Egress/Configuration/EgressPropertiesProvider.cs
jander-msft: That's not in the docs for `IConfiguration.GetChildren`: https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.iconfiguration.getchildren?view=dotnet-plat-ext-6.0

The `GetSection` method can return an empty subsection if the key doesn't exist (still not `null`), but you are not using that API here.

kelltrick: Ahh, that is what I was reading

kelltrick: Let me go remove that check then

PR #1813, opened 2022-04-27
Remove prerelease blob group

jander-msft:
Remove the unused `prerelease` blob group to avoid release process confusion.


src/Directory.Build.targets
```diff
@@ -44,7 +44,6 @@
     <!-- These are the valid BlobGroupBuildQuality values. -->
```
kkeirstead, line 44: Can't leave a comment on the exact line (41/42) - do we still need `_BlobGroupPreReleaseLabel` and `_BlobGroupPreReleaseIteration`, or are they used anywhere else?

jander-msft: Good catch. I removed more unused properties.

kkeirstead (approved):

PR #1819, opened 2022-04-27
Add ability to filter log messages based on log level

jander-msft:
Attempt read debug logger log level env var to control log level

### DIFF
--- a/src/MonitorProfiler/Environment/EnvironmentHelper.cpp
+++ b/src/MonitorProfiler/Environment/EnvironmentHelper.cpp
@@ -3,11 +3,29 @@
 // See the LICENSE file in the project root for more information.
 
 #include "EnvironmentHelper.h"
+#include "../Logging/LogLevelHelper.h"
 #include "productversion.h"
 
 using namespace std;
 
 #define IfFailLogRet(EXPR) IfFailLogRet_(pLogger, EXPR)
+
+HRESULT EnvironmentHelper::GetDebugLoggerLevel(
+    const std::shared_ptr<IEnvironment>& pEnvironment,
+    LogLevel& level)
+{
+    HRESULT hr = S_OK;
+
+    tstring tstrLevel;
+    IfFailRet(pEnvironment->GetEnvironmentVariable(
+        s_wszDebugLoggerLevelEnvVar,
+        tstrLevel
+        ));
+
+    IfFailRet(LogLevelHelper::ToLogLevel(tstrLevel, level));
+
+    return S_OK;
+}
 
 HRESULT EnvironmentHelper::SetProductVersion(
     const shared_ptr<IEnvironment>& pEnvironment,
--- a/src/MonitorProfiler/Environment/EnvironmentHelper.h
+++ b/src/MonitorProfiler/Environment/EnvironmentHelper.h
@@ -14,9 +14,17 @@
 class EnvironmentHelper final
 {
 private:
+    static constexpr LPCWSTR s_wszDebugLoggerLevelEnvVar = _T("DotnetMonitorProfiler_DebugLogger_Level");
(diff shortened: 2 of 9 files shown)


src/MonitorProfiler/Environment/EnvironmentHelper.cpp
```diff
@@ -3,12 +3,30 @@
...
 
 using namespace std;
 
 #define IfFailLogRet(EXPR) IfFailLogRet_(pLogger, EXPR)
 
+HRESULT EnvironmentHelper::GetDebugLoggerLevel(
+    const std::shared_ptr<IEnvironment>& pEnvironment,
+    LogLevel& level)
+{
+    HRESULT hr = S_OK;
```
kkeirstead, line 17: I'm assuming I'm missing something since I see this in several places - why do we seemingly set `hr = S_OK` but then never use it?

src/MonitorProfiler/Environment/EnvironmentHelper.h
```diff
@@ -14,9 +14,17 @@
 class EnvironmentHelper final
 {
 private:
+    static constexpr LPCWSTR s_wszDebugLoggerLevelEnvVar = _T("DotnetMonitorProfiler_DebugLogger_Level");
```
kkeirstead, line 17: Should this be `DotnetMonitorProfiler_DebugLogger__Level` (two underscores before Level), or is this correct as-is?

src/MonitorProfiler/Logging/LogLevelHelper.h
```diff
@@ -47,4 +47,67 @@ class LogLevelHelper final
...
+    /// <summary>
+    /// Checks if level is valid.
+    /// </summary>
+    static bool IsValid(LogLevel level)
+    {
+        return LogLevel::Trace <= level && level <= LogLevel::None;
+    }
+
+    /// <summary>
+    /// Converts a string its corresponding LogLevel
```
kkeirstead, line 73:
```suggestion
    /// Converts a string to its corresponding LogLevel
```

src/MonitorProfiler/Logging/DebugLogger.h
```diff
@@ -17,9 +18,17 @@ class DebugLogger final :
 private:
     const static int s_nMaxEntrySize = 1000;
 
+    LogLevel m_level = LogLevel::None;
```
wiktork, line 21: Should we default this to error?

jander-msft: Sure? It doesn't matter here because it gets set to something else in the ctor, either what the user specified or falls back to Information.

jander-msft: Sure? It doesn't matter here because it gets set to something else in the ctor, either what the user specified or falls back to Information.

src/MonitorProfiler/Environment/EnvironmentHelper.cpp
jander-msft: The `IfNullRet`, `IfFailRet`, `IfFailLogRet`, etc macros assume there is a variable named `hr` to which they can set the result of their condition. It makes it so that they don't have to define scoped variables every time and you can inspect the result after the macro as passed.

jander-msft: The `IfNullRet`, `IfFailRet`, `IfFailLogRet`, etc macros assume there is a variable named `hr` to which they can set the result of their condition. It makes it so that they don't have to define scoped variables every time and you can inspect the result after the macro as passed.

src/MonitorProfiler/Environment/EnvironmentHelper.h
jander-msft: These variables are not used in configuration, so the double underscore is not necessary.

PR #1829, opened 2022-05-02
Querying For Collection Rule State

### LINKED ISSUE #1539: Querying for trigger state
Once a trigger is set it would be really helpful to be able to query the state of the trigger to determine if it has fired, or how many more times it potentially could fire. Depending on the output of the trigger (does not have to be a diagnostics file) it may not always be obvious whether a trigger has fired or not. 

A new API could simply output a response describing the basic state of each of the defined triggers.

kkeirstead:
Provides an endpoint to get the state of collection rules for a given process. This does not include functionality for enabling/disabling rules through the API; this PR is solely responsible for viewing the state of already-executing collection rules.

Closes #1539


documentation/api/collectionrules.md
```diff
@@ -0,0 +1,95 @@
+# Collection Rules - Get
+
+Reports the state of the configured collection rules of a specific process.
+
+## HTTP Route
+
+```http
+GET /collectionrules?pid={pid}&uid={uid}&name={name} HTTP/1.1
```
wiktork, line 8: I'm wondering if we should have a top level list with less details and then an api to get more information on any particular rule.

src/Microsoft.Diagnostics.Monitoring.Options/CollectionRuleStateHolder.cs
```diff
@@ -0,0 +1,78 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    internal class CollectionRuleStateHolder
+    {
+        public CollectionRuleState CurrState { get; private set; } = CollectionRuleState.Running;
```
wiktork, line 9: nit: CurrentState

src/Microsoft.Diagnostics.Monitoring.Options/CollectionRuleStateHolder.cs
```diff
@@ -0,0 +1,78 @@
...
+        public string CurrStateReason { get; private set; } = CollectionRuleStateReasons.Running;
+
+        internal void BeginActionExecution()
+        {
+            CurrState = CollectionRuleState.ActionExecuting;
```
wiktork, line 14: Would it make sense to error check previous states? For example, it might not make sense to transition from Finished to Running again.

src/Microsoft.Diagnostics.Monitoring.Options/CollectionRuleStateHolder.cs
```diff
@@ -0,0 +1,78 @@
...
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    internal class CollectionRuleStateHolder
+    {
+        public CollectionRuleState CurrState { get; private set; } = CollectionRuleState.Running;
+        public string CurrStateReason { get; private set; } = CollectionRuleStateReasons.Running;
+
+        internal void BeginActionExecution()
```
wiktork, line 12: Consider making these methods public.

src/Microsoft.Diagnostics.Monitoring.Options/CollectionRuleStateReasons.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    public class CollectionRuleStateReasons
+    {
+        public const string Running = "This collection rule is active and waiting for its triggering conditions to be satisfied.";
```
wiktork, line 9: Move these to a resource, to make it easier to localize if we choose to in the future.

src/Microsoft.Diagnostics.Monitoring.WebApi/ICollectionRuleService.cs
```diff
@@ -0,0 +1,26 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    internal interface ICollectionRuleService : IHostedService, IAsyncDisposable
+    {
+        public Task ApplyRules(
```
wiktork, line 16: Remove protection level from interface methods.

src/Microsoft.Diagnostics.Monitoring.WebApi/ICollectionRuleService.cs
```diff
@@ -0,0 +1,26 @@
...
+            IEndpointInfo endpointInfo,
+            CancellationToken token);
+
+        public Task RemoveRules(
+            IEndpointInfo endpointInfo,
+            CancellationToken token);
+
+        public Dictionary<string, CollectionRuleDescription> GetCollectionRulesDescriptions(IEndpointInfo endpointInfo);
```
wiktork, line 24: consider IDictionary

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -20,17 +20,23 @@ internal class CollectionRulePipeline : Pipeline
         // The executor of the action list for the collection rule.
         private readonly ActionListExecutor _actionListExecutor;
 
-        private readonly CollectionRuleContext _context;
+        internal readonly CollectionRuleContext _context;
```
wiktork, line 23: Avoid internal fields.

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -20,17 +20,23 @@ internal class CollectionRulePipeline : Pipeline
...
         private readonly ICollectionRuleTriggerOperations _triggerOperations;
 
         // Flag used to guard against multiple invocations of _startCallback.
         private bool _invokedStartCallback = false;
 
+        internal Queue<DateTime> _executionTimestamps;
+        internal List<DateTime> _allExecutionTimestamps = new();
+        internal DateTime _pipelineStartTime;
+
+        public CollectionRuleStateHolder stateHolder = new();
```
wiktork, line 38: Most likely should be a property?

src/Microsoft.Diagnostics.Monitoring.Options/CollectionRuleStateReasons.cs
kkeirstead: This might be the correct option, but one thing I should mention is that we're accessing some of these values in FunctionalTests, which I don't believe will be able to access the Strings.resx file. That being said, when I encountered this issue previously, we just added the necessary constants in the FunctionalTests file and left a comment saying that they need to be identical to the ones in Strings.resx

jander-msft: Another option is not test the reason in the functional tests if validating the state is sufficient enough.

src/Microsoft.Diagnostics.Monitoring.WebApi/ICollectionRuleService.cs
kkeirstead: I ended up getting an error when doing that ("cannot convert lambda expression to intended delegate type because some of the return types in the block are not implicitly convertible to the delegate return type"); I noticed for `GetProcessEnvironment` we're just using `Dictionary`, so I left this as-is for now.

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleContainer.cs
```diff
@@ -32,6 +32,8 @@ internal class CollectionRuleContainer : IAsyncDisposable
         private readonly ISystemClock _systemClock;
         private readonly ICollectionRuleTriggerOperations _triggerOperations;
 
+        public List<CollectionRulePipeline> _pipelines = new();
```
wiktork, line 35: avoid public fields

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
```diff
@@ -151,6 +153,14 @@ private async Task MonitorRuleChangesAsync(CancellationToken token)
                     await rulesChangedTaskSource.Task;
                 }
 
+                foreach (var key in _containersMap.Keys)
```
wiktork, line 156: expand var to actual type

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
```diff
@@ -232,5 +242,96 @@ private async Task StopAndDisposeContainerAsync(CancellationToken token)
...
+            return (original > TimeSpan.Zero) ? TimeSpan.FromSeconds((long)original.TotalSeconds) : null; // Intentionally lose millisecond precision
+        }
+
+        private void CleanUpCompletedPipelines(IEndpointInfo key)
+        {
+            var collectionRuleNamesFrequency = new Dictionary<string, int>();
+
+            var activePipelineTimestamps = new Dictionary<string, DateTime>();
+
+            foreach (var pipeline in _containersMap[key]._pipelines)
```
wiktork, line 325: I think it would be better if we did this using eventing or something similar. For example, once a pipeline ends we callback into the service (or provide an event on the pipeline itself to indicate that it has ended) and remove the pipeline as necessary.

wiktork (changes requested):
Submitting some feedback.

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -210,16 +216,58 @@ protected override async Task OnRun(CancellationToken token)
...
+                    }
+                }
             }
         }
 
+        internal bool CheckForThrottling(int actionCountLimit, int currExecutions, TimeSpan? actionCountSWD)
+        {
+            if (!actionCountSWD.HasValue || actionCountLimit > currExecutions)
+            {
+                StateHolder.EndThrottled();
```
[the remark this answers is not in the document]
kkeirstead: My concern here is that we call this from several places, which means each caller would be responsible for having a state-change directly after it, which just feels like it's decoupling two things that are inherently coupled. There was a prior iteration of this that used cancellation tokens to get instant updates instead of manually calling this method to check if things are up-to-date, but I know Justin mentioned that would make testing using simulated time more challenging.

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
```diff
@@ -232,5 +242,96 @@ private async Task StopAndDisposeContainerAsync(CancellationToken token)
...
+
+            return collectionRulesState;
+        }
+
+        internal static CollectionRuleDescription GetCollectionRuleDescription(CollectionRulePipeline pipeline)
+        {
+            CollectionRuleLimitsOptions limitsOptions = pipeline.Context.Options.Limits;
+
+            DateTime currentTime = pipeline.Context.Clock.UtcNow.UtcDateTime;
+            CollectionRulePipeline.DequeueOldTimestamps(pipeline.ExecutionTimestamps, limitsOptions?.ActionCountSlidingWindowDuration, currentTime);
```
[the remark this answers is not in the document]
kkeirstead: Potentially clone the state to not manipulate the original version here

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
kkeirstead: Look into this more - should be able to trash all of our pipelines since we will be resetting everything (move it to when configuration changes) -> will clear out everything that's finished from a previous run.

documentation/api/collectionrules.md
jander-msft: Something like?
- `/collectionrules?pid={pid}&uid={uid}&name={name}` - returns a list of summary information for each collection for the specified process (or default process), such as an ID, name, state
- `/collectionrules/{id}` - returns detail information for a specific instance of a collection rule

The ID can be globally (within the dotnet-monitor instance) unique. So then you can use that ID to get detailed information about the rule instance. This would be easier to uniquely identify the rule as well as prevent collisions if the target process shuts down and restarts (which would have the similar process + rule name combination). This would also more easily allow "rel links" if we wanted to add those in the future.

src/Microsoft.Diagnostics.Monitoring.WebApi/CollectionRuleStateHolder.cs
```diff
@@ -0,0 +1,84 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Globalization;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    internal class CollectionRuleStateHolder
```
wiktork, line 9: Conside record for automatic copying.

documentation/api/collectionrules.md
kkeirstead: Any opposition to making the endpoints follow this pattern instead?

- `/collectionrules?pid={pid}&uid={uid}&name={name}`
- `/collectionrules?pid={pid}&uid={uid}&name={name}/NameOfMyRule`

This prevents users from _having_ to hit the generic endpoint to get an ID if they know what rule they want to look at, and should make the output cleaner of the generic endpoint by not having to include an extra field for a unique ID. Personally, I feel like it's a more intuitive/better user experience, but if there's other reasons why this is a bad idea, I can do IDs (or something else).

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -210,16 +224,55 @@ protected override async Task OnRun(CancellationToken token)
...
+                _stateHolder.RuleDurationReached();
+            }
+            catch (Exception ex)
+            {
+                _stateHolder.RuleFailure(ex.Message);
+                throw;
             }
         }
 
+        public Queue<DateTime> DequeueOldTimestamps(TimeSpan? actionCountWindowDuration, DateTime currentTimestamp)
```
kkeirstead, line 245: I know there was some discussion earlier about not wanting to have a copy of `_executionTimestamps` in `CollectionRuleService`; is there any issue with having this method interact directly with `_executionTimestamps` (instead of returning a copy), and then `CheckForThrottling` directly using its size? I think there had been concerns about updating the `CollectionRuleStateHolder` from the front-end, but is there any reason (from the perspective of needing to lock resources) that we couldn't do it this way for the timestamps?

documentation/api/collectionrules.md
jander-msft: By adding `/NameOfMyRule` to the `name` parameter, that muddles the consistent notion across all of these routes that the name parameter identifies the process name.

jander-msft: While this is unlikely and strange to see, it is also possible to add any character to a file or directory name on Unix systems, such as the forward-slash. There is no "reserved" or "illegal" character notion.

kkeirstead: > By adding `/NameOfMyRule` to the `name` parameter, that muddles the consistent notion across all of these routes that the name parameter identifies the process name.

I don't think I'm too worried about this - perhaps being more particular about the verbiage would reduce any potential confusion, but it feels relatively intuitive (for me) to say, "for this particular process (PID/UID/Name), I want to look at the rule named xyz." I like being able to keep using the same queries over and over again (or even tab to complete them), and needing to always get the ID of a rule just seems like adding unnecessary friction by making it a two-step operation. 

In regards to your second point, I'm assuming you're saying that we could encounter something like: `/collectionrules?name=firstpartofproc/secondpartofproc/FirstPartOfCRName/SecondPartOfCRName` - you would know better than I do about whether this would be a legitimate issue, or if we could check our list of collection rule names against the back-end of the string to see if there's a match (e.g. we have a CR named "HighCpuRule/2", so we check if the last 13 characters of the string are "HighCpuRule/2").

My goal is to make this intuitive/streamlined for the 99% use-case to give the bulk of customers a better user experience, but also guaranteeing it works for that last 1% by doing the parsing (if that's an option) even if you end up with something ugly like `/collectionrules?name=firstpartofproc/secondpartofproc/FirstPartOfCRName/SecondPartOfCRName`.

jander-msft: > I don't think I'm too worried about this.

I am. Consistency allows users to better predict how to use an API. If it changes slightly for arbitrary routes, that's frustrating because you have to remember "no, this route is special for some reason".

> In regards to your second point, I'm assuming you're saying that we could encounter something like: `/collectionrules?name=firstpartofproc/secondpartofproc/FirstPartOfCRName/SecondPartOfCRName` - you would know better than I do about whether this would be a legitimate issue, or if we could check our list of collection rule names against the back-end of the string to see if there's a match (e.g. we have a CR named "HighCpuRule/2", so we check if the last 13 characters of the string are "HighCpuRule/2").

Using ends-with notions leads to problems with accidental collisions and no way to disambiguate. This syntax also is hard to report errors more specific than "could not find" because we don't know where the process name ends and where the rule starts.

> My goal is to make this intuitive/streamlined for the 99% use-case to give the bulk of customers a better user experience, but also guaranteeing it works for that last 1% by doing the parsing (if that's an option) even if you end up with something ugly like `/collectionrules?name=firstpartofproc/secondpartofproc/FirstPartOfCRName/SecondPartOfCRName`.

REST-like APIs are probably the most intuitive. The operations route follows this pattern where you `GET /operations` to list all of the operations and `GET /operations/{id}` to get details of a particular operation. My proposal follows that exactly. My additional comments in my original response also enable better behaviors:
- This would be easier to uniquely identify the rule as well as prevent collisions if the target process shuts down and restarts (which would have the similar process + rule name combination).
- This would also more easily allow "rel links" if we wanted to add those in the future.

Additionally, we cannot represent two different structured data formats for the same verb (GET) and HTTP response code (200) pairing in the openapi document. So we cannot represent a JSON object with collection rule names as the properties and another JSON object with the detail fields as its properties for the `GET /collectionrules -> 200` pairing.

jander-msft: Alternatively, something like `GET /collectionrules/{name}?pid={pid}&uid={uid}&name={name}` would allow you get the collection rule instance for the matching process or the default process if `GET /collectionrules/{name}` as issued.

kkeirstead: > Alternatively, something like `GET /collectionrules/{name}?pid={pid}&uid={uid}&name={name}` would allow you get the collection rule instance for the matching process or the default process if `GET /collectionrules/{name}` is issued.

I had considered this, but wasn't sure if that would be more jarring by not having the process determination first. Purely for the purposes of not having to repeatedly look up a changing ID, I prefer this though over the initial suggestion - but I'll ultimately defer to what you think makes the most sense for the API, I just wanted to voice my concerns.

src/Microsoft.Diagnostics.Monitoring.Options/CollectionRuleStateHolder.cs
kkeirstead: Debug.Assert (will only be for our own internal purposes)

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
kkeirstead: Move to `CollectionRuleStateHolder` (holds a copy)

src/Microsoft.Diagnostics.Monitoring.WebApi/ICollectionRuleService.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        Task ApplyRules(
+            IEndpointInfo endpointInfo,
+            CancellationToken token);
+
+        Task RemoveRules(
+            IEndpointInfo endpointInfo,
+            CancellationToken token);
+
+        Dictionary<string, CollectionRuleDescription> GetCollectionRulesDescriptions(IEndpointInfo endpointInfo);
```
wiktork, line 24: nit: IDictionary

wiktork, line 24: nit: IDictionary

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -65,14 +71,16 @@ protected override async Task OnRun(CancellationToken token)
 
             CancellationToken linkedToken = linkedCancellationSource.Token;
 
-            TimeSpan? actionCountWindowDuration = _context.Options.Limits?.ActionCountSlidingWindowDuration;
-            int actionCountLimit = (_context.Options.Limits?.ActionCount).GetValueOrDefault(CollectionRuleLimitsOptionsDefaults.ActionCount);
-            Queue<DateTime> executionTimestamps = new(actionCountLimit);
+            TimeSpan? actionCountWindowDuration = Context.Options.Limits?.ActionCountSlidingWindowDuration;
+            int actionCountLimit = (Context.Options.Limits?.ActionCount).GetValueOrDefault(CollectionRuleLimitsOptionsDefaults.ActionCount);
+            ExecutionTimestamps = new(actionCountLimit);
```
wiktork, line 76: nit: Let's use explicit constructors here, it's not very obvious what you are creating here.

wiktork, line 76: nit: Let's use explicit constructors here, it's not very obvious what you are creating here.

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
```diff
@@ -232,5 +239,120 @@ private async Task StopAndDisposeContainerAsync(CancellationToken token)
...
+        }
+
+        internal static CollectionRuleDescription GetCollectionRuleDescription(CollectionRulePipeline pipeline)
+        {
+            CollectionRuleLimitsOptions limitsOptions = pipeline.Context.Options.Limits;
+
+            DateTime currentTime = pipeline.Context.Clock.UtcNow.UtcDateTime;
+            pipeline.DequeueOldTimestamps(limitsOptions?.ActionCountSlidingWindowDuration, currentTime);
+
+            CollectionRuleStateHolder stateHolderCopy = pipeline.GetStateHolderCopy();
```
wiktork, line 279: We may have discussed this already, but we should consider locking during the copy operation. Without locking, it would be possible to get a new state but an old reason for example.

src/Microsoft.Diagnostics.Monitoring.WebApi/ICollectionRuleService.cs
kkeirstead: Duplicate, see response in https://github.com/dotnet/dotnet-monitor/pull/1829#discussion_r866998280

src/Microsoft.Diagnostics.Monitoring.WebApi/CollectionRuleStateHolder.cs
kkeirstead: I was having an issue locking that appeared after switching to using `record`, so I'm currently omitting it.

src/Microsoft.Diagnostics.Monitoring.WebApi/CollectionRuleStateHolder.cs
```diff
@@ -0,0 +1,135 @@
...
+    {
+        public CollectionRuleState CurrentState { get; private set; } = CollectionRuleState.Running;
+        public string CurrentStateReason { get; private set; } = Strings.Message_CollectionRuleStateReason_Running;
+
+        private readonly object _lock = new object();
+
+        public CollectionRuleStateHolder(CollectionRuleStateHolder other)
+        {
+            lock (_lock)
```
jander-msft, line 19: This is in the ctor so it doesn't make sense to lock itself. What you really want is to lock the other's _lock field since you are copying from it: `lock (other._lock)`

src/Microsoft.Diagnostics.Monitoring.WebApi/ICollectionRuleService.cs
```diff
@@ -0,0 +1,28 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    internal interface ICollectionRuleService : IHostedService, IAsyncDisposable
+    {
+        Task ApplyRules(
```
jander-msft, line 16: ApplyRules and RemoveRules are not necessary for WebApi to have access to getting the states.

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -241,5 +302,38 @@ private void InvokeStartCallback()
                 _invokedStartCallback = true;
             }
         }
+
+        protected override async Task OnCleanup()
```
jander-msft, line 306: remove

src/Tools/dotnet-monitor/ServiceCollectionExtensions.cs
```diff
@@ -125,8 +125,8 @@ public static IServiceCollection ConfigureCollectionRules(this IServiceCollectio
             services.AddSingleton<IOptionsMonitorCache<CollectionRuleOptions>, DynamicNamedOptionsCache<CollectionRuleOptions>>();
 
             services.AddSingleton<ActionListExecutor>();
-            services.AddSingleton<CollectionRuleService>();
-            services.AddHostedServiceForwarder<CollectionRuleService>();
+            services.AddSingleton<ICollectionRuleService, CollectionRuleService>();
```
jander-msft, line 128: Please revert these changes and add a call to AddSingletonForwarder

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -65,14 +75,16 @@ protected override async Task OnRun(CancellationToken token)
 
             CancellationToken linkedToken = linkedCancellationSource.Token;
 
-            TimeSpan? actionCountWindowDuration = _context.Options.Limits?.ActionCountSlidingWindowDuration;
-            int actionCountLimit = (_context.Options.Limits?.ActionCount).GetValueOrDefault(CollectionRuleLimitsOptionsDefaults.ActionCount);
-            Queue<DateTime> executionTimestamps = new(actionCountLimit);
+            _actionCountWindowDuration = Context.Options.Limits?.ActionCountSlidingWindowDuration;
+            _actionCountLimit = (Context.Options.Limits?.ActionCount).GetValueOrDefault(CollectionRuleLimitsOptionsDefaults.ActionCount);
+            _executionTimestamps = new Queue<DateTime>(_actionCountLimit);
```
jander-msft, line 80: I would like to see at least `_extensionTimestamps` and `AllExecutionTimestamps` moved to `CollectionRuleStateHolder` as well as move the `DequeOldTimestamp` and `CheckForThottling` logic moved there as well. Then expose something like `CanExecuteActions(ISystemClock)` and `ActionExecutionComplete(bool success)` methods. These would be called by where you currently call `CheckForThrottling`. The `BeginThrottled`, `EndThrottled`, `ActionExecutionFailed`, `ActionExecutionSucceeded` can be removed and become implementation details of two new methods.

You might have to initialize the state holder with the collection rule limits.

The benefit of this is that all of the execution state is now in the state holder and doesn't require calling static methods on `CollectionRulePipeline`. The webapi can effectively get a copy of the state holder and not have to ever touch the pipeline again.

src/Microsoft.Diagnostics.Monitoring.WebApi/CollectionRulePipelineStateHolder.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    internal record class CollectionRulePipelineStateHolder
```
jander-msft, line 10: I think this should be combined with `CollectionRuleStateHolder`. I don't think there is a need to keep them separate.

src/Microsoft.Diagnostics.Monitoring.WebApi/CollectionRuleStateHolder.cs
```diff
@@ -0,0 +1,135 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics;
+using System.Globalization;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    internal class CollectionRuleStateHolder
```
jander-msft, line 10: Let's just call it `CollectionRuleState` or `CollectionRulePipelineState`

src/Microsoft.Diagnostics.Monitoring.WebApi/CollectionRulePipelineState.cs
```diff
@@ -0,0 +1,214 @@
...
+                ExecutionTimestamps = other.ExecutionTimestamps;
+                AllExecutionTimestamps = other.AllExecutionTimestamps;
+                CurrentState = other.CurrentState;
+                CurrentStateReason = other.CurrentStateReason;
+            }
+        }
+
+        public CollectionRulePipelineState(int actionCountLimit, TimeSpan? actionCountSlidingWindowDuration, TimeSpan? ruleDuration)
+        {
+            lock (_lock)
```
jander-msft, line 40: The lock and the Debug.Assert are not necessary in the ctor.

src/Microsoft.Diagnostics.Monitoring.WebApi/CollectionRulePipelineState.cs
```diff
@@ -0,0 +1,209 @@
...
+            }
+        }
+
+        public bool CanExecuteActions(DateTime currentTime)
+        {
+            lock (ExecutionTimestamps)
+            {
+                DequeueOldTimestamps(ExecutionTimestamps, ActionCountSlidingWindowDuration, currentTime);
+            }
+            bool canExecuteActions = !CheckForThrottling(ActionCountLimit, ActionCountSlidingWindowDuration, ExecutionTimestamps.Count);
```
wiktork, line 153: Should CheckForThrottling also be done under a lock?

src/Microsoft.Diagnostics.Monitoring.WebApi/CollectionRulePipelineState.cs
```diff
@@ -0,0 +1,209 @@
...
+        public List<DateTime> AllExecutionTimestamps { get; set; }
+        public TimeSpan? ActionCountSlidingWindowDuration { get; private set; }
+        public TimeSpan? RuleDuration { get; private set; }
+        public int ActionCountLimit { get; private set; }
+
+        private readonly object _lock = new object();
+
+        public CollectionRulePipelineState(CollectionRulePipelineState other)
+        {
+            lock (other._lock)
```
wiktork, line 26: Can you add some comments here?

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -136,38 +142,25 @@ protected override async Task OnRun(CancellationToken token)
...
-                    DateTime currentTimestamp = _context.Clock.UtcNow.UtcDateTime;
+                    DateTime currentTimestamp = Context.Clock.UtcNow.UtcDateTime;
 
-                    // If rule has an action count window, Remove all execution timestamps that fall outside the window.
-                    if (actionCountWindowDuration.HasValue)
+                    if (_stateHolder.CanExecuteActions(currentTimestamp))
                     {
-                        DateTime windowStartTimestamp = currentTimestamp - actionCountWindowDuration.Value;
-                        while (executionTimestamps.Count > 0)
+                        lock (_stateHolder.ExecutionTimestamps)
```
wiktork, line 149: We lock around ExecutionTimeStamps here but around _lock when copying. Can we just use a single lock?

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -136,38 +142,25 @@ protected override async Task OnRun(CancellationToken token)
...
-                            }
+                            _stateHolder.ExecutionTimestamps.Enqueue(currentTimestamp);
                         }
-                    }
 
-                    // Check if executing actions has been throttled due to count limit
-                    if (actionCountLimit > executionTimestamps.Count)
-                    {
-                        executionTimestamps.Enqueue(currentTimestamp);
+                        _stateHolder.AllExecutionTimestamps.Add(currentTimestamp);
```
wiktork, line 154: Since we clone AllexecutionTimestamps as well, we probably want to lock around this addition as well.

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -178,24 +171,31 @@ protected override async Task OnRun(CancellationToken token)
...
+                            _stateHolder.CanExecuteActions(Context.Clock.UtcNow.UtcDateTime);
+
                             // The collection rule has executed the action list the maximum
                             // number of times as specified by the limits and the action count
                             // window was not specified. Since the pipeline can no longer execute
                             // actions, the pipeline can complete.
-                            completePipeline = actionCountLimit <= executionTimestamps.Count &&
-                                !actionCountWindowDuration.HasValue;
-                        }
+                            completePipeline = _stateHolder.ActionCountLimit <= _stateHolder.ExecutionTimestamps.Count &&
```
wiktork, line 185: Consider pushing logic that operates against _stateHolder fields to Instance methods on StateHolder that you can then lock against.

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -241,5 +249,24 @@ private void InvokeStartCallback()
                 _invokedStartCallback = true;
             }
         }
+
+        public CollectionRulePipelineState GetPipelineState()
+        {
+            CollectionRulePipelineState pipelineStateCopy = new(_stateHolder.ActionCountLimit, _stateHolder.ActionCountSlidingWindowDuration, _stateHolder.RuleDuration);
```
wiktork, line 255: I think this line needs to be removed?

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -241,5 +249,24 @@ private void InvokeStartCallback()
...
+
+            lock (_stateHolder)
+            {
+                pipelineStateCopy = new CollectionRulePipelineState(_stateHolder);
+            }
+
+            lock (_stateHolder.ExecutionTimestamps)
+            {
+                pipelineStateCopy.ExecutionTimestamps = new Queue<DateTime>(_stateHolder.ExecutionTimestamps);
```
wiktork, line 264: Maybe we can push all this logic into the copy constructor instead of having two under different locks?

wiktork (commented):
Leaving some more feedback for the locking.

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -241,5 +249,24 @@ private void InvokeStartCallback()
                 _invokedStartCallback = true;
             }
         }
+
+        public CollectionRulePipelineState GetPipelineState()
+        {
+            CollectionRulePipelineState pipelineStateCopy = new(_stateHolder.ActionCountLimit, _stateHolder.ActionCountSlidingWindowDuration, _stateHolder.RuleDuration);
+
+            lock (_stateHolder)
```
wiktork, line 257: Remove lock(_stateHolder)

src/Microsoft.Diagnostics.Monitoring.WebApi/CollectionRulePipelineState.cs
```diff
@@ -0,0 +1,209 @@
...
+                Debug.Assert(CurrentState != CollectionRuleState.Finished);
+
+                CurrentState = CollectionRuleState.Finished;
+                CurrentStateReason = string.Format(CultureInfo.InvariantCulture, Strings.Message_CollectionRuleStateReason_Finished_Failure, errorMessage);
+            }
+        }
+
+        public bool CanExecuteActions(DateTime currentTime)
+        {
+            lock (ExecutionTimestamps)
```
wiktork, line 149: Remove locking around ExecutionTimestamps and use the state lock

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
wiktork, line 255: Another consideration here:
- Instead of having a copied object that is the same type (CollectionRulePipelineState), create a new object type that has all the necessary data (deep copied) but none of the locking since it won't be required.

src/Microsoft.Diagnostics.Monitoring.WebApi/CollectionRulePipelineState.cs
kkeirstead: Discussed

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
kkeirstead: Since we're not operating on `ExecutionTimestamps` (and just getting its count), I don't believe we'll need to lock - but I can move this logic into `CollectionRulePipelineState`, which will let us make `ActionCountReached()` a private method.

src/Microsoft.Diagnostics.Monitoring.WebApi/CollectionRulePipelineState.cs
wiktork: While the cloned version of the state can do this without locking, your _stateHolder object calls this method as well. The calls to ExecutionTimestamps.Count and the other properties should be under a lock.

kkeirstead: Maybe I misunderstood - when we talked about this on Friday, I thought we said we wouldn't need to lock around this since we're just getting the count for `ExecutionTimestamps` (and `ActionCountLimit` and `ActionCountSlidingWindowDuration` shouldn't change). Regardless, I can add that into the `lock`.

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -241,5 +239,14 @@ private void InvokeStartCallback()
                 _invokedStartCallback = true;
             }
         }
+
+        public CollectionRulePipelineState GetPipelineState()
+        {
+            CollectionRulePipelineState pipelineStateCopy = new CollectionRulePipelineState(_stateHolder);
+
+            pipelineStateCopy.CanExecuteActions(Context.Clock.UtcNow.UtcDateTime);
```
wiktork, line 247: Can you add some comments here? Another option is to possibly separate CanExecuteActions into two apis, one that Dequeues the timestamps and one that evaluates if an action can be executed.

src/Microsoft.Diagnostics.Monitoring.WebApi/CollectionRulePipelineState.cs
```diff
@@ -0,0 +1,231 @@
...
+            {
+                ActionExecutionSucceeded();
+            }
+
+            return success;
+        }
+
+        public bool CheckForActionCountLimitReached()
+        {
+            bool limitReached = ActionCountLimit <= ExecutionTimestamps.Count && !ActionCountSlidingWindowDuration.HasValue;
```
wiktork, line 193: While this is likely to work ok, strictly speaking `Queue<T>.Count` is not thread safe. You should still lock around it.

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
kkeirstead: Alternatively, maybe we could use a name that's more indicative of the fact that this is responsible for updating state and not just checking a condition? I personally prefer the unification of these two tasks, because it means the caller doesn't need to remember to call dequeue - they can assume that `ExecutionTimestamps` is always up-to-date when it's operated against.

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -20,25 +20,29 @@ internal class CollectionRulePipeline : Pipeline
...
         private readonly Action _startCallback;
-        
+
         // Operations for getting trigger information.
         private readonly ICollectionRuleTriggerOperations _triggerOperations;
 
         // Flag used to guard against multiple invocations of _startCallback.
         private bool _invokedStartCallback = false;
 
+        public DateTime PipelineStartTime { get; private set; }
```
jander-msft, line 34: Move this to CollectionRulePipelineState since it's not used by anything in CollectionRulePipeline.

src/Microsoft.Diagnostics.Monitoring.WebApi/CollectionRulePipelineState.cs
```diff
@@ -0,0 +1,238 @@
...
+    {
+        public CollectionRuleState CurrentState { get; private set; } = CollectionRuleState.Running;
+        public string CurrentStateReason { get; private set; } = Strings.Message_CollectionRuleStateReason_Running;
+        public Queue<DateTime> ExecutionTimestamps { get; set; }
+        public List<DateTime> AllExecutionTimestamps { get; set; }
+        public TimeSpan? ActionCountSlidingWindowDuration { get; private set; }
+        public TimeSpan? RuleDuration { get; private set; }
+        public int ActionCountLimit { get; private set; }
+
+        private readonly object _lock = new object();
```
jander-msft, line 22: I think this lock object should be in CollectionRulePipeline: the pipeline locks whenever it needs to mutate the state object and when it is making a copy of the state object to hand out to external callers. Then the object itself doesn't need to have any synchronization since we don't have multiple threads or callers accessing the same instance.

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
```diff
@@ -136,38 +142,20 @@ protected override async Task OnRun(CancellationToken token)
...
-                        }
-                    }
+                    DateTime currentTimestamp = Context.Clock.UtcNow.UtcDateTime;
 
-                    // Check if executing actions has been throttled due to count limit
-                    if (actionCountLimit > executionTimestamps.Count)
+                    if (!_stateHolder.CheckForThrottling(currentTimestamp))
                     {
-                        executionTimestamps.Enqueue(currentTimestamp);
+                        _stateHolder.AddTimestamp(currentTimestamp);
```
jander-msft, line 149: Ideally, this is part of CheckForThrottling so that way there is no race condition between checking and adding. I think in another comment you mentioned possibly renaming this method so it's clear that this is doing more than just checking. Maybe `bool BeginActionExecution(DateTime)` then you can fold the existing BeginActionExecution into the previously named CheckForThrottling.

src/Microsoft.Diagnostics.Monitoring.WebApi/CollectionRulePipelineState.cs
```diff
@@ -0,0 +1,238 @@
...
+            }
+            else
+            {
+                BeginThrottled();
+            }
+
+            return isThrottled;
+        }
+
+        public bool ActionExecutionSucceeded(bool success)
```
jander-msft, line 179:
```suggestion
        public bool ActionExecutionCompleted(bool success)
```

src/Microsoft.Diagnostics.Monitoring.WebApi/CollectionRulePipelineState.cs
```diff
@@ -0,0 +1,238 @@
...
+            {
+                if (CurrentState == CollectionRuleState.Throttled)
+                {
+                    CurrentState = CollectionRuleState.Running;
+                    CurrentStateReason = Strings.Message_CollectionRuleStateReason_Running;
+                }
+            }
+        }
+
+        public void StartupTriggerCompleted()
```
jander-msft, line 106: This, RuleDurationReached, ActionCountReached are the same thing but with different explanations. Can we merge these into one method with some parameters to determine which reason to specify?

src/Microsoft.Diagnostics.Monitoring.WebApi/CollectionRulePipelineState.cs
jander-msft: Discussed in person: add comments as to why its preferable to have it here

wiktork (dismissed):

src/Tools/dotnet-monitor/CollectionRules/CollectionRulePipeline.cs
kkeirstead: The only thing that's a bit tricky is that we also call `CheckForThrottling` when we get a copy of the state, where we aren't adding an action execution. I ended up doing something that I think is a bit different than what you suggested, but is in the same spirit - you can let me know if you want me to adjust it.

PR #1830, opened 2022-05-02
Add warning when collection rules are configured but not running in Listen mode

### LINKED ISSUE #1774: Add warning when collection rules are configured but not running in listen mode
We have at least one report of collection rules not working, but I believe it's because listen mode was not configured. Since this is a requirement of collection rules, the tool should show a warning when this situation occurs and a brief way to remedy this issue (e.g. aka.ms link to docs).

kkeirstead:
Currently, if a user is running in `Connect` mode and has collection rules configured, we don't give a warning indicating the issue. Now, a Warning logger message gives a brief description of what to do to resolve the issue. The check also occurs in two separate places since changes to the diagnostic port aren't respected once dotnet monitor is running - the first check will occur during validation for the diagnostic port, and then a check will occur every time the collection rule configuration changes. This will **not** fail validation, as that would be a breaking change.

Closes #1774

### DIFF
--- a/src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
+++ b/src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
@@ -159,7 +159,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules
 
                 _logger.CollectionRuleConfigurationChanged();
 
-                if (!(_portOptions.ConnectionMode == DiagnosticPortConnectionMode.Listen && !string.IsNullOrEmpty(_portOptions.EndpointName)))
+                if (_portOptions.ConnectionMode != DiagnosticPortConnectionMode.Listen || string.IsNullOrEmpty(_portOptions.EndpointName))
                 {
                     _logger.LogWarning(Strings.ErrorMessage_DiagnosticPortNotInListenModeForCollectionRules);
                 }
--- a/src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
+++ b/src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
@@ -159,9 +159,9 @@ namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules
 
                 _logger.CollectionRuleConfigurationChanged();
 
-                if (_portOptions.ConnectionMode != DiagnosticPortConnectionMode.Listen || string.IsNullOrEmpty(_portOptions.EndpointName))
+                if (DiagnosticPortOptionsExtensions.GetConnectionMode(_portOptions) != DiagnosticPortConnectionMode.Listen)
                 {
-                    _logger.LogWarning(Strings.ErrorMessage_DiagnosticPortNotInListenModeForCollectionRules);
+                    _logger.DiagnosticPortNotInListenModeForCollectionRules();
                 }
 
                 // Get a copy of the container list to avoid having to
--- a/src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
+++ b/src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
@@ -7,6 +7,7 @@ using Microsoft.Diagnostics.Monitoring.WebApi;
 using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Configuration;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -27,6 +28,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules
         private readonly ChannelWriter<CollectionRuleContainer> _containersToRemoveWriter;
         private readonly ILogger<CollectionRuleService> _logger;
         private readonly CollectionRulesConfigurationProvider _provider;
+        private readonly DiagnosticPortOptions _portOptions;
         private readonly IServiceProvider _serviceProvider;
 
         private long _disposalState;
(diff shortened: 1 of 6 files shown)


src/Tools/dotnet-monitor/DiagnosticPortValidateOptions.cs
```diff
@@ -12,8 +14,28 @@ namespace Microsoft.Diagnostics.Tools.Monitor
...
         public ValidateOptionsResult Validate(string name, DiagnosticPortOptions options)
         {
+            var collectionRuleNames = _provider.GetCollectionRuleNames();
+
+            if (collectionRuleNames != null && collectionRuleNames.Count > 0)
+            {
+                if (options.ConnectionMode != DiagnosticPortConnectionMode.Listen
+                    || string.IsNullOrEmpty(options.EndpointName))
+                {
+                    _logger.LogWarning(Strings.ErrorMessage_DiagnosticPortNotInListenModeForCollectionRules);
```
kkeirstead, line 35: I recognize that this isn't the same as the actual validation that follows it, and so it maybe shouldn't be in this file - in the meantime I left it here due to its similar nature, but something different can be done if this feels incorrect.

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
```diff
@@ -155,6 +159,11 @@ private async Task MonitorRuleChangesAsync(CancellationToken token)
 
                 _logger.CollectionRuleConfigurationChanged();
 
+                if (_portOptions.ConnectionMode != DiagnosticPortConnectionMode.Listen || string.IsNullOrEmpty(_portOptions.EndpointName))
```
jander-msft, line 162: The [IValidateOptions<DiagnosticPortOptions> implementation](https://github.com/dotnet/dotnet-monitor/blob/main/src/Tools/dotnet-monitor/DiagnosticPortValidateOptions.cs) already validates that the endpoint was specified if the connection mode is `Listen`. Thus, you will not be able to get an instance of `DiagnosticPortOptions` through the options system if this condition is not satisfied. You do not need to check this condition here and can just check the connection mode. I would suggest using [DiagnosticPortOptionsExtensions.GetConnectionMode](https://github.com/dotnet/dotnet-monitor/blob/main/src/Microsoft.Diagnostics.Monitoring.Options/DiagnosticPortOptionsExtensions.cs) to get its understood value.

src/Tools/dotnet-monitor/DiagnosticPortValidateOptions.cs
```diff
@@ -12,8 +14,28 @@ namespace Microsoft.Diagnostics.Tools.Monitor
...
+        {
+            _provider = provider;
+            _logger = logger;
+        }
+
         public ValidateOptionsResult Validate(string name, DiagnosticPortOptions options)
         {
+            var collectionRuleNames = _provider.GetCollectionRuleNames();
+
+            if (collectionRuleNames != null && collectionRuleNames.Count > 0)
```
jander-msft, line 30: These changes should not exist here. This class should only concern itself with validating the `DiagnosticPortOptions` instance. The changes in the `CollectionRuleService` should be sufficient to warn about the condition.

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -133,6 +133,10 @@
     <value>In 'Listen' mode, the diagnostic port endpoint name must be specified.</value>
     <comment>Gets a string similar to "In 'Listen' mode, the diagnostic port endpoint name must be specified.".</comment>
   </data>
+  <data name="ErrorMessage_DiagnosticPortNotInListenModeForCollectionRules" xml:space="preserve">
+    <value>When using collection rules, the diagnostic port must be in 'Listen' mode and the diagnostic port endpoint name must be specified. To use collection rules, please restart dotnet monitor with the updated configuration.</value>
```
jander-msft, line 137:
```suggestion
    <value>When using collection rules, the diagnostic port must be in 'Listen' mode. Please correct the configuration and restart dotnet monitor.</value>
```

More concise:
- We don't need to explain the full valid condition for using `listen` mode. Saying that the mode is required should be sufficient in order to spur the user to figure out what 'listen' mode is (presumably by looking at documentation) and how to configure it.
- I changed the second sentence around to imply order: first change the configuration, then restart dotnet monitor.

jander-msft (changes requested):

jander-msft:
There might be benefit to add a unit test (or functional test, if that is too hard) that validates that this message is logged under these circumstances. It will help remember to remove the warning if we relax the 'listen' mode requirement.

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
```diff
@@ -155,6 +159,11 @@ private async Task MonitorRuleChangesAsync(CancellationToken token)
...
+                {
+                    _logger.LogWarning(Strings.ErrorMessage_DiagnosticPortNotInListenModeForCollectionRules);
```
jander-msft, line 164: This needs to be moved to LoggingExtensions.

src/Tools/dotnet-monitor/DiagnosticPortValidateOptions.cs
kkeirstead: I would tend to agree - I didn't really want to stick it here, but my understanding is that the changes to `CollectionRuleService` will only print when the configuration is changed (not on first load). If someone starts up `dotnet monitor` in `Connect` mode and never tries to change their config file during its execution, presumably they'll never get the warning, which feels like we aren't actually fixing the problem in a lot of user scenarios. I had previously looked at putting the warning in `ServerEndpointInfoSource.cs` when we do the check within `ExecuteAsync`, but felt like that file shouldn't need to have knowledge of the existence of collection rules.

src/Tools/dotnet-monitor/DiagnosticPortValidateOptions.cs
```diff
@@ -3,6 +3,8 @@
 // See the LICENSE file in the project root for more information.
 
 using Microsoft.Diagnostics.Monitoring.WebApi;
+using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Configuration;
```
jander-msft, line 6: Revert the changes in this file and then I think the PR is good to go.

src/Tools/dotnet-monitor/DiagnosticPortValidateOptions.cs
jander-msft: I see. The GH UI made me think the condition check and logging this was in the ctor. I'll add another comment for what I think should be changed over there.

src/Tools/dotnet-monitor/CollectionRules/CollectionRuleService.cs
```diff
@@ -155,6 +159,11 @@ private async Task MonitorRuleChangesAsync(CancellationToken token)
 
                 _logger.CollectionRuleConfigurationChanged();
 
+                if (DiagnosticPortOptionsExtensions.GetConnectionMode(_portOptions) != DiagnosticPortConnectionMode.Listen)
```
jander-msft, line 162: Couple of things:
- Move this to a private method.
- Call it from the ctor and from this location.
- Update the code to check if there are any collection rules before logging the warning. If the user doesn't have collection rules, no point in logging the warning.

jander-msft (changes requested):

jander-msft (approved):

PR #1831, opened 2022-05-02
Take FI into `feature/egressExtension`

kelltrick:


jander-msft (approved):
Remember to merge, not squash

PR #1832, opened 2022-05-03
[Egress Extensibility] Add services to perform egress

kelltrick:
[Work in progress]
- Adds Extension discovery infrastructure and folder enumeration
- Adds centralized infrastructure for calling an extension
- Adds `OutputParser` to read protocol from extension
- Adds many logging events for debugging extensions

### DIFF
--- a/src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
+++ b/src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
@@ -3,15 +3,10 @@
 // See the LICENSE file in the project root for more information.
 
 using Microsoft.Diagnostics.Tools.Monitor.Egress.Configuration;
-using Microsoft.Diagnostics.Tracing.Parsers.MicrosoftWindowsTCPIP;
+using Microsoft.Diagnostics.Tools.Monitor.Extensibility;
 using Microsoft.Extensions.Logging;
 using System;
-using System.Collections.Generic;
-using System.Diagnostics;
-using System.Globalization;
 using System.IO;
-using System.Text;
-using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -27,11 +22,13 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Egress
         EgressProvider<ExtensionEgressProviderOptions>
     {
         private readonly IEgressPropertiesProvider _propertyProvider;
+        private readonly IExtensionDiscoverer _extensionDiscoverer;
 
-        public ExtensionEgressProvider(IEgressPropertiesProvider propertyProvider, ILogger<ExtensionEgressProvider> logger)
+        public ExtensionEgressProvider(IEgressPropertiesProvider propertyProvider, IExtensionDiscoverer extensionDiscoverer, ILogger<ExtensionEgressProvider> logger)
             : base(logger)
         {
             _propertyProvider = propertyProvider;
+            _extensionDiscoverer = extensionDiscoverer;
         }
 
         public override async Task<string> EgressAsync(
@@ -50,57 +47,15 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Egress
                 ProviderName = providerName,
             };
 
-            /* [TODOs]
-             * 1. Add a new service to dynamically find these extension(s)
-             * 2. Remove all raw logging statements from this method and refactor into LoggingExtensions
-             * 3. Stream StdOut and StdErr async in the process so their streams don't need to end before we can return
-             * 4. Refactor WaitForExit to do an async wait
-             * 5. Add well-factored protocol for returning information from an extension
-             */
-            const string extensionProcessPath = "MyPathTo\\Extension.exe";
-            ProcessStartInfo pStart = new ProcessStartInfo()
+            IEgressExtension ext = _extensionDiscoverer.FindExtension<IEgressExtension>(providerType);
+            EgressArtifactResult result = await ext.EgressArtifact(payload, action, token);
+
+            if (!result.Succeeded)
             {
-                FileName = extensionProcessPath,
-                RedirectStandardInput = true,
-                RedirectStandardOutput = true,
-                RedirectStandardError = true,
-                CreateNoWindow = true,
-                UseShellExecute = false,
-            };
-            Process p = new Process()
-            {
-                StartInfo = pStart,
-            };
-
-            Logger?.LogInformation("Starting process...");
-            p.Start();
-            Logger?.LogInformation("starting stream...");
-
-            await JsonSerializer.SerializeAsync<ExtensionEgressPayload>(p.StandardInput.BaseStream, payload, options: null, token);
-            await p.StandardInput.WriteLineAsync();
-            await p.StandardInput.BaseStream.FlushAsync(token);
-
-            await action(p.StandardInput.BaseStream, token);
-            await p.StandardInput.BaseStream.FlushAsync(token);
-
-            Logger?.LogInformation("Stream done, sending close...");
-
-            p.StandardInput.Close();
-
-            Logger?.LogInformation("Waiting for exit...");
-
-            string procStdOutContent = await p.StandardOutput.ReadToEndAsync();
-            string procErrorContent = await p.StandardError.ReadToEndAsync();
-
-            await p.WaitForExitAsync(token);
-            Logger?.LogInformation($"Exited with code: {p.ExitCode}");
-
-            if (!string.IsNullOrWhiteSpace(procErrorContent))
-            {
-                Logger?.LogWarning($"Process Error output: {procErrorContent}");
+                throw new EgressException(Strings.ErrorMessage_EgressExtensionFailed);
             }
 
-            return procStdOutContent;
+            return result.ArtifactPath;
         }
     }
 }
--- a/src/Tools/dotnet-monitor/Extensibility/ExtensionDiscoverer.cs
+++ b/src/Tools/dotnet-monitor/Extensibility/ExtensionDiscoverer.cs
@@ -19,6 +19,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
             _logger = logger;
         }
 
+        /// <inheritdoc/>
         public IExtension FindExtension(string extensionMoniker)
         {
             _logger.ExtensionProbeStart(extensionMoniker);
--- a/src/Tools/dotnet-monitor/Extensibility/ExtensionDiscoverer.cs
+++ b/src/Tools/dotnet-monitor/Extensibility/ExtensionDiscoverer.cs
@@ -6,6 +6,8 @@ using Microsoft.Extensions.Logging;
 using System.Collections.Generic;
 using System.Linq;
 
+#nullable enable
+
 namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
 {
     internal class ExtensionDiscoverer : IExtensionDiscoverer
@@ -20,13 +22,13 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
         }
 
         /// <inheritdoc/>
-        public IExtension FindExtension(string extensionMoniker)
+        public TExtensionType? FindExtension<TExtensionType>(string extensionMoniker) where TExtensionType : class, IExtension
         {
             _logger.ExtensionProbeStart(extensionMoniker);
             foreach (IExtensionRepository repo in _extensionRepos)
             {
                 _logger.ExtensionProbeRepo(extensionMoniker, repo);
-                IExtension result = repo.FindExtension(extensionMoniker);
+                TExtensionType? result = repo.FindExtension<TExtensionType>(extensionMoniker);
                 if (result != null)
                 {
                     _logger.ExtensionProbeSucceeded(extensionMoniker, result);
--- a/src/Tools/dotnet-monitor/Extensibility/ExtensionDiscoverer.cs
+++ b/src/Tools/dotnet-monitor/Extensibility/ExtensionDiscoverer.cs
@@ -1,0 +1,39 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Extensions.Logging;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    internal class ExtensionDiscoverer : IExtensionDiscoverer
+    {
+        private readonly IOrderedEnumerable<IExtensionRepository> _extensionRepos;
+        private readonly ILogger<IExtensionDiscoverer> _logger;
+
+        public ExtensionDiscoverer(IEnumerable<IExtensionRepository> extensionRepos, ILogger<IExtensionDiscoverer> logger)
+        {
+            _extensionRepos = extensionRepos.OrderBy(eRepo => eRepo.ResolvePriority);
+            _logger = logger;
+        }
+
+        public IExtension FindExtension(string extensionMoniker)
+        {
+            _logger.ExtensionProbeStart(extensionMoniker);
+            foreach (IExtensionRepository repo in _extensionRepos)
+            {
+                _logger.ExtensionProbeRepo(extensionMoniker, repo);
+                IExtension result = repo.FindExtension(extensionMoniker);
+                if (result != null)
+                {
+                    _logger.ExtensionProbeSucceeded(extensionMoniker, result);
+                    return result;
+                }
+            }
+            _logger.ExtensionProbeFailed(extensionMoniker);
+            return null;
+        }
+    }
+}
--- a/src/Tools/dotnet-monitor/Extensibility/ExtensionRepository.cs
+++ b/src/Tools/dotnet-monitor/Extensibility/ExtensionRepository.cs
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+#nullable enable
+
 namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
 {
     internal abstract class ExtensionRepository : IExtensionRepository
@@ -19,6 +21,6 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
 
         public string Name => _name;
 
-        public abstract IExtension FindExtension(string extensionMoniker);
+        public abstract TExtensionType? FindExtension<TExtensionType>(string extensionMoniker) where TExtensionType : class, IExtension;
     }
 }
--- a/src/Tools/dotnet-monitor/Extensibility/ExtensionRepository.cs
+++ b/src/Tools/dotnet-monitor/Extensibility/ExtensionRepository.cs
@@ -1,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    internal abstract class ExtensionRepository : IExtensionRepository
+    {
+        private readonly int _resolvePriority;
+        private readonly string _name;
+
+        public ExtensionRepository(int resolvePriority, string name)
+        {
+            _resolvePriority = resolvePriority;
+            _name = name;
+        }
+
+        public int ResolvePriority => _resolvePriority;
+
+        public string Name => _name;
+
+        public abstract IExtension FindExtension(string extensionMoniker);
+    }
+}
--- a/src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
+++ b/src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
@@ -8,6 +8,8 @@ using System;
 using System.Globalization;
 using System.IO;
 
+#nullable enable
+
 namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
 {
     internal class FolderExtensionRepository : ExtensionRepository
@@ -27,26 +29,29 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
             _loggerFactory = loggerFactory;
         }
 
-        public override IExtension FindExtension(string extensionMoniker)
+        public override TExtensionType FindExtension<TExtensionType>(string extensionMoniker)
         {
             // Pass 1: Require an exact case sensitive match
-            IExtension exactMatch = ScanExtensionDir(extensionMoniker, StringComparison.Ordinal);
+            IExtension? exactMatch = ScanExtensionDir(extensionMoniker, StringComparison.Ordinal);
             if (exactMatch != null)
             {
-                return exactMatch;
+                TExtensionType resultExact = (TExtensionType)exactMatch;
+                return resultExact;
             }
 
             // Pass 2: Relax case sensitivity
-            IExtension caseVarianceMatch = ScanExtensionDir(extensionMoniker, StringComparison.OrdinalIgnoreCase);
+            IExtension? caseVarianceMatch = ScanExtensionDir(extensionMoniker, StringComparison.OrdinalIgnoreCase);
             if (caseVarianceMatch != null)
             {
-                return caseVarianceMatch;
+                TExtensionType result = (TExtensionType)caseVarianceMatch;
+                return result;
             }
 
-            return null;
+            // [TODO] figure out how to return null and keep the base class happy
+            throw new NotImplementedException();
         }
 
-        private IExtension ScanExtensionDir(string extensionMoniker, StringComparison comparisionType)
+        private IExtension? ScanExtensionDir(string extensionMoniker, StringComparison comparisionType)
         {
             IDirectoryContents extensionRoot = _fileSystem.GetDirectoryContents(_targetFolder);
 
--- a/src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
+++ b/src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.Logging;
 using System;
 using System.Globalization;
 using System.IO;
@@ -13,12 +14,17 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
     {
         private readonly string _targetFolder;
         private readonly IFileProvider _fileSystem;
+        private readonly ILoggerFactory _loggerFactory;
 
-        public FolderExtensionRepository(IFileProvider fileSystem, int resolvePriority, string targetFolder)
+        public FolderExtensionRepository(IFileProvider fileSystem, ILoggerFactory loggerFactory, int resolvePriority, string targetFolder)
             : base(resolvePriority, string.Format(CultureInfo.CurrentCulture, Strings.Message_FolderExtensionRepoName, targetFolder))
         {
             _fileSystem = fileSystem;
+
+            // This is the root that we will start looking at relative to _fileSystem, _fileSystem is already scoped to the correct folder, so we use an empty string
             _targetFolder = String.Empty;
+
+            _loggerFactory = loggerFactory;
         }
 
         public override IExtension FindExtension(string extensionMoniker)
@@ -55,7 +61,8 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
                     {
                         if (file.Exists && !file.IsDirectory && string.Equals(file.Name, expectedFile, comparisionType))
                         {
-                            return new ProgramExtension(Path.Combine(_targetFolder, extDir.Name, file.Name));
+                            ILogger<ProgramExtension> logger = _loggerFactory.CreateLogger<ProgramExtension>();
+                            return new ProgramExtension(Path.Combine(_targetFolder, extDir.Name, file.Name), logger);
                         }
                     }
                 }
--- a/src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
+++ b/src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
@@ -1,0 +1,74 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Extensions.FileProviders;
+using System;
+using System.Globalization;
+using System.IO;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    internal class FolderExtensionRepository : ExtensionRepository
+    {
+        private readonly string _targetFolder;
+        private readonly IFileProvider _fileSystem;
+
+        public FolderExtensionRepository(IFileProvider fileSystem, int resolvePriority, string targetFolder)
+            : base(resolvePriority, string.Format(CultureInfo.CurrentCulture, Strings.Message_FolderExtensionRepoName, targetFolder))
+        {
+            _fileSystem = fileSystem;
+            _targetFolder = String.Empty;
+        }
+
+        public override IExtension FindExtension(string extensionMoniker)
+        {
+            // Pass 1: Require an exact case sensitive match
+            IExtension exactMatch = ScanExtensionDir(extensionMoniker, StringComparison.Ordinal);
+            if (exactMatch != null)
+            {
+                return exactMatch;
+            }
+
+            // Pass 2: Relax case sensitivity
+            IExtension caseVarianceMatch = ScanExtensionDir(extensionMoniker, StringComparison.OrdinalIgnoreCase);
+            if (caseVarianceMatch != null)
+            {
+                return caseVarianceMatch;
+            }
+
+            return null;
+        }
+
+        private IExtension ScanExtensionDir(string extensionMoniker, StringComparison comparisionType)
+        {
+            IDirectoryContents extensionRoot = _fileSystem.GetDirectoryContents(_targetFolder);
+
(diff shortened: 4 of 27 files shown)


src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
```diff
@@ -0,0 +1,83 @@
...
+            if (caseVarianceMatch != null)
+            {
+                TExtensionType result = (TExtensionType)caseVarianceMatch;
+                return result;
+            }
+
+            return null;
+        }
+
+        private IExtension ScanExtensionDir(string extensionMoniker, StringComparison comparisionType)
```
kkeirstead, line 51: nit: `comparisonType` instead of `comparisionType`

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -544,6 +547,73 @@
...
+  <data name="LogFormatString_ExtensionProbeSucceeded" xml:space="preserve">
+    <value>Found extension '{extensionMoniker}' at {extension}.</value>
+    <comment>Gets the format string that is printed in the 63:ExtensionProbeSucceeded event.
+1 Format Parameter:
+1. extensionMoniker: The name of the extension or moniker
+2. extension: The friendly name of the extension from the IExtension interface</comment>
+  </data>
+  <data name="LogFormatString_ExtensionStarting" xml:space="preserve">
+    <value>Starting extension '{extensionPath} {cmdArgs}'.</value>
+    <comment>Gets the format string that is printed in the 65:ExtesnionStarting event.
```
kkeirstead, line 612:
```suggestion
    <comment>Gets the format string that is printed in the 65:ExtensionStarting event.
```

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
```diff
@@ -50,57 +47,15 @@ public ExtensionEgressProvider(IEgressPropertiesProvider propertyProvider, ILogg
...
-            string procStdOutContent = await p.StandardOutput.ReadToEndAsync();
-            string procErrorContent = await p.StandardError.ReadToEndAsync();
-
-            await p.WaitForExitAsync(token);
-            Logger?.LogInformation($"Exited with code: {p.ExitCode}");
+            IEgressExtension ext = _extensionDiscoverer.FindExtension<IEgressExtension>(providerType);
+            EgressArtifactResult result = await ext.EgressArtifact(payload, action, token);
 
-            if (!string.IsNullOrWhiteSpace(procErrorContent))
+            if (!result.Succeeded)
```
kkeirstead, line 53: Based on your TODO comment, I think you're still working on the `EgressArtifact` method (so you can feel free to just resolve this if it's not relevant), but depending on how that gets implemented, should we have a null check on `result` before trying to check if it succeeded?

src/Tools/dotnet-monitor/Extensibility/ExtensionDiscoverer.cs
```diff
@@ -0,0 +1,40 @@
...
+                _logger.ExtensionProbeRepo(extensionMoniker, repo);
+                TExtensionType result = repo.FindExtension<TExtensionType>(extensionMoniker);
+                if (result != null)
+                {
+                    _logger.ExtensionProbeSucceeded(extensionMoniker, result);
+                    return result;
+                }
+            }
+            _logger.ExtensionProbeFailed(extensionMoniker);
+            return null;
```
wiktork, line 37: Should this throw?

src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            _fileSystem = fileSystem;
+
+            // This is the root that we will start looking at relative to _fileSystem, _fileSystem is already scoped to the correct folder, so we use an empty string
+            _targetFolder = String.Empty;
+
+            _loggerFactory = loggerFactory;
+        }
+
+        public override TExtensionType FindExtension<TExtensionType>(string extensionMoniker) where TExtensionType : class
```
jander-msft, line 30:
```suggestion
        public override bool TryFindExtension<TExtensionType>(string extensionMoniker, out TExtensionType extension) where TExtensionType : class
```

src/Tools/dotnet-monitor/Extensibility/FileProvider/FolderFileProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        private IFileProvider BaseProvider
+        {
+            get
+            {
+                if (_fileProvider == null)
+                {
+                    if (Directory.Exists(_rootPath))
+                    {
+                        _fileProvider = new PhysicalFileProvider(_rootPath);
```
wiktork, line 30: Any unusual behavior on a mounted drive?

src/Tools/dotnet-monitor/Extensibility/FileProvider/FolderFileProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+            {
+                yield break;
+            }
+        }
+
+        private class MissingFile : IFileInfo
+        {
+            public bool Exists => false;
+
+            public bool IsDirectory => throw new NotImplementedException();
```
wiktork, line 97: FileNotFoundException?

src/Tools/dotnet-monitor/Extensibility/ExtensionRepository.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    internal abstract class ExtensionRepository : IExtensionRepository
+    {
+        private readonly int _resolvePriority;
+        private readonly string _name;
```
wiktork, line 10: Add comments to indicate name is a friendly name, not an identifier.

src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            // Pass 1: Require an exact case sensitive match
+            IExtension exactMatch = ScanExtensionDir(extensionMoniker, StringComparison.Ordinal);
+            if (exactMatch != null)
+            {
+                TExtensionType resultExact = (TExtensionType)exactMatch;
+                return resultExact;
+            }
+
+            // Pass 2: Relax case sensitivity
```
wiktork, line 40: Just use ordinal.

src/Tools/dotnet-monitor/Extensibility/IExtensionRepository.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    /// <summary>
+    /// Interface for service used to lookup an extension from a specific source
+    /// </summary>
+    internal interface IExtensionRepository
```
wiktork, line 10: Can we just keep the abstract class or interface? Or do we need both?

src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
```diff
@@ -0,0 +1,83 @@
...
+                    }
+                }
+            }
+
+            return null;
+        }
+
+        private string GetOSSpecificExecutableFormat(string executableName)
+        {
+            string format = "{0}.exe";
```
wiktork, line 78: How do we deal with exe stubs for dotnet extension implementations? Cross plat builds of extensions?

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
kelltrick: You are correct, I moved this TODO to `ProgramExtension.cs` which will actually handle `EgressArtifact`.

kelltrick: Based on the discussion last week I made this throw rather than returning `null`

src/Tools/dotnet-monitor/Extensibility/ExtensionRepository.cs
kelltrick: renamed these to `DisplayName`

src/Tools/dotnet-monitor/Extensibility/IExtensionRepository.cs
kelltrick: Removed the `IExtensionRepository` and used `ExtensionRepository` directly. Also removed `IExtensionDiscoverer` and replaced with using `ExtensionDiscoverer` directly.

src/Tools/dotnet-monitor/Extensibility/FileProvider/FolderFileProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+            }
+        }
+
+        private class MissingDirectoryContents : IDirectoryContents
+        {
+            public bool Exists => false;
+
+            public IEnumerator<IFileInfo> GetEnumerator()
+            {
+                yield break;
```
jander-msft, line 84: `return Enumerable.Empty<IFileInfo>().GetEnumerator();` might be faster.

src/Tools/dotnet-monitor/Extensibility/FileProvider/FolderFileProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+            }
+
+            IEnumerator IEnumerable.GetEnumerator()
+            {
+                yield break;
```
jander-msft, line 89: `return Enumerable.Empty<IFileInfo>().GetEnumerator();` might be faster.

src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
```diff
@@ -0,0 +1,48 @@
...
+            _targetFolder = targetFolder;
+
+            _loggerFactory = loggerFactory;
+        }
+
+        public override bool TryFindExtension(string extensionMoniker, out IExtension extension)
+        {
+            IFileInfo extensionDir = _fileSystem.GetFileInfo(extensionMoniker);
+
+            if (extensionDir.Exists && extensionDir.IsDirectory)
```
jander-msft, line 33: If you are not specially handling the negation of this case (e.g. logging when it doesn't exist), then it shouldn't be necessary because it'll be checked with the combined path below.

src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
```diff
@@ -0,0 +1,48 @@
...
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.Logging;
+using System.Globalization;
+using System.IO;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    internal class FolderExtensionRepository : ExtensionRepository
+    {
+        private const string ExtensionDefinitionFile = "extension.jsonc";
```
jander-msft, line 14: Just make it `extension.json`; JSONC isn't really a standard

src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            IFileInfo extensionDir = _fileSystem.GetFileInfo(extensionMoniker);
+
+            if (extensionDir.Exists && extensionDir.IsDirectory)
+            {
+                IFileInfo defFile = _fileSystem.GetFileInfo(Path.Combine(extensionMoniker, ExtensionDefinitionFile));
+                if (defFile.Exists && !defFile.IsDirectory)
+                {
+                    ILogger<ProgramExtension> logger = _loggerFactory.CreateLogger<ProgramExtension>();
+                    extension = new ProgramExtension(Path.Combine(_targetFolder, extensionMoniker, ExtensionDefinitionFile), logger);
```
jander-msft, line 39: I assume this is a temporary call to only return this extension. In the future, I would imagine that you are going to read the extension.json file, look up the extension name from the content, determine what type of extension it is, and return the corresponding implementation?

src/Tools/dotnet-monitor/Extensibility/OutputParser.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        private void ParseStdOut(object sender, DataReceivedEventArgs eventArgs)
+        {
+            if (eventArgs.Data != null)
+            {
+                try
+                {
+                    // Check if the object is a TResult
+                    TResult result = JsonSerializer.Deserialize<TResult>(eventArgs.Data);
+                    _resultCompletionSource.SetResult(result);
```
jander-msft, line 52: TrySetResult

src/Tools/dotnet-monitor/Extensibility/ExtensionDeclaration.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    internal class ExtensionDeclaration
```
jander-msft, line 5: This doesn't seem to be used

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
```diff
@@ -0,0 +1,96 @@
...
+            _logger.ExtensionStarting(_path, pStart.Arguments);
+            p.Start();
+            await JsonSerializer.SerializeAsync<ExtensionEgressPayload>(p.StandardInput.BaseStream, configPayload, options: null, token);
+            await p.StandardInput.WriteAsync(NewLine, token);
+            await p.StandardInput.BaseStream.FlushAsync(token);
+            _logger.ExtensionConfigured(_path, p.Id);
+
+            await getStreamAction(p.StandardInput.BaseStream, token);
+            await p.StandardInput.BaseStream.FlushAsync(token);
+            p.StandardInput.Close();
```
jander-msft, line 87: Is this necessary, especially after you've explicitly flushed the stream?

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            Process p = new Process()
+            {
+                StartInfo = pStart,
+            };
+
+            using OutputParser<EgressArtifactResult> parser = new(p, _logger);
+
+            _logger.ExtensionStarting(_path, pStart.Arguments);
+            p.Start();
```
jander-msft, line 79: You need to check if it actually started and throw an appropriate exception if it did not.

src/Tools/dotnet-monitor/Extensibility/OutputParser.cs
```diff
@@ -0,0 +1,75 @@
...
+
+using Microsoft.Extensions.Logging;
+using System;
+using System.Diagnostics;
+using System.Text.Json;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    internal class OutputParser<TResult> : IDisposable where TResult : class
```
jander-msft, line 13: Is this going to be used in other places? If not, inline it with the ProgramExtension class.

jander-msft, line 13: Much of this could be replaced with EventTaskSource: https://github.com/dotnet/diagnostics/blob/main/src/Microsoft.Diagnostics.Monitoring.EventPipe/EventTaskSource.cs

Something like:
```cs
EgressArtifactResult result;
using EventTaskSource<DataReceivedEventHandler> resultSource = new(
    complete => (s, e) => { ... }, // This does the deserialization and invokes complete when the deserialization successfully finishes
    handler => process.OutputDataReceived += handler,
    handler => process.OutputDataReceived -= handler
)

...

await resultSource.Task;

```

src/Tools/dotnet-monitor/Extensibility/PreDefinedExtensionRepositories.cs
```diff
@@ -0,0 +1,69 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Reflection;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    internal static class PreDefinedExtensionRepositories
+    {
+        public static IEnumerable<Func<IServiceProvider, IExtensionRepository>> GetExtensionRepositoryDelegates(HostBuilderSettings settings)
```
jander-msft, line 17: Rather than return a list of Func, add a `IServiceCollection` parameter and just add the services directly from this method. You could probably remove the `FolderExtensionRepoDefinition` notion. Maybe rename the method to `AddExtensionRepositories`.

src/Tools/dotnet-monitor/Extensibility/FileProvider/FolderFileProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.IO;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    /// <summary>
+    /// This class provides an implementation of <see cref="IFileProvider"/> with a root directory that does not need to exist at construction time.
+    /// </summary>
+    public class FolderFileProvider : IFileProvider
```
jander-msft, line 17: Please add unit tests

src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
```diff
@@ -0,0 +1,48 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.Logging;
+using System.Globalization;
+using System.IO;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    internal class FolderExtensionRepository : ExtensionRepository
```
jander-msft, line 12: Please add unit tests

src/Tools/dotnet-monitor/dotnet-monitor.csproj
```diff
@@ -77,9 +77,9 @@
     </ItemGroup>
     <ItemGroup>
       <!-- Pack the profiler library for each platform if it exists. -->
-      <TfmSpecificPackageFile Include="@(MonitorProfilerLibraryFile->Exists())" />
+      <TfmSpecificPackageFile Include="@(MonitorProfilerLibraryFile-&gt;Exists())" />
```
jander-msft, line 80: Please revert these changes

src/Tools/dotnet-monitor/Extensibility/ExtensionDiscoverer.cs
```diff
@@ -0,0 +1,53 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Extensions.Logging;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    internal class ExtensionDiscoverer
```
jander-msft, line 11: Please add unit tests

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
```diff
@@ -0,0 +1,96 @@
...
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Text.Json;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    internal class ProgramExtension : IExtension, IEgressExtension
```
jander-msft, line 16: Please add unit tests

jander-msft (changes requested):

src/Tools/dotnet-monitor/Extensibility/ExtensionDiscoverer.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// <exception cref="ExtensionNotFoundException"></exception>
+        public TExtensionType FindExtension<TExtensionType>(string extensionMoniker) where TExtensionType : class, IExtension
+        {
+            _logger.ExtensionProbeStart(extensionMoniker);
+            foreach (ExtensionRepository repo in _extensionRepos)
+            {
+                bool found = repo.TryFindExtension(extensionMoniker, out IExtension genericResult);
+                if (found)
+                {
+                    bool isOfType = genericResult.TryGetTypedExtension(out TExtensionType result);
```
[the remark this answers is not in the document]
kelltrick: Because the extension declaration will declare if it supports the given type. While `ProgramExtension` will support all the interfaces, the specific instance may not support that mode, hence the method.

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
kelltrick: Yep, this is necessary. This close is what notifies the extension that the stream is complete.

src/Tools/dotnet-monitor/Extensibility/PreDefinedExtensionRepositories.cs
```diff
@@ -0,0 +1,69 @@
...
+                            new FolderFileProvider(folderProvider.ExtensionDirectory),
+                            serviceProvider.GetRequiredService<ILoggerFactory>(),
+                            folderProvider.Priority,
+                            folderProvider.ExtensionDirectory));
+            }
+
+            return services;
+        }
+
+        private class FolderExtensionRepoDefinition
```
jander-msft, line 55: This class is not necessary. Just create a static local function that takes the service collection, the priority, and the path and does the servicing registration and call it three times.

src/Tools/dotnet-monitor/Extensibility/ExtensionDiscoverer.cs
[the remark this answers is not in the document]
jander-msft: Without an alternate usage that requires this, I would ask you to use `genericResult is TExtensionType result` for now.

src/Tools/dotnet-monitor/Extensibility/ExtensionLaunchFailure.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Globalization;
+using Microsoft.Diagnostics.Monitoring;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    internal class ExtensionLaunchFailure : MonitoringException
```
jander-msft, line 10: Let's have only one exception type (e.g. `ExtensionException : MonitoringException`) for the whole extension system. If you want to put static methods for each type of exception scenario on it, I would be fine with that e.g. `ExtensionException.ThrowNotFound(string name)` and `ExtensionException.ThrowStartFailure(string name)`.

src/Tools/dotnet-monitor/Extensibility/ExtensionRepository.cs
```diff
@@ -0,0 +1,33 @@
...
+            DisplayName = displayName;
+        }
+
+        /// <summary>
+        /// Gets the integer that describes the order used to probe multiple repositories
+        /// </summary>
+        public int ResolvePriority { get; private init; }
+
+        /// <summary>
+        /// Gets a friendly name to describe this instance of an <see cref="IExtensionRepository"/>. This is used in Logs.
```
jander-msft, line 21: The `IExtensionRepository` interface doesn't exist anymore.

src/Tools/dotnet-monitor/Extensibility/FileProvider/FolderFileProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    /// <summary>
+    /// This class provides an implementation of <see cref="IFileProvider"/> with a root directory that does not need to exist at construction time.
+    /// </summary>
+    public class FolderFileProvider : IFileProvider
```
jander-msft, line 18: Rather than having a new class that completely defers to either the `PhysicalFileProvider` or `MissingDirectoryProvider`, could we have a static method that determines this and returns the appropriate one?

src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
jander-msft: For both your `MissingDirectoryProvider` and the `PhysicalFileProvider`, it will not throw when calling `GetFileInfo` with a directory that does not exist.

src/Tools/dotnet-monitor/Extensibility/FileProvider/FolderFileProvider.cs
kelltrick: Adding with next PR

src/Tools/dotnet-monitor/Extensibility/ExtensionRepository.cs
```diff
@@ -0,0 +1,36 @@
...
+        public ExtensionRepository(int resolvePriority, string displayName)
+        {
+            _resolvePriority = resolvePriority;
+            _displayName = displayName;
+        }
+
+        /// <summary>
+        /// Gets the integer that describes the order used to probe multiple repositories
+        /// </summary>
+        public int ResolvePriority => _resolvePriority;
```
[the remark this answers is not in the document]
kelltrick: That's a nice language feature, didn't know of these

src/Tools/dotnet-monitor/Extensibility/ExtensionDiscoverer.cs
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        /// <summary>
+        /// Attempts to locate an extension with the given moniker and return it in the provided type.
+        /// </summary>
+        /// <typeparam name="TExtensionType">The type of the extension that must be found.</typeparam>
+        /// <param name="extensionMoniker">The string moniker used to reffer to the extension</param>
+        /// <returns></returns>
+        /// <exception cref="ExtensionNotFoundException"></exception>
+        public TExtensionType FindExtension<TExtensionType>(string extensionMoniker) where TExtensionType : class, IExtension
```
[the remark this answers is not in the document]
kelltrick: `ExtensionDiscoverer` doesn't really care about the type, but having the `ExtensionDiscoverer` do the type query solves the problem of "you have 2 folders with the same named extension, where one extension does not support a given type and the second extension does support the given type". It also nicely puts some boilerplate code all in 1 function that you'll be using anyways.

src/Tools/dotnet-monitor/Extensibility/FileProvider/FolderFileProvider.cs
kelltrick: The reason for this class is for the `ExtensionRepository` to be initialized, it needs an `IFileProvider`, the `PhysicalFileProvider` checks upfront if the parent directory exists. This is undesirable, because we want the extension folder to be added after boot up; this class allows the folder to not exist.

kelltrick:
>For both your `MissingDirectoryProvider` and the `PhysicalFileProvider`, it will not throw when calling `GetFileInfo` with a directory that does not exist.
@jander-msft 

I don't love this behavior of `IFileProvider`. I'm following the behavior as described in the doc: https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.fileproviders.ifileprovider.getfileinfo?view=dotnet-plat-ext-6.0  Although it does not explicitly call out the difference between directory doesn't exist and file doesn't exist.

src/Tools/dotnet-monitor/Extensibility/FileProvider/FolderFileProvider.cs
jander-msft: > because we want the extension folder to be added after boot up

Why? It's much easier to require that the extension folder exists before starting dotnet-monitor.

If we get into the business of allowing extensions to be added during dotnet-monitor lifetime, wouldn't you expect to be able to remove them as well? What if you delete the content of the extension folder and add new content? I don't think we want to have to deal with various "extension did or did not exist now, and may or may not exist later" scenarios.

kelltrick: Sure, I got rid of this whole thing and just use the integrated physical providers as needed.

kelltrick:
>Without an alternate usage that requires this, I would ask you to use genericResult is TExtensionType result for now.
- @jander-msft 

I will also got rid of the `TryGetTypedExtension` method and add a check when the `EgressArtifact(..)` method is called

src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
kelltrick: This is all gone now.

src/Tools/dotnet-monitor/Extensibility/ExtensionDiscoverer.cs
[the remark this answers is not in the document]
kelltrick: Removed.

src/Tools/dotnet-monitor/Extensibility/ExtensionDeclaration.cs
```diff
@@ -0,0 +1,22 @@
...
+        /// </summary>
+        public string Id { get; set; }
+
+        /// <summary>
+        /// The version of the extension. This should be SemVer 2.0.
+        /// </summary>
+        public Version Version { get; set; }
+
+        /// <summary>
+        /// This is the command that will be executed in the
```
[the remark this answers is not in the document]
kelltrick: Fixed.

src/Tools/dotnet-monitor/Extensibility/FileProvider/FolderFileProvider.cs
kelltrick: Not that I'm aware of right now, but I'll test this further when I add tests.

src/Tools/dotnet-monitor/Extensibility/ExtensionDiscoverer.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    internal class ExtensionDiscoverer
+    {
+        private readonly IOrderedEnumerable<ExtensionRepository> _extensionRepos;
+        private readonly ILogger<ExtensionDiscoverer> _logger;
+
+        public ExtensionDiscoverer(IEnumerable<ExtensionRepository> extensionRepos, ILogger<ExtensionDiscoverer> logger)
+        {
+            _extensionRepos = extensionRepos.OrderBy(eRepo => eRepo.ResolvePriority);
```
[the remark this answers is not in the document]
kelltrick: Converted this to `Array`.

src/Tools/dotnet-monitor/LoggingExtensions.cs
```diff
@@ -333,6 +335,90 @@ internal static class LoggingExtensions
...
+        private static readonly Action<ILogger, string, string, Exception> _extensionProbeSucceeded =
+            LoggerMessage.Define<string, string>(
+                eventId: LoggingEventIds.ExtensionProbeSucceeded.EventId(),
+                logLevel: LogLevel.Information,
+                formatString: Strings.LogFormatString_ExtensionProbeSucceeded);
+
+        private static readonly Action<ILogger, string, Exception> _extensionProbeFailed =
+            LoggerMessage.Define<string>(
+                eventId: LoggingEventIds.ExtensionProbeFailed.EventId(),
+                logLevel: LogLevel.Warning,
```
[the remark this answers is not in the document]
kelltrick: Sure, I changed all these to `Error`s

src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        private const string ExtensionDefinitionFile = "extension.json";
+        private readonly string _targetFolder;
+        private readonly IFileProvider _fileSystem;
+        private readonly ILoggerFactory _loggerFactory;
+
+        public FolderExtensionRepository(IFileProvider fileSystem, ILoggerFactory loggerFactory, int resolvePriority, string targetFolder)
+            : base(resolvePriority, string.Format(CultureInfo.CurrentCulture, Strings.Message_FolderExtensionRepoName, targetFolder))
```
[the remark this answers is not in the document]
kelltrick: I tried, `GetFileInfo(".")` and `GetFileInfo("")` return a null `PhysicalPath` and empty `Name` parameter. This works if there is something inside the folder but not for the root.

src/Tools/dotnet-monitor/Strings.resx
kelltrick: Fixing in next PR

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        private ExtensionDeclaration GetExtensionDeclaration()
+        {
+            try
+            {
+                IFileInfo declFile = _fileSystem.GetFileInfo(_declarationPath);
+                if (!declFile.Exists || declFile.IsDirectory)
+                {
+                    LogBrokenDeclaration(null);
```
jander-msft, line 128: This should be removed since the filter clause already does this

jander-msft (approved):

jander-msft (approved):

src/Tools/dotnet-monitor/Extensibility/OutputParser.cs
kelltrick: The idea is this will hold all parsing logic in the future.

PR #1838, opened 2022-05-03
Do not warn about unconfigured auth when --no-auth is used

### LINKED ISSUE #1624: Do not warn about unconfigured auth when `--no-auth` is used
### Description

When passing the `--no-auth` flag, `dotnet monitor` warns that "API key authentication is not configured. See https://go.microsoft.com/fwlink/?linkid=2187444\u0026clcid=0x409 for more details."

### Expected

When explicitly not using authentication, this warning should not be shown.

### Configuration

7.0 Preview 2 and higher

### Regression?

Yes

### LINKED ISSUE #1625: First startup of dotnet-monitor includes MonitorApiKey configuration info
First output of:
`dotnet-monitor collect --no-auth`

Is:
```
03:46:48 info: Microsoft.Diagnostics.Tools.Monitor.MonitorApiKeyConfigurationObserver[25]
MonitorApiKey settings have changed. The new settings have passed validation.
03:46:48 warn: Microsoft.Diagnostics.Tools.Monitor.Startup[13]
WARNING: Authentication has been disabled. This can pose a security risk and is not intended for production environments.
03:46:48 warn: Microsoft.AspNetCore.Server.Kestrel[0]
Overriding address(es) '[http://localhost:5678'.]() Binding to endpoints defined via IConfiguration and/or UseKestrel() instead.
03:46:48 info: Microsoft.Hosting.Lifetime[14]
Now listening on: http://localhost:5678/
03:46:48 info: Microsoft.Hosting.Lifetime[14]
Now listening on: http://localhost:52325/
03:46:48 info: Microsoft.Hosting.Lifetime[0]
03:46:48 info: Microsoft.Hosting.Lifetime[0]
03:46:48 info: Microsoft.Hosting.Lifetime[0]
Content root path: /home/noah/.dotnet/tools/.store/dotnet-monitor/6.1.0/dotnet-monitor/6.1.0/tools/net6.0/any/
```

We should get rid of the 
```
03:46:48 info: Microsoft.Diagnostics.Tools.Monitor.MonitorApiKeyConfigurationObserver[25]
MonitorApiKey settings have changed. The new settings have passed validation.
```

kkeirstead:
Introduces a no-auth check to prevent the warning from appearing when `--no-auth` is present. ~~**NOTE:** I kept this in-line with the pre-regression behavior where "MonitorApiKey settings have changed. The new settings have passed validation." is printed when the no-auth flag is present. If we don't want this, I can modify the check to only show that message when the user omits the flag.~~

Closes #1624
Closes #1625

### DIFF
--- a/src/Tools/dotnet-monitor/Auth/MonitorApiKeyConfigurationObserver.cs
+++ b/src/Tools/dotnet-monitor/Auth/MonitorApiKeyConfigurationObserver.cs
@@ -33,10 +33,10 @@ namespace Microsoft.Diagnostics.Tools.Monitor
 
         public void Initialize()
         {
-            _changeRegistration = _options.OnChange(OnMonitorApiKeyOptionsChanged);
-
             if (_authConfigurationOptions.KeyAuthenticationMode != KeyAuthenticationMode.NoAuth)
             {
+                _changeRegistration = _options.OnChange(OnMonitorApiKeyOptionsChanged);
+
                 // Write out current validation state of options when starting the tool.
                 CheckMonitorApiKeyOptions(_options.CurrentValue);
             }
--- a/src/Tools/dotnet-monitor/Auth/MonitorApiKeyConfigurationObserver.cs
+++ b/src/Tools/dotnet-monitor/Auth/MonitorApiKeyConfigurationObserver.cs
@@ -64,7 +64,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor
                     _logger.ApiKeyAuthenticationOptionsValidated();
                 }
             }
-            else if (!options.Configured && !noAuthEnabled)
+            else
             {
                 _logger.MonitorApiKeyNotConfigured();
             }
--- a/src/Tools/dotnet-monitor/Auth/MonitorApiKeyConfigurationObserver.cs
+++ b/src/Tools/dotnet-monitor/Auth/MonitorApiKeyConfigurationObserver.cs
@@ -35,8 +35,11 @@ namespace Microsoft.Diagnostics.Tools.Monitor
         {
             _changeRegistration = _options.OnChange(OnMonitorApiKeyOptionsChanged);
 
-            // Write out current validation state of options when starting the tool.
-            CheckMonitorApiKeyOptions(_options.CurrentValue);
+            if (_authConfigurationOptions.KeyAuthenticationMode != KeyAuthenticationMode.NoAuth)
+            {
+                // Write out current validation state of options when starting the tool.
+                CheckMonitorApiKeyOptions(_options.CurrentValue);
+            }
         }
 
         public void Dispose()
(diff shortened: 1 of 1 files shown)


kkeirstead:
I just saw #1625 - if I'm interpreting that correctly, I should be preventing "MonitorApiKey settings have changed. The new settings have passed validation." from printing when `--no-auth` is passed. I'll make that change now.

jander-msft:
> I just saw #1625 - if I'm interpreting that correctly, I should be preventing "MonitorApiKey settings have changed. The new settings have passed validation." from printing when `--no-auth` is passed. I'll make that change now.

Not quite right... we shouldn't show "MonitorApiKey settings have changed. The new settings have passed validation." when the tool is started because the settings have not changed. Subsequent auth changes should show this message.

jander-msft:
> > > I just saw #1625 - if I'm interpreting that correctly, I should be preventing "MonitorApiKey settings have changed. The new settings have passed validation." from printing when `--no-auth` is passed. I'll make that change now.
> > 
> > 
> > Not quite right... we shouldn't show "MonitorApiKey settings have changed. The new settings have passed validation." when the tool is started because the settings have not changed. Subsequent auth changes should show this message.
> 
> Are you sure? If we have the `--no-auth` flag, my understanding from some quick debugging through it is that we don't run validation - in which case, we don't actually validate those changes since we aren't using auth.

I suppose you are right.

I also want us to fix the fact that "MonitorApiKey settings have changed." is shown even though settings haven't changed, but I suppose that should be done separately.

src/Tools/dotnet-monitor/Auth/MonitorApiKeyConfigurationObserver.cs
```diff
@@ -49,6 +51,11 @@ private void OnMonitorApiKeyOptionsChanged(MonitorApiKeyConfiguration options)
 
         private void CheckMonitorApiKeyOptions(MonitorApiKeyConfiguration options)
         {
+            if (_authConfigurationOptions.KeyAuthenticationMode == KeyAuthenticationMode.NoAuth)
```
jander-msft, line 54: I think you can move this to the Initialize method so that way it doesn't even have to listen to the changed at all.

src/Tools/dotnet-monitor/Auth/MonitorApiKeyConfigurationObserver.cs
```diff
@@ -17,24 +17,29 @@ internal class MonitorApiKeyConfigurationObserver :
...
+            _authConfigurationOptions = authConfigurationOptions;
         }
 
         public void Initialize()
         {
             _changeRegistration = _options.OnChange(OnMonitorApiKeyOptionsChanged);
 
-            // Write out current validation state of options when starting the tool.
-            CheckMonitorApiKeyOptions(_options.CurrentValue);
+            if (_authConfigurationOptions.KeyAuthenticationMode != KeyAuthenticationMode.NoAuth)
```
jander-msft, line 38: Because "no auth vs auth" configuration cannot change during the lifetime of `dotnet monitor`, I would also push the `_changeRegistration = ...` line inside the condition. There's no point in monitoring for authentication changes if the tool is running in no-auth mode.

jander-msft (approved):

kelltrick (approved):

PR #1853, opened 2022-05-05
Unhandled exception detection

jander-msft:
These changes add the ability to detect unhandled exceptions in the most commons cases. The algorithm is described (along with its known caveats) and coded into the [ExceptionTracker class](https://github.com/dotnet/dotnet-monitor/compare/release/6.x...jander-msft:dev/jander/exceptions?expand=1#diff-5eb9c3b5e67efbf830637957e3fa121a43d4ebdb4cde8b9d23ad566a7cccf518). No action is taken in these changes when an unhandled exception does occur; these changes are purely about detection.

### DIFF
--- a/src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
+++ b/src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
@@ -201,28 +201,30 @@ HRESULT ExceptionTracker::GetFullyQualifiedMethodName(FunctionID functionId, tst
         (IUnknown**)&pMetadataImport
     ));
 
-    // Get Module Name
-    ULONG count = 0;
+    // Get Module Name: typically the full path to the assembly
+    ULONG moduleNameCount = 0; // Includes null-terminater
     IfFailLogRet(m_pCorProfilerInfo->GetModuleInfo(
         moduleId,
         nullptr,
         0,
-        &count,
+        &moduleNameCount,
         nullptr,
         nullptr
     ));
 
-    unique_ptr<WCHAR[]> pModulePath(new (nothrow) WCHAR[count + 1]);
+    unique_ptr<WCHAR[]> pModulePath(new (nothrow) WCHAR[moduleNameCount]);
+    IfNullRet(pModulePath);
+
     IfFailLogRet(m_pCorProfilerInfo->GetModuleInfo(
         moduleId,
         nullptr,
-        count + 1,
+        moduleNameCount,
         nullptr,
         pModulePath.get(),
         nullptr
     ));
 
-    // Get Class Name
+    // Get Class Name: The namespace + type name in the form <Namespace>.<Type>
     tstring className;
     if (0 == classId)
     {
@@ -238,20 +240,23 @@ HRESULT ExceptionTracker::GetFullyQualifiedMethodName(FunctionID functionId, tst
             &typeDef
         ));
 
+        ULONG classNameCount = 0; // Includes null-terminater
         IfFailLogRet(pMetadataImport->GetTypeDefProps(
             typeDef,
             nullptr,
             0,
-            &count,
+            &classNameCount,
             nullptr,
             nullptr
         ));
 
-        unique_ptr<WCHAR[]> pClassName(new (nothrow) WCHAR[count + 1]);
+        unique_ptr<WCHAR[]> pClassName(new (nothrow) WCHAR[classNameCount]);
+        IfNullRet(pClassName);
+
         IfFailLogRet(pMetadataImport->GetTypeDefProps(
             typeDef,
             pClassName.get(),
-            count + 1,
+            classNameCount,
             nullptr,
             nullptr,
             nullptr
@@ -261,12 +266,13 @@ HRESULT ExceptionTracker::GetFullyQualifiedMethodName(FunctionID functionId, tst
     }
 
     // Get Method Name
+    ULONG methodNameCount = 0; // Includes null-terminater
     IfFailLogRet(pMetadataImport->GetMethodProps(
         methodDef,
         nullptr,
         nullptr,
         0,
-        &count,
+        &methodNameCount,
         nullptr,
         nullptr,
         nullptr,
@@ -274,12 +280,14 @@ HRESULT ExceptionTracker::GetFullyQualifiedMethodName(FunctionID functionId, tst
         nullptr
     ));
 
-    unique_ptr<WCHAR[]> pMethodName(new (nothrow) WCHAR[count + 1]);
+    unique_ptr<WCHAR[]> pMethodName(new (nothrow) WCHAR[methodNameCount]);
+    IfNullRet(pMethodName);
+
     IfFailLogRet(pMetadataImport->GetMethodProps(
         methodDef,
         nullptr,
         pMethodName.get(),
-        count + 1,
+        methodNameCount,
         nullptr,
         nullptr,
         nullptr,
@@ -290,6 +298,8 @@ HRESULT ExceptionTracker::GetFullyQualifiedMethodName(FunctionID functionId, tst
 
     tstring modulePath(pModulePath.get());
 
+    // The full method name should be in the following format: <ModuleName>!<Namespace>.<TypeName>.<MethodName>
+    // Example: ConsoleApp1.dll!ConsoleApp1.Program.Main
     fullMethodName.clear();
     fullMethodName.append(modulePath.substr(modulePath.find_last_of(_T("/\\")) + 1));
     fullMethodName.append(_T("!"));
--- a/src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
+++ b/src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
@@ -41,13 +41,11 @@ using namespace std;
     Unhandled Exception Detection Algorithm
 
     A subset of the aboved described callbacks is used to determine if an unhandled exception has occurred:
-    - ExceptionThrown: Record the originating FunctionID and ObjectID of the exception that is thrown.
+    - ExceptionThrown: Record the originating FunctionID and the existance of the exception.
     - ExceptionSearchCatcherFound: Record that the exception will be handled in a catching FunctionID.
     - ExceptionUnwindFunctionEnter: If the exception was not handled (ExceptionSearchCatcherFound was not invoked),
       then the exception is unhandled at this point. Do some extra bookkeeping in the case when the exception is
-      handled so that the exception is cleared once frame with the corresponding catching FunctionID is encountered.
-
-    Additionally, use GC callbacks, such as MovedReferences#, to update the exception ObjectID when compacting occurs.
+      handled so that the exception is cleared once the frame with the corresponding catching FunctionID is encountered.
 
     Current pitfalls:
     - Algorithm doesn't account for exceptions thrown within exception filters. These will be seen as a new set of
--- a/src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
+++ b/src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
@@ -7,6 +7,7 @@
 using namespace std;
 
 #define IfFailLogRet(EXPR) IfFailLogRet_(_logger, EXPR)
+#define IfFalseLogRet(EXPR, hr) IfFalseLogRet_(_logger, EXPR, hr)
 
 #define LogDebugV(format, ...) LogDebugV_(_logger, format, __VA_ARGS__)
 #define LogInformationV(format, ...) LogInformationV_(_logger, format, __VA_ARGS__)
@@ -71,14 +72,16 @@ ExceptionTracker::ExceptionTracker(
 void ExceptionTracker::AddProfilerEventMask(DWORD& eventsLow)
 {
     eventsLow |= COR_PRF_MONITOR::COR_PRF_ENABLE_STACK_SNAPSHOT;
-    eventsLow |= COR_PRF_MONITOR::COR_PRF_MONITOR_GC;
 }
 
 HRESULT ExceptionTracker::ExceptionThrown(ThreadID threadId, ObjectID objectId)
 {
+    // CAUTION: Do not store the exception ObjectID. It is not guaranteed to be correct
+    // outside of the ExceptionThrown callback without updating it using GC callbacks.
+
     HRESULT hr = S_OK;
 
-    IfFailLogRet(_threadDataManager->SetExceptionObject(threadId, objectId));
+    IfFailLogRet(_threadDataManager->SetHasException(threadId));
 
     // Exception throwing is common; don't pay to calculate method name if it won't be logged.
     if (_logger->IsEnabled(LogLevel::Debug))
@@ -119,9 +122,10 @@ HRESULT ExceptionTracker::ExceptionUnwindFunctionEnter(ThreadID threadId, Functi
 {
     HRESULT hr = S_OK;
 
-    ObjectID thrownObjectId;
-    FunctionID catcherFunctionId;
-    IfFailLogRet(_threadDataManager->GetException(threadId, &thrownObjectId, &catcherFunctionId));
+    bool hasException = false;
+    FunctionID catcherFunctionId = ThreadData::NoFunctionId;
+    IfFailLogRet(_threadDataManager->GetException(threadId, &hasException, &catcherFunctionId));
+    IfFalseLogRet(hasException, E_UNEXPECTED);
 
     if (ThreadData::NoFunctionId == catcherFunctionId)
     {
@@ -145,15 +149,6 @@ HRESULT ExceptionTracker::ExceptionUnwindFunctionEnter(ThreadID threadId, Functi
             LogDebugV(_T("Exception handled: %s"), methodName.c_str());
         }
     }
-
-    return S_OK;
-}
-
-HRESULT ExceptionTracker::MovedReferences(ULONG cMovedObjectIDRanges, ObjectID oldObjectIDRangeStart[], ObjectID newObjectIDRangeStart[], SIZE_T cObjectIDRangeLength[])
-{
-    HRESULT hr = S_OK;
-
-    IfFailLogRet(_threadDataManager->MovedReferences(cMovedObjectIDRanges, oldObjectIDRangeStart, newObjectIDRangeStart, cObjectIDRangeLength));
 
     return S_OK;
 }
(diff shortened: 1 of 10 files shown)


src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
```diff
@@ -0,0 +1,301 @@
...
+        nullptr,
+        nullptr
+    ));
+
+    unique_ptr<WCHAR[]> pMethodName(new (nothrow) WCHAR[count + 1]);
+    IfFailLogRet(pMetadataImport->GetMethodProps(
+        methodDef,
+        nullptr,
+        pMethodName.get(),
+        count + 1,
```
kkeirstead, line 282: Feel free to ignore this (and any other comments) since I'm not used to reading through these types of patterns - so, I believe I understand how these three 'get' sections are working after looking through some docs, but I realized we keep reusing the same `count` variable for all of them. It appears that we only read from it directly after getting its updated value for the length of a module/class/method name, but it did introduce some confusion as to what the responsibility of `count` was since it's being used for unrelated(ish) purposes - depending where you look in the file, it can _really_ mean `moduleNameLength`, `classNameLength`, or `methodNameLength` (I think). Just an idea for readability, unless there's a reason not to do it.

src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
```diff
@@ -0,0 +1,301 @@
...
+    IfFailLogRet(m_pCorProfilerInfo->GetModuleInfo(
+        moduleId,
+        nullptr,
+        0,
+        &count,
+        nullptr,
+        nullptr
+    ));
+
+    unique_ptr<WCHAR[]> pModulePath(new (nothrow) WCHAR[count + 1]);
```
kkeirstead, line 215: I'm assuming the `+1` is to be null-terminated - and then I'm also assuming that `&count` will only be the number of _actual_ characters in the name, and is not null-terminated?

src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
```diff
@@ -0,0 +1,301 @@
...
+        nullptr,
+        nullptr,
+        nullptr,
+        nullptr,
+        nullptr
+    ));
+
+    tstring modulePath(pModulePath.get());
+
+    fullMethodName.clear();
```
kkeirstead, line 293: Maybe add a comment showing what an expected `fullMethodName` is supposed to look like to help visually double check that this constructs it correctly?

src/MonitorProfiler/MainProfiler/ExceptionTracker.h
```diff
@@ -0,0 +1,50 @@
...
+    std::shared_ptr<ThreadDataManager> m_pThreadDataManager;
+
+public:
+    ExceptionTracker(
+        const std::shared_ptr<ILogger>& pLogger,
+        const std::shared_ptr<ThreadDataManager> pThreadDataManager,
+        ICorProfilerInfo2* pCorProfilerInfo);
+
+    /// <summary>
+    /// Adds profiler event masks need by class.
```
kkeirstead, line 29:
```suggestion
    /// Adds profiler event masks needed by class.
```

src/MonitorProfiler/MainProfiler/MainProfiler.h
```diff
@@ -32,3 +42,4 @@ class MainProfiler final :
 private:
     std::unique_ptr<CommandServer> _commandServer;
```
kkeirstead, line 43: Out of curiosity, why do we have multiple `private` keywords?

jander-msft: Because @wiktork added fields from his PR to the bottom when I added them to the top with the initial merge. We should agree to some standard, maybe enforced by something like an *.editorconfig file.

src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
jander-msft: Yes, these could be separate locals, but they are not used outside of the scope of allocating the array for each type of information we get. I don't think there is any actual noticeable impact if there is one vs three of these. I don't feel strongly either way, but if it makes it easier to read, then I can separate them.

src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
jander-msft: The documentation is not exactly clear. I found other examples where the extra character is not necessary and some empirical evidence shows that it is not necessary. I've removed the extra character calculation and added comments that the null terminater is already accounted.

src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
```diff
@@ -0,0 +1,311 @@
...
+{
+    m_pCorProfilerInfo = pCorProfilerInfo;
+    m_pLogger = pLogger;
+    m_pThreadDataManager = pThreadDataManager;
+}
+
+void ExceptionTracker::AddProfilerEventMask(DWORD& eventsLow)
+{
+    eventsLow |= COR_PRF_MONITOR::COR_PRF_ENABLE_STACK_SNAPSHOT;
+    eventsLow |= COR_PRF_MONITOR::COR_PRF_MONITOR_GC;
```
wiktork, line 74: I am very concerned about GC Tracking performance. It has impact on the GC mode and other issues such as https://github.com/dotnet/runtime/issues/53326.

jander-msft: I understand the concerns here.

However, we need to be able to keep a valid ObjectID for the thrown exceptions. The only callback that runs cooperatively is ExceptionThrown, which means that compacting GC can occur during the other callbacks (they are preemptive). I have an easy repro that demonstrates a thrown exception is moved as part of a compacting GC while it is unwinding a non-async call stack.

We also need to track the ObjectID outside of ExceptionThrown so that in a future change I can flag the first occurrence of a thrown exception as interesting and likely want to ignore subsequent rethrows; these rethrows have the same ObjectID as the original throwing of the exception.

I'm completely fine with using some other mechanism for tracking this information so that we can avoid the GC callbacks (e.g. maybe we can pin the exception instance from the profiler callback), but am currently not sure what will work.

jander-msft: I've remove the storage of the exception ObjectID for now since it is not strictly necessary do this in order to detect that an unhandled exception has occurred.

src/MonitorProfiler/MainProfiler/ThreadDataManager.cpp
```diff
@@ -0,0 +1,116 @@
...
+}
+
+HRESULT ThreadDataManager::GetThreadData(ThreadID threadId, shared_ptr<ThreadData>& threadData)
+{
+    lock_guard<mutex> mapLock(_dataMapMutex);
+
+    DataMapIterator iterator = _dataMap.find(threadId);
+    if (iterator == _dataMap.end())
+    {
+        return E_FAIL;
```
jander-msft, line 110: E_UNEXPECTED

src/MonitorProfiler/MainProfiler/ExceptionTracker.h
```diff
@@ -0,0 +1,50 @@
...
+    /// </summary>
+    static void AddProfilerEventMask(DWORD& eventsLow);
+
+    // Exceptions
+    HRESULT ExceptionThrown(ThreadID threadId, ObjectID objectId);
+    HRESULT ExceptionSearchCatcherFound(ThreadID threadId, FunctionID functionId);
+    HRESULT ExceptionUnwindFunctionEnter(ThreadID threadId, FunctionID functionId);
+
+    // Garbage Collection
+    HRESULT MovedReferences(ULONG cMovedObjectIDRanges, ObjectID oldObjectIDRangeStart[], ObjectID newObjectIDRangeStart[], SIZE_T cObjectIDRangeLength[]);
```
jander-msft, line 39: remove

src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
```diff
@@ -0,0 +1,304 @@
...
+    - ExceptionSearchCatcherFound: Record that the exception will be handled in a catching FunctionID.
+    - ExceptionUnwindFunctionEnter: If the exception was not handled (ExceptionSearchCatcherFound was not invoked),
+      then the exception is unhandled at this point. Do some extra bookkeeping in the case when the exception is
+      handled so that the exception is cleared once the frame with the corresponding catching FunctionID is encountered.
+
+    Current pitfalls:
+    - Algorithm doesn't account for exceptions thrown within exception filters. These will be seen as a new set of
+      callback invocations starting with throwing, searching, and unwinding, however, the callstack does not unwind
+      out of the exception filter and the original exception processing is resumed.
+    - Algorithm doesn't account for exceptions thrown within finally blocks. Presumably, these will be seen as new set
```
jander-msft, line 54: Also called eclipsing

src/MonitorProfiler/MainProfiler/ThreadDataManager.cpp
```diff
@@ -0,0 +1,116 @@
...
+    IfFailLogRet(threadData->SetHasException());
+
+    return S_OK;
+}
+
+HRESULT ThreadDataManager::SetExceptionCatcherFunction(ThreadID threadId, FunctionID catcherFunctionId)
+{
+    HRESULT hr = S_OK;
+
+    shared_ptr<ThreadData> threadData;
```
wiktork, line 93: Let's test this to make sure our locking doesn't have big impact on many concurrent threads.

src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
```diff
@@ -0,0 +1,304 @@
...
+HRESULT ExceptionTracker::ExceptionThrownStackSnapshotCallback(
+    FunctionID funcId,
+    UINT_PTR ip,
+    COR_PRF_FRAME_INFO frameInfo,
+    ULONG32 contextSize,
+    BYTE context[],
+    void* clientData)
+{
+    // Callback is only used to get the function ID of the top frame.
+    *static_cast<FunctionID*>(clientData) = funcId;
```
wiktork, line 163: If you use GetClassIDInfo2 in this callback with the frameinfo, you can probably get better information.

wiktork: This will also handle nested classes.

src/MonitorProfiler/MainProfiler/ThreadData.h
```diff
@@ -0,0 +1,37 @@
...
+/// Class representing common data for a single thread.
+/// </summary>
+class ThreadData
+{
+public:
+    static const FunctionID NoFunctionId = 0;
+
+private:
+    FunctionID _exceptionCatcherFunctionId;
+    ObjectID _hasException;
```
jander-msft, line 23: bool

src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
jander-msft: Done

src/MonitorProfiler/MainProfiler/ThreadData.h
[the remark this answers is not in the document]
jander-msft: Fixed

src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
```diff
@@ -0,0 +1,392 @@
...
+    ICorProfilerInfo2* corProfilerInfo)
+{
+    _corProfilerInfo = corProfilerInfo;
+    _logger = logger;
+    _threadDataManager = threadDataManager;
+}
+
+void ExceptionTracker::AddProfilerEventMask(DWORD& eventsLow)
+{
+    eventsLow |= COR_PRF_MONITOR::COR_PRF_ENABLE_STACK_SNAPSHOT;
```
jander-msft, line 72: @wiktork, we only use this when the logging level is Debug or lower. Do you think there is any benefit to dynamically adding this when we only need it? This might be moot if the stacks feature is using it unconditionally.

src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
```diff
@@ -0,0 +1,392 @@
...
+    // outside of the ExceptionThrown callback without updating it using GC callbacks.
+
+    HRESULT hr = S_OK;
+
+    IfFailLogRet(_threadDataManager->SetHasException(threadId));
+
+    // Exception throwing is common; don't pay to calculate method name if it won't be logged.
+    if (_logger->IsEnabled(LogLevel::Debug))
+    {
+        FunctionID functionId = 0;
```
kkeirstead, line 87: Are we using this anymore? Or is it just a remnant from a prior iteration?

src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
```diff
@@ -0,0 +1,392 @@
...
+        hr = _corProfilerInfo->DoStackSnapshot(
+            threadId,
+            LogExceptionThrownFrameCallback,
+            COR_PRF_SNAPSHOT_INFO::COR_PRF_SNAPSHOT_DEFAULT,
+            this,
+            nullptr,
+            0);
+
+        if (FAILED(hr) && hr != CORPROF_E_STACKSNAPSHOT_ABORTED)
```
kkeirstead, line 96: If `hr` is `CORPROF_E_STACKSNAPSHOT_ABORTED`, we currently just silently return `S_OK` - without knowing the full context of the `DoStackSnapshot` method, would it make sense to log anything when this situation occurs?

jander-msft: No, we intentionally ending the callback by returning `S_FALSE`, which DoStackSnapshot will return `CORPROF_E_STACKSNAPSHOT_ABORTED`, which is intended.

src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
jander-msft: Seems to be a remnant; I will remove in the next iteration

src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+    tstring methodName;
+    IfFailLogRet(GetFullyQualifiedMethodName(functionId, frameInfo, methodName));
+    LogDebugV(_T("Exception thrown: %s"), methodName.c_str());
+
+    return S_OK;
+}
+
+HRESULT ExceptionTracker::LogExceptionThrownFrameCallback(
+    FunctionID funcId,
```
kkeirstead, line 372: nit: functionId for consistency

src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
```diff
@@ -0,0 +1,392 @@
...
+    UINT_PTR ip,
+    COR_PRF_FRAME_INFO frameInfo,
+    ULONG32 contextSize,
```
kkeirstead, line 375: What are `ip` and `contextSize` doing?

src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
wiktork: I'm not aware of any significant perf impact from this flag, but I think as we add more features here we may end up with configuration that enables or disables features based on perf impact. Only adding the flags you actually need is probably the right approach.

src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
davmason: This signature is `DoStackSnapshot`'s callback, by setting various flags you can optionally get the register context for each frame, it's expensive and shouldn't be necessary here so it's safe to ignore. Without requesting the context `contextSize` and `context` should be default values.

`ip` is always passed since we have the information, it's cheap to provide and a lot of stack walkers want it, but `funcId` and `frameInfo` are sufficient to uniquely identify a method without `ip`.

src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
```diff
@@ -0,0 +1,392 @@
...
+    return S_OK;
+}
+
+HRESULT ExceptionTracker::GetFullClassName(ClassID classId, tstring& fullClassName)
+{
+    HRESULT hr = S_OK;
+
+    ModuleID moduleId;
+    mdTypeDef typeDefToken;
+    IfFailLogRet(_corProfilerInfo->GetClassIDInfo2(
```
wiktork, line 278: This may fail in cases where the failure has to be gracefully handled, such as Incomplete data, array types, or composite types.

wiktork: See https://github.com/dotnet/samples/blob/2cf486af936261b04a438ea44779cdc26c613f98/core/profiling/stacksampling/src/sampler.cpp#L217

davmason (approved):
I took a look and this is exactly what I expected the code to look like. It all looks good to me, with the caveat that this code is effectively never changed so I don't work in it day to day. There may be minute details we'll have to work through

src/MonitorProfiler/MainProfiler/ThreadData.h
```diff
@@ -0,0 +1,37 @@
...
+/// </summary>
+class ThreadData
+{
+public:
+    static const FunctionID NoFunctionId = 0;
+
+private:
+    FunctionID _exceptionCatcherFunctionId;
+    bool _hasException;
+    std::mutex _mutex;
```
wiktork, line 24: Do we actually need a per thread mutex? While the unordered map of threads certainly needs guarding, do we ever read or mutate the data cross thread boundaries?

Assuming we do, would it be possible to read/write _hasException and _exceptionCatcherFunctionId using interlocked functions to avoid locking on a per thread basis?

Assuming we still need the mutex, is the std::mutex implementation scalable across let's say thousands of threads?

wiktork (approved):
I do have some feedback regarding name reading and threading, but looks good otherwise.

src/MonitorProfiler/MainProfiler/ThreadData.h
jander-msft: I've removed the mutex and did not add the debug same-thread enforcement.

src/MonitorProfiler/MainProfiler/ExceptionTracker.cpp
jander-msft: Talked offline... @wiktork has near-future changes that will handle these particular cases, so I will decline to address them now.

src/MonitorProfiler/MainProfiler/ThreadDataManager.cpp
jander-msft: Nothing noticeable qualitatively. Will perform further quantitative investigation post merge.

PR #1858, opened 2022-05-05
Fix shipping metadata for blob files

jander-msft:
The metadata on the PackageFile items was accidentally dropped in #1834, which causes all of the files uploaded to the dotnetbuilds storage account to be marked as non-shipping. This fix includes the metadata from the PackageFile items in the PackageToPublish items to allow shipping vs non-shipping to be tracked correctly.

### DIFF
--- a/eng/Publishing.props
+++ b/eng/Publishing.props
@@ -74,7 +74,7 @@
 
   <Target Name="CalculatePackagesToPublish">
     <ItemGroup>
-      <PackageToPublish Include="%(PackageFile.Identity)"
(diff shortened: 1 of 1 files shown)


kelltrick (approved):
I assume `PackageFile`'s items will always be files that were included? That's the only difference I see here (this will break if the items are something like strings).

jander-msft:
> I assume `PackageFile`'s items will always be files that were included? That's the only difference I see here (this will break if the items are something like strings).

They should always be files. Earlier in the file, the PackageFile items should have been globes from the file system.

PR #1870, opened 2022-05-09
Enable using runtime id for profiler communication channel

wiktork:
E.g:
``` json
 "Actions": [
                {
                    "Type": "SetEnvironmentVariable",
                    "Settings": {
                        "Name": "DotnetMonitorProfiler_InstanceId",
                        "Value": "$(ProcessInfo.InstanceId)"
                    }
                },
```


src/MonitorProfiler/Logging/NullLogger.h
```diff
@@ -0,0 +1,32 @@
...
+{
+public:
+    // ILogger Members
+
+    static std::shared_ptr<NullLogger> Instance;
+
+    /// <inheritdoc />
+    STDMETHOD_(bool, IsEnabled)(LogLevel level) override
+    {
+        return true;
```
jander-msft, line 24: If the logger won't do anything, set this to `false` so it doesn't waste cpu cycles.

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
```diff
@@ -112,18 +115,35 @@ public IList<CollectionRuleActionOptions> GetActionDependencies(int actionIndex)
 
         public object SubstituteOptionValues(IDictionary<string, CollectionRuleActionResult> actionResults, int actionIndex, object settings)
         {
+            //Attempt to substitute context properties.
+            object originalSettings = settings;
+
+            foreach(PropertyInfo propertyInfo in GetPropertiesFromSettings(settings, p => true))
+            {
+                string property = (string)propertyInfo.GetValue(settings);
```
jander-msft, line 123: nit: originalPropertyValue

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
```diff
@@ -112,18 +115,35 @@ public IList<CollectionRuleActionOptions> GetActionDependencies(int actionIndex)
...
+                //If we don't have an Endpoint info (such as test scenarios) we cannot perform this substitution.
+                if (string.IsNullOrEmpty(property) || (_ruleContext.EndpointInfo == null))
+                {
+                    continue;
+                }
+                string replacement = property.Replace(RuntimeIdReference, _ruleContext.EndpointInfo.RuntimeInstanceCookie.ToString("D"), StringComparison.Ordinal);
+                if (!ReferenceEquals(replacement, property))
```
jander-msft, line 130: This is relying on string interning, correct?

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
```diff
@@ -162,6 +182,31 @@ public object SubstituteOptionValues(IDictionary<string, CollectionRuleActionRes
...
+        private bool TryCloneSettings(object originalSettings, ref object settings)
+        {
+            if (originalSettings == null)
+            {
+                return false;
+            }
+
+            if (ReferenceEquals(originalSettings, settings))
+            {
+                if (originalSettings is BaseRecordOptions baseRecord)
```
jander-msft, line 194: Could we just make every action settings options type derive from this and not have to check or log a failure (since there should not be a failure)? During configuration registration, we can require that all action settings options types have to implement this base class.

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
```diff
@@ -112,18 +115,35 @@ public IList<CollectionRuleActionOptions> GetActionDependencies(int actionIndex)
...
+
+            foreach(PropertyInfo propertyInfo in GetPropertiesFromSettings(settings, p => true))
+            {
+                string property = (string)propertyInfo.GetValue(settings);
+                //If we don't have an Endpoint info (such as test scenarios) we cannot perform this substitution.
+                if (string.IsNullOrEmpty(property) || (_ruleContext.EndpointInfo == null))
+                {
+                    continue;
+                }
+                string replacement = property.Replace(RuntimeIdReference, _ruleContext.EndpointInfo.RuntimeInstanceCookie.ToString("D"), StringComparison.Ordinal);
```
jander-msft, line 129: Can you put in a unit test that flexes this code path?

jander-msft, line 129: nit: newPropertyValue

jander-msft (changes requested):

src/MonitorProfiler/MainProfiler/MainProfiler.cpp
```diff
@@ -31,8 +31,11 @@ STDMETHODIMP MainProfiler::Initialize(IUnknown *pICorProfilerInfoUnk)
 
     //These should always be initialized first
     IfFailRet(ProfilerBase::Initialize(pICorProfilerInfoUnk));
-    IfFailRet(InitializeEnvironment());
+
+    m_pEnvironment = make_shared<ProfilerEnvironment>(m_pCorProfilerInfo);
```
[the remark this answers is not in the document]
wiktork: The environment helper now has a dependency on both the ProfilerEnvironment and the logger. The logger has a dependency on the ProfilerEnvironment. So ProfilerEnvironment has to be made first, then the logger, then the environment helper. I can create a separate initialization method to make this clear though.

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
wiktork: I am relying on Replace to give me back the original reference if it finds no things to replace. In the case that replacement does happen, the reference would be different than the original regardless of the original.

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
wiktork: I am not opposed to this, but I may make it a separate PR.

src/MonitorProfiler/Utilities/StringUtilities.h
```diff
@@ -0,0 +1,35 @@
...
+            {
+                return HRESULT_FROM_WIN32(result);
+            }
+#elif defined(TARGET_LINUX) && !defined(TARGET_ALPINE_LINUX)
+            //TODO Glibc does not support the recommened string copy functions
+            strncpy(destination, source, DestinationSize);
+#else
+            if (strlcpy(destination, source, DestinationSize) >= DestinationSize)
+            {
+                return E_FAIL;
```
jander-msft, line 29: nit: E_INVALIDARG might be better

src/MonitorProfiler/Environment/EnvironmentHelper.cpp
```diff
@@ -27,44 +31,41 @@ HRESULT EnvironmentHelper::GetDebugLoggerLevel(
...
     tstring tmpDir;
 #if TARGET_WINDOWS
-    IfFailLogRet(pEnvironment->GetEnvironmentVariable(s_wszTempEnvVar, tmpDir));
+    IfFailLogRet(_environment->GetEnvironmentVariable(TempEnvVar, tmpDir));
 #else
-    hr = pEnvironment->GetEnvironmentVariable(s_wszTempEnvVar, tmpDir);
-#endif
-
+    hr = _environment->GetEnvironmentVariable(TempEnvVar, tmpDir);
     if (FAILED(hr))
```
jander-msft, line 64: I think we should return `hr` if it is not `ERROR_ENVVAR_NOT_FOUND` rather than attempting to suppress all of the other errors.

src/MonitorProfiler/Environment/EnvironmentHelper.cpp
```diff
@@ -27,44 +31,45 @@ HRESULT EnvironmentHelper::GetDebugLoggerLevel(
...
     {
-        tmpDir = s_wszDefaultTempFolder;
+        if (hr != ERROR_ENVVAR_NOT_FOUND)
```
wiktork, line 66: This is not an HR, but the raw code.

src/MonitorProfiler/Utilities/StringUtilities.h
```diff
@@ -0,0 +1,35 @@
...
+
+#if TARGET_WINDOWS
+            int result = strncpy_s(destination, DestinationSize, source, DestinationSize - 1);
+            if (result != 0)
+            {
+                return HRESULT_FROM_WIN32(result);
+            }
+#elif defined(TARGET_LINUX) && !defined(TARGET_ALPINE_LINUX)
+            //TODO Glibc does not support the recommened string copy functions
+            strncpy(destination, source, DestinationSize);
```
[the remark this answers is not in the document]
wiktork: I'm going to push this off to another PR since it will likely just be copying the pal strncpy_s method.

jander-msft (approved):

src/MonitorProfiler/Utilities/StringUtilities.h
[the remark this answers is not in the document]
jander-msft: Can we add a task to track this?

PR #1877, opened 2022-05-10
Update OSS Versions

kkeirstead:
As part of the Security Review, we need to keep our OSS up-to-date. This is an initial pass - @jander-msft will be taking a look and making recommendations for what versions we should use. Some versions were updated (but not to the latest version), since the latest version (or its dependencies) hadn't been picked up yet. `System CommandLine` also had breaking changes, which led to changes in how we set Arguments.

### DIFF
--- a/documentation/openapi.json
+++ b/documentation/openapi.json
@@ -49,7 +49,9 @@
             "description": "Process ID used to identify the target process.",
             "schema": {
               "type": "integer",
-              "format": "int32"
+              "description": "Process ID used to identify the target process.",
+              "format": "int32",
+              "nullable": true
             }
           },
           {
@@ -58,7 +60,9 @@
             "description": "The Runtime instance cookie used to identify the target process.",
             "schema": {
               "type": "string",
-              "format": "uuid"
+              "description": "The Runtime instance cookie used to identify the target process.",
+              "format": "uuid",
+              "nullable": true
             }
           },
           {
@@ -66,7 +70,9 @@
             "in": "query",
             "description": "Process name used to identify the target process.",
             "schema": {
-              "type": "string"
+              "type": "string",
+              "description": "Process name used to identify the target process.",
+              "nullable": true
             }
           }
         ],
@@ -104,7 +110,9 @@
             "description": "Process ID used to identify the target process.",
             "schema": {
               "type": "integer",
-              "format": "int32"
+              "description": "Process ID used to identify the target process.",
+              "format": "int32",
+              "nullable": true
             }
           },
           {
@@ -113,7 +121,9 @@
             "description": "The Runtime instance cookie used to identify the target process.",
             "schema": {
               "type": "string",
-              "format": "uuid"
+              "description": "The Runtime instance cookie used to identify the target process.",
+              "format": "uuid",
+              "nullable": true
             }
           },
           {
@@ -121,7 +131,9 @@
             "in": "query",
             "description": "Process name used to identify the target process.",
             "schema": {
-              "type": "string"
+              "type": "string",
+              "description": "Process name used to identify the target process.",
+              "nullable": true
             }
           }
         ],
@@ -162,7 +174,9 @@
             "description": "Process ID used to identify the target process.",
             "schema": {
               "type": "integer",
-              "format": "int32"
+              "description": "Process ID used to identify the target process.",
+              "format": "int32",
+              "nullable": true
             }
           },
           {
@@ -171,7 +185,9 @@
             "description": "The Runtime instance cookie used to identify the target process.",
             "schema": {
               "type": "string",
-              "format": "uuid"
+              "description": "The Runtime instance cookie used to identify the target process.",
+              "format": "uuid",
+              "nullable": true
             }
           },
           {
@@ -179,7 +195,9 @@
             "in": "query",
             "description": "Process name used to identify the target process.",
             "schema": {
-              "type": "string"
+              "type": "string",
+              "description": "Process name used to identify the target process.",
+              "nullable": true
             }
           },
           {
@@ -195,7 +213,9 @@
             "in": "query",
             "description": "The egress provider to which the dump is saved.",
             "schema": {
-              "type": "string"
+              "type": "string",
+              "description": "The egress provider to which the dump is saved.",
+              "nullable": true
             }
           }
         ],
@@ -240,7 +260,9 @@
             "description": "Process ID used to identify the target process.",
             "schema": {
               "type": "integer",
-              "format": "int32"
+              "description": "Process ID used to identify the target process.",
+              "format": "int32",
+              "nullable": true
             }
           },
           {
@@ -249,7 +271,9 @@
             "description": "The Runtime instance cookie used to identify the target process.",
             "schema": {
               "type": "string",
-              "format": "uuid"
+              "description": "The Runtime instance cookie used to identify the target process.",
+              "format": "uuid",
+              "nullable": true
             }
           },
           {
@@ -257,7 +281,9 @@
             "in": "query",
             "description": "Process name used to identify the target process.",
             "schema": {
-              "type": "string"
+              "type": "string",
+              "description": "Process name used to identify the target process.",
+              "nullable": true
             }
           },
           {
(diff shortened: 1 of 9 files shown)


eng/Versions.props
```diff
@@ -67,25 +67,25 @@
...
-    <MicrosoftAspNetCoreAuthenticationJwtBearerVersion>6.0.0</MicrosoftAspNetCoreAuthenticationJwtBearerVersion>
-    <MicrosoftAspNetCoreAuthenticationNegotiateVersion>6.0.0</MicrosoftAspNetCoreAuthenticationNegotiateVersion>
+    <AzureStorageBlobsVersion>12.12.0</AzureStorageBlobsVersion>
+    <AzureStorageQueuesVersion>12.10.0</AzureStorageQueuesVersion>
+    <MicrosoftAspNetCoreAuthenticationJwtBearerVersion>6.0.4</MicrosoftAspNetCoreAuthenticationJwtBearerVersion>
+    <MicrosoftAspNetCoreAuthenticationNegotiateVersion>6.0.4</MicrosoftAspNetCoreAuthenticationNegotiateVersion>
     <MicrosoftExtensionsConfigurationAbstractionsVersion>6.0.0</MicrosoftExtensionsConfigurationAbstractionsVersion>
-    <MicrosoftExtensionsConfigurationKeyPerFileVersion>6.0.0</MicrosoftExtensionsConfigurationKeyPerFileVersion>
-    <MicrosoftExtensionsLoggingAbstractionsVersion>6.0.0</MicrosoftExtensionsLoggingAbstractionsVersion>
+    <MicrosoftExtensionsConfigurationKeyPerFileVersion>6.0.4</MicrosoftExtensionsConfigurationKeyPerFileVersion>
```
jander-msft, line 75: Some of the Microsoft.* and System.* packages were updated today, so some of them should be rev'd to next version.

eng/Versions.props
```diff
@@ -67,25 +67,25 @@
...
+    <MicrosoftIdentityModelTokensVersion>6.17.0</MicrosoftIdentityModelTokensVersion>
     <MicrosoftOpenApiReadersVersion>1.2.3</MicrosoftOpenApiReadersVersion>
-    <SystemCommandLineVersion>2.0.0-beta1.20468.1</SystemCommandLineVersion>
-    <SystemIdentityModelTokensJwtVersion>6.11.1</SystemIdentityModelTokensJwtVersion>
+    <SystemCommandLineVersion>2.0.0-beta1.21308.1</SystemCommandLineVersion>
+    <SystemIdentityModelTokensJwtVersion>6.17.0</SystemIdentityModelTokensJwtVersion>
     <SystemPrivateUriVersion>4.3.2</SystemPrivateUriVersion>
     <SystemSecurityPrincipalWindowsVersion>5.0.0</SystemSecurityPrincipalWindowsVersion>
-    <SystemTextJsonVersion>6.0.0</SystemTextJsonVersion>
+    <SystemTextJsonVersion>6.0.3</SystemTextJsonVersion>
```
jander-msft, line 85: I think the explicit STJ dependency isn't used anymore; could you remove it?

kkeirstead:
Looks like the changes to the command line knocked out some of the OpenApi descriptions (and nullables) - I'll go fix this

eng/Versions.props
```diff
@@ -67,25 +67,24 @@
...
     <MicrosoftExtensionsConfigurationAbstractionsVersion>6.0.0</MicrosoftExtensionsConfigurationAbstractionsVersion>
-    <MicrosoftExtensionsConfigurationKeyPerFileVersion>6.0.0</MicrosoftExtensionsConfigurationKeyPerFileVersion>
-    <MicrosoftExtensionsLoggingAbstractionsVersion>6.0.0</MicrosoftExtensionsLoggingAbstractionsVersion>
+    <MicrosoftExtensionsConfigurationKeyPerFileVersion>6.0.5</MicrosoftExtensionsConfigurationKeyPerFileVersion>
+    <MicrosoftExtensionsLoggingAbstractionsVersion>6.0.1</MicrosoftExtensionsLoggingAbstractionsVersion>
     <MicrosoftExtensionsLoggingConsoleVersion>6.0.0</MicrosoftExtensionsLoggingConsoleVersion>
     <MicrosoftExtensionsLoggingEventSourceVersion>6.0.0</MicrosoftExtensionsLoggingEventSourceVersion>
-    <MicrosoftIdentityModelTokensVersion>6.11.1</MicrosoftIdentityModelTokensVersion>
+    <MicrosoftIdentityModelTokensVersion>6.17.0</MicrosoftIdentityModelTokensVersion>
     <MicrosoftOpenApiReadersVersion>1.2.3</MicrosoftOpenApiReadersVersion>
```
jander-msft, line 80: Could this be update to 1.3.1?

eng/Versions.props
```diff
@@ -67,25 +67,24 @@
...
-    <SystemCommandLineVersion>2.0.0-beta1.20468.1</SystemCommandLineVersion>
-    <SystemIdentityModelTokensJwtVersion>6.11.1</SystemIdentityModelTokensJwtVersion>
+    <SystemCommandLineVersion>2.0.0-beta1.21308.1</SystemCommandLineVersion>
```
jander-msft, line 81: Let's set this to 2.0.0-beta3.22173.1; this is what the SDK shipped with in 7.0 Preview 4. After your changes are merged, let's add a darc subscription to keep this dependency up to date.

jander-msft (approved):

documentation/openapi.json
```diff
@@ -49,9 +49,7 @@
             "description": "Process ID used to identify the target process.",
             "schema": {
               "type": "integer",
-              "description": "Process ID used to identify the target process.",
-              "format": "int32",
-              "nullable": true
```
jander-msft, line 54: I believe this type of change is due to an upgrade from OAI 3.0 to 3.1, when billable was largely removed, even though the doc still says it's 3.0.1. I'm okay with this so long as the doc OAI version is corrected and we have a way of sticking with OAI 3.0.X for the `release/6.x` and `release/6.1` branches.

eng/Versions.props
jander-msft: Okay, let's keep it at 1.2.3 for now. It's strange that 1.3.1 is available, but its dependencies are not satisfiable. Maybe we can have dnceng add the missing dependency if it helps with compliance or gets us better functionality.

eng/Versions.props
jander-msft: > The latest version I can go to without getting that error is 2.0.0-beta3.22114.1; should I just put it there?

Add `https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries/nuget/v3/index.json` to the nuget.config at the repo root and it should be able to find that version.

> Also, in terms of the darc subscription, is that going to be an issue since this repo makes breaking changes?

No, we should adapt to the changes as they come in. Worst case, the insertion PRs fail and if we don't want to take up the new changes for a while, we just disable the subscription.

documentation/openapi.json
```diff
@@ -1409,44 +1241,129 @@
         "additionalProperties": false,
         "description": "Represents a partial model when enumerating all operations."
       },
-      "OperationError": {
+      "ProblemDetails": {
         "type": "object",
```
wiktork, line 1245: Are these breaking changes in the api spec or did everything just get shuffled around a lot?

src/Tools/dotnet-monitor/Program.cs
```diff
@@ -8,146 +8,171 @@
...
-                alias: "--no-http-egress",
-                description: Strings.HelpDescription_OptionNoHttpEgress
-                )
+            new Option<bool>(
+                name: "--no-http-egress",
+                description: Strings.HelpDescription_OptionNoHttpEgress,
+                getDefaultValue: () => false)
             {
-                Argument = new Argument<bool>(name: "noHttpEgress", getDefaultValue: () => false)
+                ArgumentHelpName =  "noHttpEgress"
```
wiktork, line 126: nit: extra space

wiktork (dismissed):

documentation/openapi.json
jander-msft: Maybe back out this version change so we can investigate it separately. I'm concerned it is introducing breaking changes in the OpenAPI doc even though it claims to be the same specification version.

kkeirstead: I've reverted `SwashbuckleAspNetCoreSwaggerGenVersion` to `5.6.3` - this appears to have undone the OpenApi changes, but this version won't be considered compliant for the review since it's over 18 months old (9/23/2020)

jander-msft (approved):

PR #1879, opened 2022-05-11
Fix broken link for CollectionRules documentation

jooooel:


jander-msft (approved):
Thanks for fixing this!

PR #1894, opened 2022-05-13
Update release-process.md

kkeirstead:


documentation/release-process.md
```diff
@@ -7,7 +7,7 @@
 1. In `/eng/Version.props`, ensure that `<BlobGroupBuildQuality>` is set appropriately. See the documentation next to this setting for the appropriate values. In release branches, its value should be `release`. This setting, in combination with the version settings, determine for which 'channel' the aks.ms links are created.
```
kkeirstead, line 7: I can't leave the comment on the exact line, but for Step 1 - is that advice accurate when doing a patch release? My understanding is that we wouldn't want to take all of the changes in `main`, but I also haven't done a patch release yet.

documentation/release-process.md
```diff
@@ -42,6 +42,8 @@ The `channel` value is used by the `dotnet-docker` repository to consume the cor
 
 The `dotnet-docker` repository runs an update process each day that detects the latest version of a given `dotnet-monitor` channel. During the stabilization/testing/release period for a release of `dotnet-monitor`, the update process should be changed to pick up builds for the release branch.
 
+>**Note:** you will need permission to edit pipeline variables; **what should we advise here?**.
```
kkeirstead, line 45: Just drawing attention to the question in this line

documentation/release-process.md
```diff
@@ -96,4 +98,10 @@ The remainder of the release will automatically push NuGet packages to nuget.org
 1. The `dotnet-docker` team will merge from `nightly` branch to `main` branch and wait for `dotnet-monitor` team approval. Typically, these changes are completed the day before the release date.
```
kkeirstead, line 98: Can't leave a comment on the exact line, but for Step 1 I initially planned to include the appropriate email address - I decided I'd ask before adding it to avoid revealing an email address that should not be publicly available.

documentation/release-process.md
jander-msft: For patch releases, there are no merges. Fixes are made directly in the release branch.

documentation/release-process.md
```diff
@@ -77,17 +79,17 @@ The nightly image is `mcr.microsoft.com/dotnet/nightly/monitor`. The tag list is
...
 >**Note:** this file comes from the **release** branch and is checked into the **main** branch.
-2. Start [release pipeline](https://dev.azure.com/dnceng/internal/_release?_a=releases&view=mine&definitionId=105). During creation of the release you must select the dotnet-monitor build to release from the list of available builds. This must be a build with the tag `MonitorRelease` and the associated `MonitorRelease` artifact.
-3. The release will start the stage "Pre-Release Verification" this will check that the above steps were done as expected.
+2. Start [release pipeline](https://dev.azure.com/dnceng/internal/_release?_a=releases&view=mine&definitionId=105). Allow the stages to trigger automatically (do not check the boxes in the associated dropdown). During creation of the release you must select the dotnet-monitor build to release from the list of available builds. This must be a build with the tag `MonitorRelease` and the associated `MonitorRelease` artifact (set `dotnet-monitor_build` to the pipeline run of `dotnet monitor` that is being released; set `dotnet-monitor_source` to the latest commit from `main`). The name of the release will be updated automatically.
```
jander-msft, line 82:
```suggestion
2. Start [release pipeline](https://dev.azure.com/dnceng/internal/_release?_a=releases&view=mine&definitionId=105). Allow the stages to trigger automatically (do not check the boxes in the associated dropdown). During creation of the release you must select the dotnet-monitor build to release from the list of available builds. This must be a build with the tag `MonitorRelease` and the associated `MonitorRelease` artifact (set `dotnet-monitor_build` to the pipeline run of `dotnet monitor` that is being released; set `dotnet-monitor_source` to the latest commit from `main`).
```

How about we put the name updating in the next step since that step is focused on what happens in the "Pre-Release Verification" stage/job.

documentation/release-process.md
jander-msft: I would not include that information here.

jander-msft (approved):

PR #1927, opened 2022-05-25
Prevent process pruning during GC dump

### LINKED ISSUE #1925: Prevent process pruning when capturing gcdump

jander-msft:
closes #1925

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
@@ -271,7 +271,22 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
                 return Result(
                     Utilities.ArtifactType_GCDump,
                     egressProvider,
-                    (stream, token) => GCDumpUtilities.CaptureGCDumpAsync(processInfo.EndpointInfo, stream, token),
+                    async (stream, token) =>
+                    {
+                        IDisposable operationRegistration = null;
+                        try
+                        {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -271,7 +271,22 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
...
                     Utilities.ArtifactType_GCDump,
                     egressProvider,
-                    (stream, token) => GCDumpUtilities.CaptureGCDumpAsync(processInfo.EndpointInfo, stream, token),
+                    async (stream, token) =>
+                    {
+                        IDisposable operationRegistration = null;
+                        try
+                        {
+                            if (_diagnosticPortOptions.Value.ConnectionMode == DiagnosticPortConnectionMode.Listen)
+                            {
```
wiktork, line 280: Are you sure this is still necessary? I remember the original issue was a bug in the runtime where rundown collection would cause the pipe to be unresponsive. We fixed a lot of our collection artifacts (such as gcdump and metrics) to no longer collect rundown since it's not needed.

If this is still happening it invalidates the original theory that rundown was the problem.

jander-msft: Yes, I think this is necessary. The gcdump operation forces a full garbage collection; if the heap is very large, this could take a long time. #1758 reports that gcdump sometimes takes significant amounts of time when the working set is very high; after 3 seconds, dotnet-monitor gives up on the connection and removes it from the listening server. The app container eventually is evicted due to failed health checks, presumably because the process is still doing the GC.

wiktork (approved):

PR #1935, opened 2022-05-26
[backport] Make sure unexpected errors finish operations

wiktork:


src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationService.cs
```diff
@@ -47,11 +48,18 @@ private async Task ExecuteEgressOperation(EgressRequest egressRequest, Cancellat
...
+                {
+                    var result = await egressRequest.EgressOperation.ExecuteAsync(_serviceProvider, token);
 
-                //It is possible that this operation never completes, due to infinite duration operations.
+                    //It is possible that this operation never completes, due to infinite duration operations.
 
-                _operationsStore.CompleteOperation(egressRequest.OperationId, result);
+                    _operationsStore.CompleteOperation(egressRequest.OperationId, result);
+                }
+                catch (Exception e) when (!(e is OperationCanceledException))
```
kelltrick, line 59: Where do the `OperationCanceledException`s operations get marked as canceled?

kelltrick (approved):

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationService.cs
jander-msft: I believe this is not needed, but maybe Wiktor can chime in. An egress operation is cancelled through the `DELETE /operations/{id}` verb/route, so the operations store is updated correctly (hopefully). Here, the egress operation is cancelled due to background service shutdown, which happens when dotnet-monitor is shutting down, so it's not necessary to put it into the cancelled state.

wiktork: That is correct, the DELETE verb will first place the operation in a cancelled state. Then it will trigger the token to be in a cancelled state, to interrupt the egress operation. Then we will enter this catch block.

jander-msft (approved):

PR #1953, opened 2022-06-01
Add example extension to egress data to Azure

kelltrick:
Adds code to accept data from extension interface and egress to Azure.


src/Tools/AzureBlobStorage/AzureBlobStorage.csproj
```diff
@@ -0,0 +1,67 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFrameworks>$(ToolTargetFrameworks)</TargetFrameworks>
+    <RuntimeIdentifiers>linux-x64;linux-musl-x64;win-x64</RuntimeIdentifiers>
```
jander-msft, line 5: `RuntimeIdentifiers` and `PackAsToolShimRuntimeIdentifiers` are not necessary unless you plan to publish or pack as a .NET CLI tool.

src/Tools/AzureBlobStorage/AzureBlobStorage.csproj
```diff
@@ -0,0 +1,67 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
```
jander-msft, line 1: The way the directory structure is set up and the usage of the Directory.Build.props files, I think this project should be moved to `src/AzureBlobStorage` (move it out of the Tools folder) or nest it into `src/Extensions/AzureBlobStorage` to organize future extensions into a common directory.

src/Tools/AzureBlobStorage/AzureBlobStorage.csproj
```diff
@@ -0,0 +1,67 @@
...
+
+  <PropertyGroup>
+    <TargetFrameworks>$(ToolTargetFrameworks)</TargetFrameworks>
+    <RuntimeIdentifiers>linux-x64;linux-musl-x64;win-x64</RuntimeIdentifiers>
+    <PackAsToolShimRuntimeIdentifiers>linux-x64;linux-musl-x64;win-x64</PackAsToolShimRuntimeIdentifiers>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
```
jander-msft, line 8: Not needed since you have `TargetFrameworks` specified above.

src/Tools/AzureBlobStorage/AzureBlobStorage.csproj
```diff
@@ -0,0 +1,67 @@
...
+    <RootNamespace>Microsoft.Diagnostics.Tools.Monitor.Egress.AzureBlob</RootNamespace>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>disable</Nullable>
+    <PublishToBlob>true</PublishToBlob>
```
jander-msft, line 12: I don't think we're going to produce a nupkg for this but the assembly will be packed into the dotnet-monitor package, correct? If so, then this is not necessary.

src/Tools/AzureBlobStorage/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+                    AccountKeyName = GetConfig(config.Configuration, nameof(AzureBlobEgressProviderOptions.AccountKeyName)),
+                    SharedAccessSignature = GetConfig(config.Configuration, nameof(AzureBlobEgressProviderOptions.SharedAccessSignature)),
+                    SharedAccessSignatureName = GetConfig(config.Configuration, nameof(AzureBlobEgressProviderOptions.SharedAccessSignatureName)),
+                    ContainerName = GetConfig(config.Configuration, nameof(AzureBlobEgressProviderOptions.ContainerName)),
+                    BlobPrefix = GetConfig(config.Configuration, nameof(AzureBlobEgressProviderOptions.BlobPrefix)),
+                    QueueName = GetConfig(config.Configuration, nameof(AzureBlobEgressProviderOptions.QueueName)),
+                    QueueAccountUri = GetUriConfig(config.Configuration, nameof(AzureBlobEgressProviderOptions.QueueAccountUri)),
+                };
+
+                if (options.AccountKey == null && options.AccountKeyName != null && config.Properties.ContainsKey(options.AccountKeyName))
```
jander-msft, line 36:
```suggestion
                if (string.IsNullOrEmpty(options.AccountKey) && !string.IsNullOrEmpty(options.AccountKeyName) && config.Properties.TryGetValue(options.AccountKeyName, out string accountKey))
```

src/Tools/AzureBlobStorage/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+                {
+                    options.AccountKey = config.Properties[options.AccountKeyName];
+                }
+
+                if (options.SharedAccessSignature == null && options.SharedAccessSignatureName != null && config.Properties.ContainsKey(options.SharedAccessSignatureName))
```
jander-msft, line 41: Apply same changes to here as for `options.AccountKey`

src/Tools/AzureBlobStorage/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+                {
+                    options.SharedAccessSignature = config.Properties[options.SharedAccessSignatureName];
+                }
+
+                Func<CancellationToken, Task<Stream>> action = new Func<CancellationToken, Task<Stream>>(
+                    (CancellationToken token) =>
+                    {
+                        StdInStream = Console.OpenStandardInput();
```
jander-msft, line 49: Move this outside of the anonymous method to avoid race conditions e.g. cancel keypress occurs before egress begins.

src/Tools/AzureBlobStorage/Program.cs
```diff
@@ -0,0 +1,105 @@
...
+                        return Task.FromResult(StdInStream);
+                    });
+                
+                AzureBlobEgressProvider provider = new AzureBlobEgressProvider(null);
+
+                Console.CancelKeyPress += Console_CancelKeyPress;
+
+                result.ArtifactPath = await provider.EgressAsync(EgressProviderTypes.AzureBlobStorage, config.ProfileName, options, action, config.Settings, CancellationToken.None);
```
jander-msft, line 57: Please add cancellation support via key press cancellation.

src/Tools/AzureBlobStorage/AzureBlobStorage.csproj
```diff
@@ -0,0 +1,67 @@
...
+    <Compile Include="..\..\Microsoft.Diagnostics.Monitoring.Options\ProcessFilterOptions.cs" Link="ExternalRefs\ProcessFilterOptions.cs" />
+    <Compile Include="..\..\Microsoft.Diagnostics.Monitoring.Options\IEgressProviderCommonOptions.cs" Link="ExternalRefs\IEgressProviderCommonOptions.cs" />
+    <Compile Include="..\..\Microsoft.Diagnostics.Monitoring.Options\OptionsDisplayStrings.Designer.cs" Link="ExternalRefs\OptionsDisplayStrings.Designer.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Azure.Identity" Version="$(AzureIdentityVersion)" />
+    <PackageReference Include="Azure.Storage.Blobs" Version="$(AzureStorageBlobsVersion)" />
+    <PackageReference Include="Azure.Storage.Queues" Version="$(AzureStorageQueuesVersion)" />
+    <PackageReference Include="System.CommandLine" Version="$(SystemCommandLineVersion)" />
```
jander-msft, line 64: This doesn't seem to be used

src/Tools/AzureBlobStorage/CollectionRuleDefaultsOptions.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options
+{
+    internal class CollectionRuleDefaultsOptions
```
jander-msft, line 7: What requires these two classes? Egress should be independent of collection rules.

src/Tools/AzureBlobStorage/AzureBlobStorage.csproj
```diff
@@ -0,0 +1,67 @@
...
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>disable</Nullable>
+    <PublishToBlob>true</PublishToBlob>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="..\dotnet-monitor\Strings.Designer.cs" Link="ExternalRefs\Strings.Designer.cs" />
+    <Compile Include="..\dotnet-monitor\LoggingEventIds.cs" Link="ExternalRefs\LoggingEventIds.cs" />
+    <Compile Include="..\dotnet-monitor\LoggingExtensions.cs" Link="ExternalRefs\LoggingExtensions.cs" />
+    <Compile Include="..\dotnet-monitor\ExperienceSurvey.cs" Link="ExternalRefs\ExperienceSurvey.cs" />
```
jander-msft, line 19: I think there are a lot of links that are not used. Are these brought in because of LoggingExtensions.cs?

src/Tools/AzureBlobStorage/CollectionRuleDefaultsOptions.cs
kelltrick: This was required by `RootOptions` which is pulled in from `LoggingExtension`. I pulled that out and made a new one with just the few extensions that are needed and was able to remove this and pair down usage of all of this.

src/Tools/AzureBlobStorage/AzureBlobStorage.csproj
kelltrick: I meant to add in the commandline parsing from another local copy of the test app. I copied that in and this ref is now used.

src/Tools/AzureBlobStorage/AzureBlobStorage.csproj
kelltrick: Most of this is due to `LoggingExtension`. I pulled that out and made a new one with just the few extensions that are needed and was able to remove this and pair down usage of all of this.

src/Tools/AzureBlobStorage/Program.cs
kelltrick: Done

kelltrick:
Here is the setup I've been testing with:
1. Compile `dotnet-monitor.sln`
2. Copy `[dotnet-monitor]\artifacts\bin\AzureBlobStorage\Debug\net6.0` to `[dotnet-monitor]\artifacts\bin\dotnet-monitor\Debug\net6.0\extensions\azure-storage`
3. Install these settings:
```json
{
    "$schema": "https://aka.ms/dotnet-monitor-schema",

    "DiagnosticPort": {
        "ConnectionMode": "Listen",
        "EndpointName": "dotnetmonitorpipe"
    },

    "DefaultProcess": {
        "Filters": [
            {
                "ProcessName": "AllocationTestApp"
            }
        ]
    },
    "Egress": {
        "AzureBlobStorage": {
            "monitorBlob": {
                "AccountKeyName": "AzAccountKey",
                "AccountUri": "https://pafenelodnmegress.blob.core.windows.net",
                "ContainerName": "dotnet-monitor",
                "BlobPrefix": "artifact"
            }
        },
        "FileSystem": {
            "localFile": {
                "DirectoryPath": "C:\\scratch"
            }
        },
        "Properties": {
            "AzAccountKey": "=====[REDACTED, this should be an Az Storage base 64 encoded key]======"
        },
        "azure-storage": {
            "azureTest": {
                "AccountKeyName": "AzAccountKey",
                "AccountUri": "https://pafenelodnmegress.blob.core.windows.net",
                "ContainerName": "dotnet-monitor",
                "BlobPrefix": "artifact"
            },
            "azure2Test": {
                "AccountKeyName": "AzAccountKey",
                "AccountUri": "https://companyblob.blob.core.windows.net",
                "ContainerName": "debugme",
                "BlobPrefix": "artifact"
            }
        },
        "AWS_S3": {
            "myBucketsEgress": {
                "AccountID": "sdfsdfhsldkf",
                "AccountKey": "foihj2930j0239r3322"
            }
        },
        "External": {
            "extTest": {}
        },
        "MyCustomProvider": {
            "custom": {}
        }
    }
}
```
4. Launch a target debuggee app called `AllocationTestApp.exe`
5. Launch `dotnet-monitor collect --metrics false`
6. Call the http api `https://localhost:56019/dump?type=WithHeap&egressProvider=azureTest`
[Observe dump getting egressed via external extension]

src/Extensions/AzureBlobStorage/Program.cs
```diff
@@ -0,0 +1,126 @@
...
+        private static async Task<int> Egress(string ProviderName)
+        {
+            EgressArtifactResult result = new();
+            try
+            {
+                string jsonConfig = Console.ReadLine();
+                ExtensionEgressPayload configPayload = JsonSerializer.Deserialize<ExtensionEgressPayload>(jsonConfig);
+                AzureBlobEgressProviderOptions options = BuildOptions(configPayload);
+
+                AzureBlobEgressProvider provider = new AzureBlobEgressProvider(logger: null); // TODO: Replace logger with real instance of console logger (console events on standard out will get forwarded to dotnet-monitor and written to it's console).
```
wiktork, line 39: Do we want to recommend using the GenericHost to make it easier to create the logger?

src/Extensions/AzureBlobStorage/Program.cs
```diff
@@ -0,0 +1,126 @@
...
+                result.Succeeded = true;
+            }
+            catch (Exception ex)
+            {
+                result.Succeeded = false;
+                result.FailureMessage = ex.Message;
+            }
+
+            string jsonBlob = JsonSerializer.Serialize<EgressArtifactResult>(result);
+            Console.Write(jsonBlob);
```
wiktork, line 53: Do we need to flush?

src/Extensions/AzureBlobStorage/Program.cs
```diff
@@ -0,0 +1,126 @@
...
+            EgressArtifactResult result = new();
+            try
+            {
+                string jsonConfig = Console.ReadLine();
+                ExtensionEgressPayload configPayload = JsonSerializer.Deserialize<ExtensionEgressPayload>(jsonConfig);
+                AzureBlobEgressProviderOptions options = BuildOptions(configPayload);
+
+                AzureBlobEgressProvider provider = new AzureBlobEgressProvider(logger: null); // TODO: Replace logger with real instance of console logger (console events on standard out will get forwarded to dotnet-monitor and written to it's console).
+
+                Console.CancelKeyPress += Console_CancelKeyPress;
```
wiktork, line 41: This is registered before StdInStream is assigned. Consider setting up the stream things upfront and then calling egress. This also removes the need for the GetStream function, which really should not be called from anywhere else.

src/Extensions/AzureBlobStorage/Program.cs
```diff
@@ -0,0 +1,126 @@
...
+
+            return await rootCommand.InvokeAsync(args);
+        }
+
+        private static async Task<int> Egress(string ProviderName)
+        {
+            EgressArtifactResult result = new();
+            try
+            {
+                string jsonConfig = Console.ReadLine();
```
wiktork, line 35: There's some concern here I have with ReadLine. Is newline part of the protocol? Do we change the newline character depending on the os?

kelltrick: We explicitly write the Windows termination here `\r\n`. It seems that dotnet isn't fussy about this and `ReadLine()` will consume a full newline, regardless of platform when I ran this on WSL.

src/Extensions/AzureBlobStorage/Program.cs
kelltrick: No, this will always get fully written to `StdOut`

src/Extensions/AzureBlobStorage/Program.cs
kelltrick: I think so, I was looking into converting this to a `GenericHost` but I didn't have time to finish the conversion today.

src/Extensions/AzureBlobStorage/Program.cs
kelltrick: That's probably a good idea, and I'll leave that to you to do.

kelltrick:
Todos after this PR gets done:
1. @wiktork 's comment about when the Standard In stream gets assigned.
> This is registered before StdInStream is assigned. Consider setting up the stream things upfront and then calling egress. This also removes the need for the GetStream function, which really should not be called from anywhere else.
2. Double check that `Console.ReadLine()` will fully consume the `\r\n` Windows newline on a linux build. I want to make sure this isn't something special on WSL.

PR #1954, opened 2022-06-02
Fix RuntimeId token

wiktork:
- Rename InstanceId to RuntimeId
- Fix issue with values that contain both dependencies and RuntimeId. We were replacing values in the wrong order.
- Add unit tests for above scenario.
- Refactor token replacement code to its own class so that it can be reused by egress in the future.


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ActionDependencyAnalyzerTests.cs
```diff
@@ -51,6 +53,66 @@ public ActionDependencyAnalyzerTests(ITestOutputHelper outputHelper)
             _outputHelper = outputHelper;
         }
 
+        [Fact]
+        public async Task DependenciesTest()
+        {
+            const string Output1 = nameof(Output1);
```
wiktork, line 59: This test is a copy/paste of the original.

PR #1990, opened 2022-06-09
Specify Json Configuration File Source Via Command Line

### LINKED ISSUE #1760: Provide path for config file in command-line invocation
## Proposed Feature
Hello.

I'd like to be able to provide path to configuration file into command-line invocation, like `dotnet monitor --config /my/settings.json collect`. According to [documentation](https://github.com/dotnet/dotnet-monitor/blob/main/documentation/configuration.md#Default-Process-Configuration) it's not supported atm.

I think it will be more convenient to end-user then placing file into pre-defined location or pass config through env vars. It will also allow to change settings more easily and store multiple settings files with git.

kkeirstead:
Users can now specify a configuration file via the command line (`--configuration-file-path`); if provided, this takes highest precedence but **does not** override the other configuration sources.

Closes #1760

### DIFF
--- a/src/Tools/dotnet-monitor/Commands/CollectCommandHandler.cs
+++ b/src/Tools/dotnet-monitor/Commands/CollectCommandHandler.cs
@@ -19,12 +19,12 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Commands
 {
     internal static class CollectCommandHandler
     {
-        public static async Task<int> Invoke(CancellationToken token, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, bool noHttpEgress)
+        public static async Task<int> Invoke(CancellationToken token, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, bool noHttpEgress, string configurationFilePath)
         {
             try
             {
                 AuthConfiguration authConfiguration = HostBuilderHelper.CreateAuthConfiguration(noAuth, tempApiKey);
-                HostBuilderSettings settings = HostBuilderSettings.CreateMonitor(urls, metricUrls, metrics, diagnosticPort, authConfiguration);
+                HostBuilderSettings settings = HostBuilderSettings.CreateMonitor(urls, metricUrls, metrics, diagnosticPort, authConfiguration, configurationFilePath);
 
                 IHost host = HostBuilderHelper.CreateHostBuilder(settings)
                     .ConfigureServices(authConfiguration, noHttpEgress)
--- a/src/Tools/dotnet-monitor/HostBuilderHelper.cs
+++ b/src/Tools/dotnet-monitor/HostBuilderHelper.cs
@@ -73,7 +73,6 @@ namespace Microsoft.Diagnostics.Tools.Monitor
                     {
                         try
                         {
-
                             File.OpenRead(userFilePath).Dispose(); // If this succeeds, we have read-permissions
                             builder.AddJsonFile(userFilePath, optional: true, reloadOnChange: true);
                         }
--- a/src/Tools/dotnet-monitor/HostBuilderHelper.cs
+++ b/src/Tools/dotnet-monitor/HostBuilderHelper.cs
@@ -69,17 +69,17 @@ namespace Microsoft.Diagnostics.Tools.Monitor
                     // User-specified configuration file path is considered highest precedence, but does NOT override other configuration sources
                     string userFilePath = settings.UserProvidedConfigFilePath;
 
-                    if (!string.IsNullOrEmpty(userFilePath))
+                    if (File.Exists(userFilePath))
                     {
-                        using (FileStream filestream = new FileStream(userFilePath, FileMode.Open))
+                        try
                         {
-                            if (filestream.CanRead)
-                            {
-                                builder.AddJsonFile(userFilePath, optional: true, reloadOnChange: true);
-                            }
+                            File.OpenRead(userFilePath).Dispose(); // If this succeeds, we have read-permissions
+                            builder.AddJsonFile(userFilePath, optional: true, reloadOnChange: true);
                         }
-
-                        builder.AddJsonFile(settings.UserProvidedConfigFilePath, optional: true, reloadOnChange: true);
+                        catch (Exception)
+                        {
+                            // Don't currently log to tell the user that their configuration file was not loaded
+                        }
                     }
                 })
                 //Note this is necessary for config only because Kestrel configuration
--- a/src/Tools/dotnet-monitor/HostBuilderHelper.cs
+++ b/src/Tools/dotnet-monitor/HostBuilderHelper.cs
@@ -69,10 +69,11 @@ namespace Microsoft.Diagnostics.Tools.Monitor
                     // User-specified configuration file path is considered highest precedence, but does NOT override other configuration sources
                     string userFilePath = settings.UserProvidedConfigFilePath;
 
-                    if (File.Exists(userFilePath))
+                    if (File.Exists(userFilePath) && Path.GetExtension(userFilePath).ToLower().Equals(".json"))
                     {
                         try
                         {
+
                             File.OpenRead(userFilePath).Dispose(); // If this succeeds, we have read-permissions
                             builder.AddJsonFile(userFilePath, optional: true, reloadOnChange: true);
                         }
--- a/src/Tools/dotnet-monitor/HostBuilderHelper.cs
+++ b/src/Tools/dotnet-monitor/HostBuilderHelper.cs
@@ -65,6 +65,22 @@ namespace Microsoft.Diagnostics.Tools.Monitor
                     {
                         ConfigureTempApiHashKey(builder, settings.Authentication);
                     }
+
+                    // User-specified configuration file path is considered highest precedence, but does NOT override other configuration sources
+                    string userFilePath = settings.UserProvidedConfigFilePath;
+
+                    if (!string.IsNullOrEmpty(userFilePath))
+                    {
+                        using (FileStream filestream = new FileStream(userFilePath, FileMode.Open))
+                        {
+                            if (filestream.CanRead)
+                            {
+                                builder.AddJsonFile(userFilePath, optional: true, reloadOnChange: true);
+                            }
+                        }
+
+                        builder.AddJsonFile(settings.UserProvidedConfigFilePath, optional: true, reloadOnChange: true);
+                    }
                 })
                 //Note this is necessary for config only because Kestrel configuration
                 //is not added until WebHostDefaults are added.
--- a/src/Tools/dotnet-monitor/Strings.resx
+++ b/src/Tools/dotnet-monitor/Strings.resx
@@ -340,6 +340,10 @@
     <value>Shows configuration, as if dotnet-monitor collect was executed with these parameters.</value>
     <comment>Gets the string to display in help that explains what the 'show' command does.</comment>
   </data>
+  <data name="HelpDescription_OptionConfigurationFilePath" xml:space="preserve">
+    <value>The fully qualified path and filename of the configuration file you'd like to add to the list of configuration sources.</value>
+    <comment>Gets the string to display in help that explains what the '--configuration-file-path' option does.</comment>
+  </data>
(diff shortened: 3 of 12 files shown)


src/Tools/dotnet-monitor/HostBuilderHelper.cs
```diff
@@ -65,6 +65,22 @@ public static IHostBuilder CreateHostBuilder(HostBuilderSettings settings)
...
+                    string userFilePath = settings.UserProvidedConfigFilePath;
+
+                    if (File.Exists(userFilePath) && Path.GetExtension(userFilePath).ToLower().Equals(".json"))
+                    {
+                        try
+                        {
+                            File.OpenRead(userFilePath).Dispose(); // If this succeeds, we have read-permissions
+                            builder.AddJsonFile(userFilePath, optional: true, reloadOnChange: true);
+                        }
+                        catch (Exception)
```
wiktork, line 79: This seems like a reasonable error to log. Is there a reason we wouldn't want to do this?

wiktork, line 79: This seems like a reasonable error to log. Is there a reason we wouldn't want to do this?

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -340,6 +340,10 @@
     <value>Shows configuration, as if dotnet-monitor collect was executed with these parameters.</value>
     <comment>Gets the string to display in help that explains what the 'show' command does.</comment>
   </data>
+  <data name="HelpDescription_OptionConfigurationFilePath" xml:space="preserve">
+    <value>The fully qualified path and filename of the configuration file you'd like to add to the list of configuration sources.</value>
```
wiktork, line 344: Maybe mention this must be json.

wiktork, line 344: Maybe mention this must be json.

src/Tools/dotnet-monitor/HostBuilderHelper.cs
```diff
@@ -65,6 +65,22 @@ public static IHostBuilder CreateHostBuilder(HostBuilderSettings settings)
...
+                        {
+                            // Don't currently log to tell the user that their configuration file was not loaded
```
wiktork, line 81: Take a look at `AddressListenResults` to see how to transfer error state from here to where a logger is available.

src/Tools/dotnet-monitor/HostBuilderHelper.cs
```diff
@@ -65,6 +65,22 @@ public static IHostBuilder CreateHostBuilder(HostBuilderSettings settings)
...
                     }
+
+                    // User-specified configuration file path is considered highest precedence, but does NOT override other configuration sources
+                    string userFilePath = settings.UserProvidedConfigFilePath;
+
+                    if (File.Exists(userFilePath) && Path.GetExtension(userFilePath).ToLower().Equals(".json"))
+                    {
+                        try
+                        {
+                            File.OpenRead(userFilePath).Dispose(); // If this succeeds, we have read-permissions
```
wiktork, line 76: This check seems unnecessary. The infrastructure should attempt to read this configuration and fail at some point if we don't have read access.

wiktork (commented):
Adding PR feedback.

src/Tools/dotnet-monitor/HostBuilderHelper.cs
kkeirstead: To be able to mimic that, would we want/need some sort of wrapper around `IHostBuilder` that also includes an error list? And then the caller (e.g. `CollectCommandHandler`) would be responsible for iterating over the error list and logging it using the service from the host?

src/Tools/dotnet-monitor/Commands/CollectCommandHandler.cs
```diff
@@ -19,12 +19,12 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Commands
 {
     internal static class CollectCommandHandler
     {
-        public static async Task<int> Invoke(CancellationToken token, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, bool noHttpEgress)
+        public static async Task<int> Invoke(CancellationToken token, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, bool noHttpEgress, string configurationFilePath)
```
jander-msft, line 22:
```suggestion
        public static async Task<int> Invoke(CancellationToken token, string[] urls, string[] metricUrls, bool metrics, string diagnosticPort, bool noAuth, bool tempApiKey, bool noHttpEgress, FileInfo configurationFilePath)
```

`System.CommandLine` supports automatically converting file paths to `FileInfo`. Plumb this through all the way to where it is actually used.

src/Tools/dotnet-monitor/HostBuilderHelper.cs
jander-msft: I would not check for permissions here at this time and just call `AddJsonFile`. This is technically an existing problem with our other file-based configuration as well, so I'd rather there be an all-up solution rather than a one-off. Feel free to log an issue for this.

src/Tools/dotnet-monitor/HostBuilderHelper.cs
wiktork: I was under the mistaken impression that services are available in this context, but this is just the configuration block. Let me try to play around with the host builder context to see if this is easy to do.

wiktork: One option here would be to use the `ConfigureAppConfiguration` overload that takes a `HostBuilderContext` object. You can then set `HostBuilderContext.Properties` to include errors.

Then during `Startup.Configure` you can add `HostBuilderContext` as a parameter, and retrieve the errors from `Properties`.

kkeirstead: I have a version of this working, but wanted to double check what you meant in terms of the implementation - is there an actual way to set errors for `Properties`, or did you just mean that I can just add the error as a property (or a list of errors for a key that holds all the errors) and print out the `Properties` from `Startup.Configure`?

src/Tools/dotnet-monitor/HostBuilderHelper.cs
```diff
@@ -65,6 +65,22 @@ public static IHostBuilder CreateHostBuilder(HostBuilderSettings settings)
...
+                            builder.AddJsonFile(userFilePath, optional: true, reloadOnChange: true);
+                        }
+                        catch (Exception)
+                        {
+                            // Don't currently log to tell the user that their configuration file was not loaded
+                        }
+                    }
```
kkeirstead, line 83: Do we want to have something that tells users here that their path either 1) didn't exist or 2) wasn't a json file?

src/Tools/dotnet-monitor/HostBuilderHelper.cs
wiktork: The latter - Properties here is just key/value pairs that are used to communicate information between various parts of the host building process. Normally we would use a singleton object (like AddressListenResults) but I'm not sure it's possible to add a singleton this early in the host builder.

src/Tools/dotnet-monitor/Startup.cs
```diff
@@ -117,6 +118,11 @@ public void ConfigureServices(IServiceCollection services)
                 logger.UnableToListenToAddress(result.Url, result.Exception);
             }
 
+            foreach (var warningMessage in hostBuilderContext.Properties)
+            {
+                logger.LogWarning(warningMessage.Value.ToString());
```
jander-msft, line 123: I like where this is going. However, I don't think we should be putting values directly in the properties bag since there could be other properties that are added not by our code. I would suggest creating a new object type (e.g. `HostBuilderResults`) that contains a property for warnings (e.g. `public List<string> Warnings { get; } = new()`), add this to `context.Properties` with a well-known key, add whatever issues to the warnings property and then get the object back out here and iteratively log them. This will allow for clear separation of ownership and allow us to expand upon it in the future more easily.

kkeirstead: I had thought about that issue too, and I think you're right - although, is there a reason why we're creating a new type (such as `HostBuilderResults`) just to hold a list of strings, when presumably couldn't we just have a key (e.g. "Warnings") that maps directly to a list of strings (the warning messages)?

src/Tools/dotnet-monitor/HostBuilderHelper.cs
```diff
@@ -112,6 +138,22 @@ public static IHostBuilder CreateHostBuilder(HostBuilderSettings settings)
                 });
         }
 
+        private static void UpdatePropertiesWarningList(HostBuilderContext context, string warningMessage)
+        {
+            if (context.Properties.TryGetValue(Strings.Key_WarningMessages, out object errorList))
```
wiktork, line 143: Do not make the key a localizable property; it should be a string constant.

src/Tools/dotnet-monitor/Startup.cs
wiktork: As Justin pointed out, it will be easier to add additional data (for example, if we also wanted Errors not just Warnings) in the future if we use an object instead of a List<string>.

src/Tools/dotnet-monitor/HostBuilderHelper.cs
kkeirstead: Won't that lead us to having two distinct string constants (one in `HostBuildHelper`, one in `Startup`) that need to be kept in sync?

wiktork (approved):

src/Tools/dotnet-monitor/HostBuilderHelper.cs
wiktork: I would suggest something like:
``` c#
class HostBuilderHelper
{
   public const string ResultKey = "DotnetMonitorHostBuilderResults";
}

Then you can use HostBuilderHelper.ResultKey in both classes

File: src/Tools/dotnet-monitor/HostBuilderHelper.cs [hunk 9]
```diff
@@ -112,6 +138,23 @@ public static IHostBuilder CreateHostBuilder(HostBuilderSettings settings)
                 });
         }
 
+        private static void UpdatePropertiesWarningList(HostBuilderContext context, string warningMessage)
+        {
+            if (context.Properties.TryGetValue(HostBuilderResults.ResultKey, out object resultsObject))
+            {
+                if (resultsObject is HostBuilderResults)
```
wiktork, line 145: nit: `resultsObject is HostBuilderResults hostBuilderResults`. No need for a secondary cast.

src/Tools/dotnet-monitor/Startup.cs
```diff
@@ -117,6 +118,17 @@ public void ConfigureServices(IServiceCollection services)
                 logger.UnableToListenToAddress(result.Url, result.Exception);
             }
 
+            if (hostBuilderContext.Properties.TryGetValue(HostBuilderResults.ResultKey, out object hostBuilderResult))
+            {
+                if (hostBuilderResult is HostBuilderResults)
+                {
+                    foreach (string message in ((HostBuilderResults)hostBuilderResult).Warnings)
```
wiktork, line 125: Same as above, use `is HostBuilderResults results` or similiar.

wiktork (approved):

src/Tools/dotnet-monitor/HostBuilderHelper.cs
```diff
@@ -112,6 +138,23 @@ public static IHostBuilder CreateHostBuilder(HostBuilderSettings settings)
...
+            if (context.Properties.TryGetValue(HostBuilderResults.ResultKey, out object resultsObject))
+            {
+                if (resultsObject is HostBuilderResults hostBuilderResults)
+                {
+                    hostBuilderResults.Warnings.Add(warningMessage);
+                }
+            }
+            else
+            {
+                HostBuilderResults results = new HostBuilderResults();
```
jander-msft, line 152: Could we add the HostBuilderResults instance to the properties collection upfront and then add warnings to it as necessary without having to continuously look up if the object exists?

jander-msft (approved):

jander-msft (approved):

PR #2022, opened 2022-06-22
Address "Swallow Everything" Exceptions

kkeirstead:
Part of the security review is to address "swallow everything" exceptions; this is an initial pass at adding logging to some of the places where we `catch` as well as expanding to specify which exceptions we want to catch. Things to consider:

- Where we should/shouldn't log (and what level makes the most sense)
- What exceptions we need to handle that were previously being encompassed by `Exception`
- Places where the exception handling should not be altered (and the reasoning)

If there are other places where we should make changes that aren't included here, please let me know and I'll add them in.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
@@ -82,6 +82,10 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
                 catch (InvalidOperationException)
                 {
                 }
+                catch (Exception ex) when (!(ex is OperationCanceledException))
+                {
+                    _logger.LogWarning(ex.Message);
+                }
 
                 IList<Models.ProcessIdentifier> processesIdentifiers = new List<Models.ProcessIdentifier>();
                 foreach (IProcessInfo p in await _diagnosticServices.GetProcessesAsync(processFilter: null, HttpContext.RequestAborted))
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
@@ -77,6 +77,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
                 }
                 catch (ArgumentException ex)
                 {
+                    // Unable to locate a default process; no action required
                     _logger.LogInformation(ex.Message);
                 }
                 catch (InvalidOperationException ex)
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
@@ -75,14 +75,12 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
                 {
                     defaultProcessInfo = await _diagnosticServices.GetProcessAsync(null, HttpContext.RequestAborted);
                 }
-                catch (ArgumentException ex)
+                catch (ArgumentException)
                 {
                     // Unable to locate a default process; no action required
-                    _logger.LogInformation(ex.Message);
                 }
-                catch (InvalidOperationException ex)
+                catch (InvalidOperationException)
                 {
-                    _logger.LogInformation(ex.Message);
                 }
 
                 IList<Models.ProcessIdentifier> processesIdentifiers = new List<Models.ProcessIdentifier>();
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
@@ -84,7 +84,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
                 }
                 catch (Exception ex) when (!(ex is OperationCanceledException))
                 {
-                    _logger.LogWarning(ex.Message);
+                    _logger.DefaultProcessUnexpectedFailure(ex);
                 }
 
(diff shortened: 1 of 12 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/ProcessInfoImpl.cs
```diff
@@ -92,7 +92,19 @@ public static async Task<IProcessInfo> FromEndpointInfoAsync(IEndpointInfo endpo
 
                     commandLine = await commandLineSource.Task;
                 }
-                catch
+                catch (ObjectDisposedException)
```
kkeirstead, line 95: I wasn't super confident in expanding this one, but this was my best guess based on looking at the code in the `try`.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -75,10 +75,13 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
                 {
                     defaultProcessInfo = await _diagnosticServices.GetProcessAsync(null, HttpContext.RequestAborted);
                 }
-                catch (Exception)
```
jander-msft, line 78: Let's add the general Exception catch block at the end with a filter for non OperationCanceledException instances. Consider adding logging at warning level to the catch block to notify the user when a non-OperationCanceledException exception occurs.

src/Microsoft.Diagnostics.Monitoring.WebApi/ProcessInfoImpl.cs
jander-msft: Let's remove ObjectDisposedException, ArgumentNullException, and ArgumentException catchers.

src/Tools/dotnet-monitor/Egress/FileSystem/FileSystemEgressProvider.cs
```diff
@@ -75,8 +78,9 @@ public FileSystemEgressProvider(ILogger<FileSystemEgressProvider> logger)
                             File.Delete(intermediateFilePath);
                         }
                     }
-                    catch (Exception)
+                    catch (Exception ex)
                     {
+                        _logger.LogTrace(ex.Message);
```
jander-msft, line 83: Debug level?

jander-msft: Also, add the LoggerExtensions instead of calling ILogger directly. And pass the exception through the exception parameter.

src/Tools/dotnet-monitor/FilteredEndpointInfoSource.cs
```diff
@@ -63,8 +63,9 @@ internal class FilteredEndpointInfoSource : IEndpointInfoSourceInternal
                         _runtimeInstanceCookieToFilterOut = runtimeInstanceCookie;
                     }
                 }
-                catch (Exception)
+                catch (Exception ex)
                 {
+                    clientSourceLogger.LogTrace(ex.Message);
```
jander-msft, line 68: Debug level. Also, add the LoggerExtensions instead of calling ILogger directly. And pass the exception through the exception parameter.

src/Tools/dotnet-monitor/Startup.cs
```diff
@@ -155,8 +155,9 @@ public void ConfigureServices(IServiceCollection services)
                     {
                         address = BindingAddress.Parse(url);
                     }
-                    catch (Exception)
+                    catch (FormatException ex)
                     {
+                        logger.LogInformation(ex.Message);
```
jander-msft, line 160: Warning level. Also, add the LoggerExtensions instead of calling ILogger directly. And pass the exception through the exception parameter.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -75,10 +75,17 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
...
+                catch (ArgumentException)
                 {
                     // Unable to locate a default process; no action required
                 }
+                catch (InvalidOperationException)
+                {
+                }
+                catch (Exception ex) when (!(ex is OperationCanceledException))
+                {
+                    _logger.LogWarning(ex.Message);
```
jander-msft, line 87: Please move to LoggingExtensions

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -565,6 +565,9 @@
     <comment>Gets the format string that is printed in the 14:InsecureAutheticationConfiguration event.
 0 Format Parameters</comment>
   </data>
+  <data name="LogFormatString_IntermediateFileDeletionFailed" xml:space="preserve">
```
jander-msft, line 568: All of the format strings that are used for logging should use named holes rather than indexed holes: e.g. change `{0}` to `{path}`.

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -625,6 +631,9 @@
     <comment>Gets the format string that is printed in the 19:RunningElevated event.
 0 Format Parameters</comment>
   </data>
+  <data name="LogFormatString_RuntimeInstanceCookieFailedToFilterSelf" xml:space="preserve">
+    <value>The runtime instance cookie failed to filter out the current process.</value>
```
jander-msft, line 635: Unable to get the runtime instance cookie of the current process.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
jander-msft: How about "Failed to determine the default process." and pass the exception in the exception parameter?

jander-msft (approved):

PR #2056, opened 2022-06-30
Use auth library versions that correspond with target framework

jander-msft:
This change makes it so that the dotnet-monitor for each TFM uses the corresponding Microsoft.AspNetCore.Authentication.* libraries that best match the TFM. Thus the net6.0 dotnet-monitor will use the 6.0.* versions whereas the net7.0 dotnet-monitor will use the 7.0.* versions.


src/Tools/dotnet-monitor/dotnet-monitor.csproj
```diff
@@ -58,11 +58,11 @@
   </ItemGroup>
 
   <PropertyGroup>
-    <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage);GatherNativeFilesForTfm</TargetsForTfmSpecificContentInPackage>
+    <TargetsForTargetFrameworkSpecificContentInPackage>$(TargetsForTargetFrameworkSpecificContentInPackage);GatherNativeFilesForTargetFramework</TargetsForTargetFrameworkSpecificContentInPackage>
```
jander-msft, line 61: I think this was a find-and-replace mistake; I'll revert these.

PR #2091, opened 2022-07-06
Fix System.CommandLine version information

jander-msft:
The System.CommandLine dependency hasn't been updated because the version information does not match the package. I've fixed it up to use the same version that is in [dotnet/installer](https://github.com/dotnet/sdk/blob/main/eng/Version.Details.xml#L267).

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -1,9 +1,5 @@
 <Dependencies>
   <ProductDependencies>
-    <Dependency Name="System.CommandLine.Experimental" Version="&gt;0.2.0-alpha.19213.1">
-      <Uri>https://github.com/dotnet/command-line-api</Uri>
-      <Sha>166610c56ff732093f0145a2911d4f6c40b786da</Sha>
-    </Dependency>
     <Dependency Name="Microsoft.Diagnostics.Monitoring" Version="5.0.0-preview.22354.1">
       <Uri>https://github.com/dotnet/diagnostics</Uri>
       <Sha>dfb3b8887d27640c56e701cbddeab29e153c9158</Sha>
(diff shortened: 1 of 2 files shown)


src/Tools/dotnet-monitor/Program.cs
```diff
@@ -38,10 +39,29 @@ private static Command CollectCommand()
...
+                NoHttpEgressOption,
+                ConfigurationFilePathOption
             };
 
-            command.SetHandler<CancellationToken, string[], string[], bool, string, bool, bool, bool, FileInfo>(CollectCommandHandler.Invoke, command.Children.OfType<IValueDescriptor>().ToArray());
+            command.SetHandler(async context =>
+            {
+                context.ExitCode = await CollectCommandHandler.Invoke(
+                    context.GetCancellationToken(),
+                    context.ParseResult.GetValueForOption(UrlsOption),
```
wiktork, line 56: Is this version of the Invoke command strongly typed? I.e. if there's a mismatch between the children declared above and the results passed here, what happens?

jander-msft: Yes, the Inovke method is something that we wrote. In this change, we are manually getting each corresponding value from each option and passing it into the Invoke method. If someone adds a new parameter to the Invoke method or changes the type of an existing parameter, this code will fail to compile, and it should be obvious why.

jander-msft: If another option is added to the Command instance but not consumed by the Invoke invocation, then that would not be obvious why that is. However, I would expect that someone who is adding a new option also attempted to plumb through the use of the new information somewhere and that would have had to been surfaced up at the Invoke method.

wiktork (approved):

PR #2108, opened 2022-07-12
Convert logger string format and arguments to native platform string width

jander-msft:
These changes update the logger infrastructure to convert the format and argument strings to appropriate widths that are natively understood by the target platform. This allows for using the C formatting functions that are native to each target platform. This is achieved using a variadic template function to convert each argument as appropriate. To minimize conversions, the format string is required to already be in the string width of the target platform.

Additionally, another logger has been added that will log to the stderr stream. Its default level is None, but can be changed using the `DotnetMonitorProfiler_StdErrLogger_Level` environment variable.

Finally, result checking has been added to the string formatting and stream writing API usage to validate that strings are written correctly.

Tested these changes with a .NET 6 application that throws exceptions with the stderr logger enabled on both Windows and WSL Ubuntu.

### DIFF
--- a/src/MonitorProfiler/Logging/DebugLogger.cpp
+++ b/src/MonitorProfiler/Logging/DebugLogger.cpp
@@ -6,26 +6,28 @@
 #include "LogLevelHelper.h"
 #include "../Environment/EnvironmentHelper.h"
 #include "NullLogger.h"
+#include "macros.h"
+
 using namespace std;
 
-DebugLogger::DebugLogger(const shared_ptr<IEnvironment>& pEnvironment)
+DebugLogger::DebugLogger(const shared_ptr<IEnvironment>& environment)
 {
     // Try to get log level from environment
 
-    EnvironmentHelper helper(pEnvironment, NullLogger::Instance);
-    if (FAILED(helper.GetDebugLoggerLevel(m_level)))
+    EnvironmentHelper helper(environment, NullLogger::Instance);
+    if (FAILED(helper.GetDebugLoggerLevel(_level)))
     {
         // Fallback to default level
-        m_level = s_DefaultLevel;
+        _level = DefaultLevel;
     }
 }
 
 STDMETHODIMP_(bool) DebugLogger::IsEnabled(LogLevel level)
 {
-    return LogLevelHelper::IsEnabled(level, m_level);
+    return LogLevelHelper::IsEnabled(level, _level);
 }
 
-STDMETHODIMP DebugLogger::Log(LogLevel level, const tstring format, va_list args)
+STDMETHODIMP DebugLogger::Log(LogLevel level, const lstring& message)
 {
     if (!IsEnabled(level))
     {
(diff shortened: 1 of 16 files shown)


src/MonitorProfiler/Logging/DebugLogger.h
```diff
@@ -16,13 +16,13 @@ class DebugLogger final :
     public ILogger
 {
 private:
-    const static LogLevel s_DefaultLevel = LogLevel::Information;
-    const static int s_nMaxEntrySize = 1000;
+    const static LogLevel DefaultLevel = LogLevel::Information;
+    const static size_t MaxEntrySize = 1000;
```
kkeirstead, line 20: Out of curiosity, why did we switch from `int` to `size_t`?

src/MonitorProfiler/Logging/LogLevelHelper.h
```diff
@@ -16,35 +16,36 @@ class LogLevelHelper final
...
+            strName.assign(_LS("trce"));
             return S_OK;
         case LogLevel::Warning:
-            strName.assign(_T("warn"));
+            strName.assign(_LS("warn"));
             return S_OK;
         default:
-            return E_FAIL;
+            strName.assign(_LS("ukwn"));
+            return S_OK;
```
kkeirstead, line 48: It looks like this method will return `S_OK` no matter what - is this intentional, and if so, do we even need to return `HRESULT` and use `IfFailRet` on it?

src/MonitorProfiler/Logging/StdErrLogger.h
```diff
@@ -0,0 +1,35 @@
...
+
+/// <summary>
+/// Logs messages to the stderr stream.
+/// </summary>
+class StdErrLogger final :
+    public ILogger
+{
+private:
+    const static LogLevel DefaultLevel = LogLevel::None;
+    const static int MaxEntrySize = 1000;
```
kkeirstead, line 20: It seems like in some places `MaxEntrySize` is `size_t` and others it's `int`. Is there a reason for this, and if so, is it worth documenting?

src/inc/macros.h
```diff
@@ -7,3 +7,8 @@
 #ifndef ExpectedPtr
 #define ExpectedPtr(ptr) { if (nullptr == ptr) return E_POINTER; }
 #endif
+
+#ifndef HRESULT_FROM_ERRNO
+#define HRESULT_FROM_ERRNO(value) \
+        MAKE_HRESULT(value == 0 ? SEVERITY_SUCCESS : SEVERITY_ERROR, FACILITY_NULL, HRESULT_CODE(value))
```
kkeirstead, line 13: Not sure if this is a common pattern, but based on the usage of this macro the ternary seems redundant - the caller appears to only use this when `errno != 0`, thus, the condition should never evaluate to true. That being said, I don't know if there's a future use-case where we would want to use `SEVERITY_SUCCESS`; if there isn't, should we enforce that `value` cannot be 0 here so that the caller can't inadvertently pass an `errno` that equals 0?

src/MonitorProfiler/Logging/LogLevelHelper.h
jander-msft: This was intentional. I didn't want to have to rewrite the callers.

src/MonitorProfiler/Logging/StdErrLogger.h
jander-msft: This should be changed to `size_t` to match the method signatures. I'll fix this one.

src/inc/macros.h
jander-msft: This should be fine the way it is. If a future caller forgets to check that errno != 0, then it will effectively return `(HRESULT)0`, which should be the same as `S_OK`.

src/MonitorProfiler/Logging/Logger.h
```diff
@@ -35,16 +54,51 @@ DECLARE_INTERFACE(ILogger)
...
-    STDMETHOD(Log)(LogLevel level, const tstring format, va_list args) PURE;
+    STDMETHOD(Log)(LogLevel level, const lstring& message) PURE;
+
+    template <typename... T>
+    inline STDMETHODIMP Log(LogLevel level, const lstring format, const T&... args)
+    {
+        // A cache of strings that were converted from their original width
+        // to the string width for the target platform. This prevents freeing of the
+        // converted strings before logging can complete.
+        std::vector<lstring> argStrings;
```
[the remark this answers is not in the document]
jander-msft: Added as a consideration

src/MonitorProfiler/Logging/StdErrLogger.cpp
```diff
@@ -0,0 +1,105 @@
...
+            {
+                // This is an undocumented condition.
+                hr = E_UNEXPECTED;
+            }
+
+            // While there was a failure to write to the stream, errno was not
+            // updated by the platform API call. Restore the value of errno prior
+            // to invoking the string formatting API. This is done after ferror
+            // is invoked in case ferror sets errno.
+            errno = previousError;
```
wiktork, line 93: nit: I'm wondering if we can code share some of the errno handling between loggers? Perhaps a higher order function?

wiktork (dismissed):

src/MonitorProfiler/Logging/DebugLogger.cpp
```diff
@@ -34,28 +36,36 @@ STDMETHODIMP DebugLogger::Log(LogLevel level, const tstring format, va_list args
...
+    // occurred.
+    if (result < 0 && errno != 0)
+    {
+        return HRESULT_FROM_ERRNO(errno);
+    }
+
+    // Successful invocations of platform APIs typically do not modify errno
+    // if no failure occurs. To maintain this behavior, restore the value of
+    // errno prior to invoking the string formatting API.
+    errno = previousError;
```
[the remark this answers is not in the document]
jander-msft: Done

src/MonitorProfiler/Logging/StdErrLogger.cpp
jander-msft: Done: higher order variadic template function + function pointer ftw

wiktork (approved):

PR #2119, opened 2022-07-13
Collect Live Metrics Action

### LINKED ISSUE #1809: CollectLiveMetrics action support
Add support for collecting metrics via a CollectionRule action

kkeirstead:
Added in the collect live metrics action as well as some tests.

Closes #1809


documentation/schema.json
```diff
@@ -1335,6 +1345,7 @@
       "enum": [
         "CollectDump",
         "CollectGCDump",
+        "CollectLiveMetrics",
```
jander-msft, line 1348: I know that the route name is `livemetrics`, but I'm wondering if the action should just be called `CollectMetrics`. I would think that there won't be any other type of metrics collection action that would need to be differentiated from this one.

jander-msft (approved):

documentation/schema.json
kkeirstead: Even if we aren't planning an action for metrics, I think that might introduce some confusion in terms of the expected type being returned (e.g.  we say "Metrics is Prometheus, and LiveMetrics is Json, _except_ in the case of the collection rules, where metrics is Json"). That being said, if people want me to change it, I can - just my 2 cents.

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/MetricsUtilities.cs
```diff
@@ -0,0 +1,38 @@
...
+            var client = new DiagnosticsClient(endpointInfo.Endpoint);
+         
+            await using EventCounterPipeline eventCounterPipeline = new EventCounterPipeline(client,
+                settings,
+                loggers:
+                new[] { new JsonCounterLogger(outputStream) });
+
+            Task runTask = await eventCounterPipeline.StartAsync(token);
+
+            if (null != startCompletionSource)
```
wiktork, line 27: nit: startCompletionSource?.TrySetResult(null)

wiktork (approved):

Issue #2122, opened 2022-07-14
Operations Api shows RuntimeId


wiktork:
Add RuntimeId to the /operations api.
- Should we able able to query for operations by uuid?

jander-msft:
Should we also be able to query for operations using process name or process ID for convenience?

PR #2136, opened 2022-07-19
Operations Api shows RuntimeId, ProcessId, and ProcessName

cablooshe:
This is for #2122. It updates the /operations and /operations/operationId api to show the RuntimeId, ProcessId, and ProcessName as a sub-object shown in the results.

Main question I wanted to bring up for discussion in this PR: 

Most EgressOperations created can't really take in an `IProcessInfo` object since they pull an endpoint directly, which doesn't have the ProcessName stored. There are a few ways I'm thinking we can use to fix this - 

1. Leave as-is and let most operations not have a processName that we can use.
2. If the processName isn't that relevant we can remove it from the API update so we only show RuntimeId and ProcessId.
3. We can query for the processInfo object using the found RuntimeId and get the name from that, similar to what's done [in DiagController.GetProcessInfo()](https://github.com/dotnet/dotnet-monitor/blob/b98c2d9c04746f6489cf9f2fbed2e3027a9414e9/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs#L128). Not sure how much I like this.
4. Update `IEndpointInfo` to have the processName inside it. Not sure how doable it is but I'm investigating.

Let me know what you think would be best, I'm not entirely sure myself but I'm leaning towards the 4th option, if possible.

### DIFF
--- a/documentation/api/operations-get.md
+++ b/documentation/api/operations-get.md
@@ -49,7 +49,11 @@ Content-Type: application/json
     "error": null,
     "operationId": "67f07e40-5cca-4709-9062-26302c484f18",
     "createdDateTime": "2021-07-21T06:21:15.315861Z",
-    "status": "Succeeded"  
+    "status": "Succeeded",
+    "processInfo":
+    {
+        "pid":1 "uid":"95b0202a-4ed3-44a6-98f1-767d270ec783","processName":"dotnet-monitor-demo"
+    }
 }
 ```
 
--- a/documentation/api/operations-get.md
+++ b/documentation/api/operations-get.md
@@ -54,7 +54,7 @@ Content-Type: application/json
     {
         "pid":1,
         "uid":"95b0202a-4ed3-44a6-98f1-767d270ec783",
-        "processName":"dotnet-monitor-demo"
+        "name":"dotnet-monitor-demo"
     }
 }
 ```
--- a/documentation/api/operations-get.md
+++ b/documentation/api/operations-get.md
@@ -50,9 +50,11 @@ Content-Type: application/json
     "operationId": "67f07e40-5cca-4709-9062-26302c484f18",
     "createdDateTime": "2021-07-21T06:21:15.315861Z",
     "status": "Succeeded",
-    "processInfo":
+    "process":
     {
-        "pid":1 "uid":"95b0202a-4ed3-44a6-98f1-767d270ec783","processName":"dotnet-monitor-demo"
+        "pid":1,
+        "uid":"95b0202a-4ed3-44a6-98f1-767d270ec783",
+        "processName":"dotnet-monitor-demo"
     }
 }
 ```
--- a/documentation/openapi.json
+++ b/documentation/openapi.json
@@ -1660,7 +1660,7 @@
             "type": "string",
             "format": "uuid"
           },
-          "processName": {
+          "name": {
             "type": "string",
             "nullable": true
           }
--- a/documentation/openapi.json
+++ b/documentation/openapi.json
@@ -1684,7 +1684,7 @@
           "status": {
             "$ref": "#/components/schemas/OperationState"
           },
-          "processInfo": {
+          "process": {
             "$ref": "#/components/schemas/OperationProcessInfo"
           }
         },
@@ -1719,7 +1719,7 @@
           "status": {
             "$ref": "#/components/schemas/OperationState"
           },
-          "processInfo": {
+          "process": {
             "$ref": "#/components/schemas/OperationProcessInfo"
           },
           "resourceLocation": {
--- a/documentation/openapi.json
+++ b/documentation/openapi.json
@@ -1654,13 +1654,11 @@
         "properties": {
           "pid": {
             "type": "integer",
-            "format": "int32",
-            "nullable": true
+            "format": "int32"
           },
           "uid": {
             "type": "string",
-            "format": "uuid",
-            "nullable": true
+            "format": "uuid"
           },
           "processName": {
             "type": "string",
--- a/documentation/openapi.json
+++ b/documentation/openapi.json
@@ -1649,6 +1649,27 @@
         ],
         "type": "string"
       },
+      "OperationProcessInfo": {
+        "type": "object",
+        "properties": {
+          "pid": {
+            "type": "integer",
+            "format": "int32",
+            "nullable": true
+          },
+          "uid": {
+            "type": "string",
+            "format": "uuid",
+            "nullable": true
+          },
+          "processName": {
+            "type": "string",
+            "nullable": true
+          }
+        },
+        "additionalProperties": false,
+        "description": "Represents the details of a given process used in an operation."
+      },
       "OperationSummary": {
         "type": "object",
         "properties": {
@@ -1662,6 +1683,9 @@
           },
           "status": {
             "$ref": "#/components/schemas/OperationState"
+          },
+          "processInfo": {
+            "$ref": "#/components/schemas/OperationProcessInfo"
           }
         },
         "additionalProperties": false,
@@ -1694,6 +1718,9 @@
           },
           "status": {
             "$ref": "#/components/schemas/OperationState"
+          },
+          "processInfo": {
+            "$ref": "#/components/schemas/OperationProcessInfo"
           },
           "resourceLocation": {
             "type": "string",
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
@@ -238,7 +238,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
                         token => _dumpService.DumpAsync(processInfo.EndpointInfo, type, token),
                         egressProvider,
                         dumpFileName,
-                        processInfo.EndpointInfo,
+                        processInfo,
                         ContentTypes.ApplicationOctetStream,
                         scope), limitKey: Utilities.ArtifactType_Dump);
                 }
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
@@ -299,7 +299,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
                     },
                     fileName,
                     ContentTypes.ApplicationOctetStream,
-                    processInfo.EndpointInfo);
+                    processInfo);
             }, processKey, Utilities.ArtifactType_GCDump);
         }
 
@@ -632,7 +632,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
                 },
                 fileName,
                 ContentTypes.ApplicationOctetStream,
-                processInfo.EndpointInfo);
+                processInfo);
         }
 
         private Task<ActionResult> StartLogs(
@@ -655,7 +655,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
                 (outputStream, token) => LogsUtilities.CaptureLogsAsync(null, format.Value, processInfo.EndpointInfo, settings, outputStream, token),
                 fileName,
                 contentType,
-                processInfo.EndpointInfo,
+                processInfo,
                 format != LogFormat.PlainText);
         }
 
@@ -704,10 +704,10 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
             Func<Stream, CancellationToken, Task> action,
             string fileName,
             string contentType,
-            IEndpointInfo endpointInfo,
+            IProcessInfo processInfo,
             bool asAttachment = true)
         {
-            KeyValueLogScope scope = Utilities.CreateArtifactScope(artifactType, endpointInfo);
+            KeyValueLogScope scope = Utilities.CreateArtifactScope(artifactType, processInfo.EndpointInfo);
 
             if (string.IsNullOrEmpty(providerName))
             {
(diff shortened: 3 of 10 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -238,7 +238,7 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
                         token => _dumpService.DumpAsync(processInfo.EndpointInfo, type, token),
                         egressProvider,
                         dumpFileName,
-                        processInfo.EndpointInfo,
+                        processInfo,
```
kkeirstead, line 241: Is there a reason not to follow this pattern in the `Result` method? Currently, it takes `IEndpointInfo`, but just skimming through the callers it looks like they're all doing `processInfo.EndpointInfo`. That would give you your `ProcessName` for all of those cases, unless I'm overlooking something.

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationStore.cs
```diff
@@ -2,14 +2,17 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.DependencyInjection;
 using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Security.Cryptography;
```
kkeirstead, line 11: nit: Remove and Sort Usings

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationStore.cs
```diff
@@ -128,29 +131,46 @@ public IEnumerable<Models.OperationSummary> GetOperations()
...
+                            ProcessID = processInfo.ProcessId,
+                            UID = processInfo.RuntimeInstanceCookie
+                        }
+                    };
                 }).ToList();
             }
         }
 
-        public Models.OperationStatus GetOperationStatus(Guid operationId)
+    public Models.OperationStatus GetOperationStatus(Guid operationId)
```
kkeirstead, line 153: nit: fix spacing

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
cablooshe: Most of them will work just fine, but there are a few calls that use `DumpAsync` that refer to endpoints retrieved through the `ICollectionRuleAction`, which seems to be locked into using endpoints (`newEndpointInfoTask` through a lot of the tests, a `CollectionRuleContext` inside `ActionListExecutor.cs`. The call I particularly found for this was in `CollectDumpAction.cs`).

We could change these ones in particular but then I'll have to add a 4th `EgressOperation` constructor. Thoughts?

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/EgressOperationStatus.cs
```diff
@@ -20,6 +20,21 @@ public class OperationSummary
 
         [JsonPropertyName("status")]
         public OperationState Status { get; set; }
+
+        [JsonPropertyName("processInfo")]
+        public OperationProcessInfo ProcessInfo { get; set; }
+    }
+
+    public class OperationProcessInfo
```
wiktork, line 28: Need to add to documentation.

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressRequest.cs
```diff
@@ -30,6 +31,8 @@ public EgressRequest(Guid operationId, IEgressOperation egressOperation, IDispos
 
         public Guid OperationId { get; }
 
+        public EgressProcessInfo  ProcessInfo { get; }
```
wiktork, line 34: Can just use operation.GetEgressProcessInfo instead of storing another field.

wiktork, line 34: nit: extra space

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
wiktork: I think we are ok adding a 4th ctor, if it enables logs/metrics/trace/etc.

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperation.cs
```diff
@@ -16,19 +17,31 @@ internal class EgressOperation : IEgressOperation
...
             _scope = scope;
+            _egressProcessInfo = new EgressProcessInfo(null, source.ProcessId, source.RuntimeInstanceCookie);
+        }
+
+        public EgressOperation(Func<CancellationToken, Task<Stream>> action, string endpointName, string artifactName, IEndpointInfo source, string contentType, KeyValueLogScope scope)
+        {
+            _egress = (service, token) => service.EgressAsync(endpointName, action, artifactName, contentType, source, token);
+            _egressProvider = endpointName;
+            _scope = scope;
+            _egressProcessInfo = new EgressProcessInfo(null, source.ProcessId, source.RuntimeInstanceCookie);
```
wiktork, line 44: Not really used for collection rules anyway, we could leave it null.

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperation.cs
```diff
@@ -57,11 +70,30 @@ public async Task<ExecutionResult<EgressResult>> ExecuteAsync(IServiceProvider s
...
         {
             serviceProvider
                 .GetRequiredService<IEgressService>()
                 .ValidateProvider(_egressProvider);
         }
     }
+
+    internal readonly struct EgressProcessInfo
+    {
+        public readonly string ProcessName;
```
wiktork, line 88: nit: Properties with just getters.

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperation.cs
```diff
@@ -57,11 +78,30 @@ public async Task<ExecutionResult<EgressResult>> ExecuteAsync(IServiceProvider s
...
+
         public void Validate(IServiceProvider serviceProvider)
         {
             serviceProvider
                 .GetRequiredService<IEgressService>()
                 .ValidateProvider(_egressProvider);
         }
     }
+
+    internal readonly struct EgressProcessInfo
```
wiktork, line 94: Let's actually convert this to a class, since we return this value and sometimes it's null.

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/EgressOperationStatus.cs
kkeirstead: I may have misinterpeted @wiktork 's ask here, but I was assuming that this was in reference to needing to update our docs on GitHub (so that people know the operations endpoint now includes `ProcessID`, `UID`, and `ProcessName`)

cablooshe: Missed that lol. Just updated the API documentation for the list operations and get operations options.

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationStore.cs
```diff
@@ -128,11 +125,21 @@ public IEnumerable<Models.OperationSummary> GetOperations()
...
             lock (_requests)
             {
-                return _requests.Select((kvp) => new Models.OperationSummary
+                return _requests.Select((kvp) =>
                 {
-                    OperationId = kvp.Key,
-                    CreatedDateTime = kvp.Value.CreatedDateTime,
-                    Status = kvp.Value.State
+                    EgressProcessInfo processInfo = kvp.Value.EgressRequest.EgressOperation.GetEgressProcessInfo();
+                    return new Models.OperationSummary
```
wiktork, line 131: Once you convert to class, you'll need to null check on processInfo here.

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperation.cs
```diff
@@ -57,11 +78,30 @@ public async Task<ExecutionResult<EgressResult>> ExecuteAsync(IServiceProvider s
             }, logger, token);
         }
 
+        public EgressProcessInfo GetEgressProcessInfo()
```
jander-msft, line 81: nit: make this an init property: `public EgressProcessInfo ProcessInfo { get; private init; }`

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/EgressOperationStatus.cs
```diff
@@ -20,6 +20,24 @@ public class OperationSummary
...
+        public OperationProcessInfo ProcessInfo { get; set; }
+    }
+
+    /// <summary>
+    /// Represents the details of a given process used in an operation.
+    /// </summary>
+    public class OperationProcessInfo
+    {
+        [JsonPropertyName("pid")]
+        public int? ProcessID { get; set; }
```
jander-msft, line 34:
```suggestion
        public int? ProcessId { get; set; }
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/EgressOperationStatus.cs
```diff
@@ -20,6 +20,24 @@ public class OperationSummary
...
+
+        [JsonPropertyName("uid")]
+        public Guid? UID { get; set; }
```
jander-msft, line 37:
```suggestion
        public Guid? Uid { get; set; }
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/EgressOperationStatus.cs
```diff
@@ -20,6 +20,24 @@ public class OperationSummary
 
         [JsonPropertyName("status")]
         public OperationState Status { get; set; }
+
+        [JsonPropertyName("processInfo")]
+        public OperationProcessInfo ProcessInfo { get; set; }
```
jander-msft, line 25: Do we want to call this `ProcessInfo` in the returned result? Would it be simpler to just call it `Process`. I don't think the `Info` part conveys any more meaning and it's not something we've used in the rest of the HTTP API.

I'm also thinking longer term if we somehow collect an artifact from someone other mechanism and then we can add something like `"ArtifactSource": "Process"` as a property to indicate from what the artifact was collected.

cablooshe: I'm a fan of changing it to `Process`. Done.

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperation.cs
cablooshe: I think because one of the $(TestTargetFrameworks) is netcoreapp3.1, we can't use `init` without adding a dummy class to `System.Runtime.CompilerServices` since there's only support for .NET 5+, so I think it might be better to just use `{ get; private set; }`.

documentation/api/operations-get.md
```diff
@@ -49,7 +49,11 @@ Content-Type: application/json
     "error": null,
     "operationId": "67f07e40-5cca-4709-9062-26302c484f18",
     "createdDateTime": "2021-07-21T06:21:15.315861Z",
-    "status": "Succeeded"  
+    "status": "Succeeded",
+    "processInfo":
```
jander-msft, line 53: This example and the other one need to be fixed:
- The property is named `process`
- Please format the inner object so that there is only one property per line.
- The inner object is not valid JSON (missing comma between the `"pid":1` and `"uid":"..."`)

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/EgressOperationStatus.cs
```diff
@@ -20,6 +20,24 @@ public class OperationSummary
...
+        public OperationProcessInfo Process { get; set; }
+    }
+
+    /// <summary>
+    /// Represents the details of a given process used in an operation.
+    /// </summary>
+    public class OperationProcessInfo
+    {
+        [JsonPropertyName("pid")]
+        public int? ProcessId { get; set; }
```
jander-msft, line 34: These properties should not be nullable. If the egress operation is missing process information, it's the EgressProcessInfo object that is null, not the individual properties on them.

jander-msft (dismissed):

wiktork (dismissed):

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/EgressOperationStatus.cs
```diff
@@ -20,6 +20,24 @@ public class OperationSummary
...
+    public class OperationProcessInfo
+    {
+        [JsonPropertyName("pid")]
+        public int ProcessId { get; set; }
+
+        [JsonPropertyName("uid")]
+        public Guid Uid { get; set; }
+
+        [JsonPropertyName("processName")]
+        public string ProcessName { get; set; }
```
jander-msft, line 40: Could you change this to just `Name`/`name` and update the documentation? This will make it more consistent with the rest of the API that exposes process name. Since this is already under a property named `process` and is on the OperationProcessInfo object, it should be well understood that it is the name of the process without having to call it `processName`.

documentation/openapi.json
```diff
@@ -1649,6 +1649,25 @@
...
+        "properties": {
+          "pid": {
+            "type": "integer",
+            "format": "int32"
+          },
+          "uid": {
+            "type": "string",
+            "format": "uuid"
+          },
+          "processName": {
```
jander-msft, line 1663:
```suggestion
          "name": {
```

jander-msft (approved):

Issue #2118, opened 2022-07-13
BackgroundService failed when starting dotnet-monitor in AKS environment


bcarthic:
### Description
We are hosting dotnet-monitor as a side car container in our pod, last 2 days we see huge spike of BackgroundService being failed and the container restarted many times which caused high CPU and this got mitigated and the monitor started fine. But trying to understand the root cause of this failure which caused the spike in CPU.

```json
{
    "Timestamp": "2022-07-11T07:09:34.8375370Z",
    "EventId": 9,
    "LogLevel": "Error",
    "Category": "Microsoft.Extensions.Hosting.Internal.Host",
    "Message": "BackgroundService failed",
    "Exception": "System.Threading.Tasks.TaskCanceledException: A task was canceled. at Microsoft.Diagnostics.Monitoring.WebApi.MetricsService.ExecuteAsync(CancellationToken stoppingToken) in /_/src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsService.cs:line 71    at Microsoft.Extensions.Hosting.Internal.Host.TryExecuteBackgroundServiceAsync(BackgroundService backgroundService)",
    "State": { "Message": "BackgroundService failed", "{OriginalFormat}": "BackgroundService failed" },
    "Scopes": []
}

{
    "Timestamp": "2022-07-11T07:09:34.8387837Z",
    "EventId": 10,
    "LogLevel": "Critical",
    "Category": "Microsoft.Extensions.Hosting.Internal.Host",
    "Message": "The HostOptions.BackgroundServiceExceptionBehavior is configured to StopHost. A BackgroundService has thrown an unhandled exception, and the IHost instance is stopping. To avoid this behavior, configure this to Ignore; however the BackgroundService will not be restarted.",
    "Exception": "System.Threading.Tasks.TaskCanceledException: A task was canceled.   at Microsoft.Diagnostics.Monitoring.WebApi.MetricsService.ExecuteAsync(CancellationToken stoppingToken) in /_/src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsService.cs:line 71    at Microsoft.Extensions.Hosting.Internal.Host.TryExecuteBackgroundServiceAsync(BackgroundService backgroundService)",
    "State": {
[...]

Issue #1827, opened 2022-05-01
"System.Net.Sockets.SocketException (98): Address in use" exception during rolling deployment


joelnotified:
I've got the following yaml configured for an application running in an AKS cluster, with dotnet-monitor as a sidecar. It runs fine, but during a rolling deployment (of a new application version), we got the following error: `System.Net.Sockets.SocketException (98): Address in use`. I'm attaching the logs as well. 

Not sure if this is actually reproduced on each rolling deployment. An engineer on my team notified me and removed dotnet-monitor, so I can't really test in under the same circumstances right now. I think the issue was that it actually prevented the old application instance to be removed.

Is this something you have seen or do you understand by the configuration what's going on? Looks like it's when it shutting down that it's actually getting the exception? 🤔 

(I know we shouldn't use `--no-auth` and expose the port, but we're just evaluating it in a test environment right now)

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: itemsservice
  labels:
    app: itemsservice
spec:
  replicas: 3
  selector:
    matchLabels:
      app: itemsservice
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 0
      maxSurge: 1
  template:
    metadata:
      labels:
        app: itemsservice
      annotations:
        prometheus.io/scrape: "true"
        prometheus.io/port: "9102"
    spec:
      volumes:
        - name: diagnostics
          emptyDir: {}
        - name: dumps
          emptyDir: {}
        - name: dotnet-monitor-config
          configMap:
            name: dotnet-monitor-config
      containers:
        - name: itemsservice
          image: <redacted>
          ports:
            - containerPort: 80
            - containerPort: 9102
          env:
            - name: DOTNET_DiagnosticPorts
              value: /diag/port
          volumeMounts:
            - name: diagnostics
              mountPath: /diag
            - name: dumps
              mountPath: /dumps
          resources:
            requests:
              memory: "256Mi"
              cpu: "50m"
            limits:
              memory: "2Gi"
          startupProbe:
            httpGet:
              path: /health
              port: 80
            timeoutSeconds: 10
            failureThreshold: 6
            periodSeconds: 10
          livenessProbe:
            httpGet:
              path: /health
              port: 80
            initialDelaySeconds: 0
            timeoutSeconds: 10
            failureThreshold: 3
            periodSeconds: 30
        - name: dotnet-monitor
[...]

Issue #2017, opened 2022-06-20
container keep restarting with dotnetmonitor side car and DOTNET_DiagnosticPorts configured


allanchanly:
### Description

We have a statefulset that configured with dotnetmonitor as sidecar, but after running for a few days, some container keeps restarting because of below error: 
The runtime has been configured to pause during startup and is awaiting a Diagnostics IPC ResumeStartup command from a Diagnostic Port.
DOTNET_DiagnosticPorts="/diag/port"
DOTNET_DefaultDiagnosticPortSuspend=0

### Configuration

Our statefulset is running in AKS, and dotnetmonitor is added as sidecar, environment variable DOTNET_DiagnosticPorts is set, DOTNET_DefaultDiagnosticPortSuspend is not set and it will get default value 0 per the dotnetmonitor document. 

### Regression?

In the same statefulset, most of the pods are running fine, this only happens to a few pods.

### Other information
![bugreport](https://user-images.githubusercontent.com/3281340/174677017-a91e7976-dd1c-45b0-9917-df9fb3b9cca4.png)

jander-msft:
This appears to be the same issue in #1958 that you logged a few weeks ago.

I would ask that you still provide the same information that I asked for in that issue:

> In the pods where this is occurring, please verify that:
> 1. There is a `dotnet monitor` container
> 1. It is configured to host its diagnostic port at `/diag/port`
> 1. The `/diag` folder is mounted to the same volume between the application container and the `dotnet monitor` container.

> If these are correct, check the logs of the `dotnet monitor` container to see if there was any issue.

jander-msft:
> I checked today, we do see some logs with error:
> 
> ![image](https://user-images.githubusercontent.com/3281340/174898148-67785608-1838-4d0f-8704-c45f8af97132.png)

This is likely the same issue as #1827. In your case, what's likely happening is that when you've updated or redeployed the StatefulSet, it had a stale Unix domain socket file at the same path that you are specifying for the diagnostic port. So dotnet-monitor fails to establish its server there and the target applications fail to communicate at that path and remain stuck in the "The runtime has been configured to pause during startup and is awaiting a Diagnostics IPC ResumeStartup command from a Diagnostic Port."

We have a fix (#2164) where we will enable deleting the Unix domain socket at startup before the server attempts to create a new socket file at the path. We hope to have this fix rolled out in the next dotnet-monitor update and automatically enabled in the dotnet-monitor images.

[...]

PR #2164, opened 2022-07-24
Delete endpoint on startup

wiktork:
Possible fix for https://github.com/dotnet/dotnet-monitor/issues/2118 and https://github.com/dotnet/dotnet-monitor/issues/1827

### DIFF
--- a/documentation/schema.json
+++ b/documentation/schema.json
@@ -902,6 +902,13 @@
           ],
           "description": "In 'Listen' mode, the maximum amount of connections to accept.",
           "format": "int32"
+        },
+        "ClearEndpointOnStartup": {
+          "type": [
+            "boolean",
+            "null"
+          ],
+          "description": "In reverse connection mode, deletes the domain socket file used for Diagnostic Port communication."
         }
       }
     },
--- a/documentation/schema.json
+++ b/documentation/schema.json
@@ -908,7 +908,7 @@
             "boolean",
             "null"
           ],
-          "description": "In 'listen' connection mode, deletes the domain socket file used for Diagnostic Port communication.",
+          "description": "In 'listen' connection mode, deletes the domain socket file used for diagnostic port communication.",
           "default": false
         }
       }
--- a/documentation/schema.json
+++ b/documentation/schema.json
@@ -903,12 +903,13 @@
           "description": "In 'Listen' mode, the maximum amount of connections to accept.",
           "format": "int32"
         },
-        "ClearEndpointOnStartup": {
+        "DeleteEndpointOnStartup": {
           "type": [
             "boolean",
             "null"
           ],
-          "description": "In reverse connection mode, deletes the domain socket file used for Diagnostic Port communication."
+          "description": "In 'listen' connection mode, deletes the domain socket file used for Diagnostic Port communication.",
+          "default": false
         }
       }
     },
--- a/src/Microsoft.Diagnostics.Monitoring.Options/DiagnosticPortOptions.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.Options/DiagnosticPortOptions.cs
@@ -27,7 +27,8 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
 
         [Display(
             ResourceType = typeof(OptionsDisplayStrings),
-            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_DiagnosticPortOptions_ClearEndpointOnStartup))]
-        public bool? ClearEndpointOnStartup { get; set; }
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_DiagnosticPortOptions_DeleteEndpointOnStartup))]
+        [DefaultValue(DiagnosticPortOptionsDefaults.DeleteEndpointOnStartup)]
+        public bool? DeleteEndpointOnStartup { get; set; }
     }
 }
--- a/src/Microsoft.Diagnostics.Monitoring.Options/DiagnosticPortOptions.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.Options/DiagnosticPortOptions.cs
@@ -24,5 +24,10 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
             ResourceType = typeof(OptionsDisplayStrings),
             Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_DiagnosticPortOptions_MaxConnections))]
         public int? MaxConnections { get; set; }
+
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_DiagnosticPortOptions_ClearEndpointOnStartup))]
+        public bool? ClearEndpointOnStartup { get; set; }
     }
 }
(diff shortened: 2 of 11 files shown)


src/Tools/dotnet-monitor/LoggingEventIds.cs
```diff
@@ -74,7 +74,9 @@ internal enum LoggingEventIds
         DiagnosticPortNotInListenModeForCollectionRules = 61,
         RuntimeInstanceCookieFailedToFilterSelf = 62,
         ParsingUrlFailed = 63,
-        IntermediateFileDeletionFailed = 64
+        IntermediateFileDeletionFailed = 64,
+        EndpointDeleteAttempt = 65,
```
jander-msft, line 78:
```suggestion
        DiagnosticPortDeleteAttempt = 65,
```

src/Tools/dotnet-monitor/LoggingEventIds.cs
```diff
@@ -74,7 +74,9 @@ internal enum LoggingEventIds
...
+        EndpointDeleteFailed = 66
```
jander-msft, line 79:
```suggestion
        DiagnosticPortDeleteFailed = 66
```

src/Microsoft.Diagnostics.Monitoring.Options/DiagnosticPortOptions.cs
```diff
@@ -24,5 +24,10 @@ public class DiagnosticPortOptions
             ResourceType = typeof(OptionsDisplayStrings),
             Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_DiagnosticPortOptions_MaxConnections))]
         public int? MaxConnections { get; set; }
+
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_DiagnosticPortOptions_ClearEndpointOnStartup))]
+        public bool? ClearEndpointOnStartup { get; set; }
```
kkeirstead, line 31: Should we have a `DefaultValue` here (following the pattern for `ConnectionMode`)?

src/Tools/dotnet-monitor/EndpointInfo/ServerEndpointInfoSource.cs
```diff
@@ -82,6 +84,23 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+                    File.Exists(_portOptions.EndpointName))
+                {
+                    // In some circumstances stale files from previous instances of dotnet-monitor cause
+                    // the new instance to fail binding. We need to delete the file in this situation.
+                    try
+                    {
+                        _logger.EndpointDeleteAttempt(_portOptions.EndpointName);
+                        File.Delete(_portOptions.EndpointName);
+                    }
+                    catch (Exception ex)
```
kkeirstead, line 98: Is this a situation where we want to swallow all exceptions? I know we went through this exercise as part of the security review, so I just want to make sure we're still following that pattern (where applicable).

src/Tools/dotnet-monitor/EndpointInfo/ServerEndpointInfoSource.cs
```diff
@@ -82,6 +84,23 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+                   !RuntimeInformation.IsOSPlatform(OSPlatform.Windows) &&
+                    File.Exists(_portOptions.EndpointName))
+                {
+                    // In some circumstances stale files from previous instances of dotnet-monitor cause
+                    // the new instance to fail binding. We need to delete the file in this situation.
+                    try
+                    {
+                        _logger.EndpointDeleteAttempt(_portOptions.EndpointName);
+                        File.Delete(_portOptions.EndpointName);
+                    }
```
kkeirstead, line 97: Consider: We always say that we're attempting the delete, but only say if we failed (and say nothing if we succeed) - do we want to log a message (lower than Warning level) that informs people that the deletion succeeded? May be unnecessary, but was just a thought.

src/Tools/dotnet-monitor/EndpointInfo/ServerEndpointInfoSource.cs
jander-msft: Yes, failing to delete the endpoint is not a failure mode of the tool, so all exceptions should be caught and logged. And the logging infra will log all of the exception information.

src/Microsoft.Diagnostics.Monitoring.Options/DiagnosticPortOptionsDefaults.cs
```diff
@@ -7,5 +7,6 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
     internal class DiagnosticPortOptionsDefaults
     {
         public const DiagnosticPortConnectionMode ConnectionMode = DiagnosticPortConnectionMode.Connect;
+        public const bool ClearEndpointOnStartup = false;
```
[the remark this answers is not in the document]
jander-msft: I think we should make this `true` in 7.x by default. I'm not sure if it should be `true` for the 6.x images since that would be a breaking change (although is suspect it won't be an issue for most users).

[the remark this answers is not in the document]
jander-msft: I think that it is fair to say that this is a bug fix and thus applicable to 6.x (and the 6.2 release as well).

src/Tools/dotnet-monitor/EndpointInfo/ServerEndpointInfoSource.cs
wiktork: This would be reasonable, but I mostly added these messages for our purposes since this is reactionary to a specific customer issue. I don't think any additional lines would be necessary for us to understand the logs around this issue.

src/Microsoft.Diagnostics.Monitoring.Options/DiagnosticPortOptionsDefaults.cs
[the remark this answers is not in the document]
wiktork: My concern here was that making this the default behavior will lead to a set of secondary issues that may be much harder to diagnose. People who are for some reason running two dotnet-monitors won't naturally fail due to binding but rather succeed and then get very bizarre behavior.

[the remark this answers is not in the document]
jander-msft: Would a reasonable middle ground be that we set as an environment variable in the dotnet-monitor container to `true` and leave it as `false` by default in the tool?

documentation/schema.json
```diff
@@ -902,6 +902,14 @@
           ],
           "description": "In 'Listen' mode, the maximum amount of connections to accept.",
           "format": "int32"
+        },
+        "DeleteEndpointOnStartup": {
+          "type": [
+            "boolean",
+            "null"
+          ],
+          "description": "In 'listen' connection mode, deletes the domain socket file used for Diagnostic Port communication.",
```
wiktork, line 911:
```suggestion
          "description": "In 'listen' connection mode, deletes the domain socket file used for diagnostic port communication.",
```

jander-msft (approved):

jander-msft:
This fix mitigates the problems reported in #1827, #2017, #2118

Issue #2152, opened 2022-07-22
Metrics endpoint memory leak


RomanBats:
### Description
In all my environments (different clusters) I see the same pattern of memory leaks in all dotnet-monitor containers:
![image](https://user-images.githubusercontent.com/5024978/180439136-ff9b45d4-b68f-4f48-a6e4-6bc89cd8c71b.png)
The rate of the increasing memory for any dotnet-monitor container is similar.
I use dotnet-monitor only to get Prometheus metrics, other endpoints are not used and were not called.
"/metrics" endpoint is called anonymously by the Prometheus with a 1m scrape interval.

When I tried to figure out what was going on locally in a debug mode, I only noticed a constantly increased number of objects related to JwtBearerOptions:
![image](https://user-images.githubusercontent.com/5024978/180441397-0b7a9e40-12bd-40f1-bf50-127168cbb588.png)

I was taking memory snapshots over and over and the similar increase in the number of objects of the same type was present:
![image](https://user-images.githubusercontent.com/5024978/180439967-442f1e7b-826b-4b73-bdcb-9e9b5e065280.png)
I was querying "/metrics" endpoint once per 10 seconds all the time, but before the 12 and 13 snapshots I started to call this endpoint as often as the API could answer it in a loop. Notice the increase in object number for 12-13 snapshots with a relatively short span of taking those snapshots.

With any call to "/metrics" endpoint, the "LiveJwtBearerHandler" constructor was called. And it seems like OnChange method adds event handler action every request and never GC this object:
![image](https://user-images.githubusercontent.com/5024978/180440778-f00c8405-a8db-43c6-a424-6acb37cdf9e6.png)

### Configuration
Kubernetes: EKS.
Image: 'mcr.microsoft.com/dotnet/monitor:6.2-alpine'
Authentication: enabled
Environment variables:
```yaml
- name: DOTNETMONITOR_DiagnosticPort__ConnectionMode
  value: Connect
- name: DOTNETMONITOR_DiagnosticPort__EndpointName
  value: /diag/port
- name: DOTNETMONITOR_Storage__DumpTempFolder
  value: /dumps
- name: DOTNETMONITOR_Urls
  value: 'http://+:52323'
- name: DOTNETMONITOR_Metrics__MetricCount
  value: '6'
- name: DOTNETMONITOR_GlobalCounter__IntervalSeconds
  value: '10'
- name: DOTNETMONITOR_Metrics__Providers__0__ProviderName
[...]

PR #2165, opened 2022-07-25
Replace redundant LiveJwtBearerHandler with standard JwtBearerHandler

RomanBats:
Closes https://github.com/dotnet/dotnet-monitor/issues/2152.

IOptionsMonitor<JwtBearerOptions> looks to be a singleton, LiveJwtBearerHandler, on the other hand, is a transient service.
The Handler is initialized per each request thus ctor is called and a new OnChange event handler is added to an IOptionsMonitor<JwtBearerOptions>. OnChange event handlers are never garbage collected and are stored in the IOptionMonitor together with captured parameters (a disposed instance of HttpContext). 

And the side effect of this implementation is not only memory leak but, in case the API was running for a couple of days and was constantly called by the Prometheus to scrape metrics, after configurations are changed, all event handlers (registered with OnChange for each Prometheus call) will be invoked (which in my case 1440 times for each day of being live).

I think LiveJwtBearerHandler is redundant since the InitializeAsync method of AuthenticationHandler<TOptions> is called with every request and the latest Options is being obtained:
image

I have tested it by changing MonitorApiKey configurations and with every subsequent request, a different bearer token should be provided for an API call to be authenticated, which I assume is the end goal of LiveJwtBearerHandler.


RomanBats:
@kelltrick

jander-msft:
> I think LiveJwtBearerHandler is redundant since the InitializeAsync method of AuthenticationHandler is called with every request and the latest Options is being obtained

That was my understanding too back when this was introduced and looking through the ASP.NET code that uses this says this should be transient.

jander-msft (approved):

PR #2167, opened 2022-07-26
Stacks event provider

wiktork:
Collect callstacks in ICorProfiler and serialize them to event pipe. Based on https://github.com/dotnet/samples/tree/main/core/profiling.

- Names and relationships for name data is cached in dictionaries.
- The dictionaries are then serialized to event pipe. Formatting will be performed by dotnet-monitor itself. The name cache also provides an implementation for name formatting for diagnostic purposes.
- EventPipe api is wrapped around a variadic template based class, to make data declaration and serialization easier.

TODO:

### Event pipe model:
* Need to map out remaining types, which are not used for the current data.
* We don't have a way of modeling data with no payload.
### Names and name caching:
* Consider making each naming function append to a sstream and consider adding flags to disable certain parts of the name such as the module.
* Currently using fixed max size names and fixed size generic parameters for name discovery. Consider dynamic allocation or TLS buffers to allow more flexibility.
* Use a better hash combine algorithm
### Stack walking:
* According to docs, need to block ThreadDestroyed while stack walking. Is this still a requirement?


src/MonitorProfiler/MainProfiler/MainProfiler.cpp
```diff
@@ -206,5 +209,51 @@ HRESULT MainProfiler::InitializeCommandServer()
...
+
+        IfFailLogRet(stackSampler.CreateCallstack(stackStates, nameCache));
+
+        std::unique_ptr<StacksEventProvider> eventProvider;
+        IfFailLogRet(StacksEventProvider::CreateProvider(m_pCorProfilerInfo, eventProvider));
+
+        for (auto& entry : nameCache->GetFunctions())
+        {
+            tstring name;
+            IfFailLogRet(nameCache->GetFullyQualifiedName(entry.first, name));
```
wiktork, line 231: Note this line is just for testing, I will remove it.

src/MonitorProfiler/Stacks/StacksEventProvider.h
```diff
@@ -0,0 +1,62 @@
...
+
+        //Note we will either send a ClassId or a ClassToken. For Shared generic functions, there is no ClassID.
+        const WCHAR* FunctionPayloads[6] = { _T("FunctionId"), _T("ClassId"), _T("ClassToken"), _T("ModuleId"), _T("Name"), _T("TypeArgs") };
+        std::unique_ptr<ProfilerEvent<UINT64, UINT64, UINT32, UINT64, tstring, std::vector<UINT64>>> _functionEvent;
+
+        //We cannot retrieve detailed information for some ClassIds. Flags is used to indicate these conditions.
+        const WCHAR* ClassPayloads[5] = { _T("ClassId"), _T("ModuleId"), _T("Token"), _T("Flags"), _T("TypeArgs") };
+        std::unique_ptr<ProfilerEvent<UINT64, UINT64, UINT32, UINT32, std::vector<UINT64>>> _classEvent;
+
+        const WCHAR* TokenPayloads[4] = { _T("ModuleId"), _T("Token"), _T("OuterToken"), _T("Name") };
```
wiktork, line 53: Rename to ClassToken

src/MonitorProfiler/EventProvider/ProfilerEvent.h
```diff
@@ -0,0 +1,179 @@
...
+        data[index].reserved = 0;
+    }
+    return WritePayload<index + 1, TArgs...>(data, rest...);
+}
+
+template<typename... Args>
+template<size_t index, typename T, typename... TArgs>
+HRESULT ProfilerEvent<Args...>::WritePayload(COR_PRF_EVENT_DATA* data, const std::vector<typename T::value_type>& first, TArgs... rest)
+{
+    std::vector<BYTE> buffer(0);
```
wiktork, line 131: Comment to make sure this stays in scope

src/MonitorProfiler/EventProvider/ProfilerEvent.h
```diff
@@ -0,0 +1,179 @@
...
+    template<size_t index, typename T, typename... TArgs>
+    HRESULT Initialize(const WCHAR* (&names)[sizeof...(Args)]);
+
+    template<size_t index>
+    HRESULT Initialize(const WCHAR* (&names)[sizeof...(Args)]);
+
+    template<size_t index, typename T, typename... TArgs>
+    HRESULT WritePayload(COR_PRF_EVENT_DATA* data, const T& first, TArgs... rest);
+
+    template<size_t index, typename T = std::wstring, typename... TArgs>
```
wiktork, line 43: tstring

src/MonitorProfiler/EventProvider/EventTypeMapping.h
```diff
@@ -0,0 +1,66 @@
...
+
+template<>
+void EventTypeMapping::GetType<UINT64>(COR_PRF_EVENTPIPE_PARAM_DESC& descriptor)
+{
+    descriptor.type = COR_PRF_EVENTPIPE_UINT64;
+    descriptor.elementType = 0;
+}
+
+template<>
+void EventTypeMapping::GetType<tstring>(COR_PRF_EVENTPIPE_PARAM_DESC & descriptor)
```
kkeirstead, line 55:
```suggestion
void EventTypeMapping::GetType<tstring>(COR_PRF_EVENTPIPE_PARAM_DESC& descriptor)
```

src/MonitorProfiler/EventProvider/ProfilerEvent.h
```diff
@@ -0,0 +1,179 @@
...
+    template<size_t index, typename T, typename... TArgs>
+    HRESULT Initialize(const WCHAR* (&names)[sizeof...(Args)]);
+
+    template<size_t index>
+    HRESULT Initialize(const WCHAR* (&names)[sizeof...(Args)]);
+
+    template<size_t index, typename T, typename... TArgs>
+    HRESULT WritePayload(COR_PRF_EVENT_DATA* data, const T& first, TArgs... rest);
+
+    template<size_t index, typename T = tstring, typename... TArgs>
```
kkeirstead, line 43: I know we talked about why this is written this way yesterday, but would you want to leave a comment since it's not obvious to anyone who reads it later?

src/MonitorProfiler/EventProvider/ProfilerEvent.h
```diff
@@ -0,0 +1,179 @@
...
+     }
+     return WritePayload<index + 1, TArgs...>(data, rest...);
+ }
+
+template<typename... Args>
+template<typename T>
+std::vector<BYTE> ProfilerEvent<Args...>::GetEventBuffer(const std::vector<T>& data)
+{
+    size_t offset = 0;
+    size_t bufferSize = 2 + (data.size() * sizeof(T));
```
kkeirstead, line 154: nit: magic number - I'll leave this to your discretion of whether you feel this warrants addressing.

src/MonitorProfiler/EventProvider/ProfilerEvent.h
```diff
@@ -0,0 +1,179 @@
...
+    {
+        WriteToBuffer<T>(buffer.data(), bufferSize, &offset, element);
+    }
+
+    return buffer;
+}
+
+template<typename... Args>
+template<typename T>
+void ProfilerEvent<Args...>::WriteToBuffer(BYTE* pBuffer, size_t bufferLength, size_t* pOffset, const T& value)
```
kkeirstead, line 168: It doesn't appear that we're using `bufferLength`; should we be? Or can this be removed from the signature?

src/MonitorProfiler/Stacks/StackSampler.h
```diff
@@ -0,0 +1,43 @@
...
+};
+
+class StackSampler
+{
+    public:
+        StackSampler(ICorProfilerInfo12* profilerInfo);
+        HRESULT CreateCallstack(std::vector<std::unique_ptr<StackSamplerState>>& stackStates, std::shared_ptr<NameCache>& nameCache);
+        static void AddProfilerEventMask(DWORD& eventsLow);
+    private:
+        static HRESULT __stdcall DoStackSnapshotStackSnapShotCallbackWrapper(
```
kkeirstead, line 34: I guess I'm not sure if this was intentional...is this name supposed to have `StackSnapshot` in it twice, or did you accidentally type it twice? Also, not that it _really_ matters, but technically it's currently `Snapshot` the first time, and `SnapShot` the second time (if we really do want it twice)

src/MonitorProfiler/Stacks/StackSampler.h
```diff
@@ -0,0 +1,43 @@
...
+            FunctionID funcId,
```
kkeirstead, line 35: nit: it seems like in most other places we've been using `functionId` instead of `funcId`, if we care about keeping it consistent.

src/MonitorProfiler/Utilities/NameCache.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+#include "NameCache.h"
+#include "macros.h"
+#include "corhlpr.h"
+
+const tstring NameCache::CompositeClassName = _T("_CompositeClass_");
+const tstring NameCache::ArrayClassName = _T("_ArrayClass_");
+const tstring NameCache::UnknownName = _T("_Unknown_");
+const tstring NameCache::ModuleSeparator = _T("!");
+const tstring NameCache::FunctionSeperator = _T(".");
```
kkeirstead, line 13: nit: `FunctionSeparator`

src/MonitorProfiler/Utilities/NameCache.cpp
```diff
@@ -0,0 +1,223 @@
...
+{
+    _moduleNames.emplace(moduleId, std::make_shared<ModuleData>(std::move(name)));
+}
+
+HRESULT NameCache::GetFullyQualifiedName(FunctionID id, tstring& name)
+{
+    HRESULT hr;
+
+    std::shared_ptr<FunctionData> functionData;
+    if (GetFunctionData(id, functionData))
```
kkeirstead, line 53: Presumably, if this evaluates to false, then we skip all of the logic in this method and don't construct `name`. Do we still want to return `S_OK` in this case, or is this considered a failure?

src/MonitorProfiler/Utilities/NameCache.cpp
```diff
@@ -0,0 +1,223 @@
...
+            else
+            {
+                name += GenericSeparator;
+            }
+        }
+
+        std::shared_ptr<ModuleData> moduleData;
+        if (GetModuleData(functionData->GetModuleId(), moduleData))
+        {
+            name = moduleData->GetName() + ModuleSeparator + name;
```
kkeirstead, line 91: Just out of curiosity, why do we do this last? It seems like everything else builds by appending the items in order, so I wasn't sure why this didn't follow that pattern by being checked first.

src/MonitorProfiler/Utilities/NameCache.cpp
```diff
@@ -0,0 +1,223 @@
...
+HRESULT NameCache::GetFullyQualifiedClassName(ClassID classId, tstring& name)
+{
+    HRESULT hr;
+
+    std::shared_ptr<ClassData> classData;
+    if (classId != 0 && GetClassData(classId, classData))
+    {
+        mdTypeDef token = classData->GetToken();
+
+        if (classData->GetFlags() == ClassFlags::None)
```
kkeirstead, line 107: Why is the `None` check kept separate from the switch? Is there a reason why this can't/shouldn't be a case in the switch?

src/MonitorProfiler/Utilities/NameCache.cpp
```diff
@@ -0,0 +1,223 @@
...
+                    break;
+                case ClassFlags::IncompleteData:
+                case ClassFlags::Error:
+                default:
+                    name = UnknownName;
+                    break;
+            }
+        }
+
+        for (int i = 0; i < classData->GetTypeArgs().size(); i++)
```
kkeirstead, line 129: It looks like this is a nearly identical copy-paste from the prior method; just from a code cleanliness perspective, is there any way this can be made more generic so we can reuse the `name` construction?

src/MonitorProfiler/Utilities/NameCache.h
```diff
@@ -0,0 +1,73 @@
...
+    const std::unordered_map<FunctionID, std::shared_ptr<FunctionData>>& GetFunctions();
+    const std::unordered_map<ModuleID, std::shared_ptr<ModuleData>>& GetModules();
+    const std::unordered_map<std::pair<ModuleID, mdTypeDef>, std::shared_ptr<TokenData>, PairHash<ModuleID, mdTypeDef>>& GetTypeNames();
+
+private:
+    static const tstring CompositeClassName;
+    static const tstring ArrayClassName;
+    static const tstring UnknownName;
+    static const tstring ModuleSeparator;
+    static const tstring FunctionSeperator;
```
kkeirstead, line 47: nit: `FunctionSeparator`

src/MonitorProfiler/Utilities/TypeNameUtilities.cpp
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    if (classId == 0)
+    {
+        return E_INVALIDARG;
+    }
+
+    ModuleID modId = 0;
+    mdTypeDef classToken = mdTokenNil;
+    ULONG32 nTypeArgs = 0;
```
kkeirstead, line 106: nit: Above, I believe we do the same type of thing, but we name it `typeArgsCount`. Assuming they're the same concept, let's make the names consistent for readability.

src/MonitorProfiler/Utilities/TypeNameUtilities.cpp
```diff
@@ -0,0 +1,249 @@
...
+    IfFailRet(_profilerInfo->GetModuleInfo(moduleId,
+        nullptr,
+        256,
+        &nameLength,
+        moduleFullName,
+        &assemID));
+
+    WCHAR* ptr = nullptr;
+
+    int index = nameLength - 1;
```
kkeirstead, line 226: nit: Given that we're evaluating it down below (as opposed to solely using it for the purposes of iteration), can we give this a more descriptive name?

src/MonitorProfiler/Stacks/StackSampler.h
wiktork: This is just me doing a bad job of copy pasting David Mason's profiler samples.

src/MonitorProfiler/Utilities/NameCache.cpp
wiktork: This was a side effect of two things:
- I kept changing whether the module id should be added in the Function or the Class api
- I would ultimately like some configurability around this. I.e. making the module name optional.

I can change the order for now.

src/MonitorProfiler/Utilities/NameCache.h
```diff
@@ -0,0 +1,73 @@
...
+#include <unordered_map>
+#include <vector>
+
+/// <summary>
+/// Stores mappings between Clr objects and their names.
+/// </summary>
+class NameCache
+{
+public:
+    bool GetFunctionData(FunctionID id, std::shared_ptr<FunctionData>& data);
```
jander-msft, line 23: nit: All of these should be `TryGet...`

src/MonitorProfiler/Utilities/NameCache.cpp
wiktork: There was another reason here which is my functions are a bit inconsistent; some append data others replace it. I will leave this in place for now and will add some TODO's around cleanup.

src/MonitorProfiler/Communication/CommandServer.cpp
```diff
@@ -91,10 +94,18 @@ void CommandServer::ListeningThread()
 
 void CommandServer::ClientProcessingThread()
```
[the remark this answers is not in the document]
wiktork: This method loops and processes each message. I don't really want to return if I fail handling any particular method but still want to log the failure.

src/MonitorProfiler/Communication/CommandServer.h
```diff
@@ -6,6 +6,9 @@
 
 #include "IpcCommServer.h"
 #include "Messages.h"
+#include "cor.h"
```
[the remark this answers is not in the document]
wiktork: I need corproh for ICorProfiler12, and oddly enough corprof does not include cor.h, so it fails if I remove the header.

src/MonitorProfiler/EventProvider/ProfilerEventProvider.cpp
```diff
@@ -0,0 +1,21 @@
...
+#include "ProfilerEventProvider.h"
+#include "corhlpr.h"
+
+HRESULT ProfilerEventProvider::CreateProvider(const WCHAR* providerName, ICorProfilerInfo12* profilerInfo, std::unique_ptr<ProfilerEventProvider>& provider)
+{
+    EVENTPIPE_PROVIDER eventProvider = 0;
+    HRESULT hr;
+
+    IfFailRet(profilerInfo->EventPipeCreateProvider(providerName, &eventProvider));
+    provider = std::unique_ptr<ProfilerEventProvider>(new ProfilerEventProvider(profilerInfo, eventProvider));
```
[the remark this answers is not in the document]
wiktork: I viewed the creation/assignment of the pointer itself as trivial, but I can certainly change it.

jander-msft (approved):

src/MonitorProfiler/Communication/CommandServer.cpp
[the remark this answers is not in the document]
wiktork: Per our discussion, leaving as is for now. If this gets larger, we can refactor.

jander-msft (approved):

PR #2194, opened 2022-07-27
Use the version of dotnet format shipped with the sdk

schmittjoseph:
Since merging the linter integration, the version of `dotnet format` pinned is no longer available. 

Instead of pinning a specific version, just use the version included in the .NET SDK we use to build.


jander-msft (approved):

jander-msft (dismissed):

.github/workflows/lint-csharp.yml
```diff
@@ -38,11 +38,12 @@ jobs:
           dotnet-version: ${{ env.dotnet_version }}
           include-prerelease: true
 
-      - name: Restore dotnet tools
-        run: dotnet tool restore
+      # Workaround for a bug in the dotnet-format build shipped in .NET 7.0 Preview 5. Ref: https://github.com/dotnet/core/blob/main/release-notes/7.0/known-issues.md#unhandled-exception-in-dotnet-format-app-in-net-70-preview-5
+      - name: Setup dotnet format
+        run: dotnet tool install -g dotnet-format --version "7.*" --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet7/nuget/v3/index.json
```
jander-msft, line 43: Rather than explicitly using a source here, I think it would be better to use nuget.config at the root of the repo with the `--configfile` parameter.

jander-msft (approved):

PR #2233, opened 2022-08-02
Remove .NET 5 testing and consolidate TFMs

jander-msft:
These changes:
- remove .NET 5 from the testing matrix (just as #2086 did for `main`, with an additional change to not install .NET 5)
- consolidate the TFM specification in the cross-targeting projects (just as #1352 did for `main`, but only an applicable subset)

The result is that .NET 5 is no longer tested (it is out of support), build times should be incrementally faster, and full test runs should be much faster.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -13,6 +13,12 @@
     -->
     <BlobGroupBuildQuality>daily</BlobGroupBuildQuality>
   </PropertyGroup>
+  <PropertyGroup Label="TargetFrameworks">
+    <!-- The TFMs of the dotnet-monitor tool.  -->
+    <ToolTargetFrameworks>net6.0</ToolTargetFrameworks>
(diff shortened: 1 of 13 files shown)


eng/Versions.props
```diff
@@ -13,6 +13,12 @@
     -->
     <BlobGroupBuildQuality>daily</BlobGroupBuildQuality>
   </PropertyGroup>
+  <PropertyGroup Label="TargetFrameworks">
+    <!-- The TFMs of the dotnet-monitor tool.  -->
+    <ToolTargetFrameworks>net6.0</ToolTargetFrameworks>
```
jander-msft, line 18:
```suggestion
    <ToolTargetFrameworks>netcoreapp3.1;net6.0</ToolTargetFrameworks>
```

PR #2290, opened 2022-08-10
Additional Metadata For Azure Egress

### LINKED ISSUE #926: Add additional metadata during Azure Blob egress
When egressing an artifact, additional metadata should be added to the artifact blob, such as
- the host name (which is the name of the pod when running in Kubernetes)
- the container name (if there is access to this information somehow)

When egressing an artifact from a collection rule, even more metadata should be added to the artifact blob:
- The name of the collection rule that egressed it.
- The index of the action that collected the artifact.
- The name of the action that collected the artifact. Action names are optional, so this may be empty.

kkeirstead:
This PR adds metadata for artifacts that come from collection rules (collection rule name, action name, and action list index). It also adds in an option to include the host name in the metadata (this is configured to be off by-default), as well as the ability to map metadata keys to environment variable names to load custom metadata.

closes #926

### DIFF
--- a/documentation/schema.json
+++ b/documentation/schema.json
@@ -1047,6 +1047,21 @@
           ],
           "description": "The URI of the Azure queue account.",
           "format": "uri"
+        },
+        "IncludeHostNameAsMetadata": {
+          "type": "boolean",
+          "description": "Determines if the host name should be included as blob metadata.",
+          "default": false
+        },
+        "Metadata": {
+          "type": [
+            "null",
+            "object"
+          ],
+          "description": "A mapping of metadata keys to environment variable names.",
+          "additionalProperties": {
+            "type": "string"
+          }
         }
       }
     },
--- a/documentation/schema.json
+++ b/documentation/schema.json
@@ -1058,7 +1058,7 @@
             "null",
             "object"
           ],
-          "description": "A mapping of metadata keys to environment variable names.",
+          "description": "A mapping of metadata keys to environment variable names. The values of the environment variables will be added as metadata for egressed artifacts.",
           "additionalProperties": {
             "type": "string"
           }
--- a/src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
+++ b/src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
@@ -705,7 +705,7 @@
     <comment>The description provided for the IncludeHostNameAsMetadata parameter on AzureBlobEgressProviderOptions.</comment>
   </data>
   <data name="DisplayAttributeDescription_AzureBlobEgressProviderOptions_Metadata" xml:space="preserve">
-    <value>A mapping of metadata keys to environment variable names.</value>
+    <value>A mapping of metadata keys to environment variable names. The values of the environment variables will be added as metadata for egressed artifacts.</value>
     <comment>The description provided for the Metadata parameter on AzureBlobEgressProviderOptions</comment>
   </data>
 </root>
--- a/src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
+++ b/src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
@@ -700,4 +700,8 @@
   <data name="DisplayAttributeDescription_DiagnosticPortOptions_DeleteEndpointOnStartup" xml:space="preserve">
     <value>In 'listen' connection mode, deletes the domain socket file used for diagnostic port communication.</value>
   </data>
+  <data name="DisplayAttributeDescription_AzureBlobEgressProviderOptions_IncludeHostNameAsMetadata" xml:space="preserve">
+    <value>Determines if the host name should be included as blob metadata.</value>
+    <comment>The description provided for the IncludeHostNameAsMetadata parameter on AzureBlobEgressProviderOptions.</comment>
+  </data>
 </root>
--- a/src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
+++ b/src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
@@ -704,4 +704,8 @@
     <value>Determines if the host name should be included as blob metadata.</value>
     <comment>The description provided for the IncludeHostNameAsMetadata parameter on AzureBlobEgressProviderOptions.</comment>
   </data>
+  <data name="DisplayAttributeDescription_AzureBlobEgressProviderOptions_Metadata" xml:space="preserve">
+    <value>A mapping of metadata keys to environment variable names.</value>
+    <comment>The description provided for the Metadata parameter on AzureBlobEgressProviderOptions</comment>
+  </data>
 </root>
--- a/src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
+++ b/src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
@@ -11,8 +11,10 @@ using Azure.Storage.Blobs.Specialized;
 using Azure.Storage.Queues;
 using Microsoft.Extensions.Logging;
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
@@ -122,8 +124,23 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Egress.AzureBlob
                 // Write blob headers
                 await blobClient.SetHttpHeadersAsync(CreateHttpHeaders(artifactSettings), cancellationToken: token);
 
-                // Write blob metadata
-                await blobClient.SetMetadataAsync(artifactSettings.Metadata, cancellationToken: token);
+                try
+                {
+                    Dictionary<string, string> mergedMetadata = artifactSettings.Metadata.Union(artifactSettings.CustomMetadata).ToDictionary(pair => pair.Key, pair => pair.Value);
+
+                    // Write blob metadata
+                    await blobClient.SetMetadataAsync(mergedMetadata, cancellationToken: token);
+                }
+                catch (Exception ex) when (ex is InvalidOperationException || ex is RequestFailedException)
+                {
+                    Logger.InvalidMetadata(ex);
+                    await blobClient.SetMetadataAsync(artifactSettings.Metadata, cancellationToken: token);
+                }
+                catch (ArgumentException ex)
+                {
+                    Logger.DuplicateKeyInMetadata(ex);
+                    await blobClient.SetMetadataAsync(artifactSettings.Metadata, cancellationToken: token);
+                }
 
                 string blobUriString = GetBlobUri(blobClient);
                 Logger?.EgressProviderSavedStream(EgressProviderTypes.AzureBlobStorage, blobUriString);
@@ -158,7 +175,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Egress.AzureBlob
 
             foreach (var pair in options.Metadata)
             {
-                artifactSettings.Metadata.Add(pair.Key, Environment.GetEnvironmentVariable(pair.Value));
+                artifactSettings.CustomMetadata.Add(pair.Key, Environment.GetEnvironmentVariable(pair.Value));
             }
         }
 
--- a/src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
+++ b/src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
@@ -30,20 +30,22 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Egress.AzureBlob
     {
         private int BlobStorageBufferSize = 4 * 1024 * 1024;
 
-        public AzureBlobEgressProvider(ILogger<AzureBlobEgressProvider> logger)
-            : base(logger)
+        public AzureBlobEgressProvider(IServiceProvider serviceProvider, ILogger<AzureBlobEgressProvider> logger)
+        : base(logger)
         {
         }
 
         public override async Task<string> EgressAsync(
-            AzureBlobEgressProviderOptions options,
+        AzureBlobEgressProviderOptions options,
             Func<CancellationToken, Task<Stream>> action,
             EgressArtifactSettings artifactSettings,
             CancellationToken token)
         {
             try
             {
-                string hostName = Dns.GetHostName();
+                Environment.SetEnvironmentVariable("TEST_ENV_VAR", "container name goes here");
+
+                AddConfiguredMetadata(options, artifactSettings);
 
                 var containerClient = await GetBlobContainerClientAsync(options, token);
 
@@ -89,7 +91,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Egress.AzureBlob
         {
             try
             {
-                string hostName = Dns.GetHostName();
+                AddConfiguredMetadata(options, artifactSettings);
 
                 var containerClient = await GetBlobContainerClientAsync(options, token);
 
@@ -145,6 +147,19 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Egress.AzureBlob
             catch (CredentialUnavailableException ex)
             {
                 throw CreateException(ex);
+            }
+        }
+
+        public void AddConfiguredMetadata(AzureBlobEgressProviderOptions options, EgressArtifactSettings artifactSettings)
+        {
+            if (options.IncludeHostNameAsMetadata)
+            {
+                artifactSettings.Metadata.Add("HostName", Dns.GetHostName()); // Make this a constant
+            }
+
+            foreach (var pair in options.Metadata)
+            {
+                artifactSettings.Metadata.Add(pair.Key, Environment.GetEnvironmentVariable(pair.Value));
             }
         }
 
--- a/src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
+++ b/src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
@@ -34,12 +34,12 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Egress.AzureBlob
         private readonly string HostNameMetadataKey = "HostName";
 
         public AzureBlobEgressProvider(ILogger<AzureBlobEgressProvider> logger)
-        : base(logger)
+            : base(logger)
         {
         }
 
         public override async Task<string> EgressAsync(
-        AzureBlobEgressProviderOptions options,
+            AzureBlobEgressProviderOptions options,
             Func<CancellationToken, Task<Stream>> action,
             EgressArtifactSettings artifactSettings,
             CancellationToken token)
@@ -173,9 +173,9 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Egress.AzureBlob
                 artifactSettings.Metadata.Add(HostNameMetadataKey, Dns.GetHostName());
             }
 
-            foreach (var pair in options.Metadata)
+            foreach (var metadataPair in options.Metadata)
             {
-                artifactSettings.CustomMetadata.Add(pair.Key, Environment.GetEnvironmentVariable(pair.Value));
+                artifactSettings.CustomMetadata.Add(metadataPair.Key, Environment.GetEnvironmentVariable(metadataPair.Value));
             }
         }
 
--- a/src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
+++ b/src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
@@ -13,6 +13,7 @@ using Microsoft.Extensions.Logging;
 using System;
 using System.Globalization;
 using System.IO;
+using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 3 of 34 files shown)


src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
```diff
@@ -700,4 +700,12 @@
   <data name="DisplayAttributeDescription_DiagnosticPortOptions_DeleteEndpointOnStartup" xml:space="preserve">
     <value>In 'listen' connection mode, deletes the domain socket file used for diagnostic port communication.</value>
   </data>
+  <data name="DisplayAttributeDescription_AzureBlobEgressProviderOptions_IncludeHostNameAsMetadata" xml:space="preserve">
```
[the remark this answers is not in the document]
kkeirstead: Given that the method we call is `Dns.GetHostName`, should this actually be one word? That was what I based it on

src/Tools/dotnet-monitor/LoggingEventIds.cs
```diff
@@ -79,6 +79,8 @@ internal enum LoggingEventIds
         DiagnosticPortDeleteFailed = 66,
         DiagnosticPortAlteredWhileInUse = 67,
         DiagnosticPortWatchingFailed = 68
```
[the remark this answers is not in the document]
kkeirstead: Yep, I had a merge conflict that I tried to resolve in GitHub and missed that (hence the build failures) - I'll check it in with the next commit

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -116,8 +124,23 @@ public AzureBlobEgressProvider(ILogger<AzureBlobEgressProvider> logger)
                 // Write blob headers
                 await blobClient.SetHttpHeadersAsync(CreateHttpHeaders(artifactSettings), cancellationToken: token);
 
-                // Write blob metadata
-                await blobClient.SetMetadataAsync(artifactSettings.Metadata, cancellationToken: token);
+                try
+                {
+                    Dictionary<string, string> mergedMetadata = artifactSettings.Metadata.Union(artifactSettings.CustomMetadata).ToDictionary(pair => pair.Key, pair => pair.Value);
```
jander-msft, line 130: This will throw if two keys are the same. Consider either manually verifying that all of the keys are unique and then throw a specific exception if not or allowing custom metadata to either override the precalculated metadata or is dropped. For the latter cases, we should probably log warnings.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -116,8 +124,23 @@ public AzureBlobEgressProvider(ILogger<AzureBlobEgressProvider> logger)
...
+
+                    // Write blob metadata
+                    await blobClient.SetMetadataAsync(mergedMetadata, cancellationToken: token);
+                }
+                catch (Exception ex) when (ex is InvalidOperationException || ex is RequestFailedException)
+                {
+                    Logger.InvalidMetadata(ex);
+                    await blobClient.SetMetadataAsync(artifactSettings.Metadata, cancellationToken: token);
+                }
+                catch (ArgumentException ex)
```
jander-msft, line 140: Let's not handling this exception here because it masks when other parts of the code throw ArgumentException.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -116,8 +124,23 @@ public AzureBlobEgressProvider(ILogger<AzureBlobEgressProvider> logger)
...
-                // Write blob metadata
-                await blobClient.SetMetadataAsync(artifactSettings.Metadata, cancellationToken: token);
+                try
+                {
+                    Dictionary<string, string> mergedMetadata = artifactSettings.Metadata.Union(artifactSettings.CustomMetadata).ToDictionary(pair => pair.Key, pair => pair.Value);
+
+                    // Write blob metadata
+                    await blobClient.SetMetadataAsync(mergedMetadata, cancellationToken: token);
+                }
+                catch (Exception ex) when (ex is InvalidOperationException || ex is RequestFailedException)
```
jander-msft, line 135: If you take any of my suggestions above, then remove this catcher.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -143,6 +166,19 @@ public AzureBlobEgressProvider(ILogger<AzureBlobEgressProvider> logger)
             }
         }
 
+        public void AddConfiguredMetadata(AzureBlobEgressProviderOptions options, EgressArtifactSettings artifactSettings)
+        {
+            if (options.IncludeHostNameAsMetadata)
+            {
+                artifactSettings.Metadata.Add(HostNameMetadataKey, Dns.GetHostName());
```
jander-msft: I'm wary of using the current hostname of dotnet-monitor and reporting that as the hostname of the target process. You should probably get this from the environment block of the target process; the variable name is HOSTNAME if I recall correctly.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -143,6 +166,19 @@ public AzureBlobEgressProvider(ILogger<AzureBlobEgressProvider> logger)
...
+            }
+
+            foreach (var metadataPair in options.Metadata)
+            {
+                artifactSettings.CustomMetadata.Add(metadataPair.Key, Environment.GetEnvironmentVariable(metadataPair.Value));
```
jander-msft, line 186: This will throw if the custom metadata contains a key that matches one of the built in ones e.g. hostname.

src/Tools/dotnet-monitor/Egress/EgressArtifactSettings.cs
```diff
@@ -21,6 +21,12 @@ internal sealed class EgressArtifactSettings
         public Dictionary<string, string> Metadata { get; }
             = new Dictionary<string, string>(StringComparer.Ordinal);
 
+        /// <summary>
+        /// Custom metadata of the blob to be created.
+        /// </summary>
+        public Dictionary<string, string> CustomMetadata { get; }
+            = new Dictionary<string, string>(StringComparer.Ordinal);
```
jander-msft, line 28: On Linux, environment variable names are case sensitive, but on Windows and MacOS they are not. So `HOSTNAME` and `hostname` are the same variable on Windows and MacOS. We might need to merge case-insensitive matching variables on Windows and MacOS. But the fact that you are using `StringComparer.Ordinal` here I think is correct.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
jander-msft, line 186: This should be getting the environment variable values from the target process, not from the dotnet-monitor process. Use [DiagnosticsClient.GetProcessEnvironmentAsync](https://github.com/dotnet/diagnostics/blob/main/src/Microsoft.Diagnostics.NETCore.Client/DiagnosticsClient/DiagnosticsClient.cs#L290). Be mindful that this could be an expensive call, so you'll want to probably make one call to this and use the result throughout the egress invocation.

jander-msft (changes requested):

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
kkeirstead: Are you sure? I intentionally am adding to `CustomMetadata` instead of `Metadata`, which I created mostly for this purpose.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
kkeirstead: Is the main reason for doing this so that we don't have to catch the `ArgumentException`? I had considered doing it this way but thought it might be slightly tedious, especially for something that probably will happen infrequently. That being said, I can do it this way, which would also allow us to keep custom metadata that _doesn't_ conflict (whereas right now, all custom metadata gets scrapped).

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
kkeirstead: Why are we removing this catcher? I was using this to handle invalid names (e.g. `&MyKey`), which is a separate issue from having duplicate keys

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
jander-msft: If someone adds `"HostName": "..."` as an entry in the `Metadata` property in configuration, then this is going to throw, correct?

src/Tools/dotnet-monitor/Egress/EgressArtifactSettings.cs
jander-msft: Ah, yes, you are correct. I'm reading this backwards. The mapping is from metadata name to environment variable name, whereas I was reading as from environment variable name to metadata name. So no issue here.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
jander-msft: In general, I try to do best-effort (unless it absolutely makes sense to abort an entire operation if a subset doesn't quite work).

We can easily determine if keys collide or not, and would make it much easier to understand from a user perspective what collision happened.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
kkeirstead: It shouldn't - the throw will happen around L130 when we try to merge `CustomMetadata` and `Metadata`, hence catching `ArgumentException` in the current implementation.

jander-msft: Sorry, another wrong reading on my part. You are adding the hostname metadata to `artifactSettings.Metadata` whereas the line is adding each `options.Metadata` to `artifactSettings.CustomMetadata`.

Is there a reason you are not adding the custom metadata directly to `artifactSettings.Metadata` here, which then alleviates the need for the rest of the code to understand that there is custom metadata?

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
kkeirstead: So, I am already doing this - in the catch statements, I add in _just_ the built-in metadata, since the issue is with the custom metadata. I originally tried to apply them both separately, but I think that was overwriting whatever I applied first with what I applied second (as opposed to merging them - I could be mistaken though). That being said, I can adjust the try/catch so that it only is wrapped around `SetMetadataAsync`

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
kkeirstead: That was what I did originally, but then if there was an issue with some of the custom metadata, I couldn't add any of the metadata (since the good built-in metadata was intermingled with the contaminated custom metadata). Keeping them separate allows us to add the built-in metadata if the merged metadata fails. However, if we actually go validate each item individually (which would also require manually checking if the custom metadata names are legal), we could just put everything in one big dictionary and yank out anything that doesn't belong in there.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -143,6 +176,37 @@ public AzureBlobEgressProvider(ILogger<AzureBlobEgressProvider> logger)
...
 
+        public async Task AddConfiguredMetadataAsync(AzureBlobEgressProviderOptions options, EgressArtifactSettings artifactSettings, IEndpointInfo endpointInfo, CancellationToken token)
+        {
+            DiagnosticsClient client = new DiagnosticsClient(endpointInfo.Endpoint);
+
+            Dictionary<string, string> envBlock = await client.GetProcessEnvironmentAsync(token);
+
+            if (options.IncludeHostNameAsMetadata)
+            {
+                if (envBlock.TryGetValue(HostNameMetadataKey, out string hostNameValue))
```
kkeirstead, line 187: I wasn't sure where we landed on this, so I put this in for now, checking "HOSTNAME" and then "CONTAINERNAME" instead of using `GetHostName`. I haven't pulled out the socket exception logger message yet, but if we decide we want to go this route (or something similar) I can pull that out.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -143,6 +176,37 @@ public AzureBlobEgressProvider(ILogger<AzureBlobEgressProvider> logger)
...
+        {
+            DiagnosticsClient client = new DiagnosticsClient(endpointInfo.Endpoint);
+
+            Dictionary<string, string> envBlock = await client.GetProcessEnvironmentAsync(token);
+
+            if (options.IncludeHostNameAsMetadata)
+            {
+                if (ProcessInfoImpl.IsWindowsProcess(endpointInfo) && envBlock.TryGetValue(ComputerNameMetadataKey, out string computerNameValue))
+                {
+                    artifactSettings.Metadata.Add(ComputerNameMetadataKey, computerNameValue);
```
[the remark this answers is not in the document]
kkeirstead: > If we go this route, I'd recommend checking the target ProcessInfo's operating system value to determine which environment variable to grab. `HOSTNAME` has no significance on Windows, so it'd be a valid scenario for users to assign it to something arbitrary and then we'd be grabbing the wrong value.

Looking at your prior comment, wouldn't doing this go against the idea that a Windows user could choose to set an environment variable named `HOSTNAME`? Unless we want to effectively reserve that term, even if it's just arbitrary for the sake of having a consistent naming convention.

src/Microsoft.Diagnostics.Monitoring.WebApi/ProcessInfoImpl.cs
```diff
@@ -144,6 +131,23 @@ public static async Task<IProcessInfo> FromEndpointInfoAsync(IEndpointInfo endpo
                 processName);
         }
 
+        public static bool IsWindowsProcess(IEndpointInfo endpointInfo)
+        {
+            if (string.IsNullOrEmpty(endpointInfo.OperatingSystem))
+            {
+                // If operating system is null, the process is likely .NET Core 3.1 (which doesn't have the GetProcessInfo command).
```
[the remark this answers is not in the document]
kkeirstead: I was also thinking the same thing, but decided to keep the existing logic since I don't have the full context of whether we need to use this check or not.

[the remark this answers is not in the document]
jander-msft: At this time, it's safe to assume `dotnet-monitor` does run on the same platform as the target process. The primary scenarios all require that they are running on the same machine.

The diagnostic communication channel is either a named pipe or a unix domain socket. UDS are inter-process communication within the same host machine. A Windows named pipe does technically allow for communication over a network, but that is not something that we've explicitly enabled (running `dotnet-monitor` on a machine different that the target process).

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -116,8 +127,30 @@ public AzureBlobEgressProvider(ILogger<AzureBlobEgressProvider> logger)
...
+
+                foreach (var metadataPair in artifactSettings.CustomMetadata)
+                {
+                    if (!mergedMetadata.ContainsKey(metadataPair.Key))
+                    {
+                        mergedMetadata[metadataPair.Key] = metadataPair.Value;
+                    }
+                    else
+                    {
+                        Logger.DuplicateKeyInMetadata(metadataPair.Key);
```
kkeirstead, line 140: Note for a future PR: planning to add in a prefix for our metadata (e.g. `DotnetMonitor_`), and then telling users to avoid setting environment variables to use that prefix. This will allow us to add more metadata without worrying about trampling existing user-provided metadata.

documentation/schema.json
```diff
@@ -1047,6 +1047,16 @@
           ],
           "description": "The URI of the Azure queue account.",
           "format": "uri"
+        },
+        "Metadata": {
+          "type": [
+            "null",
+            "object"
+          ],
+          "description": "A mapping of metadata keys to environment variable names. The values of the environment variables will be added as metadata for egressed artifacts.",
```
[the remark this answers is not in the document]
kkeirstead: Did you have it fail on you? I currently have a try/catch wrapped around where we set the set the metadata with a warning message when users try to add invalid metadata, and I just tried this and the artifact egressed without any issues (and I got a warning telling me that my metadata key needs to be a valid C# identifier). Unless I need to try a different illegal key?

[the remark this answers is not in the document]
kkeirstead: So, that isn't the behavior I'm seeing, but I think I might know what the confusion is. In the second EgressAsync, we're currently attempting to add the merged metadata, and if we fail, we add only the built-in metadata. However, I overlooked the first EgressAsync, so while that wouldn't fail, it actually wasn't including the custom metadata. The latest changes should now attempt to push the merged metadata (includes the custom) for both of them, and then fall back on the existing metadata if there's an invalid key.

src/Tools/dotnet-monitor/Egress/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -143,6 +153,53 @@ public AzureBlobEgressProvider(ILogger<AzureBlobEgressProvider> logger)
...
+                Logger.InvalidMetadata(ex);
+                await blobClient.SetMetadataAsync(artifactSettings.Metadata, cancellationToken: token);
+            }
+        }
+
+        public async Task AddConfiguredMetadataAsync(AzureBlobEgressProviderOptions options, EgressArtifactSettings artifactSettings, IEndpointInfo endpointInfo, CancellationToken token)
+        {
+            DiagnosticsClient client = new DiagnosticsClient(endpointInfo.Endpoint);
+
+            Dictionary<string, string> envBlock = await client.GetProcessEnvironmentAsync(token);
```
jander-msft, line 188: Rather than having this specific egress provider do this, could the egress service do this and fill in the CustomMetadata? This would eliminate the need to pass the IEndpointInfo into the egress provider, which is had no knowledge of previously.

kkeirstead: So, I can do that, although I'm not sure that's the correct level for it - we don't currently have any notion of which options type we're using at that level, and then once we get the options type we'd only add custom metadata if the type is Azure (since this doesn't apply to FileSystem). The way it's currently written, `EgressService` is kept generic and metadata is added regardless of the provider, but that would no longer be the case (unless we added a dummy field for FileSystem for custom metadata that is always empty, but that feels a bit off).

jander-msft: I was thinking more in terms of the egress provider would have a method/property on it that says "I'm interested in the environment block of the target process", then the egress service would populate the environment block on the EgressArtifactsSettings, and finally pass in the settings; the provider can do whatever it wants with the block, such as mapping to individual metadata. This gets the provider out of the business of interacting with the target process, which is going to be the case for egress extensibility.

kkeirstead: The latest change should follow what I think you're describing here - I didn't end up having the egress provider specifically say it wanted the environment block filled, so as to follow the same pattern as filling metadata (which we do regardless of egress provider), so let me know if you want me to adjust this or anything else.

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -570,6 +574,10 @@
   <data name="LogFormatString_EndpointTimeout" xml:space="preserve">
     <value>Unexpected timeout from process {processId}. Process will no longer be monitored.</value>
   </data>
+  <data name="LogFormatString_EnvironmentVariableNotFound" xml:space="preserve">
+    <value>The environment variable '{0}' could not be found on the target process.</value>
```
jander-msft, line 578: This should use a name instead of an index, such as `name`.

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -528,6 +528,10 @@
 2. providerType: The type of provider that the new provider was attempting to be registered.
 3. existingProviderType: The type of provider that the new provider name was already registered.</comment>
   </data>
+  <data name="LogFormatString_DuplicateKeyInMetadata" xml:space="preserve">
+    <value>Metadata cannot include duplicate keys; please change or remove the key '{0}'</value>
```
jander-msft, line 532: This should use a name instead of an index, such as `key`.

jander-msft (approved):

PR #2291, opened 2022-08-10
Disallow synchronous IO on output streams by default

### LINKED ISSUE #1384: Use async IO by default and enable sync IO as needed per endpoint
Related #83

schmittjoseph:
This PR turns off globally allowing synchronous IO on the output streams, updates the `/metrics` endpoint to use async IO, and updates `/logs` and `/livemetrics` to allow sync io (see #83)

Closes #1384

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.Metrics.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.Metrics.cs
@@ -3,7 +3,6 @@
 // See the LICENSE file in the project root for more information.
 
 using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Diagnostics.Monitoring.EventPipe;
 using System;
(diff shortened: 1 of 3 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/LogsUtilities.cs
```diff
@@ -53,6 +56,7 @@ public static async Task CaptureLogsAsync(TaskCompletionSource<object> startComp
             }
 
             await runTask;
+            await bufferingStream.DrainBufferAsync(outputStream, token);
```
jander-msft, line 59: This changes the existing behavior that logs/metrics are written to the response stream as they are generated to buffering everything and flushing it all at once at the end. This could prevent scenarios such as infinite logging/metrics sessions to monitor for a specific log message of metrics pattern.

Is there some way we can have a stream or pipeline receive from the StreamingLoggerProvider, buffer to disk when it reaches some threshold, but allow simultaneous reading out to write to the response stream?

jander-msft: Maybe something like: https://docs.microsoft.com/en-us/dotnet/standard/io/pipelines#pipe-basic-usage

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.Metrics.cs
```diff
@@ -50,6 +51,13 @@ partial class DiagController
                     includeDefaults: true,
                     durationSeconds: durationSeconds);
 
+                // Allow sync I/O on livemetrics routes due to JsonCounterLogger's usage.
+                var syncIOFeature = HttpContext.Features.Get<IHttpBodyControlFeature>();
```
jander-msft, line 55: Could you make this an extension method of HttpContext and Debug.Assert that the feature could be obtained?

jander-msft (approved):

PR #2293, opened 2022-08-10
AKS Video Samples

kkeirstead:
These are lightly modified versions of what was used in the second draft of the video; these can be changed further (at least over the next few days, before I re-record with the updated files), but these need to be added so that I can make aka.ms links to the right page.


samples/AKS Tutorial/README.md
```diff
@@ -0,0 +1,2 @@
+### Associated Video
+**Add link here**
```
kkeirstead, line 2: This will be updated once the video comes out and there's a stable link to it.

samples/AKS Tutorial/deploy.yaml
```diff
@@ -0,0 +1,74 @@
...
+          value: /diag/port.sock
+        volumeMounts:
+        - mountPath: /diag
+          name: diagvol
+        resources:
+          limits:
+            cpu: 250m
+            memory: 512Mi
+      - name: monitor
+        image: mcr.microsoft.com/dotnet/monitor:6.2.2
```
kkeirstead, line 33: For managed identity, we need at least 6.2, but I wasn't sure if we should use a specific version here or just the latest. My only concern with latest is breaking changes in the future that could cause things not to work for people following along, but I also don't know if we want to lock this to one specific version of 6.2.*

jander-msft: You can use `mcr.microsoft.com/dotnet/monitor:6` which points to the latest 6.*.* version.

jander-msft (approved):

PR #2312, opened 2022-08-15
Operations API - Specify PID/UID/Name to see operations for specific process

kkeirstead:
Part 2 of #2122; allows users to specify a PID/UID/Name to only see the operations list for that specific process.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationStore.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationStore.cs
@@ -2,12 +2,10 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using Microsoft.Extensions.DependencyInjection;
 using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.Diagnostics.Monitoring.WebApi
@@ -30,7 +28,6 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
         private readonly EgressOperationQueue _taskQueue;
         private readonly RequestLimitTracker _requestLimits;
         private readonly IServiceProvider _serviceProvider;
-        private readonly IDiagnosticServices _diagnosticServices;
 
         public EgressOperationStore(
             EgressOperationQueue queue,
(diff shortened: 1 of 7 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationStore.cs
```diff
@@ -121,13 +121,24 @@ public void CompleteOperation(Guid operationId, ExecutionResult<EgressResult> re
...
         {
             lock (_requests)
             {
                 return _requests.Select((kvp) =>
                 {
                     EgressProcessInfo processInfo = kvp.Value.EgressRequest.EgressOperation.ProcessInfo;
+
+                    if (processKey != null)
+                    {
+                        if (processInfo.ProcessName != processKey.Value.ProcessName
```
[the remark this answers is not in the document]
kkeirstead: In this case, we're just checking to see if either the PID, UID, or name matches what's found in the `processInfo` - we expect that two of the three fields in the `ProcessKey` will be null, so only one needs to match. If none match (but the `ProcessKey` isn't null), then we know the process described by the `ProcessKey` isn't the same as the one in `processInfo`.

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationStore.cs
```diff
@@ -121,13 +121,43 @@ public void CompleteOperation(Guid operationId, ExecutionResult<EgressResult> re
...
-        public IEnumerable<Models.OperationSummary> GetOperations()
+        public IEnumerable<Models.OperationSummary> GetOperations(ProcessKey? processKey)
         {
             lock (_requests)
             {
                 return _requests.Select((kvp) =>
                 {
                     EgressProcessInfo processInfo = kvp.Value.EgressRequest.EgressOperation.ProcessInfo;
+
+                    if (processKey != null)
```
jander-msft, line 132: I think this makes more sense as a `Where` LINQ expression before this `Select` LINQ expression so that way we can avoid the need to return null and a `Where` after the `Select`. You could also conditionally construct the `IEnumerable` such that you don't have to check if the process key was provided for each iteration:

```cs
IEnumerable<KeyValuePair<Guid, EgressEntry>> requests = _requests;
if (null != processKey)
{
    requests = requests.Where(...);
}
return requets.Select(...).ToList();
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/Utilities.cs
```diff
@@ -34,5 +34,10 @@ public static KeyValueLogScope CreateArtifactScope(string artifactType, IEndpoin
             scope.AddArtifactEndpointInfo(endpointInfo);
             return scope;
         }
+
+        public static ProcessKey? GetProcessKey(int? pid, Guid? uid, string name)
+        {
+            return (pid == null && uid == null && name == null) ? null : new ProcessKey(pid, uid, name);
```
jander-msft, line 40: I think you should use `.HasValue` and `string.IsNullOrEmpty(name)` to be consistent with the rest of the code.

jander-msft (approved):

PR #2356, opened 2022-08-22
Integrate Azurite for AzureBlobEgressProvider test coverage

schmittjoseph:
This PR integrates [Azurite](https://github.com/Azure/Azurite) into our test infrastructure and adds some basic tests for `AzureBlobEgressProvider`. More tests and comprehensive coverage can be added in a later PRs.

An `AzuriteFixture` class is created that manages the lifecycle of the storage emulator during test runs. If a local development environment does not have Azurite available, the tests will be skipped. In a pipeline environment, this will result in a test failure to ensure errors are not silently ignored.

Azurite-based tests will work out-of-the-box for developers running the tests through Visual Studio 2022, or through our GitHub dev container. Both of which automatically provide Azurite. To run the tests locally in a different environment, install Azurite on the machine via `npm install -g azurite`.

Notes:
- Currently Windows pipelines will skip the Azurite based tests due to environment configuration issues that can be addresses in a separate PR.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Egress/AzureBlob/AzureBlobEgressProviderTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Egress/AzureBlob/AzureBlobEgressProviderTests.cs
@@ -25,7 +25,7 @@ using System.Xml;
 using Xunit;
 using Xunit.Abstractions;
 
-namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests.Eggress.AzureBlob
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests.Egress.AzureBlob
 {
     [TargetFrameworkMonikerTrait(TargetFrameworkMonikerExtensions.CurrentTargetFrameworkMoniker)]
     public class AzureBlobEgressProviderTests : IClassFixture<AzuriteFixture>, IDisposable


src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Fixtures/AzuriteFixture.cs
```diff
@@ -0,0 +1,256 @@
...
+            // to avoid hiding failures in our CI.
+            //
+            // Workaround: for now allow Windows environments to skip Azurite based tests due to configuration
+            // issues in the Pipeline environment.
+            bool mustInitialize = !RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("TF_BUILD"));
+
+            Account = new AzuriteAccount()
+            {
+                Name = Guid.NewGuid().ToString("N"),
+                Key = Convert.ToBase64String(Encoding.UTF8.GetBytes(Guid.NewGuid().ToString())),
```
jander-msft, line 65: Use [RandomNumberGenerator.GetBytes](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.randomnumbergenerator.getbytes?view=net-6.0#system-security-cryptography-randomnumbergenerator-getbytes(system-int32))

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Eggress/AzureBlob/AzureBlobEgressProviderTests.cs
```diff
@@ -0,0 +1,295 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using System.Xml;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests.Eggress.AzureBlob
+{
+    [TargetFrameworkMonikerTrait(TargetFrameworkMonikerExtensions.CurrentTargetFrameworkMoniker)]
+    public class AzureBlobEgressProviderTests : IClassFixture<AzuriteFixture>, IDisposable
```
jander-msft, line 31: I believe that `IClassFixture<T>` will cause startup and teardown of Azurite on a per-test basis. Not sure if that's what you intended.

Alternatively, you can use a collection fixture to have the emulator start before any of the tests in the collection and shut down after all of the tests have finished. See our `DefaultCollectionFixture` as an example. Note, a test class cannot be part of more than one collection, but since this class doesn't belong to the default one, this should be fairly easy to apply.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Eggress/AzureBlob/AzureBlobEgressProviderTests.cs
```diff
@@ -0,0 +1,295 @@
...
+
+            AzureBlobEgressProviderOptions providerOptions = ConstructEgressProviderSettings(containerClient);
+            EgressArtifactSettings artifactSettings = ConstructArtifactSettings();
+
+            // Act
+            string identifier = await egressProvider.EgressAsync(providerOptions, UploadAction, artifactSettings, CancellationToken.None);
+
+            // Assert
+            List<BlobItem> blobs = await GetAllBlobsAsync(containerClient);
+            Assert.Single(blobs);
```
jander-msft, line 70: nit: `Assert.Single` returns the strongly typed instance of the single object if it passes the assertion. You can combine the next line with this line so that you don't have to call the `First` extension method.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Eggress/AzureBlob/AzureBlobEgressProviderTests.cs
```diff
@@ -0,0 +1,295 @@
...
+        private Task<Stream> UploadAction(CancellationToken token)
+        {
+            return Task.FromResult<Stream>(new FileStream(_testUploadFile, FileMode.Open, FileAccess.Read));
+        }
+
+        private async Task<BlobContainerClient> ConstructBlobContainerClientAsync(string containerName = null, bool create = true)
+        {
+            BlobServiceClient serviceClient = new(_azuriteFixture.Account.ConnectionString);
+
+            containerName ??= Guid.NewGuid().ToString("D");
```
jander-msft, line 169: If a container name isn't specified, it should go to the root of the account, which I believe has a sentinel container name value of `$root`.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Eggress/AzureBlob/AzureBlobEgressProviderTests.cs
```diff
@@ -0,0 +1,295 @@
...
+            TestOutputLoggerProvider loggerProvider = new(_outputHelper);
+            AzureBlobEgressProvider egressProvider = new(loggerProvider.CreateLogger<AzureBlobEgressProvider>());
+
+            BlobContainerClient containerClient = await ConstructBlobContainerClientAsync(create: false);
+
+            AzureBlobEgressProviderOptions providerOptions = ConstructEgressProviderSettings(containerClient);
+            EgressArtifactSettings artifactSettings = ConstructArtifactSettings();
+
+            // Act
+            string identifier = await egressProvider.EgressAsync(providerOptions, UploadAction, artifactSettings, CancellationToken.None);
```
jander-msft, line 66: Let's pass a cancellation token with a reasonable large timeout (e.g. 10 seconds) in order to stop the test in case some change causes the egress provider to stall.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Eggress/AzureBlob/AzureBlobEgressProviderTests.cs
```diff
@@ -0,0 +1,295 @@
...
+
+            // Assert
+            List<BlobItem> blobs = await GetAllBlobsAsync(containerClient);
+            Assert.Single(blobs);
+
+            BlobItem resultingBlob = blobs.First();
+
+            string downloadedFile = await DownloadBlobAsync(containerClient, resultingBlob.Name, CancellationToken.None);
```
jander-msft, line 74: Same here regarding token with large timeout. You can build this one into the method itself rather than having the callers create it since I would anticipate this is not something we care to vary from test to test.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Eggress/AzureBlob/AzureBlobEgressProviderTests.cs
```diff
@@ -0,0 +1,295 @@
...
+
+        private AzureBlobEgressProviderOptions ConstructEgressProviderSettings(BlobContainerClient containerClient, QueueClient queueClient = null, string sasToken = null)
+        {
+            AzureBlobEgressProviderOptions options = new()
+            {
+                AccountUri = new Uri(_azuriteFixture.Account.BlobEndpoint),
+                ContainerName = containerClient.Name,
+                QueueAccountUri = (queueClient == null) ? null : new Uri(_azuriteFixture.Account.QueueEndpoint),
+                QueueName = queueClient?.Name,
+                BlobPrefix = Guid.NewGuid().ToString("D")
```
jander-msft, line 219: We should have a test that doesn't provide a prefix.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Eggress/AzureBlob/AzureBlobEgressProviderTests.cs
jander-msft, line 66: Can you also add tests that verify the other overload of this method that takes a `Func<Stream, CancellationToken, Task>` for the "action"? Maybe refactor the tests into reusable methods and then have the actual test method prescribe how the artifact is provided.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Eggress/AzureBlob/AzureBlobEgressProviderTests.cs
jander-msft: It's fine if this is a follow up item. There are some other cases that we'll want to consider covering it time since we've made specific effort to support scenarios, such as partial blob uploads.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/TestOutputLogger.cs
```diff
@@ -25,13 +25,25 @@ public ILogger CreateLogger(string categoryName)
             return _loggers.GetOrAdd(categoryName, (name, helper) => new TestOutputLogger(helper, name), _outputHelper);
         }
 
+        public ILogger<T> CreateLogger<T>()
```
jander-msft, line 28: Does this extension method not work? https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loggerfactoryextensions.createlogger?view=dotnet-plat-ext-6.0#microsoft-extensions-logging-loggerfactoryextensions-createlogger-1(microsoft-extensions-logging-iloggerfactory)

Might need to add a `PackageReference` to `Microsoft.Extensions.Logging.Abstractions`.

jander-msft: Sorry, I probably saw `ILogger*` and thought they were the same.

jander-msft (approved):

PR #2344, opened 2022-08-19
Adding Stale PR Guidelines

kkeirstead:

PR #2374, opened 2022-08-25
Use FabricBot to add stale label and close PRs that are stale

kkeirstead:
I haven't been able to test this, but this PR is intended to do the following:

- If a PR hasn't had any activity in 4 weeks (28 days), we add a 'stale' label to it, and leave a comment saying that the PR will be closed if there is no activity
- If a PR hasn't had any activity in 10 weeks total (4 weeks + 6 weeks after the comment), we close the PR
- If a PR has the stale label and has activity, we remove the stale label. This piece I'm not confident in, because I'm not sure if this will immediately remove the stale label as soon as it's added (because it'll be seen as new activity), but I believe I've mitigated that by saying the activity sender cannot be the fabric bot.

Follow-up to #2344

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -278,6 +278,17 @@
             {
               "name": "isAction",
               "parameters": {}
+            },
+            {
+              "operator": "not",
+              "operands": [
+                {
+                  "name": "isActivitySender",
+                  "parameters": {
+                    "user": "msftbot"
+                  }
+                }
+              ]
             }
           ]
         },
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -105,6 +105,198 @@
           }
         ]
       }
+    },
+    {
+      "taskType": "scheduled",
+      "capabilityId": "ScheduledSearch",
+      "subCapability": "ScheduledSearch",
+      "version": "1.1",
+      "config": {
+        "frequency": [
+          {
+            "weekDay": 0,
+            "hours": [],
+            "timezoneOffset": -7
+          },
+          {
+            "weekDay": 1,
+            "hours": [],
+            "timezoneOffset": -7
+          },
+          {
+            "weekDay": 2,
+            "hours": [],
+            "timezoneOffset": -7
+          },
+          {
+            "weekDay": 3,
+            "hours": [],
+            "timezoneOffset": -7
+          },
+          {
+            "weekDay": 4,
+            "hours": [],
+            "timezoneOffset": -7
+          },
+          {
+            "weekDay": 5,
+            "hours": [],
+            "timezoneOffset": -7
+          },
+          {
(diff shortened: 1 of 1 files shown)


kkeirstead:
> suggestion: if you want to validate the behavior, you could check in these changes but with much shorter durations (e.g. minutes instead of days) and add a filter that only adds the `stale` label to PRs with a `test-fabric-bot` label already on them or something similar.

Good idea - I'll push an update in a minute with shorter durations and an additional label check.

jander-msft:
> * If a PR hasn't had any activity in 4 weeks (28 days), we add a 'stale' label to it, and leave a comment saying that the PR will be closed if there is no activity
> * If a PR hasn't had any activity in 10 weeks total (4 weeks + 6 weeks after the comment), we close the PR
> * If a PR has the stale label and has activity, we remove the stale label. This piece I'm not confident in, because I'm not sure if this will immediately remove the stale label as soon as it's added (because it'll be seen as new activity), but I believe I've mitigated that by saying the activity sender cannot be the fabric bot.

I know that these are the time frames that we agreed to, but it might be worth considering matching what the runtime repo has: https://github.com/dotnet/runtime/blob/main/.github/fabricbot.json#L2801
- After 14 days without activity, apply `no-recent-activity` label (sounds better than `stale`)
- After another 14 days (14 + 14 = 28 days), close the PR
- After another 30 days (14 + 14 + 30 = 58 days), lock the PR
This way there is consistency with one of the major repositories.

kkeirstead:
> > * If a PR hasn't had any activity in 4 weeks (28 days), we add a 'stale' label to it, and leave a comment saying that the PR will be closed if there is no activity
> > * If a PR hasn't had any activity in 10 weeks total (4 weeks + 6 weeks after the comment), we close the PR
> > * If a PR has the stale label and has activity, we remove the stale label. This piece I'm not confident in, because I'm not sure if this will immediately remove the stale label as soon as it's added (because it'll be seen as new activity), but I believe I've mitigated that by saying the activity sender cannot be the fabric bot.
> 
> I know that these are the time frames that we agreed to, but it might be worth considering matching what the runtime repo has: https://github.com/dotnet/runtime/blob/main/.github/fabricbot.json#L2801
> 
> * After 14 days without activity, apply `no-recent-activity` label (sounds better than `stale`)
> * After another 14 days (14 + 14 = 28 days), close the PR
> * After another 30 days (14 + 14 + 30 = 58 days), lock the PR
> 
> This way there is consistency with one of the major repositories.

I can definitely change the label, that should be easy enough - in terms of the time-frames, I know Will was advocating for long periods of time to give people lots of flexibility, so I don't know whether we should be doing that or just following the precedent. I can check this in with the test numbers in the meantime just to make sure it works, and then adjust the timeframes if that's what we decide to do.

PR #2387, opened 2022-08-25
Automatically open PRs for NuGet dependency updates

schmittjoseph:
This PR adds Dependabot integration with our NuGet packages. 
- Package versions managed by Maestro should be defined in `eng/Versions.props`
- Package versions managed by Dependabot should be defined in `eng/dependabot/Versions.props` with a corresponding reference in `eng/dependabot/Packages.props`.
- Dependabot will limit itself to only having 5 open PRs at any given time. Once one of its PRs is addressed, it'll open another one if needed. We can configure this limit if needed.

Example PR with a NuGet version update: https://github.com/schmittjoseph/dotnet-monitor/pull/27

### DIFF
--- a/eng/dependabot/Versions.props
+++ b/eng/dependabot/Versions.props
@@ -1,14 +1,10 @@
 <Project>
-  <!-- Import references updated by Dependabot. This file is for package references updated manually or by Maestro. -->
+  <!-- Import references updated by Dependabot. -->
 
   <PropertyGroup>
     <AzureIdentityVersion>1.6.0</AzureIdentityVersion>
     <AzureStorageBlobsVersion>12.13.0</AzureStorageBlobsVersion>
     <AzureStorageQueuesVersion>12.11.0</AzureStorageQueuesVersion>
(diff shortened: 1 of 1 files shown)


eng/dependabot/Versions.props
```diff
@@ -0,0 +1,29 @@
+<Project>
+  <!-- Import references updated by Dependabot. This file is for package references updated manually or by Maestro. -->
+
+  <PropertyGroup>
+    <AzureIdentityVersion>1.6.0</AzureIdentityVersion>
+    <AzureStorageBlobsVersion>12.13.0</AzureStorageBlobsVersion>
+    <AzureStorageQueuesVersion>12.11.0</AzureStorageQueuesVersion>
+    <MicrosoftAspNetCoreAuthenticationJwtBearerVersionNet6>6.0.5</MicrosoftAspNetCoreAuthenticationJwtBearerVersionNet6>
```
jander-msft, line 8: These four versions should be removed since dependabot won't understand what they are. Seems like they are unused because they still exist in `./eng/Versions.props`

jander-msft (approved):

jander-msft:
I've marked this for backport, but we can wait a few days to see how everything lands and will probably have to manually make the changes there anyway.

PR #2416, opened 2022-08-30
Load profiler into target processes in listen mode

jander-msft:
### Overview

These changes add the ability to load the dotnet-monitor profiler into target processes in listen mode during suspension if it can detect which runtime identifier corresponds to the target process. This determination currently only supports reading the `DotnetMonitorProfiler_RuntimeIdentifier` environment variable from the target process.

### Reserved for future PRs
- Requiring explicit enablement of loading the profiler via configuration
- Additional heuristics or lookups for the necessary runtime identifier


src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
```diff
@@ -0,0 +1,176 @@
...
+using System.IO;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Profiler
+{
+    internal sealed class ProfilerService : BackgroundService
+    {
+        private static readonly Guid ProfilerClsId = new Guid("6A494330-5848-4A23-9D87-0E57BBF6DE79");
```
kkeirstead, line 22: It looks like we use this specific GUID in ProfilerHelper.cs; even if we can't reuse the value and have to copy-paste it here, do we want a comment saying these should not be changed independently of one another?

src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
```diff
@@ -0,0 +1,176 @@
...
+
+                // CONSIDER: Allow alternate means of specifying or heuristically determining the runtime identifier
+                // - For Windows and OSX, build identifier from platform + architecture of target process.
+                // - Lookup same environment variable on dotnet-monitor itself.
+                // - Check libc type of dotnet-monitor.
+                env.TryGetValue(DotnetMonitorProfiler_RuntimeIdentifier, out string runtimeIdentifier);
+
+                if (string.IsNullOrEmpty(runtimeIdentifier))
+                {
+                    throw new InvalidOperationException("Unable to determine platform of the target platform.");
```
kkeirstead, line 130: Should this be, "Unable to determine platform of the target _process_."?

src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
```diff
@@ -0,0 +1,176 @@
...
+                else
+                {
+                    // This may be null if running dotnet-monitor from build output instead of from
+                    // dotnet tool installation.
+                    string sharedLibraryPath = await _libraryPathSource.Task;
+
+                    IFileProvider nativeFileProvider;
+                    if (string.IsNullOrEmpty(sharedLibraryPath))
+                    {
+                        // CONSIDER: Allow some way to of specifying the build output location of the libraries
```
kkeirstead, line 141:
```suggestion
                        // CONSIDER: Allow some way of specifying the build output location of the libraries
```

src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
```diff
@@ -0,0 +1,176 @@
...
+                        nativeFileProvider = NativeFileProvider.CreateShared(runtimeIdentifier, sharedLibraryPath);
+                    }
+
+                    string libraryName = NativeLibraryHelper.GetSharedLibraryName("MonitorProfiler");
+
+                    IFileInfo profilerFileInfo = nativeFileProvider.GetFileInfo(libraryName);
+                    if (!profilerFileInfo.Exists)
+                    {
+                        // Do not use IFileInfo.PhysicalPath as that is null of files that do not exist.
+                        throw new FileNotFoundException("Could not find profiler assembly at determined path.", profilerFileInfo.Name);
```
kkeirstead, line 156: Should this be `profilerFileInfo.FullName`, or do we specifically only want the `Name`?

jander-msft: [IFileInfo](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.fileproviders.ifileinfo?view=dotnet-plat-ext-6.0) does not have a `FullName` property. Note that this is an interface named `IFileInfo` from the `Microsoft.Extensions.FileProviders.Abstractions` assembly, not the class named `FileInfo` from the `System.Runtime` assembly.

kkeirstead: Ah, that was an oversight on my part, sorry about that.

src/Tools/dotnet-monitor/Profiler/NativeFileProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+using System;
+using System.Reflection;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Profiler
+{
+    internal sealed class NativeFileProvider : IFileProvider
+    {
+        private readonly string _nativeFileBasePath;
+
+        private NativeFileProvider(string osPlatform, string nativeFileBasePath)
```
kkeirstead, line 18: Unused `osPlatform`

src/Tools/dotnet-monitor/Profiler/NativeLibraryHelper.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Profiler
+{
+    internal static class NativeLibraryHelper
+    {
+        public static string GetSharedLibraryName(string rootName)
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                return $"{rootName}.dll";
```
wiktork, line 16: nit: Invariant

src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
```diff
@@ -0,0 +1,171 @@
...
+            _logger = logger;
+            _storageOptions = storageOptions;
+        }
+
+        protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+        {
+            using IDisposable _ = stoppingToken.Register(() => _libraryPathSource.TrySetCanceled(stoppingToken));
+
+            // Yield to allow other hosting services to start
+            await Task.Yield();
```
jander-msft, line 41: Measure how long the file copy takes in k8s and remove if its negligible.

src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
```diff
@@ -0,0 +1,171 @@
...
+            {
+                if (string.IsNullOrEmpty(sharedLibraryTargetPath))
+                {
+                    _libraryPathSource.SetResult(SharedLibrarySourcePath);
+                }
+                else
+                {
+                    try
+                    {
+                        // CONSIDER: Rather than using a version file, copy the shared assemblies into a subpath that
```
jander-msft, line 58: Let's do this now: copy to a version folder and write a sentinel file

src/Tools/dotnet-monitor/LoggingEventIds.cs
```diff
@@ -81,7 +81,10 @@ internal enum LoggingEventIds
         DiagnosticPortWatchingFailed = 68,
         InvalidMetadata = 69,
         DuplicateKeyInMetadata = 70,
-        EnvironmentVariableNotFound = 71
+        EnvironmentVariableNotFound = 71,
+        FailedInitializeSharedLibraryStorage = 73,
+        UnableToApplyProfiler = 74,
+        SharedLibraryPath
```
kkeirstead, line 87: Is there a reason this doesn't have a value? I'm also assuming skipping 72 was intentional to avoid the merge conflict with `EnvironmentBlockNotSupported `

jander-msft: Fixed by adding explicit value; the skipping was intentional, but there will probably be a merge conflict anyway

src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
jander-msft: I removed the `Task.Yield` call and used a `StopWatch` to time the initialization call (which basically checks the filesystem and does file copying) to see how long it would take:

- 182 ms
- 278 ms
- 203 ms
- 295 ms
- 190 ms

Average: 230 ms

These measurements were taken on a single Standard_DS2_v2 node that had no other user defined deployments on it.

I'm inclined to keep this `Task.Yield` in place in order to not block startup for this long.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.TestHostingStartup/BuildOutputSharedLibraryInitializer.cs
```diff
@@ -0,0 +1,41 @@
...
+using Microsoft.Extensions.Logging;
+using System.IO;
+using System.Reflection;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.TestHostingStartup
+{
+    internal sealed class BuildOutputSharedLibraryInitializer : ISharedLibraryInitializer
+    {
+        private static readonly string SharedLibrarySourcePath =
+            Path.GetFullPath(Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "..", "..", ".."));
```
wiktork, line 17: Maybe some comments that expand this path.

wiktork, line 17: Does this end up being the same for VS F5 scenarios and command line scenarios that call build.ps1?

src/Tools/dotnet-monitor/Profiler/DefaultSharedLibraryInitializer.cs
```diff
@@ -0,0 +1,96 @@
...
+            {
+                string expectedVersion = Assembly.GetExecutingAssembly().GetInformationalVersionString();
+                // Remove '+' and commit hash from version string
+                int hashSeparatorIndex = expectedVersion.IndexOf('+');
+                if (hashSeparatorIndex > 0)
+                {
+                    expectedVersion = expectedVersion.Substring(0, hashSeparatorIndex);
+                }
+
+                string versionedSharedLibraryTargetPath = Path.Combine(_sharedLibraryTargetPath, expectedVersion);
```
wiktork, line 61: It would be helpful to add an example of an expected directory structure here.

src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
```diff
@@ -0,0 +1,100 @@
...
+                {
+                    throw new InvalidOperationException(Strings.ErrorMessage_UnableToDetermineTargetPlatform);
+                }
+                else
+                {
+                    INativeFileProviderFactory fileProviderFactory = await _fileProviderFactorySource.Task;
+
+                    IFileProvider nativeFileProvider = fileProviderFactory.Create(runtimeIdentifier);
+
+                    string libraryName = NativeLibraryHelper.GetSharedLibraryName("MonitorProfiler");
```
wiktork, line 74: Let's put this in a constant somewhere.

src/Tools/dotnet-monitor/TestAssemblies.cs
```diff
@@ -0,0 +1,89 @@
...
+            AssemblyLoadContext alc = AssemblyLoadContext.GetLoadContext(Assembly.GetExecutingAssembly());
+            // Skip attempting to load if it is already loaded
+            if (!alc.Assemblies.Any(a => TestStartupHookAssemblyName.Equals(a.GetName().Name)))
+            {
+                // Only load if can compute the path correctly
+                if (TryComputeBuildOutputAssemblyPath(TestStartupHookAssemblyName, out string startupHookAssemblyPath))
+                {
+                    // Load and simulate startup hook initialization
+                    Assembly startupHookAssembly = alc.LoadFromAssemblyPath(startupHookAssemblyPath);
+                    Type startupHookType = startupHookAssembly.GetType("StartupHook");
```
wiktork, line 40: nit: I'm wondering if we can do less direct binding to these names so that renaming won't just break this at runtime. For example, using an assembly level attribute that tells dotnet-monitor which type to look for.

jander-msft: The startup hook notion is a public contract where a type named `StartupHook` has a static `Initialize` method. This is highly unlikely to ever change. See [runtime](https://github.com/dotnet/runtime/blob/6246f84506a7f0a2cf990c2fddd7225691f4bc89/src/coreclr/System.Private.CoreLib/src/System/StartupHookProvider.cs#L114) startup hook processing code, which the code that I wrote is a highly simplified version of it.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.TestHostingStartup/BuildOutputSharedLibraryInitializer.cs
jander-msft: Yes, it's the same regardless of it being built by VS or command line.

src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
```diff
@@ -0,0 +1,100 @@
...
+
+                // CONSIDER: Allow alternate means of specifying or heuristically determining the runtime identifier
+                // - For Windows and OSX, build identifier from platform + architecture of target process.
+                // - Lookup same environment variable on dotnet-monitor itself.
+                // - Check libc type of dotnet-monitor.
+                env.TryGetValue(ProfilerIdentifiers.EnvironmentVariables.RuntimeIdentifier, out string runtimeIdentifier);
+
+                if (string.IsNullOrEmpty(runtimeIdentifier))
+                {
+                    throw new InvalidOperationException(Strings.ErrorMessage_UnableToDetermineTargetPlatform);
```
wiktork, line 66: After this PR, let's make this opt in.

src/Tools/dotnet-monitor/Profiler/ProfilerEndpointInfoSourceCallbacks.cs
```diff
@@ -0,0 +1,35 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Diagnostics.Monitoring.WebApi;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Profiler
+{
+    internal sealed class ProfilerEndpointInfoSourceCallbacks : IEndpointInfoSourceCallbacks
```
jander-msft, line 11: After this PR, allow profiler to be loaded on demand or at startup as required by each feature that needs it.

wiktork (approved):

PR #2452, opened 2022-09-06
Disable exception tracking via profiler

jander-msft:
Remove exception tracking while evaluating future designs and allow the profiler to ship without it.

### DIFF
--- a/src/MonitorProfiler/CMakeLists.txt
+++ b/src/MonitorProfiler/CMakeLists.txt
@@ -22,10 +22,7 @@ set(SOURCES
     Logging/LoggerHelper.cpp
     Logging/NullLogger.cpp
     Logging/StdErrLogger.cpp
-    MainProfiler/ExceptionTracker.cpp
     MainProfiler/MainProfiler.cpp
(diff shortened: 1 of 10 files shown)


wiktork (dismissed):
Did you want to disable this with a preprocessor directive, to make it easier to flip?

src/MonitorProfiler/CMakeLists.txt
```diff
@@ -38,6 +38,9 @@ set(SOURCES
     Communication/CommandServer.cpp
     )
 
+# Include exceptions tracking feature
+#add_compile_options(-DDOTNETMONITOR_FEATURE_EXCEPTIONS)
```
kkeirstead, line 42:
```suggestion
#add_compile_options(-DOTNETMONITOR_FEATURE_EXCEPTIONS)
```

jander-msft: It is correct as-is. The `-D` specifies that it is a preprocessor definition.

wiktork (approved):

PR #2471, opened 2022-09-08
Add DefaultSharedPath option for sharing dumps, sockets, and libraries

jander-msft:
This new, optional `DefaultSharedPath` option allows a single setting for sharing file assets (dumps, sockets, libraries, etc) between dotnet-monitor and target processes. This setting will also implicitly set the tool to run in `listen` mode and establish its diagnostic port at `{DefaultSharedPath}/dotnet-monitor.sock` for non-Windows machines. If the setting is not specified, the profiler communication socket and dump creation will fall back to using `/tmp` or `%TEMP%`.

The `DumpTempFolder` and `SharedLibrariesPath` will supersede this setting if they are specified.

AKS PodSpec Sample:
```yml
containers:
- name: app
  image: mcr.microsoft.com/dotnet/samples:aspnetapp
  env:
  - name: DOTNET_DiagnosticPorts
    value: /diag/dotnet-monitor.sock
  - name: DotnetMonitorProfiler_RuntimeIdentifier
    value: linux-x64
  volumeMounts:
  - mountPath: /diag
    name: diagvol
- name: monitor
  image: mcr.microsoft.com/dotnet/nightly/monitor:6
  env:
  - name: DOTNETMONITOR_Storage__DefaultSharedPath
    value: /diag
  volumeMounts:
  - mountPath: /diag
    name: diagvol
```


src/MonitorProfiler/Environment/EnvironmentHelper.h
```diff
@@ -21,6 +21,7 @@ class EnvironmentHelper final
     static constexpr LPCWSTR DebugLoggerLevelEnvVar = _T("DotnetMonitorProfiler_DebugLogger_Level");
     static constexpr LPCWSTR ProfilerVersionEnvVar = _T("DotnetMonitorProfiler_ProductVersion");
     static constexpr LPCWSTR RuntimeInstanceEnvVar = _T("DotnetMonitorProfiler_RuntimeInstanceId");
+    static constexpr LPCWSTR SharingPathEnvVar = _T("DotnetMonitorProfiler_SharingPath");
```
wiktork, line 24: nit: SharedPath?

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/TemporaryDirectory.cs
```diff
@@ -17,7 +17,7 @@ public TemporaryDirectory(ITestOutputHelper outputhelper)
         {
             _outputHelper = outputhelper;
 
-            _directoryInfo = Directory.CreateDirectory(Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString("N")));
+            _directoryInfo = Directory.CreateDirectory(Path.Combine(Path.GetTempPath(), Path.GetRandomFileName()));
```
wiktork, line 20: Are we deleting this folder afterwards? Note that GetRandomFileName likely shares some of the limitations of GetTempFile such as a limit of 65k entries.

jander-msft: Yes, it is deleted on dispose

documentation/schema.json
```diff
@@ -1183,6 +1183,13 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
+        "DefaultSharedPath": {
+          "type": [
+            "null",
+            "string"
+          ],
+          "description": "The default path where assets (dumps, socket files, libraries) will be shared between dotnet-monitor and target processes."
```
wiktork, line 1191: I think we should expand this documentation. For example, we should mention what settings are superseded and that we switch to listen mode automatically.

jander-msft: I'm not sure how to describe all of that in this description without writing a book. How's this?

The default path where assets will be shared between dotnet-monitor and target processes. Dumps are temporarily stored under this path or in a sub folder unless DumpTempFolder is specified. Shared libraries are stored under this path or in a sub folder unless SharedLibraryPath is specified. On non-Windows, dotnet-monitor runs in listen mode with a Unix domain socket named 'dotnet-monitor.sock' immediately under this path unless the diagnostic port is specified on the command line or the DiagnosticPort options are specified.

src/Tools/dotnet-monitor/DiagnosticPortPostConfigureOptions.cs
```diff
@@ -29,6 +34,15 @@ public void PostConfigure(string name, DiagnosticPortOptions options)
...
                 options.EndpointName = diagPortSection.Value;
             }
+
+            // Create a default server socket under the default shared path if diagnostic port was not configured
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows) &&
+                !options.ConnectionMode.HasValue && string.IsNullOrEmpty(options.EndpointName) &&
+                !string.IsNullOrEmpty(_storageOptions.Value.DefaultSharedPath))
+            {
+                options.ConnectionMode = DiagnosticPortConnectionMode.Listen;
+                options.EndpointName = Path.Combine(_storageOptions.Value.DefaultSharedPath, "dotnet-monitor.sock");
```
wiktork, line 44: nit: move this name to a constant

src/Tools/dotnet-monitor/EndpointInfo/ServerEndpointInfoSource.cs
```diff
@@ -84,21 +84,25 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+                            _logger.DiagnosticPortDeleteAttempt(_portOptions.EndpointName);
+                            File.Delete(_portOptions.EndpointName);
+                        }
+                        catch (Exception ex)
+                        {
+                            _logger.DiagnosticPortDeleteFailed(_portOptions.EndpointName, ex);
+                        }
                     }
+
+                    Directory.CreateDirectory(Path.GetDirectoryName(_portOptions.EndpointName));
```
wiktork, line 105: Is this method safe if the directory already exists?

jander-msft: Yes

src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
```diff
@@ -80,6 +84,22 @@ public async Task ApplyProfiler(IEndpointInfo endpointInfo, CancellationToken ca
                         throw new FileNotFoundException(Strings.ErrorMessage_UnableToFindProfilerAssembly, profilerFileInfo.Name);
                     }
 
+                    // This optional setting instructs where the profiler should establish its socket file
+                    // and where to provide any additional files to dotnet-monitor.
+                    // CONSIDER: Include the runtime instance identifier in the path in order to keep
```
wiktork, line 89: Given that the uds files will have the runtime instance id and that we probably want to share the same profiler bits between multiple apps, I'm not sure we want the share folder to have the runtime instance.

src/Tools/dotnet-monitor/StartupLogging.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Runtime.InteropServices;
+using System.Threading;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    internal sealed class StartupLogging : IDisposable
+    {
+        private const string WindowsLocalNamedPipePrefix = @"\\.\pipe\";
+
+        private readonly IHostApplicationLifetime _lifetime;
```
wiktork, line 20: What is the difference between our current startup logging (such as the warnings around no auth) vs. using IHostApplicationLifetime? Should the other log messages be moved here?

src/Tools/dotnet-monitor/StoragePostConfigureOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    public class StoragePostConfigureOptions : IPostConfigureOptions<StorageOptions>
+    {
+        void IPostConfigureOptions<StorageOptions>.PostConfigure(string name, StorageOptions options)
+        {
+            if (string.IsNullOrEmpty(options.DumpTempFolder))
+            {
+                if (!string.IsNullOrEmpty(options.DefaultSharedPath))
+                {
+                    options.DumpTempFolder = Path.Combine(options.DefaultSharedPath, "dumps");
```
wiktork, line 19: nit: let's make constants for these subfolders.

src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
jander-msft: With the current changes, if DefaultSharedPath is set to `/diag`, then the structure looks like:

```
/diag/dotnet-monitor.sock
/diag/dumps/*
/diag/libs/<ver>/<rid>/native/*MonitorProfiler*
/diag/<rii>.sock
```

If I were to make this change in the future, I think the files that the profiler produces should go into a runtime instance ID folder. The shared library bits would not go into this folder. If the DefaultSharedPath is set to `/diag`, then it would look like this:

```
/diag/dotnet-monitor.sock
/diag/dumps/*
/diag/libs/<ver>/<rid>/native/*MonitorProfiler*
/diag/<rii>/profiler.sock
/diag/<rii>/symbols/*
```

src/Tools/dotnet-monitor/StartupLogging.cs
jander-msft: I did this to stop polluting the Startup class with logging aspects that are unrelated to ASP.NET startup. The exist logging can be moved over here to keep the purely logging aspects separate from the ASP.NET startup aspects. Want me to do that in this PR?

src/Microsoft.Diagnostics.Monitoring.Options/StorageOptions.cs
```diff
@@ -8,6 +8,11 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
 {
     internal class StorageOptions
     {
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_StorageOptions_DefaultSharedPath))]
+        public string DefaultSharedPath { get; set; }
```
wiktork, line 14: I'll adjust my managed code to use SharedLibraryPath

jander-msft: Not `SharedLibraryPath`, you'll use `DefaultSharedPath`

jander-msft: At least until yet another property is added for where the profiler should produce files e.g. its communication socket.

src/Tools/dotnet-monitor/StartupLogging.cs
wiktork: We can definitely save it for another PR.

wiktork (approved):

PR #2481, opened 2022-09-12
Clear server URLs after setting endpoints on KestrelServerOptions

jander-msft:
This change clears the server URLs from configuration since the URLs are represented as endpoints in the KestrelServerOptions. This removes the warning that is displayed about overriding URLs from configuration:

```
12:15:26 warn: Microsoft.AspNetCore.Server.Kestrel[0]
      Overriding address(es) 'https://localhost:52323'. Binding to endpoints defined via IConfiguration and/or UseKestrel() instead.
```

The above warning will no longer be logged since there are no URLs in configuration, yet the endpoints are still bound.

### DIFF
--- a/src/Tools/dotnet-monitor/HostBuilderHelper.cs
+++ b/src/Tools/dotnet-monitor/HostBuilderHelper.cs
@@ -130,6 +130,10 @@ namespace Microsoft.Diagnostics.Tools.Monitor
                             options,
                             urls,
                             metricsOptions.Enabled.GetValueOrDefault(MetricsOptionsDefaults.Enabled) ? metricUrls : Array.Empty<string>());
+
+                        // Since the endpoints have already been defined on the KestrelServerOptions, clear out the urls
+                        // so that the address binder does not log a warning about overriding urls as specified by configuration.
+                        context.Configuration[WebHostDefaults.ServerUrlsKey] = string.Empty;
                     })
                     .UseStartup<Startup>();
                 });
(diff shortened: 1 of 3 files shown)


src/Tools/dotnet-monitor/HostBuilderHelper.cs
```diff
@@ -130,6 +130,10 @@ public static IHostBuilder CreateHostBuilder(HostBuilderSettings settings)
                             options,
                             urls,
                             metricsOptions.Enabled.GetValueOrDefault(MetricsOptionsDefaults.Enabled) ? metricUrls : Array.Empty<string>());
+
+                        // Since the endpoints have already been defined on the KestrelServerOptions, clear out the urls
+                        // so that the address binder does not log a warning about overriding urls as specified by configuration.
+                        context.Configuration[WebHostDefaults.ServerUrlsKey] = string.Empty;
```
wiktork, line 136: Have you tested this with dotnet config show? I'm assuming it will still work correctly.

wiktork (approved):

src/Tools/dotnet-monitor/HostBuilderHelper.cs
jander-msft: This appears to work correctly:

```
dotnet dotnet-monitor.dll config show --urls "http://localhost:52327;http://+:52329"

{
  "urls": "http://localhost:52327;http://\u002B:52329",
   ...
}
```

src/Tools/dotnet-monitor/Startup.cs
```diff
@@ -147,28 +147,11 @@ public void ConfigureServices(IServiceCollection services)
                 {
                     logger.LogTempKey(options.TemporaryJwtKey.Token);
                 }
-                //Auth is enabled and we are binding on http. Make sure we log a warning.
 
-                string hostingUrl = Configuration.GetValue<string>(WebHostDefaults.ServerUrlsKey);
```
jander-msft, line 152: This refactor is necessary due to the value being cleared from configuration by the point this executes.

PR #2488, opened 2022-09-14
Update SDK to 7.0 RC 1 and runtimes to latest

jander-msft:
Edit: I've made all of the options properties that have a type of `IDictionary<>` to be settable to work around this issue: https://github.com/dotnet/runtime/issues/75626

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.Options/AzureBlobEgressProviderOptions.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.Options/AzureBlobEgressProviderOptions.cs
@@ -85,7 +85,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Egress.AzureBlob
         [Display(
             ResourceType = typeof(OptionsDisplayStrings),
             Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_AzureBlobEgressProviderOptions_Metadata))]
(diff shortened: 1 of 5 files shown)


src/Microsoft.Diagnostics.Monitoring.Options/AzureBlobEgressProviderOptions.cs
```diff
@@ -85,7 +85,7 @@ internal sealed partial class AzureBlobEgressProviderOptions :
         [Display(
             ResourceType = typeof(OptionsDisplayStrings),
             Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_AzureBlobEgressProviderOptions_Metadata))]
-        public IDictionary<string, string> Metadata { get; }
+        public IDictionary<string, string> Metadata { get; set; }
```
wiktork, line 88: Do you want to add comments regarding the setter?

wiktork (approved):

src/Microsoft.Diagnostics.Monitoring.Options/AzureBlobEgressProviderOptions.cs
jander-msft: The issue is going to be fixed in RC 2, so I don't think it is worthwhile to describe a situation that will be fixed soon.

PR #2502, opened 2022-09-15
Enable testing profiler with x86 host

jander-msft:


src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/DotNetHost.cs
```diff
@@ -18,13 +18,23 @@ public partial class DotNetHost
...
-            @"..\..\..\..\..\.dotnet\dotnet.exe" :
-            "../../../../../.dotnet/dotnet";
+        public static string ExeNameWithoutExtension => RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ?
+            Path.GetFileNameWithoutExtension(ExeName) :
+            ExeName;
+
+        public static string GetPath(Architecture? arch = null)
+        {
+            // e.g. <repoPath>/.dotnet
+            string dotnetDirPath = Path.Combine("..", "..", "..", "..", "..", ".dotnet");
```
wiktork, line 30: Are there cases where we don't want to use this host but rather the one on the machine itself? For example, VS unit test scenarios?

wiktork (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/DotNetHost.cs
jander-msft: I'll think about how we can handle that and save it for another PR

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/ProfilerHelper.cs
```diff
@@ -58,6 +58,28 @@ public static string GetTargetRuntimeIdentifier(Architecture? architecture)
             throw new PlatformNotSupportedException("Unable to determine OS platform.");
         }
 
+        public static IEnumerable<object[]> GetArchitecture()
+        {
+            // There isn't a good way to check which architecture to use when running unit tests.
+            // Each build job builds one specific architecture, but from a test perspective,
+            // it cannot tell which one was built. Gather all of the profilers for every architecture
```
wiktork, line 65: I realize this is a bit roundabout, since TargetRuntimeIdentifier uses the profiler path to determine the id, but this does give the impression that every test is going to run all 3 profiler scenarios on the target app. Isn't it only going to find one? Couldn't we just map TargetRuntimeIdentifier to an architecture or something similiar?

wiktork (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/ProfilerHelper.cs
jander-msft: > I realize this is a bit roundabout, since TargetRuntimeIdentifier uses the profiler path to determine the id, but this does give the impression that every test is going to run all 3 profiler scenarios on the target app. Isn't it only going to find one? 

It's only going to find as many as you built, and very likely not all three.

> Couldn't we just map TargetRuntimeIdentifier to an architecture or something similiar?

There isn't any known architecture though. The same host x64 OS for Windows is used to build arm64, x64, and x86 on different machines. Similarly, the host x64 Linux OS is used to build arm64 and x64 on different machines. The only ways to work around this are to rely on some type of build architecture environment variable or to use a test settings file, both of which are going to be prone to error and require more setup.

PR #2506, opened 2022-09-16
Add experimental flags and in-process feature options

jander-msft:
This change makes the call stacks feature and experimental feature by only allowing to be enabled if `DOTNETMONITOR_EXPERIMENTAL_FEATURE_CALLSTACKS` is set to a non-`0` and non-`false` value. Additionally, in-process features are disabled (such as call stacks) unless enabled via configuration using the `InProcessFeatures:Enabled` property (e.g. set environment variable `DotnetMonitor_InProcessFeatures__Enabled` to `true`. Functional tests have been updated to only allow the stacks feature to be enabled when testing the stacks feature.


src/Tools/dotnet-monitor/Experimental/ExperimentalFlags.cs
```diff
@@ -0,0 +1,41 @@
...
+
+#if UNITTEST
+namespace Microsoft.Diagnostics.Monitoring.TestCommon
+#else
+namespace Microsoft.Diagnostics.Tools.Monitor
+#endif
+{
+    internal static class ExperimentalFlags
+    {
+        private const string ExperimentalPrefix = "DotnetMonitor_Experimental_";
```
kkeirstead, line 15: Should this be `DotnetMonitor_Experimental__" (with two underscores at the end)?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/StacksTests.cs
```diff
@@ -200,6 +212,75 @@ public Task TestRepeatStackCalls(Architecture targetArchitecture)
...
+                    await runner.SendCommandAsync(TestAppScenarios.Stacks.Commands.Continue);
+                },
+                configureApp: runner =>
+                {
+                    runner.Architecture = targetArchitecture;
+                },
+                configureTool: runner =>
+                {
+                    runner.EnableCallStacksFeature = true;
+                    // Note that the in-process features are not been enabled
```
kkeirstead, line 283: nit: grammar

src/Tools/dotnet-monitor/Experimental/ExperimentalFlags.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        // Feature flags
+        public const string Feature_CallStacks = ExperimentalPrefix + nameof(Feature_CallStacks);
```
kkeirstead, line 18: In the same vein as the prior comment (so this may not be applicable) - if we're using double underscores to separate, then should this be "Feature__CallStacks"?

src/Microsoft.Diagnostics.Monitoring.Options/InProcessFeaturesOptions.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Diagnostics.Monitoring.Options
+{
+    internal class InProcessFeaturesOptions
```
wiktork, line 7: Add documentation and schema gen.

src/Tools/dotnet-monitor/Experimental/ExperimentalFeatures.cs
```diff
@@ -0,0 +1,22 @@
...
+    internal sealed class ExperimentalFeatures : IExperimentalFeatures
+    {
+        private readonly InProcessFeaturesOptions _options;
+
+        public ExperimentalFeatures(IOptions<InProcessFeaturesOptions> options)
+        {
+            _options = options.Value;
+        }
+
+        public bool IsCallStacksEnabled => _options.Enabled && ExperimentalFlags.IsCallStacksEnabled;
```
wiktork, line 20: Do you want to make this class more generic? For example, once stacks is not an experimental feature but inproc features are disabled, we'll still need to test the features much like we do now, but just not the experimental flag.

src/Tools/dotnet-monitor/Experimental/ExperimentalFlags.cs
```diff
@@ -0,0 +1,41 @@
...
+        private static bool IsFeatureEnabled(string environmentVariable)
+        {
+            string value = Environment.GetEnvironmentVariable(environmentVariable);
+
+            if (string.IsNullOrEmpty(value))
+            {
+                return false;
+            }
+
+            return !string.Equals(DisabledZeroValue, value, StringComparison.OrdinalIgnoreCase) &&
```
wiktork, line 35: Since this is off by default, maybe we should test for the positive case? From my understanding DotnetMonitor_Experimental_Feature_CallStacks=Potato would enable this feature.

wiktork (approved):

src/Tools/dotnet-monitor/Experimental/ExperimentalFlags.cs
jander-msft: Changed to enable using `1`/`true`.

wiktork (approved):

PR #2523, opened 2022-09-20
Only create default diagnostic port when running in listen mode

jander-msft:
As part of the stacks experimental feature, a default shared path configuration property was added to provide a unified default directory where all file communication and sharing can be performed while preserving the ability to configure individual file sharing aspects. This work also enabled the creation of a listening diagnostic port socket in the default shared path if the diagnostic port was not otherwise configured and "flips" dotnet-monitor into listen mode.

This change modifies the behavior of the default diagnostic port by having the user opt into the behavior by having them specify the connection mode to be listen. Hence, if the user configures the connection mode to be listen AND the default shared path is specified AND the diagnostic port name is not specified, then a default diagnostic port will be created under the default shared path. This change allows the tool to potentially allow sharing of files while it is running in connect mode without having to explicitly force connect mode (it is the default) when the default shared path is specified.

With this change, the following example is the minimum necessary for configuring all features that require file sharing across container boundaries with the tool running in listen mode:

```yml
containers:
- name: app
  image: mcr.microsoft.com/dotnet/samples:aspnetapp
  env:
  - name: DOTNET_DiagnosticPorts
    value: /diag/dotnet-monitor.sock
  - name: DotnetMonitorProfiler_RuntimeIdentifier
    value: linux-x64
  volumeMounts:
  - mountPath: /diag
    name: diagvol
- name: monitor
  image: mcr.microsoft.com/dotnet/nightly/monitor:6
  env:
  - name: DOTNETMONITOR_DiagnosticPort__ConnectionMode
    value: Listen
  - name: DOTNETMONITOR_Storage__DefaultSharedPath
    value: /diag
  volumeMounts:
  - mountPath: /diag
    name: diagvol

### DIFF
--- a/documentation/schema.json
+++ b/documentation/schema.json
@@ -1213,7 +1213,7 @@
             "null",
             "string"
           ],
-          "description": "The default path where assets will be shared between dotnet-monitor and target processes. Dumps are temporarily stored under this path or in a sub folder unless DumpTempFolder is specified. Shared libraries are stored under this path or in a sub folder unless SharedLibraryPath is specified. On non-Windows, dotnet-monitor runs in listen mode with a Unix domain socket named 'dotnet-monitor.sock' immediately under this path unless the diagnostic port is specified on the command line or the DiagnosticPort options are specified."
+          "description": "The default path where assets will be shared between dotnet-monitor and target processes. Dumps are temporarily stored under this path or in a sub folder unless DumpTempFolder is specified. Shared libraries are stored under this path or in a sub folder unless SharedLibraryPath is specified. On non-Windows platforms, a server diagnostic port is created with the name of 'dotnet-monitor.sock' immediately under this path if running in listen mode unless the diagnostic port is specified on the command line or the DiagnosticPort options are specified."
         },
         "DumpTempFolder": {
           "type": [
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/DiagnosticPortTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/DiagnosticPortTests.cs
@@ -5,13 +5,10 @@
 using Microsoft.Diagnostics.Monitoring.TestCommon;
 using Microsoft.Diagnostics.Monitoring.TestCommon.Options;
 using Microsoft.Diagnostics.Monitoring.TestCommon.Runners;
-using Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests.Fixtures;
 using Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests.Runners;
 using Microsoft.Diagnostics.Monitoring.WebApi;
-using Microsoft.Extensions.DependencyInjection;
 using System;
 using System.IO;
-using System.Net.Http;
 using System.Threading.Tasks;
 using Xunit;
 using Xunit.Abstractions;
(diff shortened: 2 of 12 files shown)

### CODE REVIEW

File: src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/DiagnosticPortTests.cs [hunk 1]
```diff
@@ -0,0 +1,121 @@
...
+            using TemporaryDirectory tempDir = new(_outputHelper);
+
+            await using MonitorCollectRunner toolRunner = new(_outputHelper);
+            toolRunner.ConnectionModeViaCommandLine = WebApi.DiagnosticPortConnectionMode.Connect;
+            toolRunner.ConfigurationFromEnvironment.SetDefaultSharedPath(tempDir.FullName);
+
+            await toolRunner.StartAsync();
+
+            string expectedMissingSocketPath = GetDefaultSharedSocketPath(tempDir.FullName);
+            Assert.False(File.Exists(expectedMissingSocketPath), $"Expected socket to not exist at '{expectedMissingSocketPath}'.");
```
kkeirstead, line 49: nit: Instead of just copy-pasting this error message for a bunch of tests, do we want to re-use it?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/DiagnosticPortTests.cs
```diff
@@ -0,0 +1,121 @@
...
+            _outputHelper = outputHelper;
+        }
+
+        /// <summary>
+        /// When setting the default shared path in connect mode, a server socket should not be created.
+        /// </summary>
+        [Fact]
+        public async Task DefaultDiagnosticPort_NotSupported_ConnectMode()
+        {
+            using TemporaryDirectory tempDir = new(_outputHelper);
```
kkeirstead, line 40: nit: I know this is test code, but from a readability standpoint it might be nice to give this a real name (e.g. `defaultSharedDir`) - that way when we reference it within the test it's more obvious what the intent of the `TemporaryDirectory` is

documentation/schema.json
```diff
@@ -1213,7 +1213,7 @@
             "null",
             "string"
           ],
-          "description": "The default path where assets will be shared between dotnet-monitor and target processes. Dumps are temporarily stored under this path or in a sub folder unless DumpTempFolder is specified. Shared libraries are stored under this path or in a sub folder unless SharedLibraryPath is specified. On non-Windows, dotnet-monitor runs in listen mode with a Unix domain socket named 'dotnet-monitor.sock' immediately under this path unless the diagnostic port is specified on the command line or the DiagnosticPort options are specified."
+          "description": "The default path where assets will be shared between dotnet-monitor and target processes. Dumps are temporarily stored under this path or in a sub folder unless DumpTempFolder is specified. Shared libraries are stored under this path or in a sub folder unless SharedLibraryPath is specified. On non-Windows platforms, a server diagnostic port is created with the name of 'dotnet-monitor.sock' immediately under this path if running in listen mode unless the diagnostic port is specified on the command line or the DiagnosticPort options are specified."
```
kkeirstead, line 1216: Should we also be updating our `configuration.md` docs to reflect this?

jander-msft: This doesn't exist in the docs yet.

wiktork:
> This change allows the tool to potentially allow sharing of files while it is running in connect mode without having to explicitly force connect mode (it is the default) when the default shared path is specified.

This makes sense, and I think not flipping the mode is probably the less confusing behavior.

wiktork (approved):

PR #2532, opened 2022-09-22
Avoid Kestrel override warning for .NET 7

jander-msft:
The change in #2481 was not sufficient to avoid the Kestrel warning about overriding the URLs for .NET 7. This change adds additional configuration to clear and restore the server URLs around the ASP.NET Core configuration and provides tests to validate that the warning does not appear.

### DIFF
--- a/src/Tools/dotnet-monitor/HostBuilder/HostBuilderHelper.cs
+++ b/src/Tools/dotnet-monitor/HostBuilder/HostBuilderHelper.cs
@@ -20,6 +20,8 @@ namespace Microsoft.Diagnostics.Tools.Monitor
 
         public static IHostBuilder CreateHostBuilder(HostBuilderSettings settings)
         {
+            string appUrls = string.Empty;
+            string serviceUrls = string.Empty;
             return new HostBuilder()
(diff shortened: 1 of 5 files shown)


src/Tools/dotnet-monitor/HostBuilder/HostBuilderHelper.cs
```diff
@@ -87,6 +89,18 @@ public static IHostBuilder CreateHostBuilder(HostBuilderSettings settings)
                         }
                     }
                 })
+                .ConfigureAppConfiguration((HostBuilderContext context, IConfigurationBuilder builder) =>
+                {
+                    // Save and clear URLs so ASP.NET app configuration cannot see them
+                    appUrls = context.Configuration[WebHostDefaults.ServerUrlsKey];
```
jander-msft, line 95: Might be able to do better here by checking which configuration source provides the value, save the source and the value, and clear the value on the source. Then restore the value on the source after configuring ASP.NET so that the perceived configuration provider is not changed. The current implementation should not affect product behavior, but it does change the configuration source of the value as seen by `config show --show-sources`.

PR #2535, opened 2022-09-23
Avoid Kestrel address override warning

jander-msft:
The change in #2481 was not sufficient to avoid the Kestrel warning about overriding the URLs for .NET 7 and may not be sufficient in other untested scenarios.

This change provides a comprehensive solution such that the URLs are blocked from being read at each applicable configuration phase so that Kestrel cannot read the real value of the Urls option. The block is lifted by default at the end of the configuration callbacks in the `HostBuilderHelper.CreateHostBuilder` method so that the Urls option is readable, allowing the `config show` command and tests to read the value as expected. The handler for the `collect` command re-enables the block before Kestrel is able to read out the Urls in its hosted service implementation, thus avoiding the address override warning printed to standard out.

New tests have been added to validate that the setting of the Urls option at various levels of configuration will bind to the specified URL correctly (by attempting to access the `/info` route) and check that the warning was not issued.


src/Tools/dotnet-monitor/Commands/CollectCommandHandler.cs
```diff
@@ -172,6 +172,14 @@ private static IHostBuilder ConfigureServices(this IHostBuilder builder, AuthCon
...
                 services.ConfigureInProcessFeatures(context.Configuration);
                 services.AddSingleton<IInProcessFeatures, InProcessFeatures>();
+            })
+            .ConfigureContainer((HostBuilderContext context, IServiceCollection services) =>
+            {
+                ServerUrlsBlockingConfigurationManager manager =
+                    context.Properties[typeof(ServerUrlsBlockingConfigurationManager)] as ServerUrlsBlockingConfigurationManager;
+
+                // Block reading of the Urls option so that Kestrel is unable to read it from the composed configuration.
+                manager.IsBlocking = true;
```
wiktork, line 182: can we null check this?

src/Tools/dotnet-monitor/HostBuilder/HostBuilderHelper.cs
```diff
@@ -133,12 +150,38 @@ public static IHostBuilder CreateHostBuilder(HostBuilderSettings settings)
...
+                    // The settings.json, key-per-file, and DOTNETMONITOR_* environment variables were added as part
+                    // of the app configuration phase. Before the phase is completed, add a configuration source that will
+                    // conditionally block reading of the Urls options so that they are not picked up by future Kestrel
+                    // configuration callbacks.
+                    builder.Add(new ServerUrlsBlockingConfigurationSource(manager));
+                })
+                .ConfigureContainer((HostBuilderContext context, IServiceCollection services) =>
+                {
+                    // Container configuration is the last phase of building the host before the service provider is constructed.
+                    // At this point, all configuration callbacks have been executed. Lift the block on the Urls option so that
```
wiktork, line 182: Is this order something we can rely on long term? For example what if another phase is added in the future?

jander-msft: I believe it is something we can fairly rely on. These phases haven't changed since .NET Core 2.1. All of the frameworks (runtime and aspnetcore) add all of their configuration sources during the host configuration phase. We add configuration sources to both host configuration and app configuration phases. Hence why the ServerUrlsBlockingConfigurationSource is added at the end of each of these phases. Since configurations sources cannot be added during service and container phases, don't think this will pop up unexpectedly. I've also added tests to help check if this occurs.

src/Tools/dotnet-monitor/ConfigurationExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Extensions.Configuration;
+using System.Linq;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    internal static class ConfigurationExtensions
+    {
+        public static bool TryGetProviderAndValue(this IConfiguration configuration, string key, out IConfigurationProvider provider, out string value)
```
wiktork, line 12: I'm assuming some of these will go away after the merge from main?

jander-msft: Yes, I rebased the changes and they're gone because they were included in #2550

wiktork (dismissed):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/KestrelTests.cs
```diff
@@ -0,0 +1,101 @@
...
+                // it will cause dotnet-monitor to no bind correctly and the /info route
+                // check will fail.
+                runner.DotNetUrls = "dotnet_invalid";
+                runner.AspNetCoreUrls = "aspnetcore_invalid";
+
+                runner.DotNetMonitorUrls = url;
+            });
+        }
+
+        private async Task ExecuteNoOverridWarning(Action<MonitorCollectRunner, string> configure = null)
```
kkeirstead, line 76: nit: `ExecuteNoOverrideWarning`

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/KestrelTests.cs
```diff
@@ -0,0 +1,101 @@
...
+            });
+        }
+
+        [Fact]
+        public Task UrlBinding_NoOverrideWarning_AspNetCore()
+        {
+            return ExecuteNoOverridWarning((runner, url) =>
+            {
+                // This should be overriden by the ASPNETCORE_Urls entry. If it is not,
+                // it will cause dotnet-monitor to no bind correctly and the /info route
```
kkeirstead, line 54:
```suggestion
                // it will cause dotnet-monitor to not bind correctly and the /info route
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/KestrelTests.cs
```diff
@@ -0,0 +1,101 @@
...
+            });
+        }
+
+        [Fact]
+        public Task UrlBinding_NoOverrideWarning_DotNetMonitor()
+        {
+            return ExecuteNoOverridWarning((runner, url) =>
+            {
+                // These should be overriden by the DotnetMonitor_Urls entry. If it is not,
+                // it will cause dotnet-monitor to no bind correctly and the /info route
```
kkeirstead, line 67:
```suggestion
                // it will cause dotnet-monitor to not bind correctly and the /info route
```

src/Tools/dotnet-monitor/HostBuilder/HostBuilderHelper.cs
```diff
@@ -146,12 +163,38 @@ public static IHostBuilder CreateHostBuilder(HostBuilderSettings settings)
...
+                    // the option my be read from configuration by default.
```
kkeirstead, line 196:
```suggestion
                    // the option may be read from configuration by default.
```

src/Tools/dotnet-monitor/HostBuilder/ServerUrlsBlockingConfigurationProvider.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public ServerUrlsBlockingConfigurationProvider(ServerUrlsBlockingConfigurationManager manager)
+        {
+            _manager = manager;
+        }
+
+        public override void Set(string key, string value)
+        {
+            // Overridden to prevent set of data since this provider does not
+            // provider any real data from any source.
```
kkeirstead, line 24:
```suggestion
            // provide any real data from any source.
```

wiktork (approved):

PR #2550, opened 2022-09-26
Unwrap ChainedConfigurationProvider to get real provider

jander-msft:
Currently, all of the configuration that is added as part of host configuration (this would largely be anything from the command line, the DOTNET_* variables, the ASPNETCORE_* variables) will have a reported source of `ChainedConfigurationProvider` when executing `config show --show-sources`. This is not particular helpful in understanding from where these values are provided.

This change effectively unwraps the occurrence of `ChainedConfigurationProvider` instances in the configuration writing algorithm to search within the chained configuration for the real configuration sources. Additionally, values that are provided from the command line have been changed to be represented by the `CommandLineConfigurationProvider` instead of by `MemoryConfigurationProvider`.

### DIFF
--- a/src/Tools/dotnet-monitor/HostBuilder/HostBuilderHelper.cs
+++ b/src/Tools/dotnet-monitor/HostBuilder/HostBuilderHelper.cs
@@ -242,7 +242,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor
 
         private static string FormatCmdLineArgument(string key, string value)
         {
-            return $"{key}={value}";
+            return FormattableString.Invariant($"{key}={value}");
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/Tools/dotnet-monitor/HostBuilder/HostBuilderHelper.cs
```diff
@@ -164,47 +177,72 @@ public static AuthConfiguration CreateAuthConfiguration(bool noAuth, bool tempAp
...
+        private static void AddUrlsArguments(List<string> arguments, HostBuilderSettings settings)
+        {
+            arguments.Add(FormatCmdLineArgument(
+                WebHostDefaults.ServerUrlsKey,
+                ConfigurationHelper.JoinValue(settings.Urls ?? Array.Empty<string>())));
+        }
+
+        private static string FormatCmdLineArgument(string key, string value)
+        {
+            return $"{key}={value}";
```
wiktork, line 245: Invariant?

wiktork (dismissed):

wiktork (approved):

PR #2557, opened 2022-09-27
Add support for a StoppingEvent on CollectTrace action

schmittjoseph:
###### Summary
Add support for stopping a `CollectTrace` action when it observes a specific task name & opcode from an event provider. An optional payload filter can also be specified, which supports a subset of all possible fields on the specified event's payload. 

###### Release Notes Entry
Add the option to specify a `StoppingEvent` on the `CollectTrace` action. The trace will be stopped once either the duration is reached or the specified event occurs.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/EventMonitoringPassthroughStream.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/EventMonitoringPassthroughStream.cs
@@ -85,17 +85,21 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
         /// Start processing the event stream, monitoring it for the requested event and transferring its data to the specified destination stream.
         /// This will continue to run until the event stream is complete or a stop is requested, regardless of if the requested event has been observed.
         /// </summary>
-        /// <param name="token">The cancellation token. It can only be signaled before processing has been started. After that point <see cref="DisposeAsync"/> should be called to stop processing.</param>
+        /// <param name="token">The cancellation token.</param>
         /// <returns></returns>
         public Task ProcessAsync(CancellationToken token)
         {
             return Task.Run(() =>
             {
                 _eventSource = new EventPipeEventSource(this);
+                token.ThrowIfCancellationRequested();
+                using IDisposable registration = token.Register(() => _eventSource.Dispose());
+
                 _eventSource.Dynamic.AddCallbackForProviderEvent(_providerName, _eventName, TraceEventCallback);
 
                 // The EventPipeEventSource will drive the transferring of data to the destination stream as it processes events.
                 _eventSource.Process();
+                token.ThrowIfCancellationRequested();
             }, token);
         }
 
@@ -147,7 +151,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
             // If there's no payload filter, there's nothing to do.
             if (_payloadFilter == null || _payloadFilter.Count == 0)
             {
-                _payloadFilterIndexCache = new(capacity: 0);
+                _payloadFilterIndexCache = new Dictionary<int, string>(capacity: 0);
                 return true;
             }
 
@@ -242,8 +246,8 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
         public override void CopyTo(Stream destination, int bufferSize) => throw new NotSupportedException();
         public override Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken) => throw new NotSupportedException();
 
-        public override void Flush() => _sourceStream.Flush();
-        public override Task FlushAsync(CancellationToken cancellationToken) => _sourceStream.FlushAsync(cancellationToken);
+        public override void Flush() => _destinationStream.Flush();
+        public override Task FlushAsync(CancellationToken cancellationToken) => _destinationStream.FlushAsync(cancellationToken);
 
         public override async ValueTask DisposeAsync()
         {
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/TraceUtilities.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/TraceUtilities.cs
@@ -141,6 +141,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
             if (stoppingEventHitSource.Task.IsCompleted)
             {
                 await pipeProcessor.StopAsync(token);
+                await pipelineRunTask;
             }
         }
     }


src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/TraceUtilities.cs
```diff
@@ -97,5 +99,48 @@ public static async Task CaptureTraceAsync(TaskCompletionSource<object> startCom
...
+                    outputStream,
+                    DefaultBufferSize,
+                    callOnEventOnlyOnce: true,
+                    leaveDestinationStreamOpen: true /* We do not have ownership of the outputStream */);
+
+                await eventMonitoringStream.ProcessAsync(token);
+            });
+
+            Task pipelineRunTask = pipeProcessor.RunAsync(token);
+            await Task.WhenAny(pipelineRunTask, stoppingEventHitSource.Task);
```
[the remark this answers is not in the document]
jander-msft: Alternatively, tack on an `.Unwrap()` to the end.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/CollectTraceTests.cs
```diff
@@ -0,0 +1,155 @@
...
+                            new EventPipeProvider[] {
+                                new EventPipeProvider()
+                                {
+                                    Name = EventProviderName,
+                                    Keywords = "-1"
+                                }
+                            },
+                            EgressProvider, options =>
+                            {
+                                options.Duration = duration ?? TimeSpan.Parse(ActionOptionsConstants.Duration_MaxValue);
```
jander-msft, line 127: Let's use a duration that is shorter than collection rule completion timeout but long enough where the positive validation tests can detect the stopping event.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/CollectTraceTests.cs
```diff
@@ -0,0 +1,155 @@
...
+                });
+
+            string[] files = Directory.GetFiles(tempDirectory.FullName, "*.nettrace", SearchOption.TopDirectoryOnly);
+            string traceFile = Assert.Single(files);
+            await ValidateNettraceFile(traceFile);
+        }
+
+        private async Task ValidateNettraceFile(string filePath)
+        {
+            byte[] expectedMagicToken = Encoding.UTF8.GetBytes("Nettrace");
```
jander-msft, line 146: I think it would be worthwhile to use EventPipeEventSource to validate that the expected event is in the trace. Have `ValidateNettraceFile` return a `Task<bool>`, update `StopOnEventTestCore` to have a `bool expectEvevent` parameter, and it assert whether the event existing or not is expected. Let's not rely on TaskCanceledException due to the collection rule timeout occurring since that is ambiguous as to why the collection rule didn't finish in time.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/TraceEventsScenario.cs
```diff
@@ -0,0 +1,76 @@
...
+            context.ExitCode = await ScenarioHelpers.RunScenarioAsync(async logger =>
+            {
+                TaskCompletionSource<object> stopGeneratingEvents = new();
+                Task eventEmitterTask = Task.Run(() =>
+                {
+                    Random random = new();
+                    while (!stopGeneratingEvents.Task.IsCompleted)
+                    {
+                        TestScenarioEventSource.Log.RandomNumberGenerated(random.Next());
+                        Task.Delay(100, context.GetCancellationToken());
```
jander-msft, line 56:
```suggestion
                        await Task.Delay(TimeSpan.FromMilliseconds(100), context.GetCancellationToken());
```

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/TraceEventsScenario.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            string[] acceptableCommands = new string[]
+            {
+                TestAppScenarios.TraceEvents.Commands.EmitUniqueEvent,
+                TestAppScenarios.TraceEvents.Commands.ShutdownScenario
+            };
+
+            context.ExitCode = await ScenarioHelpers.RunScenarioAsync(async logger =>
+            {
+                TaskCompletionSource<object> stopGeneratingEvents = new();
```
jander-msft, line 49: This is currently not used in an async way, so it's probably better to use a `ManualResetEvent` and `WaitOne(0)` so as not to create the illusion that something async is happening.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/TraceEventsScenario.cs
```diff
@@ -0,0 +1,76 @@
...
+                Task eventEmitterTask = Task.Run(() =>
+                {
+                    Random random = new();
+                    while (!stopGeneratingEvents.Task.IsCompleted)
+                    {
+                        TestScenarioEventSource.Log.RandomNumberGenerated(random.Next());
+                        Task.Delay(100, context.GetCancellationToken());
+                    }
+                });
```
jander-msft, line 58:
```suggestion
                }, context.GetCancellationToken());
```

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/TraceEventsScenario.cs
```diff
@@ -0,0 +1,76 @@
...
+            string[] acceptableCommands = new string[]
+            {
+                TestAppScenarios.TraceEvents.Commands.EmitUniqueEvent,
+                TestAppScenarios.TraceEvents.Commands.ShutdownScenario
+            };
+
+            context.ExitCode = await ScenarioHelpers.RunScenarioAsync(async logger =>
+            {
+                TaskCompletionSource<object> stopGeneratingEvents = new();
+                Task eventEmitterTask = Task.Run(() =>
```
jander-msft, line 50:
```suggestion
                Task eventEmitterTask = Task.Run(async () =>
```

src/Microsoft.Diagnostics.Monitoring.WebApi/LoggingExtensions.cs
```diff
@@ -51,6 +52,18 @@ internal static class LoggingExtensions
                 logLevel: LogLevel.Warning,
                 formatString: Strings.LogFormatString_DefaultProcessUnexpectedFailure);
 
+        private static readonly Action<ILogger, string, string, Exception> _stoppingTraceEventHit =
+            LoggerMessage.Define<string, string>(
+                eventId: new EventId(8, "StoppingTraceEventHit"),
+                logLevel: LogLevel.Information,
```
jander-msft, line 58:
```suggestion
                logLevel: LogLevel.Debug,
```

I don't think we need to log this normally since the action executor should be logging the completion of the action.

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectTraceAction.cs
```diff
@@ -90,7 +94,19 @@ public CollectTraceAction(IServiceProvider serviceProvider, IEndpointInfo endpoi
                     async (outputStream, token) =>
                     {
                         using IDisposable operationRegistration = _operationTrackerService.Register(EndpointInfo);
-                        await TraceUtilities.CaptureTraceAsync(startCompletionSource, EndpointInfo, configuration, duration, outputStream, token);
+                        if (null != stoppingEvent)
+                        {
+                            ILogger<CollectTraceAction> logger = _serviceProvider
+                                .GetRequiredService<ILoggerFactory>()
+                                .CreateLogger<CollectTraceAction>();
+                            using var _ = logger.BeginScope(scope);
```
jander-msft, line 102: I think this is adding a duplicated scope since the egress operation should be logging this already.

src/Microsoft.Diagnostics.Monitoring.WebApi/EventMonitoringPassthroughStream.cs
```diff
@@ -0,0 +1,254 @@
...
+            _onEvent = onEvent;
+            _onPayloadFilterMismatch = onPayloadFilterMismatch;
+            _sourceStream = sourceStream;
+            _payloadFilter = payloadFilter;
+            _callOnEventOnlyOnce = callOnEventOnlyOnce;
+
+            // Wrap a buffered stream around the destination stream
+            // to avoid slowing down the event processing with the data
+            // passthrough unless there is significant pressure.
+            _destinationStream = new BufferedStream(
```
wiktork, line 77: You may want to indicate that the destination stream has to be full duplex. For example, if this was a MemoryStream, the reads and writes would likely conflict with each other.

documentation/configuration.md
```diff
@@ -1421,6 +1421,7 @@ An action that collects a trace of the process that the collection rule is targe
 | `BufferSizeMegabytes` | int | false | The size (in megabytes) of the event buffer used in the runtime. If the event buffer is filled, events produced by event providers may be dropped until the buffer is cleared. Increase the buffer size to mitigate this or pair down the list of event providers, keywords, and level to filter out extraneous events. Only applies when `Providers` is specified. | `256` | `1` | `1024` |
 | `Duration` | TimeSpan? | false | The duration of the trace operation. | `"00:00:30"` (30 seconds) | `"00:00:01"` (1 second) | `"1.00:00:00"` (1 day) |
 | `Egress` | string | true | The named [egress provider](egress.md) for egressing the collected trace. | | | |
+| `StoppingEvent` | [TraceEventFilter](api/definitions.md#traceeventfilter)? | false | The event to watch for while collecting the trace, and once either the event is hit or the `Duration` is reached the trace will be stopped. This can only be specified if `Providers` is set. | `null` | | |
```
wiktork, line 1424: Is the Providers restriction necessary?

src/Tools/dotnet-monitor/CollectionRules/Actions/CollectTraceAction.cs
```diff
@@ -78,8 +81,9 @@ public CollectTraceAction(IServiceProvider serviceProvider, IEndpointInfo endpoi
                     EventPipeProvider[] optionsProviders = Options.Providers.ToArray();
                     bool requestRundown = Options.RequestRundown.GetValueOrDefault(CollectTraceOptionsDefaults.RequestRundown);
                     int bufferSizeMegabytes = Options.BufferSizeMegabytes.GetValueOrDefault(CollectTraceOptionsDefaults.BufferSizeMegabytes);
-
```
wiktork, line 81: Note that if we end up going the client way for this work, we'll need to add this configuration into the /trace post endpoint as well. That's probably desirable either way.

documentation/configuration.md
jander-msft: If you wanted to allow the stopping event + profiler scenario, you can look up the effective provider list by calling https://github.com/dotnet/dotnet-monitor/blob/main/src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/TraceUtilities.cs#L25

src/Microsoft.Diagnostics.Monitoring.WebApi/EventMonitoringPassthroughStream.cs
```diff
@@ -0,0 +1,254 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    /// <summary>
+    /// A stream that can monitor an event stream which is compatible with <see cref="EventPipeEventSource"/>
+    /// for a specific event while also passing along the event data to a destination stream.
+    /// </summary>
+    public class EventMonitoringPassthroughStream : Stream
```
wiktork, line 19: nit: Let's default classes to internal sealed unless there's a reason otherwise.

src/Microsoft.Diagnostics.Monitoring.WebApi/EventMonitoringPassthroughStream.cs
```diff
@@ -0,0 +1,254 @@
...
+        /// </summary>
+        public void StopMonitoringForEvent()
+        {
+            _eventSource?.Dynamic.RemoveCallback<TraceEvent>(TraceEventCallback);
+        }
+
+        /// <summary>
+        /// Stops processing the event data, data will no longer be written to the provided destination stream.
+        /// </summary>
+        public void StopProcessing()
```
wiktork, line 113: Do we actually call this externally?

src/Microsoft.Diagnostics.Monitoring.WebApi/EventMonitoringPassthroughStream.cs
```diff
@@ -0,0 +1,254 @@
...
+                return true;
+            }
+
+            if (_payloadFilter == null || _payloadFilter.Count == 0)
+            {
+                _payloadFilterIndexCache = new(capacity: 0);
+                return true;
+            }
+
+            if (obj.PayloadNames.Length < _payloadFilter.Count)
```
wiktork, line 156: Please add comment regarding this condition

src/Microsoft.Diagnostics.Monitoring.WebApi/EventMonitoringPassthroughStream.cs
```diff
@@ -0,0 +1,254 @@
...
+            {
+                if (_payloadFilter.TryGetValue(obj.PayloadNames[i], out string payloadValue))
+                {
+                    payloadFilterCache.Add(i, payloadValue);
+                }
+            }
+
+            if (_payloadFilter.Count != payloadFilterCache.Count)
+            {
+                return false;
```
wiktork, line 172: Consider returning an HRESULT and raising it in the callback.

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/CollectTraceOptions.Validate.cs
```diff
@@ -27,9 +27,20 @@ IEnumerable<ValidationResult> IValidatableObject.Validate(ValidationContext vali
...
 
             if (hasProfile)
             {
+                if (hasStoppingEvent)
+                {
+                    results.Add(new ValidationResult(
+                        string.Format(
+                            CultureInfo.InvariantCulture,
+                            Strings.ErrorMessage_TwoFieldsCannotBeSpecified,
+                            nameof(Profile),
```
wiktork, line 40: As mentioned, let's relax this in a future pr.

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/CollectTraceOptions.cs
```diff
@@ -63,5 +63,10 @@ internal sealed partial record class CollectTraceOptions : BaseRecordOptions, IE
         [ValidateEgressProvider]
```
wiktork, line 63: Eventually we'll need to change the max here, but I don't think it has to be part of this work.

src/Microsoft.Diagnostics.Monitoring.WebApi/EventMonitoringPassthroughStream.cs
wiktork, line 19: If we were to migrate this code, is EventMonitoringPassthroughStream & StreamLeaveOpenWrapper sufficient?

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/TraceUtilities.cs
```diff
@@ -97,5 +99,49 @@ public static async Task CaptureTraceAsync(TaskCompletionSource<object> startCom
...
+                () => stoppingEventHitSource.TrySetCanceled(token));
+
+            await using EventTracePipeline pipeProcessor = new(client, new EventTracePipelineSettings
+            {
+                Configuration = configuration,
+                Duration = timeout,
+            },
+            async (eventStream, token) =>
+            {
+                startCompletionSource?.TrySetResult(null);
```
wiktork, line 118: This is just a thought, not something we need to implement now, but it's very likely that the delegate code to create a new EventMonitoringPassThroughStream, pass the respective parameters, and stop tracing based on an event is going to be very boilerplate. I'm wondering if we can encapsulate that code as something the user can just pass directly into the EventTracePipeline rather than having each user re-write the boilerplate code.

wiktork (approved):

src/Microsoft.Diagnostics.Monitoring.WebApi/EventMonitoringPassthroughStream.cs
```diff
@@ -0,0 +1,256 @@
...
+                    ? new StreamLeaveOpenWrapper(destinationStream)
+                    : destinationStream,
+                bufferSize);
+        }
+
+        /// <summary>
+        /// Start processing the event stream, monitoring it for the requested event and transferring its data to the specified destination stream.
+        /// This will continue to run until the event stream is complete or a stop is requested, regardless of if the requested event has been observed.
+        /// </summary>
+        /// <param name="token">The cancellation token. It can only be signaled before processing has been started. After that point <see cref="DisposeAsync"/> should be called to stop processing.</param>
```
jander-msft, line 88: Can you register a callback on the cancellation token within the `Task.Run` anonymous function that disposes the event source and then call `token.ThrowIfCancellationRequested` after the call to `_eventSource.Process` returns? It looks like `EventPipeEventSource ` disposal is idempotent. This would then match the expected semantics for async methods.

jander-msft: You may want to add a `token.ThrowIfCancellationRequested` call after the construction of the `EventPipeEventSource` as well. The ctor frequently hangs for several seconds waiting for some events from the runtime.

src/Microsoft.Diagnostics.Monitoring.WebApi/EventMonitoringPassthroughStream.cs
```diff
@@ -0,0 +1,256 @@
...
+        {
+            if (_payloadFilterIndexCache != null)
+            {
+                return true;
+            }
+
+            // If there's no payload filter, there's nothing to do.
+            if (_payloadFilter == null || _payloadFilter.Count == 0)
+            {
+                _payloadFilterIndexCache = new(capacity: 0);
```
jander-msft, line 150: nit: Can you explicitly use the `Dictionary<int, string>` type after the `new` keyword? It's not obvious what is being constructed from this context when reading the code.

src/Microsoft.Diagnostics.Monitoring.WebApi/EventMonitoringPassthroughStream.cs
```diff
@@ -0,0 +1,256 @@
...
+
+        public override long Seek(long offset, SeekOrigin origin) => throw new NotSupportedException();
+        public override void SetLength(long value) => throw new NotSupportedException();
+        public override void Write(byte[] buffer, int offset, int count) => throw new NotSupportedException();
+
+        public override void CopyTo(Stream destination, int bufferSize) => throw new NotSupportedException();
+        public override Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken) => throw new NotSupportedException();
+
+        public override void Flush() => _sourceStream.Flush();
+        public override Task FlushAsync(CancellationToken cancellationToken) => _sourceStream.FlushAsync(cancellationToken);
```
jander-msft, line 246: Same comment about `Flush`

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/TraceUtilities.cs
```diff
@@ -97,5 +99,49 @@ public static async Task CaptureTraceAsync(TaskCompletionSource<object> startCom
...
+
+                await eventMonitoringStream.ProcessAsync(token);
+            });
+
+            Task pipelineRunTask = pipeProcessor.RunAsync(token);
+            await Task.WhenAny(pipelineRunTask, stoppingEventHitSource.Task).Unwrap();
+
+            if (stoppingEventHitSource.Task.IsCompleted)
+            {
+                await pipeProcessor.StopAsync(token);
```
jander-msft, line 143: There's possibly a slight race condition here where the stopping event is noticed, the trace starts rundown due to the duration backstop, and calling trying to stop the trace here if the duration expires between when `stoppingEventHitSource` and this `StopAsync` are executed. In actual use, this is fine. However, the tests that are expecting no rundown to be produced when stopping due to the stopping event may fail randomly if this condition is encountered. I'm not sure if there is anything we can do from a functional test standpoint.

jander-msft (approved):

PR #2568, opened 2022-09-28
Add "Was This Helpful" Buttons (Part 1)

kkeirstead:
Part 1 of 2 for integrating "Was This Helpful" buttons - this PR adds in the actual buttons at the top of the markdown files, as well as the source image that renders in the markdown. Certain markdown files (such as the releases) were omitted.


documentation/api-key-format.md
```diff
@@ -1,3 +1,6 @@
+
+[<img src=/images/WasThisHelpful.png width="200"/>](https://www.research.net/r/DGDQWXH?src=api-key-format)
```
jander-msft, line 2:
```suggestion
[<img src=/images/WasThisHelpful.png width="200"/>](https://www.research.net/r/DGDQWXH?src=%2Fdocumentation%2Fapi-key-format.md)
```

Is it possible to get the full subpath of the document so we can avoid collisions and be able to differentiate, for example, which README file they are leaving feedback on?

kkeirstead: I should be able to do that - I currently strip away the path in favor of just keeping the file name so that the parameter wouldn't look so unwieldy, but then we do have a couple collisions.

kkeirstead:
Joe suggested switching over to using text instead of images for the buttons for accessibility/scaling - I've switched everything over, and we can revert if we'd rather stick with the initial approach. If we decide to use the text-based approach, then I'll yank out the images directory and the WasThisHelpful image.

jander-msft:
> Joe suggested switching over to using text instead of images for the buttons for accessibility/scaling - I've switched everything over, and we can revert if we'd rather stick with the initial approach. If we decide to use the text-based approach, then I'll yank out the images directory and the WasThisHelpful image.

I think the textual link works well. Let's remove the image.

documentation/release-process.md
```diff
@@ -1,3 +1,6 @@
+
+### Was this documentation helpful? [Share feedback](https://www.research.net/r/DGDQWXH?src=documentation%2Frelease-process)
```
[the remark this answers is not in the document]
kkeirstead: Could you clarify this comment? I'm not sure what you mean

jander-msft (approved):

PR #2600, opened 2022-09-30
Fallback to the system's copy of dotnet when running tests under Visual Studio

schmittjoseph:
Currently if you have a cloned version of this repo and exclusively build & test it with Visual Studio, a portion of the tests will fail, as they rely on the repo's copy of `dotnet` which is only hydrated when using the scripts at the root of the repo to build and test.

Update our testing code to fallback to the system's copy of dotnet when testing under VisualStudio and the repo's copy of `dotnet` has not yet been hydrated.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/DotNetHost.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/DotNetHost.cs
@@ -35,9 +35,18 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon
             }
 
             string absoluteExecutablePath = Path.GetFullPath(Path.Combine(dotnetDirPath, ExeName));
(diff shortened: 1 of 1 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/DotNetHost.cs
```diff
@@ -33,7 +33,15 @@ public static string GetPath(Architecture? arch = null)
                 // e.g. Append "\x86" to the path
                 dotnetDirPath = Path.Combine(dotnetDirPath, arch.Value.ToString("G").ToLowerInvariant());
             }
-            return Path.GetFullPath(Path.Combine(dotnetDirPath, ExeName));
+
+            string absoluteExecutablePath = Path.GetFullPath(Path.Combine(dotnetDirPath, ExeName));
+            if (!File.Exists(absoluteExecutablePath) && RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("VSAPPIDDIR")))
```
jander-msft, line 38: Can you add comments as to why this fallback is used and why you chose these specific conditions?

PR #2601, opened 2022-10-01
Run apps with test frameworks by using customized runtimeconfig.json

jander-msft:
For a part of the .NET 6 and all of the .NET 7 cycles, the tests would execute dotnet-monitor and the test applications using the runtimes installed via the SDK. The repository also brings in newer framework versions via dependency flow, but since these are non-final versions, the runtime automatically excludes them from consideration for execution. The unit test apps were built with the RuntimeFrameworkVersion properties to force the newer frameworks to be used, but this is not an option for any assemblies that are shipped. Finally, the `--fx-version` argument that was used previously did not work for applications that have more than one framework e.g. ASP.NET Core applications.

The change creates a general-purpose solution by which a test runtimeconfig.json file is generated at build time and used by the test infrastructure if the dotnet host was restored within the repository (which is the case for CI builds and command line builds). The customized runtimeconfig.json file contains versions for the NETCore and AspNetCore frameworks provided by depedency flow. When run with this configuration file, the runtime will select the specified framework versions instead of the versions that happen to ship with the SDK.

By using the newer framework versions, the builds will be able to catch whether newer runtime versions will work and allow us to make updates to account for any changes before they are shipped.


src/Directory.Build.targets
```diff
@@ -35,4 +35,83 @@
...
+  <!--
+    Typically, the SDK will generate a runtimeconfig.json file with frameworks that it natively has included
+    with itself. This does not allow testing with newer runtime versions brought in through dependency updates.
+    The unit test apps can be rolled forward using RuntimeFrameworkVersion at build time, however we do not want
+    to do this with any binaries that are shipped e.g. dotnet-monitor. Additionally the runtime-fx argument to
+    dotnet does not work for any application that has more than one framework, such as ASP.NET apps.
+
+    The general purpose solution is to generate a custom runtimeconfig[.test].json file that instructs the
+    runtime which framework versions to use. The framework versions in this file are the ones provided by the
+    depedency flow system.
```
kkeirstead, line 54:
```suggestion
    dependency flow system.
```

src/Directory.Build.targets
```diff
@@ -35,4 +35,83 @@
...
+    with itself. This does not allow testing with newer runtime versions brought in through dependency updates.
+    The unit test apps can be rolled forward using RuntimeFrameworkVersion at build time, however we do not want
+    to do this with any binaries that are shipped e.g. dotnet-monitor. Additionally the runtime-fx argument to
+    dotnet does not work for any application that has more than one framework, such as ASP.NET apps.
+
+    The general purpose solution is to generate a custom runtimeconfig[.test].json file that instructs the
+    runtime which framework versions to use. The framework versions in this file are the ones provided by the
+    dependency flow system.
+  -->
+  <Target Name="GenerateTestRuntimeConfigurationFiles"
```
wiktork, line 56: I think we covered this, but I want to make sure that it's still possible to run unit tests via the unit test explorer

jander-msft: Yes, that is correct. The runtimeconfig.test.json file is used only if the dotnet host was restored to `<repoRoot>/.dotnet/`. This means that if you've restored the dotnet host and then sync your repo to get newer runtime versions, you will have to restore those runtimes again before building in Visual Studio. If this becomes a problem, we can either scope it down so that the runtimeconfig.test.json file is only used outside of Visual Studio or we can dynamically restore the missing runtimes.

PR #2637, opened 2022-10-06
Split process stop from disposal and use disposal helpers

jander-msft:
This change splits out the behavior of stopping the process runners from disposing the process runners. This allows consistent access to process information after the process has stopped but before it is disposed (which the process information, such as exit code, are no longer available). As it was before, owners of the process runner are NOT required to call `StopAsync` before calling `DisposeAsync`.

The `AppRunner.ExecuteAsync` extension methods have been updated to call `StopAsync` instead of `DisposeAsync` in order to maintain the symmetry with the method starting the runner rather than creating and owning the runner. Callers are now responsible for disposing the runners.

Additional changes were made to unify the disposal patterns to use the common disposal helpers.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunner.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunner.cs
@@ -32,7 +32,9 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests.Runners
 
         private readonly LoggingRunnerAdapter _adapter;
 
-        private bool _isDisposed;
+        private readonly TemporaryDirectory _tempDir;
+
+        private long _disposedState;
 
         /// <summary>
         /// Sets configuration values via environment variables.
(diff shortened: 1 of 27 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunner.cs
```diff
@@ -32,7 +32,9 @@ internal class MonitorRunner : IAsyncDisposable
 
         private readonly LoggingRunnerAdapter _adapter;
 
-        private bool _isDisposed;
+        private readonly TemporaryDirectory _tempDir;
+
+        private long _disposedState;
```
wiktork, line 37: Consider something like AsyncDisposable that has DisposeAsync and DisposeAsyncCore. The non-virtual DisposeAync can do checks against the state flag and DisposeAsyncCore can be virtual.

wiktork (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunner.cs
jander-msft: This might be a good candidate for a [source generator](https://learn.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/source-generators-overview) instead of creating a base class. I'm going to leave these changes as-is and log an issue to consider each solution.

PR #2748, opened 2022-10-20
Updates to `feature/egressExtension` Part 1

kkeirstead:
###### Summary

Part 1 of my updates to the `feature/egressExtension` branch - I pulled out some things to keep this PR smaller, as well as areas that I'm still working through. I'll leave comments throughout the code to explain different things that I think are worth pointing out or that I've cut out (for now). Builds on top of https://github.com/dotnet/dotnet-monitor/pull/1953.


src/Extensions/AzureBlobStorage/EgressArtifactSettings.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// <summary>
+        /// The metadata of the blob to be created.
+        /// </summary>
+        public Dictionary<string, string> Metadata { get; set; }
+            = new Dictionary<string, string>(StringComparer.Ordinal);
+
+        /// <summary>
+        /// Custom metadata of the blob to be created.
+        /// </summary>
+        public Dictionary<string, string> CustomMetadata { get; }
```
kkeirstead, line 23: Not currently being populated (I'll mention more about this in another comment)

src/Microsoft.Diagnostics.Monitoring.Options/ExtensionEgressProviderOptions.cs
```diff
@@ -5,15 +5,14 @@
...
-using System.Text.Json.Serialization;
 
 namespace Microsoft.Diagnostics.Tools.Monitor.Egress
 {
     /// <summary>
     /// Egress provider options for external egress providers.
     /// </summary>
     internal sealed partial class ExtensionEgressProviderOptions : 
-        Dictionary<string, string>,
+        Dictionary<string, object>,
```
kkeirstead, line 15: This isn't implemented in this PR, but in the initial iteration of this design we were assuming that we could treat all of our extension's configuration as a `Dictionary<string, string>`. The addition of the Azure provider's `Metadata` (used for custom metadata) breaks this assumption, and will require changes to how we pass the configuration to the extension. To simplify this PR, I am not currently passing the `Metadata` section of configuration, but this will be fixed in a future PR.

src/Tools/dotnet-monitor/Egress/Configuration/EgressProviderConfigureNamedOptions.cs
```diff
@@ -34,7 +34,23 @@ public void Configure(string name, TOptions options)
                 IConfigurationSection providerOptionsSection = providerTypeSection.GetSection(name);
                 if (providerOptionsSection.Exists())
                 {
-                    providerOptionsSection.Bind(options);
+                    if (options is ExtensionEgressProviderOptions extensionOptions)
```
kkeirstead, line 37: This is where we'll need to handle `Metadata` (and anything else that doesn't have a value for a child).

src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
```diff
@@ -28,15 +29,42 @@ public FolderExtensionRepository(IFileProvider fileSystem, ILoggerFactory logger
 
         public override bool TryFindExtension(string extensionName, out IExtension extension)
         {
-            IDirectoryContents extensionDir = _fileSystem.GetDirectoryContents(extensionName);
+            string extensionPath = string.Empty;
+
+            bool isDotnetToolsLocation = _targetFolder == HostBuilderSettings.ExtensionDirectoryPath;
+
+            // Special case -> need to look in particular path for dotnet tool installation
```
kkeirstead, line 36: This logic is for the `dotnet tool install` scenario, so that users don't need to manually place an extension downloaded from Nuget into the right directory. This works in my limited testing (building a nuget package locally, and then calling `dotnet tool install --add-source...` on it), but I think this area may need some more refinement going forward.

src/Tools/dotnet-monitor/LoggingEventIds.cs
```diff
@@ -66,10 +65,7 @@ internal enum LoggingEventIds
         LoadingProfiler = 53,
         SetEnvironmentVariable = 54,
         GetEnvironmentVariable = 55,
-        MonitorApiKeyNotConfigured = 56,
-        QueueDoesNotExist = 57,
-        QueueOptionsPartiallySet = 58,
-        WritingMessageToQueueFailed = 59,
+        MonitorApiKeyNotConfigured = 56, // Note the gap - from removing things related to Azure egress
```
kkeirstead, line 68: Will we want to update all of these numbers to be in-order again? I haven't for now, but we have several gaps left behind by the logging specific to the Azure extension.

src/Extensions/AzureBlobStorage.sln
```diff
@@ -0,0 +1,25 @@
+﻿
```
jander-msft, line 1: Remove this and add the AzureBlobStorage project to the dotnet-monitor.sln file if it already doesn't exist.

src/Extensions/AzureBlobStorage/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -9,36 +9,28 @@
...
     /// </summary>
     /// <remarks>
     /// Blobs created through this provider will overwrite existing blobs if they have the same blob name.
     /// </remarks>
-    internal partial class AzureBlobEgressProvider :
-        EgressProvider<AzureBlobEgressProviderOptions>
+    internal partial class AzureBlobEgressProvider
     {
-        private int BlobStorageBufferSize = 4 * 1024 * 1024;
+        private readonly string AzureBlobStorage = "AzureBlobStorage";
```
jander-msft, line 25: This should be moved to some kind of shared constant because it is used in the same way both here and in the Program class.

src/Extensions/AzureBlobStorage/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -9,36 +9,28 @@
...
+        protected ILogger Logger { get; }
```
jander-msft, line 26: This should be a private field.

src/Extensions/AzureBlobStorage/Program.cs
```diff
@@ -2,38 +2,43 @@
...
         private static Stream StdInStream = null;
         private static CancellationTokenSource CancelSource = new CancellationTokenSource();
+
+        protected static ILogger Logger { get; set; }
+
         static async Task<int> Main(string[] args)
         {
-            // Expected command line format is: AzureBlobStorage.exe Egress --Provider-Name MyProviderEndpointName
-            RootCommand rootCommand = new RootCommand("Egresses an artifact to Azure Storage.");
+            using var loggerFactory = LoggerFactory.Create(builder =>
```
jander-msft, line 21: This should be done in the Egress method and not stored as static state on the Program class.

src/Extensions/CODE-OF-CONDUCT.md
```diff
@@ -0,0 +1,6 @@
+# Code of Conduct
```
jander-msft, line 1: Do not create a separate file here. This is still part of the dotnet-monitor repo so it should be using the files at the root.

src/Extensions/LICENSE.TXT
```diff
@@ -0,0 +1,23 @@
+The MIT License (MIT)
```
jander-msft, line 1: Do not create a separate license file here. This is still part of the dotnet-monitor repo so it should be using the files at the root.

kkeirstead:
Closing this PR in favor of breaking things down even further for ease of review. I will incorporate the feedback already left on this PR into my changes.

src/Tools/dotnet-monitor/Egress/Configuration/EgressProviderConfigureNamedOptions.cs
```diff
@@ -34,7 +34,23 @@ public void Configure(string name, TOptions options)
...
+                    {
+                        var children = providerOptionsSection.GetChildren();
```
jander-msft, line 39: Why is this needed? Are you working around this issue? https://github.com/dotnet/runtime/issues/77246

src/Tools/dotnet-monitor/Egress/Extension/ExtensionEgressProvider.cs
```diff
@@ -43,8 +42,7 @@ public ExtensionEgressProvider(IEgressPropertiesProvider propertyProvider, Exten
             {
                 Settings = artifactSettings,
                 Configuration = options,
-                Properties = _propertyProvider.GetAllProperties(),
-                ProviderName = providerName,
```
jander-msft, line 47: Why is the provider name being removed?

src/Tools/dotnet-monitor/LoggingEventIds.cs
jander-msft: Do not remove or reorder anything here. Just add a comment after then that they've moved to the extension.

src/Tools/dotnet-monitor/HostBuilder/HostBuilderSettings.cs
```diff
@@ -42,6 +44,16 @@ internal sealed class HostBuilderSettings
                     Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), "." + ProductFolderName) :
                     Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), ProductFolderName));
 
+        // Location where extensions are stored by default.
+        // Windows: "%USERPROFILE%\.dotnet\Tools"
+        // Other: "%XDG_CONFIG_HOME%/dotnet/tools" OR "%HOME%/.config/dotnet/tools" -> IS THIS RIGHT?
```
jander-msft, line 49: It appears to be the same as on Windows:

```bash
jander@DESKTOP-N8NH8I6:~/.dotnet/tools$ ls
dotnet-dump  dotnet-monitor  dotnet-sos  dotnet-trace
```

src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
```diff
@@ -28,15 +29,42 @@ public FolderExtensionRepository(IFileProvider fileSystem, ILoggerFactory logger
 
         public override bool TryFindExtension(string extensionName, out IExtension extension)
         {
-            IDirectoryContents extensionDir = _fileSystem.GetDirectoryContents(extensionName);
+            string extensionPath = string.Empty;
+
+            bool isDotnetToolsLocation = _targetFolder == HostBuilderSettings.ExtensionDirectoryPath;
```
jander-msft, line 34: I think you should have a separate ExtensionRepository implementation for the dotnet tools folder since its algorithm is very much different than the FolderExtensionRepository implementation.

PR #2759, opened 2022-10-22
Updates to `feature/egressExtension` Part 1

kkeirstead:
###### Summary

I closed #2748 in favor of doing an even more limited set of changes. A couple of notes:
- A large chunk of the changes is moving logging statements from the dotnet-monitor project to the Azure Blob Storage project in preparation for completely separating them.
- The majority of newly introduced files are copies from within dotnet-monitor and don't need to be re-reviewed (unless there's a concern with the way the file was originally written).
- To keep the changes smaller, I've eliminated some refactoring that has already been completed in my personal feature branch - as a result, there are some code paths/variables in this PR that are not being exercised.


src/Extensions/AzureBlobStorage/Program.cs
```diff
@@ -14,8 +12,18 @@ internal class Program
     {
         private static Stream StdInStream = null;
         private static CancellationTokenSource CancelSource = new CancellationTokenSource();
+
+        protected static ILogger Logger { get; set; }
```
kkeirstead, line 16: @jander-msft , I know you left a comment about this on the original PR, but I left it as is for now since it's also being used for logging in `BuildOptions` - let me know how you want it and I can change it.

src/Extensions/AzureBlobStorage/AzureBlobStorage.csproj
```diff
@@ -2,27 +2,15 @@
 
   <PropertyGroup>
     <TargetFrameworks>$(ToolTargetFrameworks)</TargetFrameworks>
-    <OutputType>Exe</OutputType>
-    <RootNamespace>Microsoft.Diagnostics.Tools.Monitor.Egress.AzureBlob</RootNamespace>
+    <RootNamespace>Microsoft.Diagnostics.Monitoring.AzureStorage</RootNamespace>
```
jander-msft, line 5:
```suggestion
    <RootNamespace>Microsoft.Diagnostics.Monitoring.AzureBlobStorage</RootNamespace>
```

The primary purpose of the assembly is to store artifacts into Azure Blobs and naming it this makes it clear that this extension is for that purpose. The namespace should follow in the same vein.

src/Extensions/AzureBlobStorage/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -9,34 +9,28 @@
...
     /// </summary>
     /// <remarks>
     /// Blobs created through this provider will overwrite existing blobs if they have the same blob name.
     /// </remarks>
-    internal partial class AzureBlobEgressProvider :
-        EgressProvider<AzureBlobEgressProviderOptions>
+    internal partial class AzureBlobEgressProvider
     {
         private int BlobStorageBufferSize = 4 * 1024 * 1024;
+        private ILogger Logger { get; }
```
jander-msft, line 26: Make this a private readonly field

src/Extensions/AzureBlobStorage/AzureBlobStorage.csproj
```diff
@@ -2,27 +2,15 @@
...
-    <Compile Include="..\..\Tools\dotnet-monitor\Egress\Extension\EgressArtifactResult.cs" Link="ExternalRefs\EgressArtifactResult.cs" />
-    <Compile Include="..\..\Microsoft.Diagnostics.Monitoring.Options\IEgressProviderCommonOptions.cs" Link="ExternalRefs\IEgressProviderCommonOptions.cs" />
-    <Compile Include="..\..\Microsoft.Diagnostics.Monitoring.Options\OptionsDisplayStrings.Designer.cs" Link="ExternalRefs\OptionsDisplayStrings.Designer.cs" />
-    <Compile Include="..\..\Tools\dotnet-monitor\Extensibility\IExtensionResult.cs" Link="ExternalRefs\IExtensionResult.cs" />
-    <Compile Include="..\..\Microsoft.Diagnostics.Monitoring.Options\AzureBlobEgressProviderOptions.cs" Link="ExternalRefs\AzureBlobEgressProviderOptions.cs" />
-  </ItemGroup>
+    <Description>Azure Blob Storage extension for dotnet-monitor</Description>
+    <PackageTags>Diagnostic</PackageTags>
+    <PackageReleaseNotes>$(Description)</PackageReleaseNotes>
+    <ToolCommandName>AzureBlobStorage</ToolCommandName>
```
jander-msft, line 12: IIRC, this property is used for creating the host executable name when installing the tool such that a user would then invoke `dotnet azureblobstorage`. I don't think this name is specific enough to the scenario and it wouldn't really be invoked in any meaningful user scenario. Maybe a more appropriate name would be `dotnet-monitor-egress-azureblobstorage`

src/Extensions/AzureBlobStorage/Program.cs
jander-msft: This is okay for now.

In the long term (before merging back to `main`), I would like to see this rewritten using ServiceCollection/ServiceProvider (see https://github.com/dotnet/dotnet-monitor/blob/main/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/ScenarioHelpers.cs#L29 as an example) and have the main method register the necessary services, pull out the egress provider and invoke it.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Egress/AzureBlob/AzureBlobEgressProviderTests.cs
```diff
@@ -12,7 +12,7 @@
 using Microsoft.Diagnostics.Monitoring.TestCommon.Fixtures;
 using Microsoft.Diagnostics.Monitoring.WebApi;
 using Microsoft.Diagnostics.Tools.Monitor.Egress;
-using Microsoft.Diagnostics.Tools.Monitor.Egress.AzureBlob;
+//using Microsoft.Diagnostics.Tools.Monitor.Egress.AzureBlob;
```
jander-msft, line 15: Remove if it's not used

src/Extensions/AzureBlobStorage/extension.json
```diff
@@ -1,8 +1,8 @@
 {
-  "Id": "Microsoft.Diagnositcs.Monitoring.AzureStorage",
+  "Id": "Microsoft.Diagnostics.Monitoring.AzureStorage",
   "Version": "1.0.0",
   "Program": "AzureBlobStorage.exe",
```
jander-msft, line 4:
```suggestion
  "Program": "dotnet-monitor-egress-azureblobstorage.exe",
```

I assume this needs to be changed to this?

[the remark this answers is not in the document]
jander-msft: I think this is okay for now and should be addressed later.

jander-msft (approved):

PR #2790, opened 2022-10-26
Updates to `feature/egressExtension` Part 2

kkeirstead:
###### Summary

These changes primarily remove unused code paths that were left in for [Part 1](https://github.com/dotnet/dotnet-monitor/pull/2759) to simplify the review process. I will include comments around certain changes that need additional context.


src/Extensions/AzureBlobStorage/EgressArtifactSettings.cs
```diff
@@ -6,11 +6,6 @@ namespace Microsoft.Diagnostics.Monitoring.AzureBlobStorage
 {
     internal sealed class EgressArtifactSettings
     {
-        /// <summary>
-        /// The content encoding of the blob to be created.
-        /// </summary>
-        public string ContentEncoding { get; set; }
```
kkeirstead, line 12: I noticed last week that this is never set - when we were assigning `headers.ContentEncoding = artifactSettings.ContentEncoding`, this never has a value. Since this is something we're not currently doing, the idea was to take it out so that it's not part of the extension contract for `EgressArtifactSettings` if it's empty; however, if we end up using it, then it could be integrated again. Alternatively, we can leave this in if people prefer - I just wanted to call attention to the fact that it's not currently used.

src/Tools/dotnet-monitor/Egress/EgressArtifactSettings.cs
```diff
@@ -5,27 +5,17 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Egress
...
         /// <summary>
         /// The metadata of the blob to be created.
         /// </summary>
         public Dictionary<string, string> Metadata { get; }
             = new Dictionary<string, string>(StringComparer.Ordinal);
 
-        /// <summary>
-        /// Custom metadata of the blob to be created.
-        /// </summary>
-        public Dictionary<string, string> CustomMetadata { get; }
```
kkeirstead, line 26: The concept of `CustomMetadata` only pertains to the `AzureBlobStorage` extension, so this should not be part of `dotnet-monitor`'s extension contract.

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
```diff
@@ -54,8 +54,6 @@ public async Task<EgressArtifactResult> EgressArtifact(ExtensionEgressPayload co
                 ExtensionException.ThrowWrongType(_extensionName, _declarationPath, typeof(IEgressExtension));
             }
 
-            // This _should_ only be used in this method, it can get moved to a constants class if that changes
-            const string CommandArgProviderName = "--Provider-Name";
```
kkeirstead, line 58: I know Justin made a comment on my initial PR about [this](https://github.com/dotnet/dotnet-monitor/pull/2748#discussion_r1002185585); IIRC, there was a discussion during one of my design reviews where we talked about removing this. It was originally included so that extensions could control their own configuration if they wanted to; however, to keep the experience consistent for users we weren't sure we wanted to encourage that. At this time, the `ProviderName` is still being included as part of `ExtensionEgressPayload` and I've only stripped it out as a command line arg; I'm open to suggestions on what course of action we should take in terms of leaving this in, or completely ripping it out.

src/Extensions/AzureBlobStorage/Program.cs
```diff
@@ -23,24 +23,19 @@ static async Task<int> Main(string[] args)
 
             Logger = loggerFactory.CreateLogger<Program>();
 
-            // Expected command line format is: AzureBlobStorage.exe Egress --Provider-Name MyProviderEndpointName
+            // Expected command line format is: dotnet-monitor-egress-azureblobstorage.exe Egress
             RootCommand rootCommand = new RootCommand("Egresses an artifact to Azure Storage.");
 
-            var providerNameOption = new Option<string>(
```
kkeirstead, line 29: See [comment](https://github.com/dotnet/dotnet-monitor/pull/2790/files#r1006173015) below for context

src/Extensions/AzureBlobStorage/AzureBlobStorage.csproj
```diff
@@ -10,6 +10,7 @@
     <PackageTags>Diagnostic</PackageTags>
     <PackageReleaseNotes>$(Description)</PackageReleaseNotes>
     <ToolCommandName>dotnet-monitor-egress-azureblobstorage</ToolCommandName>
+    <AssemblyName>dotnet-monitor-egress-azureblobstorage</AssemblyName>
```
kkeirstead, line 13: I inadvertently didn't update this when updating the `ToolCommandName` in the last set of changes.

src/Extensions/AzureBlobStorage/AzureBlob/AzureBlobEgressProvider.cs
```diff
@@ -80,74 +77,6 @@ public AzureBlobEgressProvider(ILogger logger)
             }
         }
 
-        public async Task<string> EgressAsync(
```
jander-msft, line 83: I think I would prefer that we keep this variant of the `EgressAsync` method because it supports writing blocks rather than requiring the entire stream to be written. This helps in failure modes where we can write out partial artifacts (e.g. logs, metrics, etc). @wiktork, do you have an opinion on which one is kept?

src/Extensions/AzureBlobStorage/EgressArtifactSettings.cs
jander-msft: If it's not used, then removing it fine. We can always add it back later.

src/Extensions/AzureBlobStorage/AzureBlob/AzureBlobEgressProvider.cs
kkeirstead: I switched this out to use the other method - let me know if anything doesn't look right or if you decide you want to go back to the prior way.

src/Extensions/AzureBlobStorage/Program.cs
```diff
@@ -53,7 +49,14 @@ private static async Task<int> Egress(string ProviderName)
 
                 Console.CancelKeyPress += Console_CancelKeyPress;
 
-                result.ArtifactPath = await provider.EgressAsync(Constants.AzureBlobStorageProviderName, configPayload.ProviderName, options, GetStream, configPayload.Settings, CancelSource.Token);
+                result.ArtifactPath = await provider.EgressAsync(options,
+                    async (stream, token) =>
```
jander-msft, line 53: This can be simplified by just passing `GetStream` instead of creating an anonymous async function that merely forwards the arguments.

jander-msft (approved):

jander-msft (approved):

PR #2791, opened 2022-10-26
Faster logs unit tests and eventing workaround updates

jander-msft:
###### Summary

- CollectLogs actions unit tests have been updated to proactively validate the log stream while logs are produced and immediately end when the sentinel category is encountered. This is done by adding a test-only egress provider that writes log data to a pipe writer and is consumed by the validation immediately instead of egressing to the file system.
- The event buffering workarounds have been updated to send several flushing events with brief waits between each iteration to help unblock buffering behaviors in the runtime and in the trace event library.
- Moved the knowledge of the sentinel and flushing categories to the testing infrastructure instead of the tests.

On my local machine, each one of the CollectLogs action unit tests used to take about 10-11 seconds; with these changes, each test takes about 1 second.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectLogsActionTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectLogsActionTests.cs
@@ -6,16 +6,22 @@ using Microsoft.Diagnostics.Monitoring.Options;
 using Microsoft.Diagnostics.Monitoring.TestCommon;
 using Microsoft.Diagnostics.Monitoring.TestCommon.Options;
 using Microsoft.Diagnostics.Monitoring.TestCommon.Runners;
+using Microsoft.Diagnostics.Monitoring.Tool.UnitTests.Egress.Pipe;
 using Microsoft.Diagnostics.Monitoring.WebApi;
 using Microsoft.Diagnostics.Monitoring.WebApi.Models;
+using Microsoft.Diagnostics.Tools.Monitor;
 using Microsoft.Diagnostics.Tools.Monitor.CollectionRules;
 using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Actions;
 using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Actions;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.Memory;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.IO.Pipelines;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Channels;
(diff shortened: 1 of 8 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/LoggerScenario.cs
```diff
@@ -73,11 +74,15 @@ public static async Task ExecuteAsync(InvocationContext context)
 
                 // The log entries above may get stuck in buffers in the runtime eventing infra or
                 // in the trace event library event processor due to their close proximity in being emitted.
-                // To mitigate this, wait a short time and send another log entry, which will cause the buffer
-                // to flush the existing entries. This log entry should be ignored by the logs tests.
-                await Task.Delay(CommonTestTimeouts.EventSourceBufferAvoidanceTimeout);
+                // To mitigate this, repeatedly wait a short time and send another log entry, which will cause the buffer
+                // to flush the existing entries. These log entry should be ignored by the logs tests.
```
kkeirstead, line 78:
```suggestion
                // to flush the existing entries. These log entries should be ignored by the logs tests.
```

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/LoggerScenario.cs
```diff
@@ -73,11 +74,15 @@ public static async Task ExecuteAsync(InvocationContext context)
...
+                ILogger flushCategory = loggerFactory.CreateLogger(TestAppScenarios.Logger.Categories.FlushCategory);
+                for (int i = 0; i < 3; i++)
```
kkeirstead, line 80: Any reason specifically for using 3 here (and if so, should it be documented)?

jander-msft: I reduced it down to two and further clarified why that number was chosen.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/LoggerScenario.cs
```diff
@@ -73,11 +74,22 @@ public static async Task ExecuteAsync(InvocationContext context)
 
                 // The log entries above may get stuck in buffers in the runtime eventing infra or
                 // in the trace event library event processor due to their close proximity in being emitted.
-                // To mitigate this, wait a short time and send another log entry, which will cause the buffer
-                // to flush the existing entries. This log entry should be ignored by the logs tests.
-                await Task.Delay(CommonTestTimeouts.EventSourceBufferAvoidanceTimeout);
+                // To mitigate this, repeatedly wait a short time and send another log entry, which will cause the buffer
```
wiktork, line 77: Github issue to track this, in case you want to reference it here: https://github.com/dotnet/runtime/issues/76704

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectLogsActionTests.cs
```diff
@@ -239,27 +253,41 @@ public Task LogsWildcardActionTest(TargetFrameworkMoniker tfm, LogFormat logForm
...
-
-                    using FileStream logsStream = new(egressPath, FileMode.Open, FileAccess.Read);
-                    Assert.NotNull(logsStream);
-
-                    await LogsTestUtilities.ValidateLogsEquality(logsStream, callback, logFormat, _outputHelper);
                 });
+            },
+            services =>
+            {
+                services.RegisterProvider<PipeEgressOptions, PipeEgressProvider>(PipeEgressProvider.Name);
```
wiktork, line 278: After egress extensibility, will this still be an inproc implementation? It seems like we would want to accommodate that option for unit testing.

wiktork (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectLogsActionTests.cs
jander-msft: It can also be out-of-proc that forwards its stdin to a Unix domain socket if we want to avoid special casing in-proc implementations.

wiktork (approved):

PR #2795, opened 2022-10-27
Add support for speedscope format

wiktork:
###### Summary
Add support for speedscope format when capturing stacks. 

###### Release Notes Entry
Add support for [speedscope](https://speedscope.app) format when capturing stacks.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/ContentTypeUtilities.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/ContentTypeUtilities.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
         public static readonly MediaTypeHeaderValue NdJsonHeader = new MediaTypeHeaderValue(ContentTypes.ApplicationNdJson);
         public static readonly MediaTypeHeaderValue JsonSequenceHeader = new MediaTypeHeaderValue(ContentTypes.ApplicationJsonSequence);
         public static readonly MediaTypeHeaderValue TextPlainHeader = new MediaTypeHeaderValue(ContentTypes.TextPlain);
-        public static readonly MediaTypeHeaderValue SpeedScopeJsonHeader = new MediaTypeHeaderValue(ContentTypes.ApplicationSpeedscopeJson);
+        public static readonly MediaTypeHeaderValue SpeedscopeJsonHeader = new MediaTypeHeaderValue(ContentTypes.ApplicationSpeedscopeJson);
 
         public static StackFormat? ComputeStackFormat(IList<MediaTypeHeaderValue> acceptedHeaders)
         {
@@ -24,10 +24,10 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
             if (acceptedHeaders == null) return null;
             if (acceptedHeaders.Contains(TextPlainHeader)) return StackFormat.PlainText;
             if (acceptedHeaders.Contains(ApplicationJsonHeader)) return StackFormat.Json;
-            if (acceptedHeaders.Contains(SpeedScopeJsonHeader)) return StackFormat.Speedscope;
+            if (acceptedHeaders.Contains(SpeedscopeJsonHeader)) return StackFormat.Speedscope;
             if (acceptedHeaders.Any(TextPlainHeader.IsSubsetOf)) return StackFormat.PlainText;
             if (acceptedHeaders.Any(ApplicationJsonHeader.IsSubsetOf)) return StackFormat.Json;
-            if (acceptedHeaders.Any(SpeedScopeJsonHeader.IsSubsetOf)) return StackFormat.Speedscope;
+            if (acceptedHeaders.Any(SpeedscopeJsonHeader.IsSubsetOf)) return StackFormat.Speedscope;
             return null;
         }
 
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/ContentTypeUtilities.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/ContentTypeUtilities.cs
@@ -1,0 +1,45 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Net.Http.Headers;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    internal static class ContentTypeUtilities
+    {
+        public static readonly MediaTypeHeaderValue ApplicationJsonHeader = new MediaTypeHeaderValue(ContentTypes.ApplicationJson);
+        public static readonly MediaTypeHeaderValue NdJsonHeader = new MediaTypeHeaderValue(ContentTypes.ApplicationNdJson);
+        public static readonly MediaTypeHeaderValue JsonSequenceHeader = new MediaTypeHeaderValue(ContentTypes.ApplicationJsonSequence);
+        public static readonly MediaTypeHeaderValue TextPlainHeader = new MediaTypeHeaderValue(ContentTypes.TextPlain);
+        public static readonly MediaTypeHeaderValue SpeedScopeJsonHeader = new MediaTypeHeaderValue(ContentTypes.ApplicationSpeedscopeJson);
+
+        public static StackFormat? ComputeStackFormat(IList<MediaTypeHeaderValue> acceptedHeaders)
+        {
+            // CONSIDER We currently don't factor quality into account.
+            // CONSIDER Centralize content negotiation across logs, stacks, and any other types that need it.
+
+            if (acceptedHeaders == null) return null;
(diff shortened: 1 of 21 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/ContentTypeUtilities.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    internal static class ContentTypeUtilities
+    {
+        public static readonly MediaTypeHeaderValue ApplicationJsonHeader = new MediaTypeHeaderValue(ContentTypes.ApplicationJson);
+        public static readonly MediaTypeHeaderValue NdJsonHeader = new MediaTypeHeaderValue(ContentTypes.ApplicationNdJson);
+        public static readonly MediaTypeHeaderValue JsonSequenceHeader = new MediaTypeHeaderValue(ContentTypes.ApplicationJsonSequence);
+        public static readonly MediaTypeHeaderValue TextPlainHeader = new MediaTypeHeaderValue(ContentTypes.TextPlain);
+        public static readonly MediaTypeHeaderValue SpeedScopeJsonHeader = new MediaTypeHeaderValue(ContentTypes.ApplicationSpeedscopeJson);
```
kkeirstead, line 17: nit: it seems like there's a mix of `Speedscope` and `SpeedScope` - do we want to make this consistent?

src/Microsoft.Diagnostics.Monitoring.WebApi/ContentTypeUtilities.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public static StackFormat? ComputeStackFormat(IList<MediaTypeHeaderValue> acceptedHeaders)
+        {
+            // CONSIDER We currently don't factor quality into account.
+            // CONSIDER Centralize content negotiation across logs, stacks, and any other types that need it.
+
+            if (acceptedHeaders == null) return null;
+            if (acceptedHeaders.Contains(TextPlainHeader)) return StackFormat.PlainText;
+            if (acceptedHeaders.Contains(ApplicationJsonHeader)) return StackFormat.Json;
+            if (acceptedHeaders.Contains(SpeedScopeJsonHeader)) return StackFormat.Speedscope;
```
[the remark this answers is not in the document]
wiktork: I would prefer to keep this as close to real content negotiation as possible, which means more specific accept headers supersede subsets.

src/Microsoft.Diagnostics.Monitoring.WebApi/Stacks/SpeedScopeStacksFormatter.cs
```diff
@@ -0,0 +1,128 @@
...
+                    Name = string.Format(CultureInfo.CurrentCulture, Strings.CallstackThreadHeader, stack.ThreadId),
+                };
+
+                speedScopeResult.Profiles.Add(profile);
+
+                foreach (CallStackFrame frame in stack.Frames)
+                {
+                    if (frame.FunctionId == 0)
+                    {
+                        var profileEvent = new ProfileEvent { At = FixedStart, Frame = 0, Type = ProfileEventType.O };
```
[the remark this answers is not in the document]
wiktork: Good catch. I will see about adding a unit test to make sure well known frames are handled.

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/SpeedScopeStackResults.cs
```diff
@@ -0,0 +1,109 @@
...
+        public List<SharedFrame> Frames { get; set; }
+    }
+
+    public class SharedFrame
+    {
+        [JsonPropertyName("name")]
+        public string Name { get; set; }
+
+        [JsonPropertyName("col")]
+        public int? Col { get; set; }
```
jander-msft, line 43: nit: Let's avoid abbreviations in the property names:

```suggestion
        public int? Column { get; set; }
```

src/MonitorProfiler/MainProfiler/MainProfiler.cpp
```diff
@@ -290,6 +291,11 @@ HRESULT MainProfiler::ProcessCallstackMessage()
         IfFailLogRet(eventProvider->WriteCallstack(stackState->GetStack()));
     }
 
+    //HACK See https://github.com/dotnet/runtime/issues/76704
+    // We sleep here for 200ms to ensure that our event is timestamped. Since we are on a dedicated message
+    // thread we should not be interfering with the app itself.
+    ThreadUtilities::Sleep(200);
```
kkeirstead, line 297: I'm assuming this value (`200`) should be the same as https://github.com/dotnet/dotnet-monitor/pull/2791/files#diff-f2ab9c8a7408cb463a090d005d24483a25ede050ee8f4ffe43693945d01e7e3cR99 (`250`)

src/Microsoft.Diagnostics.Monitoring.WebApi/Stacks/SpeedScopeStacksFormatter.cs
```diff
@@ -0,0 +1,127 @@
...
+                    Name = string.Format(CultureInfo.CurrentCulture, Strings.CallstackThreadHeader, stack.ThreadId),
+                };
+
+                speedscopeResult.Profiles.Add(profile);
+
+                foreach (CallStackFrame frame in stack.Frames)
+                {
+                    if (frame.FunctionId == 0)
+                    {
+                        profile.Events.Add(new ProfileEvent { At = FixedStart, Frame = 0, Type = ProfileEventType.O });
```
jander-msft, line 75: Given that all of these properties are statically known, could you either create a readonly static field for this or at least only create this once per invocation of `FormatStack` and reuse the same instance?

src/Microsoft.Diagnostics.Monitoring.WebApi/Stacks/SpeedScopeStacksFormatter.cs
```diff
@@ -0,0 +1,127 @@
...
+
+                    }
+                    else if (cache.FunctionData.TryGetValue(frame.FunctionId, out FunctionData functionData))
+                    {
+                        var builder = new StringBuilder();
```
jander-msft, line 80: The construction of and appendation to the builder should only be done if the shared frame doesn't exist, otherwise this looks like a waste of compute. I think this will be fairly common occurrence (shared frames), especially for async call frames.

src/Microsoft.Diagnostics.Monitoring.WebApi/Stacks/SpeedScopeStacksFormatter.cs
```diff
@@ -0,0 +1,127 @@
...
+                            Frame = mapping,
+                            Type = ProfileEventType.O,
+                        };
+
+                        profile.Events.Add(profileEvent);
+
+                    }
+                    else
+                    {
+                        profile.Events.Add(new ProfileEvent { At = FixedStart, Frame = 1, Type = ProfileEventType.O });
```
jander-msft, line 107: Same static readonly or single initialization treatment as above.

src/Microsoft.Diagnostics.Monitoring.WebApi/Stacks/SpeedScopeStacksFormatter.cs
```diff
@@ -0,0 +1,127 @@
...
+                // Speedscope requires a close event for each open event.
+                for (int i = profile.Events.Count - 1; i >= 0; i--)
+                {
+                    profile.Events.Add(new ProfileEvent { At = FixedEnd, Frame = profile.Events[i].Frame, Type = ProfileEventType.C });
+                }
+            }
+
+            JsonSerializerOptions options = new JsonSerializerOptions
+            {
+#if NET6_0_OR_GREATER
```
jander-msft, line 120: This shouldn't be necessary anymore since this should only be building for .NET 6+.

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/StackUtilities.cs
```diff
@@ -11,6 +11,13 @@
 
 namespace Microsoft.Diagnostics.Monitoring.WebApi
 {
+    internal enum StackFormat
```
jander-msft, line 14: Can `CallStackFormat` from the options be used rather than making a copy here and having to convert between the two?

src/MonitorProfiler/MainProfiler/MainProfiler.cpp
wiktork: 150 is the absolute minimum, since that's the actual timer in the framework. However, these are slightly different scenarios. For stacks, this will stall the scenario by this much time. For log testing, the idea is to speed up tests by creating a fake log value. I will consider changing this value if our stacks tests are very slow.

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/StackUtilities.cs
wiktork: All of our collection rule options are currently defined as part of dotnet-monitor.

PR #2828, opened 2022-11-02
Support gracefully stopping traces on-demand

schmittjoseph:
###### Summary

This PR introduces support for stopping a requested trace on demand while also collecting rundown, regardless of if an egress provider is specified. The following high-level changes were made to support this:
- Operations now have a new `Stopping` state. This state indicates that the user requested the operation to stop, but it will take an indeterminate amount of time. Operations in this state can still be immediately cancelled.
  - To help users distinguish what operations can, and cannot, be gracefully stopped a new `isStoppable` property is added onto operations. 
- For API consistency, *all* API routes that supported sending the artifact back via the HTTP response now also record it in the operations table and supply the operations URL in the `Location` header.
  - To help users distinguish between operations, the `egressProviderName` is now recorded and reported for each operation.
  - As a result, request limit handling is now entirely handled by the operations table. There is no longer a need for the request tracking middleware or attributes.
-  Http response egressed based operations are now plumbed into the egress operations service. Using this approach, we can ensure operation state consistency since these operations have multiple ways of being cancelled or otherwise controlled. e.g. an HTTP based response can be cancelled via the operations API, or by the user stopping the request.

###### Release Notes Entry
Add support for gracefully stopping a requested trace on demand while also collecting rundown, via [Stop Operation](https://github.com/dotnet/dotnet-monitor/blob/main/documentation/api/operations-stop.md). All artifact requests are now also recorded under the [Operations API](https://github.com/dotnet/dotnet-monitor/blob/main/documentation/api/operations.md), regardless of if they have an egress provider specified.


documentation/api/dump.md
```diff
@@ -45,12 +45,14 @@ Allowed schemes:
...
-| 200 OK | stream | A managed dump of the process. | `application/octet-stream` |
-| 202 Accepted | | When an egress provider is specified, the Location header containers the URI of the operation for querying the egress status. | |
+| 200 OK | stream | A managed dump of the process when no egress provider is specified. | `application/octet-stream` |
+| 202 Accepted | | When an egress provider is specified, the artifact has begun being collected. | |
 | 400 Bad Request | [ValidationProblemDetails](definitions.md#validationproblemdetails) | An error occurred due to invalid input. The response body describes the specific problem(s). | `application/problem+json` |
 | 401 Unauthorized | | Authentication is required to complete the request. See [Authentication](./../authentication.md) for further information. | |
 | 429 Too Many Requests | | There are too many dump requests at this time. Try to request a dump at a later time. | |
 
+> **NOTE:** Regardless if an egress provider is specified if the request was successful (response codes 200 or 202), the Location header containers the URI of the operation. This can be used to query the status of the operation or change its state.
```
kkeirstead, line 54: nit: Grammar, "containers" -> "contains"

documentation/api/operations-delete.md
```diff
@@ -3,7 +3,8 @@
 
 # Operations - Delete
 
-Cancel a running operation. Only valid against operations in the `Running` state. Transitions the operation to `Cancelled` state.
+Cancel a running operation. Only valid against operations in the `Running` or `Stopping` state. Transitions the operation to `Cancelled` state. Cancelling an operation may result in an incomplete or unreadable artifact. To stop an operation early while still producing a valid artifact using [Stop Operation](operations-stop.md).
```
kkeirstead, line 6:
```suggestion
Cancel a running operation. Only valid against operations in the `Running` or `Stopping` state. Transitions the operation to `Cancelled` state. Cancelling an operation may result in an incomplete or unreadable artifact. To stop an operation early while still producing a valid artifact, use the [Stop Operation](operations-stop.md).
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.Metrics.cs
```diff
@@ -25,7 +25,6 @@ partial class DiagController
         [ProducesWithProblemDetails(ContentTypes.ApplicationJsonSequence)]
         [ProducesResponseType(typeof(ProblemDetails), StatusCodes.Status429TooManyRequests)]
         [ProducesResponseType(typeof(void), StatusCodes.Status202Accepted)]
-        [RequestLimit(LimitKey = Utilities.ArtifactType_Metrics)]
```
jander-msft, line 28: I know that we want to route everything through the operations service eventually, but removing these without doing that leaves all of the routes in a state of allowing unlimited requests. I would ask that you add these attributes and the middleware back until the route is actually registered in the operation service.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/OperationsController.cs
```diff
@@ -72,14 +72,27 @@ public IActionResult GetOperationStatus(Guid operationId)
         [HttpDelete("{operationId}")]
         [ProducesWithProblemDetails(ContentTypes.ApplicationJson)]
         [ProducesResponseType(typeof(void), StatusCodes.Status200OK)]
-        public IActionResult CancelOperation(Guid operationId)
+        [ProducesResponseType(typeof(void), StatusCodes.Status202Accepted)]
+        public IActionResult CancelOperation(
+            Guid operationId,
+            [FromQuery]
+            bool gracefulStop = false)
```
jander-msft, line 79: I would just call this `stop`; we're never going to add an `ungracefulStop`/`abruptStop` parameter.

src/Tools/dotnet-monitor/Trace/AbstractTraceOperation.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    internal abstract class AbstractTraceOperation : PipelineArtifactOperation<EventTracePipeline>
+    {
+        // Buffer size matches FileStreamResult
+        protected const int DefaultBufferSize = 0x10000;
+
+        protected readonly EventTracePipelineSettings _settings;
+
+        public AbstractTraceOperation(IEndpointInfo endpointInfo, EventTracePipelineSettings settings, ILogger logger)
+            : base(logger, Utils.ArtifactType_Trace, endpointInfo, isStoppable: true)
```
jander-msft, line 21:
```suggestion
            : base(logger, Utils.ArtifactType_Trace, endpointInfo)
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -636,29 +634,21 @@ private string GetDiagnosticPortName()
...
-            string fileName = TraceUtilities.GenerateTraceFileName(processInfo.EndpointInfo);
+            IArtifactOperation traceOperation = _traceOperationFactory.Create(
+                processInfo.EndpointInfo,
+                configuration,
+                duration);
+
+            if (_diagnosticPortOptions.Value.ConnectionMode == DiagnosticPortConnectionMode.Listen)
+            {
+                IDisposable operationRegistration = _operationTrackerService.Register(processInfo.EndpointInfo);
+                HttpContext.Response.RegisterForDispose(operationRegistration);
```
jander-msft, line 645: nit: This changes the scope of when this registration is disposed (it was created and disposed in the same function previously). This is probably okay so long as we are guaranteed that this will be disposed, even in the case of something catastrophic happening when handling the http request.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/EgressTests.cs
```diff
@@ -101,6 +102,42 @@ public async Task EgressCancelTest()
...
+                appValidate: async (appRunner, apiClient) =>
+                {
+                    int processId = await appRunner.ProcessIdTask;
+
+                    OperationResponse response = await apiClient.EgressTraceAsync(processId, durationSeconds: -1, FileProviderName);
+                    Assert.Equal(HttpStatusCode.Accepted, response.StatusCode);
+
+                    OperationStatusResponse operationResult = await apiClient.GetOperationStatus(response.OperationUri);
+                    Assert.Equal(HttpStatusCode.OK, operationResult.StatusCode);
+                    Assert.Equal(OperationState.Running, operationResult.OperationStatus.Status);
```
jander-msft, line 132: Also validate that this is stoppable before attempting to stop it?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/EgressTests.cs
```diff
@@ -226,6 +267,121 @@ public async Task SharedConcurrencyLimitTest()
...
+
+                    // Cancel the trace operation
+                    HttpStatusCode deleteStatus = await apiClient.CancelEgressOperation(operationUri);
+                    Assert.Equal(HttpStatusCode.OK, deleteStatus);
+
+                    operationResult = await apiClient.GetOperationStatus(operationUri);
+                    Assert.Equal(HttpStatusCode.OK, operationResult.StatusCode);
+                    Assert.Equal(OperationState.Cancelled, operationResult.OperationStatus.Status);
+
+                    await Assert.ThrowsAsync<IOException>(async () => await drainResponseTask);
```
jander-msft, line 303:
```suggestion
                    await Assert.ThrowsAsync<IOException>(() => drainResponseTask);
```

In general, if a function is of the form `async (...) => await ...`, then the `async` and `await` are not necessary; these create a state machine object that is not needed. The exception is if some kind of disposal/unregistration needs to occur before the function returns.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/EgressTests.cs
```diff
@@ -226,6 +267,121 @@ public async Task SharedConcurrencyLimitTest()
...
+                    using ResponseStreamHolder responseBox = await apiClient.HttpEgressTraceAsync(processId, durationSeconds: -1);
+
+                    Uri operationUri = responseBox.Response.Headers.Location;
+                    Assert.NotNull(operationUri);
+
+                    // Start saving the stream
+                    string traceFile = Path.Combine(tempDir.FullName, "test.nettrace");
+                    using FileStream traceFileWriter = File.OpenWrite(traceFile);
+
+                    using MemoryStream dstStream = new();
```
jander-msft, line 336: This is not used

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/EgressTests.cs
```diff
@@ -226,6 +267,121 @@ public async Task SharedConcurrencyLimitTest()
...
+                    // Make sure the operation exists
+                    OperationStatusResponse operationResult = await apiClient.GetOperationStatus(operationUri);
+                    Assert.Equal(HttpStatusCode.OK, operationResult.StatusCode);
+                    Assert.True(operationResult.OperationStatus.Status == OperationState.Running);
+
+                    // Stop the trace operation
+                    HttpStatusCode deleteStatus = await apiClient.StopEgressOperation(operationUri);
+                    Assert.Equal(HttpStatusCode.Accepted, deleteStatus);
+
+                    await drainResponseTask;
```
jander-msft, line 348: If the http response doesn't end for some reason (e.g. stopping mechanism is broken), then this will hang the test. Consider `.WaitAsync(...)` with some reasonable timeout.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/EgressTests.cs
```diff
@@ -226,6 +267,121 @@ public async Task SharedConcurrencyLimitTest()
...
+
+                    await appRunner.SendCommandAsync(TestAppScenarios.AsyncWait.Commands.Continue);
+                },
+                configureTool: (toolRunner) =>
+                {
+                    toolRunner.WriteKeyPerValueConfiguration(new RootOptions().AddFileSystemEgress(FileProviderName, _tempDirectory.FullName));
+                });
+        }
+
+        private static Task<bool> DoesNettraceFileHaveRundown(string filePath)
```
jander-msft, line 366: Was this copied from another test (it seems familiar)? Can you refactor into some utility method and share it?

documentation/api/definitions.md
```diff
@@ -281,6 +282,8 @@ Detailed information about an operation.
 | `operationId` | guid | Unique identifier for the operation. |
 | `createdDateTime` | datetime string | UTC DateTime string of when the operation was created. |
 | `status` | [OperationState](#operationstate) | The current status of operation. |
+| `egressProviderName` | string | The name of the egress provider that the artifact is being sent to. This will be null if the artifact is being sent directly back to the user from an HTTP request. |
```
jander-msft, line 285: For all of the new functionality, let's add the `8.0+` notion so we can call out these are not applicable to the 7.0 release that just went out.

jander-msft (approved):

PR #2829, opened 2022-11-02
Updates to `feature/egressExtension` Part 3

kkeirstead:
###### Summary

These changes enable the `dotnet tool install` scenario, automatically placing the extension in a location where `dotnet monitor` knows to look. Additionally, due to the nature of potentially wanting a different package name for an extension (e.g. `dotnet-monitor-egress-azureblobstorage`) and a friendly name to use in configuration (e.g. `AzureBlobStorage`, to avoid breaking changes), we now look through all the directories in the target locations for a `DisplayName` within each `extension.json` file - if the `DisplayName` matches the name used in configuration, we use that extension, regardless of the name of the directory it's under.


src/Tools/dotnet-monitor/Extensibility/ToolsExtensionRepository.cs
```diff
@@ -0,0 +1,58 @@
...
+
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.Logging;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    internal class ToolsExtensionRepository : ExtensionRepository
```
kkeirstead, line 13: This is very similar to `FolderExtensionRepository` - these may be refactored later, but Justin recommended we keep them separate for the time being.

src/Tools/dotnet-monitor/Extensibility/ExtensionRepositoryUtilities.cs
```diff
@@ -0,0 +1,28 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Extensions.FileProviders;
+using System.IO;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    public static class ExtensionRepositoryUtilities
+    {
+        public static bool ExtensionDefinitionExists(IFileProvider fileSystem, string extensionPath)
```
jander-msft, line 12:
```suggestion
        public static bool TryGetExtensionDefinitionPath(this IFileProvider fileProvider, string extensionName, out string definitionPath)
```

1. Make this an extension method instead of yet another utility class.
2. Let's more all of the path calculations inside this method as much as possible so that callers don't have to understand the pathing structure.

src/Tools/dotnet-monitor/Extensibility/ExtensionRepositoryUtilities.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Extensions.FileProviders;
+using System.IO;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    public static class ExtensionRepositoryUtilities
```
jander-msft, line 10:
```suggestion
    public static class FileProviderExtensions
```

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
```diff
@@ -21,18 +21,26 @@ internal partial class ProgramExtension : IExtension, IEgressExtension
         private readonly string _extensionName;
         private readonly string _targetFolder;
         private readonly string _declarationPath;
+        private readonly string _exePath;
         private readonly IFileProvider _fileSystem;
         private readonly ILogger<ProgramExtension> _logger;
-        private Lazy<ExtensionDeclaration> _extensionDeclaration;
+        public Lazy<ExtensionDeclaration> ExtensionDeclaration { get; }
```
jander-msft, line 27: Why is this public? There's already a private property that you can make public instead.

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
```diff
@@ -21,18 +21,26 @@ internal partial class ProgramExtension : IExtension, IEgressExtension
...
         public ProgramExtension(string extensionName, string targetFolder, IFileProvider fileSystem, string declarationPath, ILogger<ProgramExtension> logger)
         {
             _extensionName = extensionName;
             _targetFolder = targetFolder;
             _fileSystem = fileSystem;
             _declarationPath = declarationPath;
+            _exePath = declarationPath;
             _logger = logger;
-            _extensionDeclaration = new Lazy<ExtensionDeclaration>(() => { return GetExtensionDeclaration(); }, LazyThreadSafetyMode.ExecutionAndPublication);
+            ExtensionDeclaration = new Lazy<ExtensionDeclaration>(() => { return GetExtensionDeclaration(); }, LazyThreadSafetyMode.ExecutionAndPublication);
```
jander-msft, line 37:
```suggestion
            ExtensionDeclaration = new Lazy<ExtensionDeclaration>(GetExtensionDeclaration, LazyThreadSafetyMode.ExecutionAndPublication);
```

src/Tools/dotnet-monitor/Extensibility/ExtensionDeclaration.cs
```diff
@@ -23,6 +23,11 @@ internal class ExtensionDeclaration
         /// </summary>
         public string Program { get; set; }
 
+        /// <summary>
+        /// This is the name that users specify in configuration to refer to the extension.
+        /// </summary>
+        public string DisplayName { get; set; }
```
jander-msft, line 29: `DisplayName` is not a good name for this. A property named "DisplayName" is typically only used for presenting to an end user with a human readable name, not for identification purposes. I would just class this `Name` for now.

I think we'll have to revisit the format of the extension.json file because it feels like it wants to be partly generic multi-extension manifest and partly specific for egress extensions.

src/Tools/dotnet-monitor/HostBuilder/HostBuilderSettings.cs
```diff
@@ -42,6 +44,16 @@ internal sealed class HostBuilderSettings
                     Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), "." + ProductFolderName) :
                     Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), ProductFolderName));
 
+        // Location where extensions are stored by default.
+        // Windows: "%USERPROFILE%\.dotnet\Tools"
+        // Other: "%XDG_CONFIG_HOME%/.dotnet/tools" OR "%HOME%/.dotnet/tools" -> THIS HAS NOT BEEN TESTED YET ON LINUX
+        public static readonly string DotnetToolsExtensionDirectoryPath =
```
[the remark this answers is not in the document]
kkeirstead: One thought about this - I'm not doing it yet, but in my testing PR that I'm working on I'm changing the paths using `HostBuilderSettings`, and I intend to do that for this path (instead of directly using its value). In that case, would it make more sense to leave it where it is?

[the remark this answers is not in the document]
jander-msft: Being able to change the value for testing is still very much an option if this is moved to `ToolsExtensionRepository`. You would still use an environment variable to override it. Possibly better, it can be implemented as a separate service that provides special file system information; the tests just provide a replacement of the service implementation using [HostingStartup](https://github.com/dotnet/dotnet-monitor/blob/main/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.TestHostingStartup/HostingStartup.cs#L14)/unit test service registration.

src/Extensions/Directory.Build.props
```diff
@@ -0,0 +1,13 @@
+<Project>
+  <Import Project="$(MSBuildThisFileDirectory)..\Directory.Build.props"/>
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <GenerateDocumentationFile>false</GenerateDocumentationFile>
+    <IsPackable>true</IsPackable>
+    <IsShipping>true</IsShipping>
+    <PackAsTool>true</PackAsTool>
+    <PackAsToolShimRuntimeIdentifiers>win-x64;win-x86;win-arm64;win-arm;osx-x64;linux-x64;linux-musl-x64;linux-arm64;linux-musl-arm64;linux-arm</PackAsToolShimRuntimeIdentifiers>
```
jander-msft, line 10:
```suggestion
    <PackAsToolShimRuntimeIdentifiers>win-x64;win-x86;win-arm64;osx-x64;linux-x64;linux-musl-x64;linux-arm64;linux-musl-arm64</PackAsToolShimRuntimeIdentifiers>
```

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
```diff
@@ -31,14 +32,21 @@ public ProgramExtension(string extensionName, string targetFolder, IFileProvider
             _targetFolder = targetFolder;
             _fileSystem = fileSystem;
             _declarationPath = declarationPath;
+            _exePath = declarationPath;
             _logger = logger;
-            _extensionDeclaration = new Lazy<ExtensionDeclaration>(() => { return GetExtensionDeclaration(); }, LazyThreadSafetyMode.ExecutionAndPublication);
+            _extensionDeclaration = new Lazy<ExtensionDeclaration>(GetExtensionDeclaration, LazyThreadSafetyMode.ExecutionAndPublication);
+        }
+
+        public ProgramExtension(string extensionName, string targetFolder, IFileProvider fileSystem, string declarationPath, string exePath, ILogger<ProgramExtension> logger)
```
jander-msft, line 40: I think this ctor is unnecessary, namely the `_exePath` parameter. Even in the dotnet tool case, the executable is located next to extension.json file, so I don't see how this is different than the existing folder based lookup.

jander-msft: As I understand it, in the dotnet tool case, `_exePath` is set to "AzureBlobStorage", which isn't even the executable name nor a path.

src/Tools/dotnet-monitor/Extensibility/ToolsExtensionRepository.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Linq;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    internal class ToolsExtensionRepository : ExtensionRepository
+    {
+        private readonly string _targetFolder;
+        private readonly IFileProvider _fileSystem;
+        private readonly ILoggerFactory _loggerFactory;
```
jander-msft, line 19: This isn't used other than for creating a single `ILogger` instance. Can we change it so that the ctor takes `ILogger<ProgramExtension>` and have the service creation get that instead of getting `ILoggerFactory`?

src/Tools/dotnet-monitor/ServiceCollectionExtensions.cs
```diff
@@ -220,26 +220,25 @@ public static IServiceCollection ConfigureExtensions(this IServiceCollection ser
...
+                || string.IsNullOrWhiteSpace(dotnetToolsFolder))
             {
                 throw new InvalidOperationException();
             }
 
             // Add the folders we search to get extensions from
-            services.AddExtensionRepository(1000, nextToMeFolder);
-            services.AddExtensionRepository(2000, progDataFolder);
-            services.AddExtensionRepository(3000, settingsFolder);
+            services.AddFolderExtensionRepository(1000, nextToMeFolder);
```
jander-msft, line 233: Can we remove the priority notion? We'll get them in the order in which they are registered in the service collection, which is already in the order that we want.

src/Tools/dotnet-monitor/Extensibility/FileProviderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Extensions.FileProviders;
+using System.IO;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    public static class FileProviderExtensions
+    {
+        public static bool TryGetExtensionDefinitionPath(this IFileProvider fileProvider, string extensionName, out string definitionPath)
```
jander-msft, line 12: The `extensionName` parameter isn't really just a name, it's a sub path under the file provider; consider renaming to `extensionSubPath`.

src/Tools/dotnet-monitor/Extensibility/ToolsExtensionRepository.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        public override bool TryFindExtension(string extensionName, out IExtension extension)
+        {
+            const string storeDirectory = ".store";
+            IDirectoryContents toolsStoreDir = _fileSystem.GetDirectoryContents(storeDirectory);
+            ILogger<ProgramExtension> logger = _loggerFactory.CreateLogger<ProgramExtension>();
+
+            foreach (IFileInfo tool in toolsStoreDir)
+            {
+                string toolVersion = _fileSystem.GetDirectoryContents(Path.Combine(storeDirectory, tool.Name)).First().Name;
```
jander-msft, line 50: Let's be a bit more careful here in case the folder structure isn't what we expect. Something like:

```
IFileInfo versionDirInfo = _fileSystem.GetDirectoryContents(Path.Combine(storeDirectory, tool.Name)).FirstOrDefault();
if (null == versionDirInfo)
{
    continue;
}
string toolVersion = versionDirInfo.Name;
```

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
kkeirstead: You're correct in that `_exePath` is currently being (erroneously) set to "AzureBlobStorage" - this wasn't properly addressed when I made the change to allow the executable to have a different name from the name in configuration, though it doesn't make a difference given that we just get the `DirectoryName` which is empty. Unless I'm missing something though, aren't executables for dotnet tools located at `.dotnet/tools/name-of-extension.exe`, whereas the `extension.json` file is located at `.dotnet/tools/.store/.../any/`? Given my previous statement about the actual value of `_exePath` not being relevant, we probably don't need a value for it - however, if I'm correct about the location of the executable being different, then we would still need to special-case this from the other extension locations

src/Tools/dotnet-monitor/Extensibility/FileProviderExtensions.cs
kkeirstead: Agreed, this is a path and not a name - I actually originally had it set to `extensionPath` but changed it when I made your suggestion https://github.com/dotnet/dotnet-monitor/pull/2829#discussion_r1015774783 - I'll switch it to `extensionSubPath`.

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
jander-msft: > You're correct in that `_exePath` is currently being (erroneously) set to "AzureBlobStorage" - this wasn't properly addressed when I made the change to allow the executable to have a different name from the name in configuration, though it doesn't make a difference given that we just get the `DirectoryName` which is empty. Unless I'm missing something though, aren't executables for dotnet tools located at `.dotnet/tools/name-of-extension.exe`, whereas the `extension.json` file is located at `.dotnet/tools/.store/.../any/`? Given my previous statement about the actual value of `_exePath` not being relevant, we probably don't need a value for it - however, if I'm correct about the location of the executable being different, then we would still need to special-case this from the other extension locations

Mostly correct, however, the executable at `.dotnet/tools/name-of-extension.exe` is targeted to run one of the possibly multiple TFMs in the tool package e.g. if you install it with the .NET 6 SDK, it will target `net6.0`; if you install it with .NET 7 SDK, it will target `net7.0`; there's a `-f` switch on `dotnet tool install` to force the TFM as well. Now, this might not be an issue because if it installed for some TFM, then we probably don't care which one it is so long as it runs.

You'll either need to remove the ctor (since it doesn't make sense as-is) or fix it along with the executable determination to actually point to the correct location for dotnet tools.

jander-msft (approved):
I'm approving, but I think [this](https://github.com/dotnet/dotnet-monitor/pull/2829/files#r1020534584) should be fixed shortly.

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
kkeirstead: Resolving for now - will likely revisit this later to finalize a solution, but for now the current iteration works and will unblock some work around testing.

PR #2836, opened 2022-11-04
Collect dumps via artifact operation

jander-msft:
###### Summary

Create an `IArtifactOperation` implementation for collecting dumps. This change reduces the complexity for collecting dumps via the `/dump` route and the CollectDumpAction.


src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -223,32 +225,14 @@ public Task<ActionResult<IEnumerable<Models.ProcessIdentifier>>> GetProcesses()
         {
             ProcessKey? processKey = Utilities.GetProcessKey(pid, uid, name);
 
-            return InvokeForProcess(async processInfo =>
-            {
-                string dumpFileName = DumpUtilities.GenerateDumpFileName();
-
-                if (string.IsNullOrEmpty(egressProvider))
-                {
-                    Stream dumpStream = await _dumpService.DumpAsync(processInfo.EndpointInfo, type, HttpContext.RequestAborted);
```
kkeirstead, line 232: nit: Are we using `_dumpService` in this file anymore? If not, we probably don't need it anymore.

jander-msft: Good catch. Removed it.

PR #2858, opened 2022-11-08
Remove synchronous IO for metrics

jander-msft:
###### Summary

This change removes the need for synchronous IO when collecting metrics.


src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/JsonCounterLogger.cs
```diff
@@ -3,21 +3,41 @@
...
         {
+            _stream = stream;
+            _bufferWriter = new(InitialBufferCapacity);
         }
 
-        protected override void SerializeCounter(Stream stream, ICounterPayload counter)
+        protected override async Task SerializeAsync(ICounterPayload counter)
         {
-            stream.WriteByte(StreamingLogger.JsonSequenceRecordSeparator);
-            using (var writer = new Utf8JsonWriter(stream, new JsonWriterOptions { Indented = false }))
```
wiktork, line 20: Related topic: https://github.com/dotnet/runtime/issues/30401

Note the approach in that issue would force some aggressive flushing on our part so I'm not suggesting we use it.

src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/StreamingCounterLogger.cs
```diff
@@ -3,39 +3,71 @@
...
-            SerializeCounter(_stream, counter);
+            _channelWriter.TryWrite(counter);
         }
 
         public void PipelineStarted()
         {
+            _ = ReadAndSerializeAsync();
         }
 
         public void PipelineStopped()
```
wiktork, line 50: This may be beyond the scope of this change, but we can alter this callback to better support async writing (perhaps a ValueTask api)

src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/StreamingCounterLogger.cs
```diff
@@ -3,39 +3,71 @@
...
         {
-            Cleanup();
+            _channelWriter.Complete();
+
+            _finishedSerialization.WaitOne(FinishSerializationTimeout);
```
wiktork, line 54: This still seems like we're doing synchronous I/O: blocking an async worker thread until an IO operation completes (or times out).

wiktork: If we keep the Log call synchronous, but change PipelineStopped/PipelineStarted to async, we should be able to make this fully async.

jander-msft: > This still seems like we're doing synchronous I/O: blocking an async worker thread until an IO operation completes (or times out).

There's no synchronous I/O in terms of writing to the target stream. Whether the worker thread blocks due to a synchronization primitive or not is orthogonal to I/O.

> If we keep the Log call synchronous, but change PipelineStopped/PipelineStarted to async, we should be able to make this fully async.

Sure, that would be nicer, and afford some minimal extra throughput on the thread calling `PipelineStarted` and `PipelineStopped`, but we'd still want to probably only await for some short-ish amount of time in case the reading/serialization loop is blocked for some reason; it's better to give up on waiting for completion (and probably log when that happens) instead of indefinitely blocking. I suppose that could be controlled by the caller via a `CancalletionToken` if we are able to plumb that through correctly.

src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/StreamingCounterLogger.cs
```diff
@@ -3,39 +3,71 @@
...
         }
 
         public void Log(ICounterPayload counter)
         {
-            //CONSIDER
-            //Ideally this would be an asynchronous api, but making this async would extend the lifetime of writing to the stream
-            //beyond the lifetime of the Counters pipeline.
-
-            SerializeCounter(_stream, counter);
+            _channelWriter.TryWrite(counter);
```
wiktork, line 42: Do we have enough to propagate an ILogger here? We could count up the amount of DropWrite's and during pipeline end report a warning saying we dropped that many events.

src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/StreamingCounterLogger.cs
```diff
@@ -3,39 +3,71 @@
...
+        {
+            try
+            {
+                ChannelReader<ICounterPayload> reader = _channel.Reader;
+                while (await reader.WaitToReadAsync())
+                {
+                    await SerializeAsync(await reader.ReadAsync());
+                }
+            }
+            finally
```
wiktork, line 67: As with the other call, can we do any logging here during failures?

src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/StreamingCounterLogger.cs
wiktork: Per offline discussion, let's keep it this way and then change the PipelineFinished/PipelineStarted callbacks once we change the counter to support System.Diagnostics.Metrics

src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/StreamingCounterLogger.cs
```diff
@@ -3,39 +3,71 @@
...
+using System.Threading;
+using System.Threading.Channels;
+using System.Threading.Tasks;
 
 namespace Microsoft.Diagnostics.Monitoring.WebApi
 {
     internal abstract class StreamingCounterLogger : ICountersLogger
     {
-        private readonly Stream _stream;
+        private const int BoundedChannelSize = 1000;
```
wiktork, line 15: Can you test if we drop anything in a resource constrained environment such as k8 with our resource recommendations + 1 sec metric interval? I'm expecting that we don't but it might help guide the optimal size of this buffer.

wiktork (dismissed):

src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/StreamingCounterLogger.cs
jander-msft: Yep, with the changes in #2845, it'll be much easier to do this.

src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/StreamingCounterLogger.cs
jander-msft: I'd like to defer this until after the changes are merged at this point, but I will definitely check this, probably add some more diagnostic logging to see how full the channel gets with different providers and numbers of counters.

PR #2863, opened 2022-11-09
Update breaking changes with details

jander-msft:
###### Summary

Add more details about the (few) breaking changes in 7.0

### DIFF
--- a/documentation/compatibility/7.0/README.md
+++ b/documentation/compatibility/7.0/README.md
@@ -40,7 +40,7 @@ When overriding the `CMD` instruction, the existing arguments will need to be re
 
 ### Egress: Built-In Metadata Is Prefixed for Azure Blob Egress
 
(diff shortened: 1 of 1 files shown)


documentation/compatibility/7.0/README.md
```diff
@@ -5,10 +5,50 @@
...
+```docker
+ENTRYPOINT [ "dotnet-monitor" ]
+CMD [ "collect", "--urls", "https://+:52323", "--metricUrls", "http://+:52325" ]
+```
+
+When overriding the `CMD` instruction, the existing arguments will need to be repeated or replaced with the desired values in addition to any additional arguments that need to be passed.
+
+### Egress: Built-In Metadata Is Prefixed for Azure Blob Egress
+
+When egressing an artifact to Azure blob storage, several metadata keys and values are applied to Auzre blob, such as `dotnet monitor` activity information, collection rule execution information, and process information. Starting in 7.0, these metadata keys are now prefixed with `DotnetMonitor_`. This change was made in order to avoid potential collisions with metadata acquired from the `metadata` description on the [Azure blob storage](../../configuration.md#azure-blob-storage-egress-provider) egress provider. The names are changed as follows:
```
kkeirstead, line 43:
```suggestion
When egressing an artifact to Azure blob storage, several metadata keys and values are applied to the Azure blob, such as `dotnet monitor` activity information, collection rule execution information, and process information. Starting in 7.0, these metadata keys are now prefixed with `DotnetMonitor_`. This change was made in order to avoid potential collisions with metadata acquired from the `metadata` description on the [Azure blob storage](../../configuration.md#azure-blob-storage-egress-provider) egress provider. The names are changed as follows:
```

kkeirstead (approved):

PR #2871, opened 2022-11-10
Refactor collection rule action tests

jander-msft:
###### Summary

This change is the first part in hopefully solving some of the instability issues around the collection rule action tests. The cause seems to be thread pool starvation such that scheduled `Task` callbacks are not executed for some time, causing the tests to timeout. The solution that I'm proposing here is to move all of the tests that use the product's collection rule actions into a new assembly and have them execute serially such that they don't overwhelm the thread pool (xUnit runs each unit test assembly in a separate process; serial test execution can be accomplished with test collections: https://xunit.net/docs/running-tests-in-parallel.html).

The current changes are focused on creating the new assembly and moving code:

1. Create new unit test assembly (named `CollectionRuleActionUnitTests`) focused on testing the product's collection rules actions.
2. Move all unit tests that use the product collection rule actions to this new assembly.
3. Move all code common to the new assembly and the existing tool unit test assembly into a new assembly (named `Microsoft.Diagnostics.Monitoring.Tool.UnitTestCommon`).
4. Adjust namespaces for code movements.
5. Fixed minor code style violations flagged by Visual Studio.
6. Fixed unit test app path calculations due to code movement.

A subsequent PR will make the adjustments to run the tests serially and to fix up the test collection rule actions (such as `PassThroughAction`) to not be susceptible to the thread pool starvation.


src/Tests/CollectionRuleActions.UnitTests/Egress/Pipe/PipeEgressOptions.cs
```diff
@@ -4,7 +4,7 @@
 
 using System.IO.Pipelines;
 
-namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests.Egress.Pipe
+namespace CollectionRuleActions.UnitTests
```
kkeirstead, line 7: Just wanted to double check that trimming off the namespace on a few of these files was intentional, feel free to ignore.

jander-msft: Yes, intentional, to match the name of the assembly to which it was moved.

kkeirstead (approved):

wiktork (approved):

PR #2914, opened 2022-11-18
Rework test infrastructure to fix flakiness

schmittjoseph:
###### TL;DR
Eliminate most/all test failures that are from flakiness (the ~60% test suite pass rate is now 100%) by changing how we run tests, relaxing timeouts, and fixing several race conditions in our tests.

###### Summary

Areas that impacted test reliability and have been fixed in this PR:
- Disable test assembly parallelization. 
- There were several race conditions in our test code and dependent libraries which were being hidden by the other flaky tests.
  - There appears to be a race condition in chained configuration building, which could manifest itself in our log test scenario.
  - https://github.com/dotnet/dotnet-monitor/pull/2909
  - `MultiProcessIdentificationTest` has a race condition around process termination.
  - `DumpTests.*` has race conditions around process discovery.
  - `StacksTests.*` has race conditions around both sending requests too quickly and process discovery.
  - `CollectionRulePipeline_EventCounterTriggerTest` is nondeterministic as it relies on the test host being able to devote enough CPU to the scenario app to trigger the CPU usage threshold. This isn't guaranteed to happen, especially on resource constrained systems. Disable it for now; one alternative is to mock out the CPU data and ensure the test data is deterministic.
 - Our timeouts are too aggressive on resource constrained systems. Significantly relax them to 3 minutes in most scenarios.
   -  Despite some tests having more tolerant timeouts, the underlying http client used still enforced a 100s timeout. This PR also relaxes the http client timeout, tying it to the test timeout.
 
TODO:
- [x] Update disabled sections of tests with a link to a tracking GitHub issue.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunnerExtensions.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunnerExtensions.cs
@@ -86,7 +86,7 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests.Runners
 
             string address = await runner.GetDefaultAddressAsync(cancellation.Token);
 
-            return await runner.CreateHttpClientAsync(factory, address, name, cancellation.Token, timeout);
+            return await runner.CreateHttpClientAsync(factory, address, name, cancellation.Token, TestTimeouts.HttpApi);
         }
 
         /// <summary>


eng/build.yml
```diff
@@ -185,10 +175,9 @@ jobs:
         -configuration ${{ parameters.configuration }}
         -architecture ${{ parameters.architecture }}
         $(_CrossBuildArgs)
-        $(_TestArgs)
```
jander-msft, line 188: The test parallelism issue due to using msbuild to run the tests should be controllable by using `ProjectToBuild` items with the `BuildInParallel` metadata set to false if we break out build from test (we probably want to still allow parallel builds). See https://github.com/dotnet/arcade/blob/main/src/Microsoft.DotNet.Arcade.Sdk/tools/Build.proj#L42 and https://github.com/dotnet/arcade/blob/main/src/Microsoft.DotNet.Arcade.Sdk/tools/Build.proj#L69. These would need to be only set when we know we are testing on the build machines or maybe controlled via a property that the yaml sets by default which would allow us to attain the same behavior locally if desired.

jander-msft (commented):
You might want to consider splitting the test fixes from the infrastructure changes. I mostly agree with the test fixes, but I think there is lots of room for getting our desired abilities for running tests using arcade.

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/ConsoleOutputHelper.cs
```diff
@@ -3,24 +3,33 @@
...
 namespace Microsoft.Diagnostics.Monitoring.TestCommon
 {
     public sealed class ConsoleOutputHelper : ITestOutputHelper
     {
-        public ConsoleOutputHelper()
+        private readonly bool _stdout;
+
+        public ConsoleOutputHelper(bool stdout = true)
         {
+            _stdout = stdout;
```
jander-msft, line 17: Could this store the `TextWriter` instance instead of constantly looking it up based on the `bool` field?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunnerExtensions.cs
```diff
@@ -84,7 +86,7 @@ public static async Task<HttpClient> CreateHttpClientDefaultAddressAsync(this Mo
 
             string address = await runner.GetDefaultAddressAsync(cancellation.Token);
 
-            return await runner.CreateHttpClientAsync(factory, address, name, cancellation.Token);
+            return await runner.CreateHttpClientAsync(factory, address, name, cancellation.Token, timeout);
```
jander-msft, line 89:
```suggestion
            return await runner.CreateHttpClientAsync(factory, address, name, cancellation.Token, TestTimeouts.HttpApi);
```

I think this should be set to the HttpApi timeout explicitly. While in practice the `timeout` value is probably source from this same static value, the parameter shouldn't be used for both the timeout associated with the creation of the `HttpClient` and the timeout for a HTTP invocation.

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/ConsoleOutputHelper.cs
jander-msft: Yeah, I think it's okay to cache that without disposing it.

jander-msft (approved):

PR #2919, opened 2022-11-21
Add Tags for API Requests

### LINKED ISSUE #2783: Tag For API Requests
## Background and Motivation

Currently, users need to know the `operationId` or timestamp for an operation in order to reference it via the `/operations` endpoint. This would allow users to optionally add a `tag` to their API requests to make it easier to identify specific operations.

## Usage Examples

```http
GET /trace?tag=my_tagged_trace
```

```http
GET /operations


HTTP/1.1 200 OK
Content-Type: application/json

[
    {
        "operationId": "67f07e40-5cca-4709-9062-26302c484f18",
        "createdDateTime": "2021-07-21T06:21:15.315861Z",
        "status": "Succeeded", 
        "process":
        {
            "pid":1,
            "uid":"95b0202a-4ed3-44a6-98f1-767d270ec783",
            "name":"dotnet-monitor-demo"
        },
        "tag": "my_tagged_trace"
    }
]
```

kkeirstead:
###### Summary

This change allows users to add a tag to an operation to make it easier to identify when looking at the operations table.

Closes #2783 

###### Release Notes Entry

Add Tags For API Requests


documentation/api/definitions.md
```diff
@@ -297,6 +299,12 @@ Detailed information about an operation.
     "status": "Succeeded",
     "egressProviderName": "monitorBlob",
     "isStoppable": false,
+    "process": {
```
kkeirstead, line 302: Note that this wasn't changed in this PR - I just noticed that this was missing while reviewing my own change.

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationStore.cs
```diff
@@ -212,7 +223,8 @@ public IEnumerable<Models.OperationSummary> GetOperations(ProcessKey? processKey
                                 Name = processInfo.ProcessName,
                                 ProcessId = processInfo.ProcessId,
                                 Uid = processInfo.RuntimeInstanceCookie
-                            } : null
```
wiktork, line 215: Just to be clear, this allows us to specify Process, tag, or both correct?

wiktork (dismissed):

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationStore.cs
kkeirstead: Correct - you can now narrow down your query by specifying one (or both) of them

jander-msft:
Consider a design where an operation has a `tags` property that is an array of `string` so that it can be future proof for having multiple tags, even if it only has zero or one entries in it for now.

wiktork (dismissed):

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationStore.cs
```diff
@@ -161,12 +164,22 @@ public void CompleteOperation(Guid operationId, ExecutionResult<EgressResult> re
...
             {
                 IEnumerable<KeyValuePair<Guid, EgressEntry>> requests = _requests;
 
+                if (!string.IsNullOrEmpty(tags))
+                {
+                    ISet<string> tagsSet = Utilities.SplitTags(tags);
+
+                    requests = requests.Where((kvp) =>
+                    {
+                        return tagsSet.All(tag => kvp.Value.Tags.Contains(tag));
```
jander-msft, line 179:
```suggestion
                        return tagsSet.IsSubsetOf(kvp.Value.Tags);
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/Utilities.cs
```diff
@@ -40,5 +42,15 @@ public static KeyValueLogScope CreateArtifactScope(string artifactType, IEndpoin
...
         }
+
+        public static ISet<string> SplitTags(string tags)
+        {
+            if (string.IsNullOrEmpty(tags))
+            {
+                return new HashSet<string>();
+            }
+
+            return tags.Split(',', StringSplitOptions.RemoveEmptyEntries).ToHashSet();
```
jander-msft, line 53:
```suggestion
            return tags.Split(',', StringSplitOptions.RemoveEmptyEntries).ToHashSet(StringComparer.Ordinal);
```

assuming casing should matter e.g. `ThisTag` != `thisTag`

jander-msft (dismissed):
Assuming that we don't want to allow commas to be part of a tag, then this looks good to me.

jander-msft (approved):

PR #2939, opened 2022-11-29
Add Metadata Support For Metrics

### LINKED ISSUE #430: Allow Prometheus metric labels
## Background and Motivation

It is currently not possible to publish metrics with labels.

It would be useful to provide a mechanism for `DiagnosticCounter`  to expose labels.

## Proposed Feature

Parse out the `Name` property of `ICounterPayload` to extract the labels according the Prometheus native notation.

See [Prometheus doc](https://prometheus.io/docs/concepts/data_model/).
 
## Usage Examples

A `DiagnosticCounter` with the name *my_counter_name{label1="value1", label2="value2"}*.

I have done a quick prototype for my experimentation and the result is nice.

kkeirstead:
###### Summary

These changes are the second part of addressing #430 (first-part can be found in the [diagnostics repo](https://github.com/dotnet/diagnostics/pull/3498)). This allows dotnet-monitor to show metadata in metrics and in live metrics.

Closes #430 

###### Release Notes Entry

Add Metadata Support For Metrics

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsStore.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsStore.cs
@@ -95,7 +95,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
                 string metricName = PrometheusDataModel.GetPrometheusNormalizedName(metricInfo.Provider, metricInfo.Name, metricInfo.Unit);
                 string metricType = "gauge";
 
-                var keyValuePairs = from pair in metricInfo.Metadata select pair.Key + "=" + "\"" + pair.Value + "\"";
+                var keyValuePairs = from pair in metricInfo.Metadata select PrometheusDataModel.GetPrometheusNormalizedLabel(pair.Key, pair.Value);
                 string metricLabels = string.Join(", ", keyValuePairs);
 
                 //TODO Some clr metrics claim to be incrementing, but are really gauges.
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsStore.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsStore.cs
@@ -95,7 +95,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
                 string metricName = PrometheusDataModel.GetPrometheusNormalizedName(metricInfo.Provider, metricInfo.Name, metricInfo.Unit);
                 string metricType = "gauge";
 
-                var keyValuePairs = from pair in metricInfo.Metadata select pair.Key + ":" + pair.Value;
+                var keyValuePairs = from pair in metricInfo.Metadata select pair.Key + "=" + "\"" + pair.Value + "\"";
                 string metricLabels = string.Join(", ", keyValuePairs);
 
                 //TODO Some clr metrics claim to be incrementing, but are really gauges.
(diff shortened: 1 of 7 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsStore.cs
```diff
@@ -95,6 +95,9 @@ public async Task SnapshotMetrics(Stream outputStream, CancellationToken token)
                 string metricName = PrometheusDataModel.GetPrometheusNormalizedName(metricInfo.Provider, metricInfo.Name, metricInfo.Unit);
                 string metricType = "gauge";
 
+                var keyValuePairs = from pair in metricInfo.Metadata select pair.Key + "=" + "\"" + pair.Value + "\"";
```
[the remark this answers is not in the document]
kkeirstead: Wanted to get people's opinions on this - limiting characters for the keys hasn't been an issue using the existing `NormalizeString` method (other than the quirk of potentially having two distinct keys become identical when replacing illegal characters with underscores). However, for the values (according to https://prometheus.io/docs/instrumenting/exposition_formats/), I'm concerned about cases where there might be some ambiguity about how the users do their formatting. Specifically, if a user wants a newline, their value would include `\n`; however, it seems that the expectation is that users aren't necessarily going to escape other things (like a single `\`), hence the need to do cleaning. 

**Example**:
Tag Value: naughty\nice
Interpretation 1: naughty(newline)ice, which should be left as `naughty\nice`
Interpretation 2: naughty(slash)nice, which should be transformed into `naughty\\nice`

There are some other edge cases like that since we don't know what kind of escaping a user will/won't have already done...unless I'm overlooking something, which is entirely possible.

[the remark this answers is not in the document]
jander-msft: I don't think there is any way we can reason about the difference between the interpretations in the example provided. I think it's best that we just take the value, escape it according to the Prometheus docs, and write it out.

src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/PrometheusDataModel.cs
```diff
@@ -58,6 +74,45 @@ public static string GetPrometheusNormalizedValue(string unit, double value)
...
+                    }
+                    else if (i < value.Length - 1 && value[i + 1] == QuotationChar)
+                    {
+                        builder.Append(QuotationChar);
+                        ++i; // Skip over the '"'
+                    }
+                    else
+                    {
+                        builder.Append(SlashChar);
+                    }
```
jander-msft, line 102: This doesn't seem quite right. I think it should just escape a backslash as two backslashes, similar to how you are handling the double quote. Separately, you need to handle when `value[i] == '\n'`, and write that out as a literal backslash and a literal `n`.

src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/PrometheusDataModel.cs
```diff
@@ -38,17 +42,30 @@ public static string GetPrometheusNormalizedName(string metricProvider, string m
...
+                builder.Append(SeparatorChar);
                 NormalizeString(builder, baseUnit, isProvider: false);
             }
 
             return builder.ToString();
         }
 
+        public static string GetPrometheusNormalizedLabel(string key, string value)
+        {
+            StringBuilder builder = new StringBuilder(key.Length + 2 * value.Length + 3); // Includes =,", and ", as well as extra padding for potential escape characters in the value
```
jander-msft, line 58: nit: We might want to reuse the `StringBuilder` between invocations if there are many metadata pairs that are written out; this method would then call `EnsureCapacity` to make sure it has enough room for the current metadata item.

jander-msft (approved):

src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/PrometheusDataModel.cs
kkeirstead: I figure we'll want to keep this consistent with whatever we do in `GetPrometheusNormalizedName` (which is what I modeled this on), and it looks like there's already a note about potentially optimizing by using Span/stackalloc there. With that in mind, I think I'll merge this as-is, but if you're still concerned about performance, there can be another PR for investigating changing our logic in these methods.

PR #2941, opened 2022-11-29
Add support for thread names stack artifacts.

wiktork:
###### Summary
Add support for thread names in the callstack artifact.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Strings.resx
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Strings.resx
@@ -118,7 +118,7 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="CallstackThreadHeader" xml:space="preserve">
-    <value>Thread: (0x{0:X}) {1}</value>
+    <value>Thread: (0x{0:X})</value>
   </data>
   <data name="ErrorMessage_CanNotGetEnvironment" xml:space="preserve">
     <value>Unable to get process environment.</value>
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Strings.resx
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Strings.resx
@@ -118,7 +118,7 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="CallstackThreadHeader" xml:space="preserve">
-    <value>Thread: (0x{0:X})</value>
+    <value>Thread: (0x{0:X}) {1}</value>
   </data>
   <data name="ErrorMessage_CanNotGetEnvironment" xml:space="preserve">
     <value>Unable to get process environment.</value>
--- a/src/MonitorProfiler/EventProvider/ProfilerEvent.h
+++ b/src/MonitorProfiler/EventProvider/ProfilerEvent.h
@@ -111,21 +111,10 @@ template<typename... Args>
 template<size_t index, typename T, typename... TArgs>
 HRESULT ProfilerEvent<Args...>::WritePayload(COR_PRF_EVENT_DATA* data, const tstring& first, TArgs... rest)
 {
-    // Must be valid when next function is called
-    tstring empty;
-
-    if (first.size() == 0)
-    {
-        data[index].ptr = reinterpret_cast<UINT64>(empty.c_str());
-        data[index].size = sizeof(WCHAR);
-        data[index].reserved = 0;
-    }
-    else
-    {
-        data[index].ptr = reinterpret_cast<UINT64>(first.c_str());
-        data[index].size = static_cast<UINT32>((first.size() + 1) * sizeof(WCHAR));
-        data[index].reserved = 0;
-    }
+    //Note this works for empty strings.
+    data[index].ptr = reinterpret_cast<UINT64>(first.c_str());
+    data[index].size = static_cast<UINT32>((first.size() + 1) * sizeof(WCHAR)); // + 1 for null terminator.
+    data[index].reserved = 0;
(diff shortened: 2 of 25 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/StacksTests.cs
```diff
@@ -315,8 +320,10 @@ public Task TestCollectStacksAction(Architecture targetArchitecture)
```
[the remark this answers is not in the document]
wiktork: My plan was to reenable this after Joe's test stability fixes go through.

src/Microsoft.Diagnostics.Monitoring.WebApi/Strings.resx
```diff
@@ -118,7 +118,7 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="CallstackThreadHeader" xml:space="preserve">
```
jander-msft, line 120: nit: Consider that we may not want to write out an extra space on the end if the thread happens to not have a name; this will look out of place when writing it out as JSON. We may need two resource strings for this.

src/MonitorProfiler/EventProvider/ProfilerEvent.h
```diff
@@ -111,10 +111,13 @@ template<typename... Args>
 template<size_t index, typename T, typename... TArgs>
 HRESULT ProfilerEvent<Args...>::WritePayload(COR_PRF_EVENT_DATA* data, const tstring& first, TArgs... rest)
 {
+    // Must be valid when next function is called
+    tstring empty;
```
jander-msft, line 115: nit: Consider statically allocating this.

src/MonitorProfiler/EventProvider/ProfilerEvent.h
```diff
@@ -111,10 +111,13 @@ template<typename... Args>
...
+
     if (first.size() == 0)
     {
-        data[index].ptr = 0;
-        data[index].size = 0;
+        data[index].ptr = reinterpret_cast<UINT64>(empty.c_str());
+        data[index].size = sizeof(WCHAR);
```
jander-msft, line 120: This feels kind of like magic constant... Is this `sizeof(WCHAR)` because an empty string is null-terminated without any other characters? Maybe add a comment for it.

src/MonitorProfiler/EventProvider/ProfilerEvent.h
jander-msft: And why does a new string object need to be used? Why can't `first` be used? Is that on the stack?

src/MonitorProfiler/MainProfiler/MainProfiler.h
```diff
@@ -9,7 +9,9 @@
 #include "../Environment/EnvironmentHelper.h"
 #include "../Logging/Logger.h"
 #include "../Communication/CommandServer.h"
+#include "../Utilities//ThreadNameCache.h"
```
jander-msft, line 12:
```suggestion
#include "../Utilities/ThreadNameCache.h"
```

src/MonitorProfiler/Stacks/StackSampler.h
```diff
@@ -10,6 +10,7 @@
 #include "tstring.h"
 #include "Stack.h"
 #include "../Utilities/NameCache.h"
+#include "../Utilities//ThreadNameCache.h"
```
jander-msft, line 13:
```suggestion
#include "../Utilities/ThreadNameCache.h"
```

src/MonitorProfiler/Utilities/ThreadNameCache.cpp
```diff
@@ -0,0 +1,30 @@
+// Licensed to the.NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#include "ThreadNameCache.h"
+
+void ThreadNameCache::Set(ThreadID id, tstring&& name)
```
jander-msft, line 7: nit: Seeing this in the rest of the code, it would not be obvious to me that the name that is passed in would be reset after this function exits. Should be rename this function to be clearer about it? I don't have a good suggestion, nor do I feel strongly about this; I just think this might trip us up later if we try to use the thread name after calling this function.

jander-msft (approved):

src/MonitorProfiler/Utilities/ThreadNameCache.cpp
```diff
@@ -0,0 +1,35 @@
...
+    std::lock_guard<std::mutex> lock(_mutex);
+    _threadNames[id] = std::move(name);
+}
+
+void ThreadNameCache::Set(ThreadID id, const tstring& name)
+{
+    Set(id, tstring(name));
+}
+
+bool ThreadNameCache::Get(ThreadID id, std::wstring& name)
```
jander-msft, line 18:
```suggestion
bool ThreadNameCache::Get(ThreadID id, tstring& name)
```

jander-msft (dismissed):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/StacksTests.cs
```diff
@@ -310,7 +315,7 @@ public Task TestInProcessFeaturesNotEnabled(Architecture targetArchitecture)
                 });
         }
 
-        [Theory(Skip = "Disable unstable tests.")]
+        [Theory]
```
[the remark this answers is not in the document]
jander-msft: I just put out a PR that does this check for all tests that use `ScenarioHelper.SingleTarget`, which includes this test: #2990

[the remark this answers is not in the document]
wiktork: Since this is a trigger-based test that suspends the application, I'm not sure this is the issue, but I will do some testing around it. I will disable the test for this PR.

jander-msft (approved):

PR #3035, opened 2022-12-07
Egress Extensibility Testing - Part 1

kkeirstead:
###### Summary

This is the first piece of testing for egress extensibility. It covers finding extensions in each of the extension locations and starting up a test extension (`EgressExtensibilityApp`) and getting a success/failure response from it.


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EgressExtensibilityTests.cs
```diff
@@ -0,0 +1,194 @@
...
+        private static readonly TimeSpan DefaultTimeout = TimeSpan.FromMinutes(3);
+
+        private ITestOutputHelper _outputHelper;
+
+        private const string ExtensionsFolder = "extensions";
+        public const string SampleArtifactPath = "sample\\path";
+        public const string SampleFailureMessage = "the extension failed";
+        private const string ProviderName = "TestingProvider"; // Must match the name in extension.json
+        private const string AppName = "Microsoft.Diagnostics.Monitoring.EgressExtensibilityApp";
+        private const string AppExe = AppName + ".exe";
```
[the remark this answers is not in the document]
kkeirstead: Ah, good call - I had only run it locally and didn't think of that. I'll update this.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EgressExtensibilityTests.cs
```diff
@@ -0,0 +1,194 @@
...
+        private const string DotnetToolsExtensionDir = ".store\\tool-name\\7.0\\tool-name\\7.0\\tools\\net7.0\\any";
```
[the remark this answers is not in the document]
kkeirstead: I'll add a TODO for this - I currently have this fixed to net7.0 in ToolsExtensionRepository, which will also need to be changed to be dynamic.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EgressExtensibilityTests.cs
```diff
@@ -0,0 +1,194 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    public sealed class EgressExtensibilityTests
+    {
+        private static readonly TimeSpan DefaultTimeout = TimeSpan.FromMinutes(3);
```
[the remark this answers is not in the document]
kkeirstead: I'll add a TODO for this as well - my feature branch doesn't have that change yet

src/Tests/Microsoft.Diagnostics.Monitoring.EgressExtensibilityApp/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            return rootCommand.Invoke(args);
+        }
+
+        private static int Egress()
+        {
+            EgressArtifactResult result = new();
+            try
+            {
+                string jsonConfig = Console.ReadLine();
```
wiktork, line 34: Not directly related to your changes, but it would be great if we had a test and example of how we want extensibility writers to consume configuration and artifacts.

src/Tools/dotnet-monitor/HostBuilder/HostBuilderSettings.cs
```diff
@@ -42,6 +55,22 @@ internal sealed class HostBuilderSettings
...
+        // Other: "%XDG_CONFIG_HOME%/.dotnet/tools" OR "%HOME%/.dotnet/tools" -> THIS HAS NOT BEEN TESTED YET ON LINUX
+        public static readonly string DotnetToolsExtensionDirectoryPath =
+            GetEnvironmentOverrideOrValue(
+                DotnetToolsExtensionDirectoryOverrideEnvironmentVariable,
+                RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ?
+                    Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), "." + DotnetFolderName, ToolsFolderName) :
+                    Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "." + DotnetFolderName, ToolsFolderName));
+
+        // Location for dotnet-monitor's executing assembly.
+        public static readonly string ExecutingAssemblyDirectoryPath =
```
jander-msft, line 69: I don't think this override is necessary: since "next to me", the local user settings, and the shared settings extension discoverers are the same implementation, you can choose to drop the extension in the already existing directory specified by DotnetMonitorTestSettings__UserConfigDirectoryOverride; testing this is the equivalent of testing any of the others that use the same implementation.

src/Tools/dotnet-monitor/HostBuilder/HostBuilderSettings.cs
```diff
@@ -42,6 +55,22 @@ internal sealed class HostBuilderSettings
                     Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), "." + ProductFolderName) :
                     Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), ProductFolderName));
 
+        // Location where extensions are stored by default.
+        // Windows: "%USERPROFILE%\.dotnet\Tools"
+        // Other: "%XDG_CONFIG_HOME%/.dotnet/tools" OR "%HOME%/.dotnet/tools" -> THIS HAS NOT BEEN TESTED YET ON LINUX
+        public static readonly string DotnetToolsExtensionDirectoryPath =
```
jander-msft, line 61: I don't think this override is necessary; use the test hosting startup to inject an override:
1. Change this to be a service (e.g. IDotnetToolsFileSystem.Path) that the ToolsExtensionRepository will consume.
2. Register this new service with TryAddSingleton (e.g. https://github.com/dotnet/dotnet-monitor/blob/main/src/Tools/dotnet-monitor/ServiceCollectionExtensions.cs#L244)
3. Provide an override implementation in the test hosting startup (e.g. https://github.com/dotnet/dotnet-monitor/blob/main/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.TestHostingStartup/HostingStartup.cs#L19)

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
```diff
@@ -117,6 +134,14 @@ public async Task<EgressArtifactResult> EgressArtifact(ExtensionEgressPayload co
             _logger.ExtensionEgressPayloadCompleted(p.Id);
 
             EgressArtifactResult result = await parser.ReadResult();
+
+            p.Kill();
```
jander-msft, line 138: Consider allowing the program to exit gracefully by using `Process.WaitForExitAsync`. You can use `CancellationTokenSource.CreateLinkedTokenSource` if you want to use a timeout as well so you don't indefinitely wait on it before attempting to kill it.

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
```diff
@@ -90,6 +106,7 @@ public async Task<EgressArtifactResult> EgressArtifact(ExtensionEgressPayload co
                 UseShellExecute = false,
             };
             pStart.ArgumentList.Add(ExtensionTypes.Egress);
+            _processEnvironmentVariables.ToList().ForEach(x => pStart.EnvironmentVariables[x.Key] = x.Value);
```
jander-msft, line 109:
```suggestion
            foreach ((string key, string value) in _processEnvironmentVariables)
            {
                _pStart.Environment.Add(key, value);
            }
```

Or write a `DictionaryExtensions.AddRange<TKey, TValue>(this IDictionary<TKey, TValue> target, IDictionary<TKey, TValue> source)` extension method that does this and call it here if you think this is reusable.

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
```diff
@@ -65,7 +73,15 @@ public async Task<EgressArtifactResult> EgressArtifact(ExtensionEgressPayload co
             // This is really weird, yes, but this is one of 2 overloads for [Stream].WriteAsync(...) that supports a CancellationToken, so we use a ReadOnlyMemory<char> instead of a string.
             ReadOnlyMemory<char> NewLine = new ReadOnlyMemory<char>("\r\n".ToCharArray());
 
-            string programRelPath = Path.Combine(Path.GetDirectoryName(_exePath), Declaration.Program);
+            string exeName = Declaration.Program;
+
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && Path.GetExtension(exeName) == ".exe")
```
jander-msft, line 78: I think I would rather have it the opposite way: extensions declare their executable without a file extension and then we tack on ".exe" for the Windows platform.

src/Tests/Microsoft.Diagnostics.Monitoring.EgressExtensibilityApp/Program.cs
```diff
@@ -0,0 +1,83 @@
...
```
jander-msft, line 34: `Console.ReadLine` and `Console.WriteLine` use line endings that are platform dependent e.g. Windows is `\r\n` and Unix-like (includes Linux and MacOS) is typically `\n`. I'm not sure that we want the protocol to be platform dependent. It might be okay since this is stdin/stdout on the same machine, but if we for some reason decide to remote the protocol usage (e.g. over the network), then the two machines aren't necessarily the same platform.

src/Tests/Microsoft.Diagnostics.Monitoring.EgressExtensibilityApp/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+                }
+            }
+            catch (Exception ex)
+            {
+                result.Succeeded = false;
+                result.FailureMessage = ex.Message;
+            }
+
+            string jsonBlob = JsonSerializer.Serialize(result);
+            Console.Write(jsonBlob);
```
jander-msft, line 57:
```suggestion
            Console.WriteLine(jsonBlob);
```

By writing the newline, it's more clear to the consumer that the entire response was written.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EgressExtensibilityTests.cs
```diff
@@ -0,0 +1,214 @@
...
+                    return Path.GetExtension(path) == string.Empty;
+                }
+            }
+
+            return false;
+        }
+
+        private HostBuilderSettings CreateHostBuilderSettings()
+        {
+            TemporaryDirectory executingAssemblyDir = new(_outputHelper);
```
jander-msft, line 174: All of these are disposable, so you are leaking temporary directories for each invocation of this method.

jander-msft (changes requested):

src/Tests/Microsoft.Diagnostics.Monitoring.EgressExtensibilityApp/Program.cs
kkeirstead: Not sure if you wanted me to take action on this point, but I did want to point out that this particular example is for testing, not the actual extension - we can discuss this further if it's a concern though.

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
kkeirstead: Let me know if what I did looks right for this - I tried to reference some similar existing instances in our code.

src/Tools/dotnet-monitor/HostBuilder/HostBuilderSettings.cs
kkeirstead: I wasn't sure exactly what you wanted here, so I took my best guess and ripped out the functionality surrounding setting this via environment variables since it wasn't needed - I can change this if you want something else.

src/Tools/dotnet-monitor/HostBuilder/HostBuilderSettings.cs
kkeirstead: It may have been a result of my implementation, but this ended up being a bit intrusive. Previously, in `ServiceCollectionExtensions`, we used the `targetExtensionFolder` when we add the extension repository - now, if we're getting that from the `IServiceProvider`, my understanding is that we don't know what that value is when we're configuring our services. The main fallout of this was having to move some checks around for whether the directory exists, as well as adding a null check within `FindExtension` in `ExtensionDiscoverer` in case the `targetExtensionFolder` doesn't exist. I wasn't sure if you were expecting me to need to work around these things as part of the change, or if there's a cleaner way to do it.

src/Tools/dotnet-monitor/HostBuilder/HostBuilderSettings.cs
```diff
@@ -4,6 +4,7 @@
 
 using System;
```
jander-msft, line 5: I think all of the changes in this file should be undone.

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
```diff
@@ -117,6 +140,27 @@ public async Task<EgressArtifactResult> EgressArtifact(ExtensionEgressPayload co
...
+
+                using CancellationTokenSource linkedTokenSource = CancellationTokenSource.CreateLinkedTokenSource(token, timeoutSource.Token);
+
+                await p.WaitForExitAsync(linkedTokenSource.Token).ConfigureAwait(false);
+            }
+            catch (OperationCanceledException) when (timeoutSource.IsCancellationRequested)
+            {
+                p.Kill();
+
+                while (!p.HasExited)
```
jander-msft, line 158: I think you can use `await p.WaitForExitAsync(token);` here instead of polling.

src/Tools/dotnet-monitor/Extensibility/DefaultDotnetToolsFileSystem.cs
```diff
@@ -0,0 +1,26 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    internal sealed class DefaultDotnetToolsFileSystem :
+        IDotnetToolsFileSystem
+    {
+        private const string DotnetFolderName = "dotnet";
```
jander-msft, line 13: Make this `.dotnet` so you don't have to prepend `.` at every usage.

src/Tools/dotnet-monitor/ServiceCollectionExtensions.cs
```diff
@@ -216,24 +215,23 @@ public static IServiceCollection ConfigureExtensions(this IServiceCollection ser
         {
             // Add the services to discover extensions
             services.AddSingleton<ExtensionDiscoverer>();
+            services.TryAddSingleton<IDotnetToolsFileSystem, DefaultDotnetToolsFileSystem>();
```
jander-msft, line 218: I think all of the changes in this file should be undone _except_ for this line. The folder location that is next to the assembly shouldn't be configurable by test IMO. Like I mentioned before, testing the user and shared locations shares an equivalency with the "next to assembly" location, so it doesn't need to be explicitly tested, which simplifies the override knobs.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EgressExtensibilityTests.cs
```diff
@@ -0,0 +1,214 @@
...
+                    return Path.GetExtension(path) == string.Empty;
+                }
+            }
+
+            return false;
+        }
+
+        private HostBuilderSettings CreateHostBuilderSettings()
+        {
+            using TemporaryDirectory executingAssemblyDir = new(_outputHelper);
```
jander-msft, line 176: Disposing these here is causing a usage after disposal problem (the directories no longer exist once something actually tries to read them off the `HostBuilderSettings` instance. You will probably have to inline the TemporaryDirectory usage and the settings creation with each test method.

src/Tests/Microsoft.Diagnostics.Monitoring.EgressExtensibilityApp/Program.cs
jander-msft: I suppose that's okay to push down the road, but I don't want this to proliferate to additional copies and others start playing with extensions assuming this is okay.

src/Tests/Microsoft.Diagnostics.Monitoring.EgressExtensibilityApp/Microsoft.Diagnostics.Monitoring.EgressExtensibilityApp.csproj
```diff
@@ -0,0 +1,25 @@
...
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>$(ToolTargetFrameworks)</TargetFrameworks>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.CommandLine" Version="$(SystemCommandLineVersion)" />
+    <PackageReference Include="xunit.assert" Version="$(XUnitVersion)" />
+  </ItemGroup>
+
+  <ItemGroup>
```
jander-msft, line 13: Why are any of these project references needed? Real world usage won't be taking references to these assemblies; it's hard to understand what is actually being used. I suspect you are only using a very small subset, which then I would recommend linking in those source files so we can more easily understand their usage if/when we refactor them out in the future.

src/Tools/dotnet-monitor/HostBuilder/HostBuilderSettings.cs
jander-msft: > The main fallout of this was having to move some checks around for whether the directory exists, as well as adding a null check within `FindExtension` in `ExtensionDiscoverer` in case the `targetExtensionFolder` doesn't exist. I wasn't sure if you were expecting me to need to work around these things as part of the change, or if there's a cleaner way to do it.

I would push that particular directory check into `ToolsExtensionRepository` and have it preemptively fail to return extensions if they don't exist.

kkeirstead: That was my initial thought too, but when I did that I got an error when the `createDelegate` in `ServiceCollectionExtensions` would attempt to create a `PhysicalFileProvider` using a path that didn't exist. Technically, I guess the responsibility for creating the `IFileProvider` could be pushed down to the `*ExtensionRepository` level as part of `TryFindExtension` if it doesn't already exist, but I'm still concerned about what I referenced in my second comment above.

jander-msft: > I'm still concerned about what I referenced in my second comment above.

I can see the individual configuration file notion for the user and shared settings path fairly easily. I'm not sure that would translate to the dotnet tools directory though since it's not a simple "take all json files in this folder" approach. It would have to use the same algorithm that your discovery code uses and likely be implemented as a custom configuration source.

jander-msft: > When I did that I got an error when the `createDelegate` in `ServiceCollectionExtensions` would attempt to create a `PhysicalFileProvider` using a path that didn't exist.

You can return a `NullFileProvider` if the directory doesn't exist, so maybe a static method that checks if it exists and then returns wither `PhysicalFileProvider` or `NullFileProvider`.

src/Tools/dotnet-monitor/Extensibility/IDotnetToolsFileSystem.cs
```diff
@@ -0,0 +1,14 @@
...
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    /// <summary>
+    /// Interface used for calling an Egress Extension
+    /// </summary>
+    internal interface IDotnetToolsFileSystem
+    {
+        public string Path { get; set; }
```
jander-msft, line 12:
```suggestion
        string Path { get; set; }
```

src/Tests/Microsoft.Diagnostics.Monitoring.EgressExtensibilityApp/Microsoft.Diagnostics.Monitoring.EgressExtensibilityApp.csproj
```diff
@@ -0,0 +1,27 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>$(ToolTargetFrameworks)</TargetFrameworks>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.CommandLine" Version="$(SystemCommandLineVersion)" />
+    <PackageReference Include="xunit.assert" Version="$(XUnitVersion)" />
```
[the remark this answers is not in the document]
kkeirstead: I was using it before, but not in the current iteration. Removed.

jander-msft (approved):

PR #3040, opened 2022-12-08
Merge Main into feature/egressExtension

kkeirstead:
###### Summary

Merging main into feature branch for egress extensibility. As a result, this integrated the S3 changes - however, these changes have not been incorporated into the extensibility model yet and were not refactored as part of this merge. 

Most manual changes made as part of this merge were related to code cleanliness (e.g. `sealed`, default null values, etc.).


src/Tests/CollectionRuleActions.UnitTests/CollectLogsActionTests.cs
```diff
@@ -43,7 +51,7 @@ public CollectLogsActionTests(ITestOutputHelper outputHelper)
         /// at the log level specified in the request body.
         /// </summary>
         /// 
-        [Theory]
+        [Theory(Skip = "Haven't integrated PipeEgress into egress extensibility model yet")]
```
kkeirstead, line 54: @jander-msft These tests were failing due to the addition of the `PipeEgressProvider` - my thinking is that it's probably best to temporarily disable them since it would be hard to review the changes required to support it as part of the merge PR, but I'll leave that up to you.

jander-msft: Makes sense. One way to re-enable this (not necessarily now) is to make this egress provider an extension and have it communicate back over named pipe or Unix domain socket.

PR #3043, opened 2022-12-09
Pull requests only run .NET 8 tests

jander-msft:
###### Summary

For pull requests, only run .NET 8 tests. For CI builds, run all tests. Remove scheduled build since CI builds will now run all tests.


jander-msft:
> Mind also updating `testResultTfms` in `eng/build.yml`? It looks like net8.0 is missing there

Fixing in #3044

PR #3082, opened 2022-12-12
[release/7.x] Allow pivoting on released vs development versions of diagnostics libraries

jander-msft:
###### Summary

This change allows easy pivoting of the diagnostic library versions between public releases and development versions. Change `UseMicrosoftDiagnosticsMonitoringShippedVersion` to `true` in order to use the publicly shipped version or `false` to use the development version.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -1,12 +1,8 @@
 <Dependencies>
   <ProductDependencies>
-    <Dependency Name="Microsoft.Diagnostics.Monitoring" Version="5.0.328102">
+    <Dependency Name="Microsoft.Diagnostics.Monitoring" Version="6.0.0-preview.22609.1">
       <Uri>https://github.com/dotnet/diagnostics</Uri>
-      <Sha>ab657cc109f93e48d52f6b12cd7f136bb2d0b311</Sha>
-    </Dependency>
-    <Dependency Name="Microsoft.Diagnostics.Monitoring.EventPipe" Version="5.0.328102">
-      <Uri>https://github.com/dotnet/diagnostics</Uri>
-      <Sha>ab657cc109f93e48d52f6b12cd7f136bb2d0b311</Sha>
+      <Sha>5002b6c90d80f205fad75674f159dee74d4ff19a</Sha>
     </Dependency>
     <Dependency Name="System.CommandLine" Version="2.0.0-beta4.22568.1">
       <Uri>https://github.com/dotnet/command-line-api</Uri>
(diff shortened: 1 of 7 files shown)


eng/Versions.props
```diff
@@ -16,6 +16,7 @@
       - 'release': sets the blob group release name to 'release'. Can be used for prereleases and full releases.
     -->
     <BlobGroupBuildQuality>release</BlobGroupBuildQuality>
+    <UseMicrosoftDiagnosticsMonitoringShippedVersion>false</UseMicrosoftDiagnosticsMonitoringShippedVersion>
```
wiktork, line 19: It would probably be a good idea to add a step in our release process, something along the lines of "Ensure UseMicrosoftDiagnosticsMonitoringShippedVersion = true"

wiktork (dismissed):

eng/Version.Details.xml
```diff
@@ -1,12 +1,8 @@
 <Dependencies>
   <ProductDependencies>
-    <Dependency Name="Microsoft.Diagnostics.Monitoring" Version="5.0.328102">
+    <Dependency Name="Microsoft.Diagnostics.Monitoring" Version="6.0.0-preview.22609.1">
       <Uri>https://github.com/dotnet/diagnostics</Uri>
-      <Sha>ab657cc109f93e48d52f6b12cd7f136bb2d0b311</Sha>
-    </Dependency>
-    <Dependency Name="Microsoft.Diagnostics.Monitoring.EventPipe" Version="5.0.328102">
```
wiktork, line 7: Was this removed intentionally?

eng/Version.Details.xml
```diff
@@ -1,12 +1,8 @@
 <Dependencies>
   <ProductDependencies>
-    <Dependency Name="Microsoft.Diagnostics.Monitoring" Version="5.0.328102">
```
wiktork, line 3: Will this cause any issues with darc for release build prep?

eng/Version.Details.xml
jander-msft: It was, but I can add it back so that development versions are used independently.

eng/Version.Details.xml
jander-msft: I don't think it will.

eng/Versions.props
jander-msft: Done

eng/Version.Details.xml
jander-msft: I've separated the versions for each of the libraries so that they are specified independently for darc updates and identically for dependabot updates.

wiktork (approved):

PR #2654, opened 2022-10-11
Functional Tests For HTTP Trace Followed By AspNet* Triggers

kkeirstead:
Added two test cases that exercise the API followed by a collection rule; these tests will fail (and are currently turned off) until https://github.com/dotnet/diagnostics/pull/3425 is integrated.

PR #3103, opened 2022-12-13
[release/7.x] Functional Tests For HTTP Trace Followed By AspNet* Triggers (#2654)

wiktork:


src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/TraceTestUtilities.cs
```diff
@@ -1,5 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
kkeirstead, line 3: Use the new two-liner license

PR #3126, opened 2022-12-15
Remove component governance exclusion for Alpine builds

jander-msft:
###### Summary

Example build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2067800&view=results


jander-msft:
The example build shows that GC still does not work on Alpine x64 builds.

PR #3204, opened 2023-01-03
Create platform and TFM specific archives

jander-msft:
###### Summary

These changes add the ability to produce archives (zip for Windows, gzip-compressed tarballs for non-Windows) as part of build. The pipelines have been updated to produce and sign these archives but not to publish them (they will only be available as build artifacts until all of the outstanding issues are resolved).

- Add `-archive` switch to build scripts to optionally produce archives after building the managed and native components.
- The archive is built for the current platform, current build architecture, and latest target framework (namely .NET 7).
- The archives are produced from the `dotnet-monitor-archive.proj` file using Arcade's `Microsoft.DotNet.Build.Tasks.Archives` build task
  - It builds the `dotnet-monitor.csproj` project as a framework dependent, RID-specific (e.g. `win-x64`) output excluding the apphost file (e.g. `dotnet-monitor.exe`)
  - It produces a product archive with just the product binaries and related files it in (no symbol files) and a separate symbols archive that only contains symbol files (no libraries, assemblies, etc).
- The archive includes the profiler assembly only for the corresponding platform and build architecture e.g. Windows x64 archive only contains the `win-x64` profiler.

Example archive name: `dotnet-monitor-7.1.0-rtm.23052.1-win-x64.zip`
Example internal build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2077707&view=results

Outstanding issues to be fixed in future PRs:
- ~~The package version should be `7.1.0` instead of `7.1.0-rtm.23052.1`.~~ Because the archives are marked as non-shipping, they do not get final version numbers.
- ~~The package versioning on non-Windows platforms is not respecting the `-ci` flag when building, so it produces versions with a `-dev` label e.g. `7.1.0-dev.23052.1`~~ Fixed: The `-ci` flag is captured in the ` $__ArcadeScriptArgs` variable, which has been added back to the `build.sh` script.
- The files inside the tarballs have inconsistent permissions.
- The third-party notice (TPN) file is not included in the archives due to build ordering.
- Addition of a LICENSE.txt file with the MIT license in it.
- Consider including the profiler for other platforms that are emulatable e.g. Windows x64 archive should contain `win-x64` and `win-x86` profilers, Windows ARM64 archive should contain `win-arm64`, `win-x64`, and `win-x86` profilers, etc.

The intent is to backport this to `6.x` an `main` with minor adjustments (the TFM) so that way there is a consistent acquisition experience across all in-support versions.

### DIFF
--- a/eng/Signing.props
+++ b/eng/Signing.props
@@ -4,7 +4,15 @@
     <FileSignInfo Include="Newtonsoft.Json.Bson.dll" CertificateName="3PartySHA2" />
   </ItemGroup>
   <ItemGroup>
+    <!--
+      Tarballs are currently not signed nor are their contents.
+      This entry allows them to be signed automatically when tooling allows for it.
+      -->
     <ItemsToSign Include="$(ArtifactsPackagesDir)**\*.tar.gz" />
+    <!--
+      The zip file itself is not signed but their contents are.
+      The zip is expanded, the contents are signed, and then rezipped.
+      -->
     <ItemsToSign Include="$(ArtifactsPackagesDir)**\*.zip" />
   </ItemGroup>
 </Project>
--- a/eng/Signing.props
+++ b/eng/Signing.props
@@ -3,4 +3,8 @@
     <FileSignInfo Include="Newtonsoft.Json.dll" CertificateName="3PartySHA2" />
     <FileSignInfo Include="Newtonsoft.Json.Bson.dll" CertificateName="3PartySHA2" />
   </ItemGroup>
+  <ItemGroup>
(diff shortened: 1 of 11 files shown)


eng/Signing.props
```diff
@@ -3,4 +3,8 @@
     <FileSignInfo Include="Newtonsoft.Json.dll" CertificateName="3PartySHA2" />
     <FileSignInfo Include="Newtonsoft.Json.Bson.dll" CertificateName="3PartySHA2" />
   </ItemGroup>
+  <ItemGroup>
```
jander-msft, line 6: Add comments as to what is actually signed

PR #3239, opened 2023-01-06
Implement strncpy_s when on glibc

schmittjoseph:
###### Summary

Port the PAL's implementation of `strncpy_s` for use on glibc builds of the profiler. The PAL uses a single generic algorithm for all `strncpy_s` variants via heavy macro usage, since we only need the basic case this PR ports over it in a simplified form with minimal extra code. There are the following differences from the PAL's:
- The only functional difference is that in debug builds this ported version will not fill any remaining space in the dst buffer with `0xFE`.
- Expanded/replaced macros.

### DIFF
--- a/src/MonitorProfiler/Utilities/StringUtilities.h
+++ b/src/MonitorProfiler/Utilities/StringUtilities.h
@@ -59,6 +59,7 @@ class StringUtilities
             /* validation section */
             if (dst == NULL || size == 0)
             {
+                errno = EINVAL;
                 return EINVAL;
             }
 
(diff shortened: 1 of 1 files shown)


src/MonitorProfiler/Utilities/StringUtilities.h
```diff
@@ -31,4 +40,75 @@ class StringUtilities
...
+            if (src == NULL)
+            {
+                *dst = 0;
+                errno = EINVAL;
+                return EINVAL;
+            }
+
+            p = dst;
+            available = size;
+            if (count == _TRUNCATE)
```
wiktork, line 82: is _TRUNCATE already defined somewhere in glibc?

wiktork (approved):

PR #3241, opened 2023-01-07
Initial stab at creating swagger for dotnet-monitor

samsp-msft:
###### Summary

Adds swagger UI to dotnet-monitor as the default endpoint. That helps by giving you documentation of the endpoint, and also the ability to use it.

image

## Issues
* The default runtime generation of the api info was causing issues, so I am using a static version done at build time - which probably makes more sense for this scenario anyway.
* The swagger files are generated in a wwwroot folder as part of the build, and despite having an itemgroup for it, they are not copied to the output artifacts folder, which is necessary for the UI. If you manually copy them over, it works.
  * This uses the static files middleware to deliver the files. I don't know much about the deployment of dotnet-monitor and if static files is the right approach. It may be better to have a middleware that can deliver the files from embedded resources.
* The files I am generating don't have the output of the xml comments in them - not sure why - that needs to be debugged
* You are already generating an openapi.json file somehow - which is what the UI is keyed off of, and yours has the descriptions from the xml comments, so I am doing excess work not as well as you.
* I think we can do more annotations on the properties to make using them easier, providing default values etc.


samsp-msft:
Added screenshot 
image

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/SwaggerFilesController.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        /// <summary>
+        /// Get the swagger files embedded into the assembly
+        /// </summary>
+        [HttpGet("{filename}")]
+        [ProducesResponseType(typeof(string), StatusCodes.Status200OK)]
+        public FileStreamResult GetFile(string filename)
+        {
+            if (filename != "swagger.json") { throw new Exception(string.Format(Strings.ErrorMessage_ResourceNotFound, filename)); }
+            var stream = Assembly.GetExecutingAssembly().GetManifestResourceStream("Microsoft.Diagnostics.Monitoring.WebApi.swagger.v1." + filename);
```
wiktork, line 34: nit: nameof or similar

wiktork (dismissed):

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/SwaggerFilesController.cs
samsp-msft: I tried assembly.name, but that ended up being the full name with hash etc. is there a way to just get the assembly name?

jander-msft: It's not actually the assembly name; embedded resources use the default namespace of the assembly, which isn't really gettable in a direct manner without tying it to a type that happens to be directly in that namespace.

jander-msft (approved):

PR #3257, opened 2023-01-09
Add an Alpine-based dev container

schmittjoseph:
###### Summary

Add an Alpine-based dev container that people can optionally select when creating a Codespace for a musl libc development environment. It has all the necessary dependencies to build & run `dotnet-monitor` from the repo. The default devcontainer  will remain the Debian-based one.

This configuration can be selected in the advanced Codespace creation:
image

**Limitations**:
Most of GitHub dev container's "features" are heavily reliant on `apt` and other Debian-based system notions, as a result this new dev container offers limited support for workflows outside of simply building, running, and testing `dotnet-monitor` locally. To help communicate this a dedicated readme explaining this will automatically be opened up for users that create an Alpine-based codespace.

### DIFF
--- a/.devcontainer/musl/devcontainer.json
+++ b/.devcontainer/musl/devcontainer.json
@@ -2,10 +2,7 @@
 {
     "name": "Development (Alpine)",
     "build": {
-        "dockerfile": "Dockerfile",
-        "args": {
-            "VARIANT": "3.17"
-        }
+        "dockerfile": "Dockerfile"
     },
     "hostRequirements": {
         "cpus": 4


.devcontainer/musl/Dockerfile
```diff
@@ -0,0 +1,4 @@
+ARG VARIANT="3.17"
```
wiktork, line 1: nit: can we avoid duplicate specification of VARIANT?

.vscode/settings.json
```diff
@@ -0,0 +1,39 @@
+{
+    "files.associations": {
```
wiktork, line 2: won't this propagate to vscode editing of dotnet-monitor outside the dev container scenario, or is that the intent?

wiktork (approved):

jander-msft (approved):

PR #3291, opened 2023-01-12
Enable checksum and publishing of archives

jander-msft:
###### Summary

The following changes allow archive packages to participate in the publishing process (inclusion in the BAR manifest so that they are published to dotnetbuilds feed during Maestro promotion).

Overview of changes:
- Update `src/Directory.Build.targets` to produce projectpath and projectprops (these are new in this PR to support specifying per-project properties when the project is invoked during publish) files for both product archives and symbols archives.
- Update `src/Directory.Build.targets` to report correct archive name from `GetPackageFileName` target, which is invoked during publish.
- Update `eng/Publishing.props` include archive file types, to use the projectprops file for each package file (if it exists), and to only apply `Category=OTHER` to nupkg files (this categorization forces publishing to blob feeds instead of NuGet feeds; all other 'binary layout' types are published to blob feeds automatically).

Example build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2086815&view=results

### DIFF
--- a/eng/Publishing.props
+++ b/eng/Publishing.props
@@ -13,7 +13,11 @@
 
   <ItemGroup>
     <PackageFile Include="$(ArtifactsShippingPackagesDir)**/*.nupkg" IsShipping="true" />
+    <PackageFile Include="$(ArtifactsShippingPackagesDir)**/*.tar.gz" IsShipping="true" />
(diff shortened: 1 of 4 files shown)


jander-msft:
I want to point out that this PR does not actually enable the archives to be published yet (it's still maked as `IsShipping=false`) but updates the targets to allow it to be published. The example build shows the behavior when the archives are marked as shipping.

PR #3295, opened 2023-01-12
Cleanup spelling in source files

schmittjoseph:
###### Summary

Cleanup spelling errors in non-user-facing strings. User-facing strings will be fixed in a second PR.


cspell.json
```diff
@@ -1,31 +1,73 @@
...
+        "Impls",
+        "JITID",
         "JWTs",
         "libc",
         "livemetrics",
         "minidump",
         "MSRC",
+        "ndjson",
         "newtonsoft",
+        "Notfication",
```
kkeirstead, line 44: Should this be included? It appears to just be a misspelling of "Notification"

jander-msft: We'll probably pick up a new version of the header file this release so we can remove the exclusion at that time.

jander-msft (approved):

PR #3301, opened 2023-01-13
Exception identification

jander-msft:
###### Summary

Small building block for grouping thrown exceptions together by exception type and throwing location (method + IL offset).

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/ExceptionIdentifier.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/ExceptionIdentifier.cs
@@ -1,0 +1,41 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Diagnostics;
+using System.Reflection;
+
+namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions
+{
+    /// <summary>
+    /// Class representing a type of exception thrown from the specific location
+    /// in code. This class may be used as an identifier for determining whether
+    /// exception usages are equivalent (same exception type thrown from the same location).
+    /// </summary>
+    internal sealed record class ExceptionIdentifier
+    {
+        public ExceptionIdentifier(Exception ex)
+        {
+            ArgumentNullException.ThrowIfNull(ex);
+
+            ExceptionType = ex.GetType();
+
+            StackTrace stackTrace = new(ex, fNeedFileInfo: false);
+            foreach (StackFrame stackFrame in stackTrace.GetFrames())
+            {
+                ThrowingMethod = stackFrame.GetMethod();
+                if (null != ThrowingMethod)
+                {
+                    ILOffset = stackFrame.GetILOffset();
+                    break;
+                }
+            }
+        }
+
+        public Type ExceptionType { get; }
+
+        public MethodBase? ThrowingMethod { get; }
+
+        public int ILOffset { get; }
+    }
+}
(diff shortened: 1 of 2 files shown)


src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/ExceptionIdentifier.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// </summary>
+    internal sealed record class ExceptionIdentifier
+    {
+        public ExceptionIdentifier(Exception ex)
+        {
+            ArgumentNullException.ThrowIfNull(ex);
+
+            ExceptionType = ex.GetType();
+
+            StackTrace stackTrace = new(ex, fNeedFileInfo: false);
```
wiktork, line 23: In many cases we will actually have the fileinfo (pdb) here. Are you excluding this for perf reasons?

wiktork, line 23: I'm assuming that there's minimal overhead here compared to calling Exception.StackTrace correct?

jander-msft: Correct, turning it off avoids reading in source line information and processors.

jander-msft: Correct, turning it off avoids reading in source line information and processors.

jander-msft: `Exception.StackTrace` effectively calls 'new StackTrace(this, fNeedFileInfo: true)`, and we don't need the source line information. Also, `Exception.StackTrace` is a string and I don't want to parse that.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/ExceptionIdentifier.cs
```diff
@@ -0,0 +1,41 @@
...
+            foreach (StackFrame stackFrame in stackTrace.GetFrames())
```
wiktork, line 24: I'm concerned that in practice this is going to cause a lot of collisions between exceptions that are the same at the throwing method but have vastly different callstacks underneath. For example, if the library is throwing an exception from a data structure, it would look the same from completely different callsites.

wiktork: One possible way to address this is to do immunity for certain frames and continue walking the callstack. Another would be to take more of the frames as part of the exception identity.

jander-msft: Agreed.

However, I don't know if we have enough data or a good design yet to address that yet. This gets into filtering frames and modules. One reasonable approach is to say anything that is in `System.Private.CoreLib` is excluded (or we search down the callstack until we get to a non-`System.Private.CoreLib` module). Or we have a module list that is excluded; or we do a JMC style concept.

But these can also lead to collisions of unrelated exceptions. For example, if `MethodA` calls `MethodB` in `System.Private.CoreLib`, `MethodB` calls `MethodC` in `System.Private.CoreLib` and `MethodC` throws an exception, it is caught, and `MethodB` later throws an exception of the same type, then the resolution of walking down the stack until we find user code would show these two separate exceptions as equivalent. Now, maybe that's okay because to the app developer, they don't even see the first exception, and possibly not even the second exception.

Extending this to a realistic scenario, it seems that sockets throw lots of exceptions when you do either a read or write on them. If those operations are throwing the same exception type and handling them all before it reaches user code, and you have this "walk down the stack until you find user code" resolution, then you are likely to see a grouping of many of these exceptions that have equivalent `ExceptionId` instances.

I think it's best to start without the filtering, gather some data, and add in the ability to filter or walk down the stack in the near future.

wiktork (approved):

PR #3380, opened 2023-01-18
Use templates for dependabot and configure public-feed dependencies

schmittjoseph:
###### Summary

This PR does 2 things:
- Add a new Dependabot project for dependencies that should only ever pull from public feeds (e.g. nuget.org). This is primarily to address the shipping `dotnet/diagnostics` dependency.
- Leverage [ytt](https://carvel.dev/ytt/) to template `dependabot.yml` to avoid having to manually manage this file. There is a new VSCode/Codespaces task to help with this, and Codespaces will automatically install the needed tooling.
  - This currently has to be done manually, but we can look into PR automation later on to help with this.


.github/dependabot.template.yml
```diff
@@ -0,0 +1,36 @@
...
+  - package-ecosystem: "nuget"
+    directory: "/eng/dependabot"
+    schedule:
+      interval: "daily"
+    target-branch: #@ branch
+    ignore:
+      - dependency-name: "Microsoft.Extensions.*"
+        update-types: [ "version-update:semver-major" ]
+  - package-ecosystem: "nuget"
+    directory: "/eng/dependabot/public-feed"
```
jander-msft, line 22: Does "public feed" mean nuget.org? If so, maybe rename it to "nuget.org" so it's more clear as to where its getting package information. When I read/hear "public-feed" in terms of the dotnet repos, I think of https://dev.azure.com/dnceng/public/_artifacts/feed/dotnet-public

jander-msft (approved):

jander-msft (approved):

PR #3406, opened 2023-01-20
Exceptions event source and exception identifier cache

jander-msft:
###### Summary

- Add event source for producing exception information so that external process (e.g. dotnet-monitor) can get exception information that is grouped and identified by the `ExceptionIdentifier` class.
- Add exception identifier cache that invokes callbacks when new exception identification information is added to it. This allows filtering of repeated exceptions so that the event source doesn't have to repeat the same exception, method, type, and module information.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
@@ -1,0 +1,215 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Diagnostics.Tracing;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Eventing
+{
+    [EventSource(Name = SourceName)]
+    internal sealed class ExceptionsEventSource : EventSource
+    {
+        public const int ExceptionIdEventId = 1;
+        public const int ExceptionEventId = 2;
+        public const int ClassDescriptionEventId = 3;
+        public const int FunctionDescriptionEventId = 4;
+        public const int ModuleDescriptionEventId = 5;
+        public const int TokenDescriptionEventId = 6;
+
+        public const string SourceName = "Microsoft.Diagnostics.Monitoring.Exceptions";
+
+        [Event(ExceptionIdEventId)]
+        public void WriteExceptionId(
+            ulong ExceptionId,
+            ulong ExceptionClassId,
+            ulong ThrowingMethodId,
+            int ILOffset)
+        {
+            Span<EventData> data = stackalloc EventData[4];
+
+            SetValue(ref data[ExceptionEvents.ExceptionIdPayloads.ExceptionId], ExceptionId);
+            SetValue(ref data[ExceptionEvents.ExceptionIdPayloads.ExceptionClassId], ExceptionClassId);
+            SetValue(ref data[ExceptionEvents.ExceptionIdPayloads.ThrowingMethodId], ThrowingMethodId);
+            SetValue(ref data[ExceptionEvents.ExceptionIdPayloads.ILOffset], ILOffset);
+
+            WriteEventCore(ExceptionIdEventId, data);
+        }
+
+        [Event(ExceptionEventId)]
+        public void WriteException(
+            ulong ExceptionId,
+            string? ExceptionMessage)
+        {
+            Span<EventData> data = stackalloc EventData[2];
+            using PinnedData namePinned = PinnedData.Create(ExceptionMessage);
+
+            SetValue(ref data[ExceptionEvents.ExceptionPayloads.ExceptionId], ExceptionId);
+            SetValue(ref data[ExceptionEvents.ExceptionPayloads.ExceptionMessage], namePinned);
+
+            WriteEventCore(ExceptionEventId, data);
+        }
+
+        [Event(ClassDescriptionEventId)]
+        [SuppressMessage("Performance", "CA1822:Mark members as static", Justification = "Instance method required for event source manifest generation.")]
+        private unsafe void WriteClassDescription(
+            ulong ClassId,
+            ulong ModuleId,
+            uint Token,
+            uint Flags,
+            byte* TypeArgs)
+        {
+        }
+
+        [NonEvent]
+        public void WriteClassDescription(
+            ulong ClassId,
+            ulong ModuleId,
+            uint Token,
+            uint Flags,
+            ulong[] TypeArgs)
+        {
+            Span<EventData> data = stackalloc EventData[5];
+            using PinnedData typeArgsPinned = PinnedData.Create(TypeArgs);
+
+            SetValue(ref data[NameIdentificationEvents.ClassDescPayloads.ClassId], ClassId);
+            SetValue(ref data[NameIdentificationEvents.ClassDescPayloads.ModuleId], ModuleId);
+            SetValue(ref data[NameIdentificationEvents.ClassDescPayloads.Token], Token);
+            SetValue(ref data[NameIdentificationEvents.ClassDescPayloads.Flags], Flags);
+            SetValue(ref data[NameIdentificationEvents.ClassDescPayloads.TypeArgs], typeArgsPinned);
+
+            WriteEventCore(ClassDescriptionEventId, data);
+        }
+
+        [Event(FunctionDescriptionEventId)]
+        [SuppressMessage("Performance", "CA1822:Mark members as static", Justification = "Instance method required for event source manifest generation.")]
+        private unsafe void WriteFunctionDescription(
+            ulong FunctionId,
+            ulong ClassId,
+            uint ClassToken,
+            ulong ModuleId,
+            string Name,
+            byte* TypeArgs)
+        {
+        }
+
+
+        [NonEvent]
+        public void WriteFunctionDescription(
+            ulong FunctionId,
+            ulong ClassId,
+            uint ClassToken,
+            ulong ModuleId,
+            string Name,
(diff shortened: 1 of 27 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/Stacks/NameCache.cs
```diff
@@ -0,0 +1,60 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Concurrent;
+using System.Diagnostics;
+
+#if STARTUPHOOK
+namespace Microsoft.Diagnostics.Monitoring.StartupHook
```
wiktork, line 8: Are you avoiding dependencies for the startup hook dll?

jander-msft: Yes, I'm trying to put as little code (within reason) into any assembly that we want to load into the target application.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        [NonEvent]
+        private static void SetValue(ref EventData data, in PinnedData value)
+        {
+            data.DataPointer = value.Address;
+            data.Size = value.Size;
+        }
+
+        private ref struct PinnedData
```
wiktork, line 177: Should this implement IDisposable?

jander-msft: Ref structs cannot implement interfaces. However, this doesn't _really_ need to be a ref struct; I think it might have had some pointers or refs in earlier iterations. I can change this to be a regular struct and have it implement `IDisposable`.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
```diff
@@ -0,0 +1,215 @@
...
+using System.Diagnostics.Tracing;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Eventing
+{
+    [EventSource(Name = SourceName)]
+    internal sealed class ExceptionsEventSource : EventSource
+    {
+        public const int ExceptionIdEventId = 1;
```
wiktork, line 15: To my understanding, if the OnMetadata* events are not observed, we will have no way of seeing them again correct? I understand we probably want to start listening from the beginning of the app, but it would be nice to be able to enable/disable this and invalidate the cache as necessary so that the client can receive it again.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
```diff
@@ -0,0 +1,215 @@
...
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Diagnostics.Tracing;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Eventing
+{
+    [EventSource(Name = SourceName)]
+    internal sealed class ExceptionsEventSource : EventSource
```
wiktork, line 13: Any issues with https://github.com/dotnet/runtime/issues/76704? Should we send a marker event to ensure that all the data is flushed?

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
jander-msft: I was thinking that we might want that, but I'm not sure how we'll want to do that just yet (e.g. command on the UDS channel to invalidate the cache, invalidate when the event source is disabled, etc). Invalidating when the event source is disable has a chicken-and-egg problem since the cache is the one that calls the event source already. I'd like to think about this more and save it for another PR.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
jander-msft: I do need to add something that does this flushing but probably in a separate PR. I'm thinking something along the lines of a timer callback that will fire after a short time, which will be reset for every event, and finally send off the marker event when the timer expires. I'm open to other ideas too, but I don't want to always wait 100s of milliseconds after each event or exception is sent.

jander-msft: Note that there aren't any tests that consume this event source from outside of the target process yet, so we don't run into the issue yet.

src/Microsoft.Diagnostics.Monitoring.WebApi/Stacks/NameFormatter.cs
```diff
@@ -0,0 +1,131 @@
...
+            else
+            {
+                BuildClassName(builder, cache, functionData.ModuleId, functionData.ParentToken);
+            }
+        }
+
+        public static void BuildClassName(StringBuilder builder, NameCache cache, ulong classId)
+        {
+            string className = UnknownClass;
+            if (cache.ClassData.TryGetValue(classId, out ClassData? classData))
```
kkeirstead, line 43: I noticed that in the existing code we aren't using nullable when we try to get a value (e.g. `ClassData` instead of `ClassData?`), but we are in all of the new code it looks like - what's the reason?

src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/Exceptions/Identification/ExceptionIdentifierCacheTests.cs
```diff
@@ -0,0 +1,192 @@
...
+            Assert.NotEqual(InvalidId, throwingMethodData.ParentClass);
+            Assert.NotEqual(InvalidToken, throwingMethodData.ParentToken);
+            Assert.Empty(throwingMethodData.TypeArgs);
+        }
+
+        private static uint ValidateTokenAndGetOuterToken(NameCache cache, ulong moduleId, uint token, Type expectedType)
+        {
+            Assert.True(cache.TokenData.TryGetValue(new ModuleScopedToken(moduleId, token), out TokenData? tokenData));
+            Assert.NotNull(tokenData);
+            string? expectedTypeName = null == expectedType.DeclaringType ?
```
kkeirstead, line 182: This is technically slightly different than the way this is implemented earlier (`null == expectedType.DeclaringType ? expectedType.FullName ?? expectedType.Name : expectedType.Name`) - should this be standardized for consistency?

src/Microsoft.Diagnostics.Monitoring.WebApi/Stacks/NameFormatter.cs
jander-msft: The StartupHook project has nullability enforcement enabled whereas the existing projects do not. I don't want to go through the rest of dotnet-monitor and make that kind of change as part of this work.

src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/Exceptions/Identification/ExceptionIdentifierCacheTests.cs
jander-msft: I don't see anywhere else in the codebase that has `null == expectedType.DeclaringType`; can you point me directly to what you are referencing?

kkeirstead: That was in reference to this: https://github.com/dotnet/dotnet-monitor/pull/3406/files#diff-aea088e1c05d63b61267f1dadf91ab9b32525969da9b1f6a272dd9934079d09fR160 - it's possible I just noticed they looked similar but aren't meant to behave the same way.

jander-msft: These don't need to be the same. The product code is handling all cases where `Type.FullName` _may be_ null (https://learn.microsoft.com/en-us/dotnet/api/system.type.fullname?view=net-7.0#property-value) whereas the test code knows its a fully resolved type that is an Exception type.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
jander-msft: Made this a non-ref struct and implemented IDisposable.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
```diff
@@ -0,0 +1,215 @@
...
+        public const string SourceName = "Microsoft.Diagnostics.Monitoring.Exceptions";
+
+        [Event(ExceptionIdEventId)]
+        public void WriteExceptionId(
+            ulong ExceptionId,
+            ulong ExceptionClassId,
+            ulong ThrowingMethodId,
+            int ILOffset)
+        {
+            Span<EventData> data = stackalloc EventData[4];
```
wiktork, line 31: Test suggestions: n parallel threads, all throwing exceptions.

wiktork (approved):

PR #3433, opened 2023-01-24
Add test run option to release notes workflow

schmittjoseph:
###### Summary

Add a new run option for the generate release notes workflow. Setting this will skip checking for a matching release build and open a draft PR instead of a regular one.


.github/workflows/generate-release-notes.yml
```diff
@@ -1,8 +1,13 @@
 name: 'Generate release notes'
-run-name: '[${{ github.ref_name }}] Generate release notes'
+run-name: '[${{ github.ref_name }}] Generate release notes (test: ${{ inputs.test_run }})'
 
 on:
   workflow_dispatch:
+    inputs:
+      test_run:
```
jander-msft, line 7: Should this have a default value, maybe true?

jander-msft (approved):

PR #3480, opened 2023-01-26
Trigger updates to dotnet-docker for release builds

schmittjoseph:
###### Summary

This PR reworks how `dotnet-docker` is updated for both release builds and nightly builds of `main`.  There is a corresponding PR to `dotnet-docker` here:  https://github.com/dotnet/dotnet-docker/pull/4363

With these changes our `dotnet-docker` pipeline will automatically reacts to our release builds, gathers information from it, and updates the correct version.  **NOTE** it will only react to release builds if we check the `Update dotnet-docker?` checkbox when queuing a release build.

This affords us a few advantages:
1. Concurrent releases of `dotnet-monitor` of multiple minor versions in the same major version (e.g. `7.0.X` and `7.1.X`) is now supported and requires no extra work.
2. We no longer need to update the `dotnet-docker` pipeline's variables before and after a release.
3. We no longer need to update the `dotnet-docker` pipeline each time we changes the major/minor version of what is in our `main` branch.

Overview of changes:
- All `dotnet-monitor` builds will now produce an artifact named `Build_Info` which contains the full build version. We can add more to this artifact as needed. This provides a reliable and consistent way of other pipelines gathering information about the build produced by a given pipeline run across all branches and build qualities.
- An `update-docker` tag is added to release builds if requested by the person queuing the build. This will cause the `dotnet-docker` pipeline to run when our build finishes and act accordingly.
-  When our `dotnet-docker` pipeline runs according to its nightly schedule, it will pick up the latest build from `main` (it does not need to have the `update-docker` tag as these are not release builds and updates every day).

### DIFF
--- a/dotnet-monitor.yml
+++ b/dotnet-monitor.yml
@@ -39,7 +39,7 @@ variables:
   value: DotNetCore
 - name: _TPNFile
   value: THIRD-PARTY-NOTICES.TXT
-  
+
 - ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
   # DotNet-Diagnostics-SDL-Params provides Tsa* variables for SDL checks.
   - group: DotNet-Diagnostics-SDL-Params
--- a/eng/pipelines/jobs/pack-sign-publish.yml
+++ b/eng/pipelines/jobs/pack-sign-publish.yml
@@ -48,7 +48,7 @@ jobs:
         targetType: inline
         script: |
             New-Item "$(Build.ArtifactStagingDirectory)/buildInfo" -Type Directory
-            Copy-Item "$(Build.SourcesDirectory)/artifacts/packages/Release/Shipping/*.buildversion" "$(Build.ArtifactStagingDirectory)/buildInfo/buildversion"
+            Copy-Item "$(Build.SourcesDirectory)/artifacts/packages/Release/Shipping/dotnet-monitor.*.nupkg.buildversion" "$(Build.ArtifactStagingDirectory)/buildInfo/dotnet-monitor.nupkg.buildversion"
     - task: PublishPipelineArtifact@1
       displayName: Publish Build Info
       inputs:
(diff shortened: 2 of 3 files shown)


dotnet-monitor.yml
```diff
@@ -121,3 +125,5 @@ stages:
         - 'PackageArtifacts'
 # This sets up the bits to do a Release.
 - template: /eng/pipelines/stages/preparerelease.yml
+  parameters:
+    updateDocker: ${{ parameters.updateDocker }}
```
jander-msft, line 129: nit: each of these files should have a newline at the end

eng/pipelines/jobs/pack-sign-publish.yml
```diff
@@ -42,3 +42,15 @@ jobs:
         /p:DotNetPublishUsingPipelines=true
         /p:ThirdPartyNoticesFilePath='$(Build.SourcesDirectory)/$(_TPNFile)'
       displayName: Pack, Sign, and Publish
+    - task: powershell@2
+      displayName: Generate Build Info
+      inputs:
+        targetType: inline
+        script: |
+            New-Item "$(Build.ArtifactStagingDirectory)/buildInfo" -Type Directory
+            Copy-Item "$(Build.SourcesDirectory)/artifacts/packages/Release/Shipping/*.buildversion" "$(Build.ArtifactStagingDirectory)/buildInfo/buildversion"
```
jander-msft, line 51: We may want to copy the non-shipping build version information too (there shouldn't be any collisions), but that's not necessary to do right now.

jander-msft (approved):

eng/pipelines/jobs/pack-sign-publish.yml
jander-msft, line 51: Btw, there are multiple build version files; now, they should all have the same value _except_ if we start versioning some package versions differently (e.g. diagnostics repo does this for the NETCore.Client library where it has a version of 0.2.0). Maybe this should copy the files as-is without rename and then let the consume pick which file it wants to key off of. If we need to construct build version files with just the package name, I'm okay with that. It's likely we'll do away with the existing *.buildversion and *.version files once this change is in place.

jander-msft (approved):

PR #3506, opened 2023-01-30
Sign published artifacts before archiving

jander-msft:
###### Summary

The SignTool task only knows how to unpack *.nuget and *.zip files in order to sign their contents. This leaves all of the files *.tar.gz unsigned.

This change adds an additional job after build but before archiving that signs all of the archive content in one pass.

Example build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2100455&view=results

### DIFF
--- a/eng/Signing.props
+++ b/eng/Signing.props
@@ -4,7 +4,7 @@
     <FileSignInfo Include="Newtonsoft.Json.Bson.dll" CertificateName="3PartySHA2" />
     <FileSignInfo Include="Swashbuckle.AspNetCore.SwaggerUI.dll" CertificateName="3PartySHA2" />
   </ItemGroup>
-  <ItemGroup>
+  <ItemGroup Condition="'$(SignAllBinaries)' != 'true'">
     <!--
       Tarballs are currently not signed nor are their contents.
       This entry allows them to be signed automatically when tooling allows for it.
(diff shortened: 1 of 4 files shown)


eng/Signing.props
```diff
@@ -16,4 +16,12 @@
       -->
     <ItemsToSign Include="$(ArtifactsPackagesDir)**\*.zip" />
   </ItemGroup>
+  <ItemGroup Condition="'$(SignAllBinaries)' == 'true'">
+    <!--
+      Sign all binaries in the artifacts directory.
+      -->
+    <ItemsToSign Remove="@(ItemsToSign)" />
+    <ItemsToSign Include="$(ArtifactsDir)bin\**\*" />
```
wiktork, line 24: Will this sign json files properly?

eng/Signing.props
```diff
@@ -16,4 +16,12 @@
...
+    <ItemsToSign Include="$(ArtifactsDir)pub\**\*" />
```
wiktork, line 25: Are there duplications between pub and bin? Do we need to sign both places?

wiktork (approved):

eng/Signing.props
jander-msft: `bin` contains the profiler assemblies only in the current usage; `pub` contains the results of `dotnet publish` for dotnet-monitor for each platform. There isn't much duplication except for the dependencies; however the signing task seems to be smart about it when submitting for signing by hashing the files to determine uniqueness, only submitting the unique set, and copying over duplicates with the same signed files.

eng/Signing.props
jander-msft: No, these are not currently signed either in this new mechanism or in the existing nupkgs. As a follow up, I can look into seeing if signing the json files (and possibly the xml files) (1) actually works (it should), and (2) if the runtime is okay with *.deps.json and *.runtimeconfig.json files being signed.

PR #3549, opened 2023-02-02
Automatically create branches that track the latest releases

schmittjoseph:
###### Summary

This PR introduces automation that will automatically create/update a `shipped/v{major}.{minor}` branch each time a release is made. The HEAD of these branches will be the latest shipped commit for that major/minor version. (e.g. `shipped/v7.0`).

The purpose of these branches is to give us an easier mechanism for Component Governance (CG) to identify if a shipping version contains any potential CVEs in its dependencies and notify us accordingly.  This will also require an additional small pipeline to register these branches (`shipped/*`) with CG but that can be done in a separate PR.

### DIFF
--- a/.github/workflows/track-shipping-version.yml
+++ b/.github/workflows/track-shipping-version.yml
@@ -6,7 +6,7 @@ permissions:
   contents: write
 
 env:
-  shipping_branch_prefix: 'shipping'
+  shipping_branch_prefix: 'shipped'
 
 jobs:
   update-shipping-branch:


.github/workflows/track-shipping-version.yml
```diff
@@ -0,0 +1,34 @@
+on:
+  release:
+    types: [released]
+
+permissions:
+  contents: write
+
+env:
+  shipping_branch_prefix: 'shipping'
```
jander-msft, line 9: nit: 
```suggestion
  shipping_branch_prefix: 'shipped'
```

"shipping" makes it sound like it the shipping procedure is currently active whereas "shipped" indicates that the shipping procedure is complete

.github/workflows/track-shipping-version.yml
```diff
@@ -0,0 +1,34 @@
...
+              echo "Unexpected release tag: $release_version."
+              exit 1
+          fi
+
+          major_minor_version=${release_version%.*}
+          shipping_branch_name="${{ env.shipping_branch_prefix }}/$major_minor_version"
+
+          # This is a shallow clone so we will always create a new local branch even if it already exists on the remote
+          git checkout -b "$shipping_branch_name"
+          git push --force --set-upstream origin "HEAD:$shipping_branch_name"
```
jander-msft, line 34: We'll also have to mirror these branches with a registration here: https://github.com/dotnet/versions/blob/main/Maestro/subscriptions.json#L447

jander-msft (approved):

.github/workflows/track-shipping-version.yml
```diff
@@ -0,0 +1,34 @@
...
+          if [[ ! "$release_version" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+              echo "Unexpected release tag: $release_version."
+              exit 1
+          fi
+
+          major_minor_version=${release_version%.*}
+          shipping_branch_name="${{ env.shipping_branch_prefix }}/$major_minor_version"
+
+          # This is a shallow clone so we will always create a new local branch even if it already exists on the remote
+          git checkout -b "$shipping_branch_name"
```
wiktork, line 33: The branch and the tags we already create will match correct?

jander-msft: My understanding is that a GitHub release with tag "v7.0.2" would translate into a branch named "shipped/7.0".

jander-msft (approved):

PR #3591, opened 2023-02-07
Allow per provider interval specification

wiktork:
###### Summary
Requires https://github.com/dotnet/diagnostics/pull/3639

### DIFF
--- a/documentation/configuration/metrics-configuration.md
+++ b/documentation/configuration/metrics-configuration.md
@@ -8,6 +8,8 @@
 Due to limitations in event counters, `dotnet monitor` supports only **one** refresh interval when collecting metrics. This interval is used for
 Prometheus metrics, livemetrics, triggers, traces, and trigger actions that collect traces. The default interval is 5 seconds, but can be changed in configuration.
 
+[7.1+] For EventCounter providers, is possible to specify a different interval for each provider. See [Per provider intervals](#per-provider-intervals-71).
+
 <details>
   <summary>JSON</summary>
 
@@ -34,6 +36,48 @@ Prometheus metrics, livemetrics, triggers, traces, and trigger actions that coll
   ```yaml
   - name: DotnetMonitor_GlobalCounter__IntervalSeconds
     value: "10"
+  ```
+</details>
+
+## Per provider intervals (7.1+)
+
+It is possible to override the global interval on a per provider basis. Note this forces all scenarios (triggers, live metrics, prometheus metrics, traces) that use a particular provider to use that interval. Metrics that are `System.Diagnostics.Metrics` based always use global interval.
+
+<details>
+  <summary>JSON</summary>
+
+  ```json
+  {
+      "GlobalCounter": {
+        "IntervalSeconds": 5,
+        "Providers": {
+            "System.Runtime": {
+              "IntervalSeconds": 10
+            }
+          }
+      }
+  }
+  ```
+</details>
+
+<details>
+  <summary>Kubernetes ConfigMap</summary>
+  
+  ```yaml
+  GlobalCounter__IntervalSeconds: "5"
+  GlobalCounter__Providers__System.Runtime__IntervalSeconds: "10"
+  ```
+</details>
+
+<details>
+  <summary>Kubernetes Environment Variables</summary>
+  
+  ```yaml
+  - name: DotnetMonitor_GlobalCounter__IntervalSeconds
+    value: "5"
+  - name: DotnetMonitor_GlobalCounter__Providers__System.Runtime__IntervalSeconds
+    value: "10"
+
   ```
 </details>
 
--- a/documentation/schema.json
+++ b/documentation/schema.json
@@ -900,16 +900,30 @@
           "maximum": 2147483647.0,
           "minimum": 1.0
         },
-        "ProviderIntervals": {
+        "Providers": {
           "type": [
             "null",
             "object"
           ],
-          "description": "Dictionary of provider names and their specific intervals.",
+          "description": "Dictionary of provider names and their global configuration.",
           "additionalProperties": {
-            "type": "number",
-            "format": "float"
+            "$ref": "#/definitions/GlobalProviderOptions"
           }
+        }
+      }
+    },
+    "GlobalProviderOptions": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "IntervalSeconds": {
+          "type": [
+            "null",
+            "number"
+          ],
+          "format": "float",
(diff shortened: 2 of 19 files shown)


src/Microsoft.Diagnostics.Monitoring.Options/GlobalCounterOptions.cs
```diff
@@ -32,6 +33,11 @@ public class GlobalCounterOptions
         [DefaultValue(GlobalCounterOptionsDefaults.MaxTimeSeries)]
         [Range(1, int.MaxValue)]
         public int? MaxTimeSeries { get; set; }
+
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_GlobalCounterOptions_ProviderIntervals))]
+        public System.Collections.Generic.IDictionary<string, float> ProviderIntervals { get; set; } = new Dictionary<string, float>(StringComparer.OrdinalIgnoreCase);
```
jander-msft, line 40: Please rename to `ProviderIntervalSeconds` for consistency with the existing `IntervalSeconds` option.

jander-msft: Or `PerProviderIntervalSeconds`

documentation/schema.json
```diff
@@ -899,6 +899,17 @@
...
+        "ProviderIntervals": {
+          "type": [
+            "null",
+            "object"
+          ],
+          "description": "Dictionary of provider names and their specific intervals.",
+          "additionalProperties": {
+            "type": "number",
+            "format": "float"
+          }
```
kkeirstead, line 912: Don't forget to update the config docs (and label it for the correct version)

src/Microsoft.Diagnostics.Monitoring.Options/GlobalCounterOptions.cs
jander-msft, line 40: Let's nest the intervals under an object to allow for future default properties e.g.:

```json
"GlobalCounter": {
  "Providers": {
    "System.Runtime": {
      "Interval": 10
    }
  }
}
```

kkeirstead, line 40: Let's make sure we do validation to prevent negative intervals

documentation/configuration/metrics-configuration.md
```diff
@@ -8,6 +8,8 @@
 Due to limitations in event counters, `dotnet monitor` supports only **one** refresh interval when collecting metrics. This interval is used for
 Prometheus metrics, livemetrics, triggers, traces, and trigger actions that collect traces. The default interval is 5 seconds, but can be changed in configuration.
 
+[7.1+] For EventCounter providers, is possible to specify a different interval for each provider. See [Per provider intervals](#per-provider-intervals-71).
```
jander-msft, line 11: Are we going to hold 7.1 until all runtime issues are resolved regarding metrics? If not, then this should be `8.0+`.

src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsSettingsFactory.cs
```diff
@@ -61,6 +65,14 @@ public static MetricsPipelineSettings CreateSettings(GlobalCounterOptions counte
                 eventPipeCounterGroups.Add(new EventPipeCounterGroup { ProviderName = MonitoringSourceConfiguration.GrpcAspNetCoreServer, Type = CounterGroupType.EventCounter });
             }
 
+            foreach (EventPipeCounterGroup counterGroup in eventPipeCounterGroups)
+            {
+                if (intervalMap.TryGetValue(counterGroup.ProviderName, out GlobalProviderOptions providerInterval))
+                {
+                    counterGroup.IntervalSeconds = providerInterval.IntervalSeconds;
```
jander-msft, line 72: Should this code throw if `counterGroup.IntervalSeconds` is not null AND does not match `providerInterval.IntervalSeconds` value? In addition to this code if `counterGroup.IntervalSeconds` was null.

src/Tools/dotnet-monitor/CollectionRules/Triggers/AspNetRequestDurationTriggerFactory.cs
```diff
@@ -44,7 +44,8 @@ public ICollectionRuleTrigger Create(IEndpointInfo endpointInfo, Action callback
                 SlidingWindowDuration = options.SlidingWindowDuration ?? TimeSpan.Parse(AspNetRequestDurationOptionsDefaults.SlidingWindowDuration, CultureInfo.InvariantCulture),
             };
 
-            var aspnetTriggerSourceConfiguration = new AspNetTriggerSourceConfiguration(_counterOptions.CurrentValue.GetIntervalSeconds());
+            //HACK we get the provider specific interval for the configuration
```
jander-msft, line 47: How is this a hack?

jander-msft (dismissed):

documentation/configuration/metrics-configuration.md
wiktork: My hope is that we can ship metrics in 7.1 as soon as we get a backport for the most egregious issue.

src/Tools/dotnet-monitor/CollectionRules/Triggers/AspNetRequestDurationTriggerFactory.cs
wiktork: If AspNetTriggerSourceConfiguration uses a different provider or changes some implementation details, the caller code will also have to be adjusted.

src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsSettingsFactory.cs
wiktork: We have just constructed counterGroup, so it will always start as null.

jander-msft: > We have just constructed counterGroup, so it will always start as null.

This function does not know that; that is only necessarily true for the default groups. The `createCounterGroups` callback can construct whatever it wants, including non-null values for this property.

If you don't want to do that, at least `Debug.Assert` that each of these are null.

jander-msft (approved):

jander-msft (approved):

PR #3626, opened 2023-02-09
Add exceptions pipeline and tests

jander-msft:
###### Summary

- Add exceptions pipeline to establish event pipe session for extracting exception information from the exceptions event source and place it into a store. The store has a callback mechanism that allows for reacting to when an exception instance is stored.
- Add unit tests for the pipeline and the store.
- Update the exceptions event source to be more friendly to event pipe readers. This required using self-describing events, removing the methods with the `byte*` parameter in favor of the actual method. This comes at the expense of in-proc EventListeners not being able to read out the TypeArgs correctly; the decoding mechanism in the in-proc dispatcher doesn't handle non-`byte` element arrays and just reports them as strings.


src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
```diff
@@ -148,7 +145,29 @@ internal sealed class ExceptionsEventSource : EventSource
...
+        {
+            Flush();
+        }
+
+        [NonEvent]
+        private void RestartFlushingEventTimer()
+        {
+            // This will reset the timer to fire after the specified time period. If the timer is already
+            // started, it will be reset. If it already finished, it will be started again.
+            _flushEventsTimer.Change(TimeSpan.FromMilliseconds(100), Timeout.InfiniteTimeSpan);
```
kkeirstead, line 170: Does the `100` hold special significance here?

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
```diff
@@ -167,13 +186,49 @@ private unsafe void WriteEventCore(int eventId, Span<EventData> data)
...
+        private static int GetArrayDataSize<T>(T[] data) where T : unmanaged
+        {
+            // Arrays are written with a length prefix + the data as bytes
+            return ArrayLengthFieldSize + Unsafe.SizeOf<T>() * data.Length;
+        }
+
+        private static unsafe void FillArrayData<T>(Span<byte> target, T[] source) where T : unmanaged
+        {
+#if DEBUG
+            if (target.Length != GetArrayDataSize(source))
```
kkeirstead, line 212: Just for my own understanding - when might this scenario occur? My initial thought was if the `stackalloc` didn't allocate enough space the `target.Length` could be too small, but my understanding is that if we tried to allocate more than what's available we would get a `StackOverflowException` (and we'd never get here). I haven't worked with `stackalloc` before so I figured I'd clarify.

jander-msft: Dev authoring error. If you try to pass a Span that wasn't sized to what was expected. Should never be an issue at runtime.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
jander-msft: Yes, I'll change it to a constant.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
```diff
@@ -148,7 +155,29 @@ internal sealed class ExceptionsEventSource : EventSource
...
             SetValue(ref data[NameIdentificationEvents.TokenDescPayloads.Name], namePinned);
 
-            WriteEventCore(TokenDescriptionEventId, data);
+            WriteEventCore(ExceptionEvents.EventIds.TokenDescription, data);
+
+            RestartFlushingEventTimer();
+        }
+
+        [Event(ExceptionEvents.EventIds.Flush)]
+        public void Flush()
```
wiktork, line 164: Should this be public? We would want to either force all the clients to flush after a batch of events, or do it automatically with a timer, but not both.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
```diff
@@ -148,7 +155,29 @@ internal sealed class ExceptionsEventSource : EventSource
...
+        {
+            Flush();
+        }
+
+        [NonEvent]
+        private void RestartFlushingEventTimer()
+        {
+            // This will reset the timer to fire after the specified time period. If the timer is already
+            // started, it will be reset. If it already finished, it will be started again.
+            _flushEventsTimer.Change(EventSourceBufferAvoidanceTimeout, Timeout.InfiniteTimeSpan);
```
wiktork, line 180: Is EventSource disposable? We should be disposing _flushEventsTimer

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
jander-msft: I'll have to check if the event source manifest generation considers internal/private methods.

jander-msft: Changed to private

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
jander-msft: It is and now disposing the timer

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
```diff
@@ -167,13 +202,51 @@ private unsafe void WriteEventCore(int eventId, Span<EventData> data)
             data.Size = sizeof(T);
         }
 
+        [NonEvent]
+        private static unsafe void SetValue(ref EventData data, in Span<byte> value)
+        {
+            data.DataPointer = (nint)Unsafe.AsPointer(ref value.GetPinnableReference());
```
wiktork, line 208: Can you put some comments here regarding the expected type of input? For example, if you forget to pin reference data and send it here as a span, a GC could destroy the validity of the pointer since it's not actually pinned.

wiktork (approved):

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
jander-msft: Done

wiktork (approved):

PR #3637, opened 2023-02-10
Generate openapi.json at runtime (aad-auth pt 1)

schmittjoseph:
###### Summary

This is the first in a series of PRs to enable AzureAd auth (aad). The combined changes for aad auth needed are unwieldy for a single code review, so I'm breaking it up across several given that its impacting our authentication code.

We currently embed a prebuilt `openapi.json` file which limits our ability to customize its authentication options depending on how the user has configured dotnet-monitor (e.g. the auth section in swagger should look different between no-auth, api key, and aad). This PR makes dotnet-monitor generate the `openapi.json` file at runtime.

**NOTE**: We will still have a checked-in `openapi.json`, but it no longer contains information about how to authenticate against an arbitrary dotnet-monitor installation. This is the only behavioral difference in this PR.

### DIFF
--- a/src/Tools/dotnet-monitor/Swagger/SwaggerGenOptionsExtensions.cs
+++ b/src/Tools/dotnet-monitor/Swagger/SwaggerGenOptionsExtensions.cs
@@ -17,8 +17,6 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Swagger
     {
         public static void ConfigureMonitorSwaggerGen(this SwaggerGenOptions options)
         {
-            ConfigureMonitorSwaggerGenSecurity(options);
-
             options.DocumentFilter<BadRequestResponseDocumentFilter>();
             options.DocumentFilter<UnauthorizedResponseDocumentFilter>();
             options.DocumentFilter<TooManyRequestsResponseDocumentFilter>();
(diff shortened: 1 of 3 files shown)


src/Tools/dotnet-monitor/Swagger/SwaggerGenOptionsExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+            options.DocumentFilter<TooManyRequestsResponseDocumentFilter>();
+
+            options.OperationFilter<BadRequestResponseOperationFilter>();
+            options.OperationFilter<RemoveFailureContentTypesOperationFilter>();
+            options.OperationFilter<TooManyRequestsResponseOperationFilter>();
+            options.OperationFilter<UnauthorizedResponseOperationFilter>();
+
+            string documentationFile = $"{typeof(DiagController).Assembly.GetName().Name}.xml";
+            string documentationPath = Path.Combine(AppContext.BaseDirectory, documentationFile);
+            options.IncludeXmlComments(documentationPath);
```
jander-msft, line 31: Ideally, the `Microsoft.Diagnostics.Monitoring.WebApi.xml` file is not shipped (but it currently is) and we don't rely on it here but I think this is okay for now.

src/Tools/dotnet-monitor/Startup.cs
```diff
@@ -81,6 +81,7 @@ public static void Configure(IApplicationBuilder app, IWebHostEnvironment env, I
                 app.UseHsts();
             }
 
+            app.UseSwagger();
             app.UseSwaggerUI(options =>
             {
                 options.SwaggerEndpoint("/swagger/v1/swagger.json", "dotnet-monitor v1.0");
```
jander-msft, line 87: This seems like it's still loading the statically produced document.

jander-msft (approved):

PR #3665, opened 2023-02-13
Refactor authentication initialization to support additional modes (aad-auth pt 2)

schmittjoseph:
###### Summary

This is the second in a series of PRs to enable AzureAd auth.

This PR refactors how authentication is initialized, creating a simple way to add additional authentication modes in the future. 

High level changes: 
1. A key concept in this PR is the notion of a `StartupAuthenticationMode` which can be `Deferred`. When the user hasn't explicitly set what auth mode they want via a command line option, we now defer deciding which auth mode until we can process the user's configuration.
2. Authentication specific code is now isolated into dedicated classes (type: `IAuthenticationConfigurator`) with a single factory that determines which auth mode to use based on command line options and the user's configuration.

Behavioral changes:
1. Temporary API Key:
   -  The user-supplied `--temp-apikey` flag currently can be effectively ignored by us and the temp bearer token we provide to the user is non-functional in two different scenarios (**fixed by this PR**):
      - The user has a configuration json file with `Authentication.MonitorApiKey` already set.
      - The user updates their configuration file's `Authentication.MonitorApiKey` while `dotnet-monitor` is running.
   - Before this PR, `config show --temp-apikey` would overwrite the user's `Authentication.MonitorApiKey` config section with the temporary key. This PR changes that and preserves the user's original configuration.
2. Swagger / OpenAPI.json
   - We now dynamically configure Swagger with the correct security requirements depending on the authentication mode used. e.g. In `--no-auth` mode, no authentication options are described or presented to the user.

###### Release Notes Entry
The `config show --temp-apikey` command no longer shows the temporary API key information in the `Authentication.MonitorApiKey` config section.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunner.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunner.cs
@@ -216,13 +216,13 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests.Runners
 
         public async Task WriteUserSettingsAsync(RootOptions options)
         {
-            await WriteSettingsFileAsync(options, UserSettingsFilePath);
+            await WriteSettingsFileAsync(options, UserSettingsFilePath).ConfigureAwait(false);
             _outputHelper.WriteLine("Wrote user settings.");
         }
 
         public async Task WriteExplicitlySetSettingsFileAsync(RootOptions options)
         {
-            await WriteSettingsFileAsync(options, ExplicitlySetSettingsFilePath);
+            await WriteSettingsFileAsync(options, ExplicitlySetSettingsFilePath).ConfigureAwait(false);
             _useExplicitlySetConfiguration = true;
             _outputHelper.WriteLine("Wrote settings file.");
         }


src/Tools/dotnet-monitor/Auth/IAuthHandler.cs
```diff
@@ -0,0 +1,20 @@
...
+using System;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Auth
+{
+    internal interface IAuthHandler
+    {
+        void ConfigureApiAuth(IServiceCollection services, HostBuilderContext context);
+        void ConfigureSwaggerGenAuth(SwaggerGenOptions options);
+        void ConfigureSwaggerUI(SwaggerUIOptions options);
+        void LogStartup(ILogger logger, IServiceProvider serviceProvider);
```
jander-msft, line 18: All of the other methods adjust configuration whereas this logs information post-configuration. I would be cleaner if this were either (a) folded into the ConfigureApiAuth where a IStartupLogger implementation is provided, or (b) refactored to `IStartupLogger CreateStartupLogger(ILogger, IServiceProvider);` so that the IStartupLogger instance can be inserted into the correct place in the `ConfigureStartupLoggers` method. This then allows `IAuthHandler` to not be registered as a service since it no longer executes any implementation itself for post-configuration executable and can be purely an implementation detail of configuration.

src/Tools/dotnet-monitor/Auth/AuthenticationStartupLogger.cs
```diff
@@ -1,51 +1,42 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using Microsoft.Diagnostics.Tools.Monitor.Auth;
 using Microsoft.Extensions.Logging;
+using System;
 
 namespace Microsoft.Diagnostics.Tools.Monitor
 {
     internal sealed class AuthenticationStartupLogger :
```
jander-msft, line 10: If IAuthHandler.LogStartup is changed as suggested, this becomes greatly simplified. In fact, it can be folded into the existing `AddressListenResultsStartupLogger` class since it is already responsible for logging information based on the `AddressListenResults`.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunner.cs
```diff
@@ -130,6 +135,12 @@ public virtual async Task StartAsync(string command, string[] args, Cancellation
                 argsList.AddRange(args);
             }
 
+            if (ExplicitlySetConfigurationFilePath)
```
jander-msft, line 138: Instead of overloading the use of the `UserSettingsFilePath`, please add another method similar to `WriteUserSettingsAsync` for writing out a configuration file that would be passed via the command line. You can write it into directory under `_tempDir` so that it is cleaned up when the runner is disposed.

src/Tools/dotnet-monitor/HostBuilder/HostBuilderHelper.cs
```diff
@@ -75,17 +76,6 @@ public static IHostBuilder CreateHostBuilder(HostBuilderSettings settings)
                     builder.AddKeyPerFile(path, optional: true, reloadOnChange: true);
                     builder.AddEnvironmentVariables(ToolIdentifiers.StandardPrefix);
 
-                    if (settings.Authentication.KeyAuthenticationMode == KeyAuthenticationMode.TemporaryKey)
```
jander-msft, line 78: Is there a technical reason why this cannot be moved to right after `FileInfo userFilePath = settings.UserProvidedConfigFilePath;` to fix the intended behavior (`--temp-apikey` always wins), keep the rest of the refactor as-is, and avoid the technical breaking change?

jander-msft: Spoke offline; we'll keep the "breaking" change for 8.0. If we backport to `release/7.x`, we'll make an affordance to add back the temporary API key to `config show`.

jander-msft (approved):

Issue #3644, opened 2023-02-10
S3 Egress Provider Produces Empty Logs


kkeirstead:
### Description

When collecting logs and egressing them to S3 Storage, `dotnet monitor` is able to egress a file to storage; however, it is always empty. It has been confirmed that logs are being produced/captured by `dotnet-monitor` (by egressing locally and to Azure Blob Storage), and that the issue is isolated to the S3 egress provider.

### Other information

It appears that the `Write` override method within `MultiPartUploadStream` is required to egress logs (other artifacts use `WriteAsync`) - more investigation is needed.

jander-msft:
This will be fixed by the new egress extension architecture that we plan to target for 8.0. We won't be fixing this in the `main` branch until the egress extensibility is complete, unless there is overwhelming need to do so before its completion.

kkeirstead:
This issue should now be resolved as of [8.0 Preview 4](https://github.com/dotnet/dotnet-monitor/releases/tag/v8.0.0-preview.4.23260.4)

Issue #3643, opened 2023-02-10
S3 Egress Fails While Uploading Parts 


kkeirstead:
### Description

When egressing an artifact to S3, the following exception occurs: `Microsoft.Diagnostics.Tools.Monitor.Egress.EgressException: S3 storage egress failed: S3 storage egress failed: Part number must be an integer between 1 and 10000, inclusive` within `DoWriteAsync`. It appears that `_parts.Count` is 0 at this point, which is causing the exception.

### Other information

As a workaround, using `_parts.Count + 1` when uploading a part seems to resolve the issue (when `_parts.Count` is 0, this translates to uploading part 1) - more investigation is needed.

Issue #3642, opened 2023-02-10
Setting `RegionName` for S3 Egress Not Working


kkeirstead:
### Description

When setting the `RegionName` for AWS S3 Egress (e.g. "us-east-2"), `dotnet monitor` is unable to find the bucket and the egress operation does not complete (`dotnet monitor` does not log a success/failure message). 

### Other information

Setting the configuration's `RegionEndpoint` (instead of the `AuthenticationRegion`) appears to resolve this issue - more investigation is needed.

PR #3702, opened 2023-02-16
[Egress Extensibility] Convert S3 to An Egress Extension (Pt 1)

kkeirstead:
###### Summary

This is the first part of converting the S3 egress functionality into an extension (similar to how Azure Blob Storage is currently set up in this feature branch). For the most part, there is no "new" code here - portions of the code have been transferred from the core product to an extension, and other supporting files have been copied from the Azure extension. 

At this time, **this extension does not have the fixes for #3642 #3643 #3644 and thus is not fully functional**. Currently, the plan is for the fixes for these issues to be ported over to this branch after being made first in `main`. Using the workarounds specified in the aforementioned issues, this extension can egress artifacts to S3 storage.

Future work that still needs to be done around this extension includes:

- Cleaning up unused code paths / functions that were left in here for a simpler code review
- Improvements to validation and error-handling
- Re-enabling (and adding additional) testing
- Fixing docs


src/Extensions/S3/EgressArtifactResult.cs
```diff
@@ -0,0 +1,29 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics;
+
+namespace Microsoft.Diagnostics.Monitoring.S3
+{
+    [DebuggerDisplay("{Succeeded?\"Succeeded\":\"Failed\",nq}: {Succeeded?ArtifactPath:FailureMessage}")]
+    internal sealed class EgressArtifactResult
```
jander-msft, line 9: For all of these files that are copies from the other extension, please:
- Create a new project (e.g. Microsoft.Diagnostics.Monitoring.Extension.Common)
- Add these classes to that project; make sure all of the types are internal
- Reference this project in each of the extension projects.

src/Extensions/S3/S3/S3StorageEgressProvider.cs
```diff
@@ -2,42 +2,39 @@
 // The .NET Foundation licenses this file to you under the MIT license.
```
jander-msft, line 2: This file should be moved up one directory so that it is not nested in two `S3` directories.

src/Extensions/S3/Program.cs
```diff
@@ -0,0 +1,113 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.CommandLine;
+using System.CommandLine.Parsing;
+using System.Text.Json;
+
+namespace Microsoft.Diagnostics.Monitoring.S3
+{
+    internal sealed class Program
```
jander-msft, line 10: Because this is >95% a copy of what's in the Azure extension, I'd like to see this moved to the Common project as a utility method that you can parameterize (passing it the args from the Main entrypoint, specifying the egress provide via generic parameter, etc).

src/Extensions/S3/S3.csproj
```diff
@@ -0,0 +1,35 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
```
jander-msft, line 1:
```suggestion
﻿<Project Sdk="Microsoft.NET.Sdk">
```

src/Extensions/S3/EgressArtifactResult.cs
kkeirstead: Do you want this as part of this PR, or in a follow-up? In theory it shouldn't be too much extra since it's just renamespacing and moving some files, but I want to make sure this one doesn't become too unwieldy from a review perspective.

src/Extensions/S3/Program.cs
kkeirstead: Same question as above

src/Extensions/S3/EgressArtifactResult.cs
jander-msft: I think you should do it as part of this PR; there won't be as many new files because they'll look like moves with minor changes in them

src/Extensions/S3/Program.cs
jander-msft: Same response as above

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/SharedEntrypoint.cs
```diff
@@ -0,0 +1,122 @@
...
+{
+    internal sealed class SharedEntrypoint<T> where T : class, IEgressProviderOptions, new()
+    {
+        private static Stream StdInStream;
+        private static CancellationTokenSource CancelSource = new CancellationTokenSource();
+
+        private static EgressProvider _provider;
+        private static Action<ExtensionEgressPayload, T> _configureOptions;
+
+        public static async Task<int> Entrypoint(string[] args, EgressProvider provider, Action<ExtensionEgressPayload, T> configureOptions)
```
jander-msft, line 25: I think this whole method should be moved back to the Program classes for each extension. This particular class should focus on egressing whereas the extension can focus on translating command line to invoking this class (separation of concerns). Maybe keep the egress command creation part in this class; follow the pattern we use for the unit test app: https://github.com/dotnet/dotnet-monitor/blob/main/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/AsyncWaitScenario.cs

I would rename this class to something like `EgressHelper`.

The `Egress` method in this class can be updated to be `private static async Task<int> Egress<TOptions>(EgressProvider provider, Action<ExtensionEgressPayload, TOptions> configureOptions = null)`

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/SharedEntrypoint.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            _provider = provider;
+            _configureOptions = configureOptions;
+
+            using var loggerFactory = LoggerFactory.Create(builder =>
```
jander-msft, line 30: This logger factory doesn't do anything so it should be removed.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/Utilities.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Diagnostics.Monitoring.Extension.Common
+{
+    internal static class Utilities
+    {
+        public static ILogger CreateLogger<T>() where T : class, IEgressProviderOptions, new()
```
jander-msft, line 10: This creates a logger from a source that is immediately disposed. For this particular logger, it won't be a crashing issue (it might be for others). It however basically just writes the message to the console directly.

However, this doesn't seem to be a new issue introduced by this change, so this is okay for now.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/SharedEntrypoint.cs
```diff
@@ -0,0 +1,122 @@
...
+        private static async Task<int> Egress()
+        {
+            EgressArtifactResult result = new();
+            try
+            {
+                string jsonConfig = Console.ReadLine();
+                ExtensionEgressPayload configPayload = JsonSerializer.Deserialize<ExtensionEgressPayload>(jsonConfig);
+                IEgressProviderOptions options = BuildOptions(configPayload);
+
+                Console.CancelKeyPress += Console_CancelKeyPress;
```
jander-msft, line 56: I'd like to see the cancellation hook moved to a higher layer, as well as the source of the stdin and stdout streams; this will allow for easier testing of components. This does not have to be done in this change.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/SharedEntrypoint.cs
kkeirstead: Accidental left-over - removed.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressProviderOptions.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.Diagnostics.Monitoring.Extension.Common
+{
+    internal interface IEgressProviderOptions
```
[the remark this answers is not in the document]
kkeirstead: I like this a lot better as well - it's a lot cleaner than how I had it originally.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/SharedEntrypoint.cs
kkeirstead: This should be resolved, but I wasn't entirely sure how you wanted the command work split out. Let me know if you had something different in mind and I can make some more adjustments.

src/Extensions/AzureBlobStorage/AzureBlobEgressProvider.cs
```diff
@@ -94,7 +94,7 @@ public AzureBlobEgressProvider(ILogger logger)
                 throw CreateException(ex);
             }
         }
-
+
```
jander-msft, line 97: nit: remove whitespace

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/Microsoft.Diagnostics.Monitoring.Extension.Common.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <IsShippingAssembly>true</IsShippingAssembly>
```
jander-msft, line 4: This shouldn't be necessary since the Directory.Build.props sets IsShipping to true.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/Microsoft.Diagnostics.Monitoring.Extension.Common.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
```
jander-msft, line 3: You'll want to add `<PackAsTool>false</PackAsTool>`.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/Microsoft.Diagnostics.Monitoring.Extension.Common.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
```
jander-msft, line 1:
```suggestion
﻿<Project Sdk="Microsoft.NET.Sdk">
```

jander-msft: You might have to pull in other libraries (e.g. Microsoft.Extensions.*) to get this working after changing the SDK.

src/Extensions/S3Storage/S3Storage.csproj
```diff
@@ -0,0 +1,45 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>$(ToolTargetFrameworks)</TargetFrameworks>
+    <RootNamespace>Microsoft.Diagnostics.Monitoring.Extension.S3Storage</RootNamespace>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>disable</Nullable>
```
jander-msft, line 7:
```suggestion
```

src/Extensions/S3Storage/S3Storage.csproj
```diff
@@ -0,0 +1,45 @@
...
+
+    <Description>S3 Storage extension for dotnet-monitor</Description>
+    <PackageTags>Diagnostic</PackageTags>
+    <PackageReleaseNotes>$(Description)</PackageReleaseNotes>
+    <AssemblyName>dotnet-monitor-egress-s3storage</AssemblyName>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="AWSSDK.S3" Version="$(AwsSdkS3Version)" />
+    <PackageReference Include="System.CommandLine" Version="$(SystemCommandLineVersion)" />
```
jander-msft, line 17: This reference shouldn't be necessary since it is transitively acquired through the M.D.M.Extension.Common project.

jander-msft (approved):

src/Extensions/AzureBlobStorage/Program.cs
```diff
@@ -1,155 +1,73 @@
...
-            }
 
-            // If shared access signature (SAS) was not provided but the name was provided,
-            // lookup the SAS property value from EgressOptions.Properties
-            if (string.IsNullOrEmpty(options.SharedAccessSignature) && !string.IsNullOrEmpty(options.SharedAccessSignatureName))
-            {
-                if (configPayload.Properties.TryGetValue(options.SharedAccessSignatureName, out string sasSig))
+                // If shared access signature (SAS) was not provided but the name was provided,
+                // lookup the SAS property value from EgressOptions.Properties
+                if (string.IsNullOrEmpty(options.SharedAccessSignature) && !string.IsNullOrEmpty(options.SharedAccessSignatureName))
```
wiktork, line 36: I'm assuming there is still validation happening on the dotnet-monitor side? Will the exceptions/logs thrown from the program be propagated to dotnet-monitor logs?

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -0,0 +1,104 @@
...
+
+                result.Succeeded = true;
+            }
+            catch (Exception ex)
+            {
+                result.Succeeded = false;
+                result.FailureMessage = ex.Message;
+            }
+
+            string jsonBlob = JsonSerializer.Serialize<EgressArtifactResult>(result);
```
wiktork, line 53: nit: we could avoid the intermediate string with JsonSerializer.SerializeAsync

src/Extensions/S3Storage/S3StorageEgressProviderOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public string SecretAccessKey { get; set; }
+
+        public string AwsProfileName { get; set; }
+
+        public string AwsProfilePath { get; set; }
+
+        public bool GeneratePreSignedUrl { get; set; }
+
+        [Range(typeof(TimeSpan), "00:01:00", "1.00:00:00")]
```
wiktork, line 31: I don't think these attributes are checked here correct? Should we remove them?

wiktork (approved):

src/Extensions/AzureBlobStorage/Program.cs
jander-msft: stdout is logged [here](https://github.com/dotnet/dotnet-monitor/blob/feature/egressExtension/src/Tools/dotnet-monitor/Extensibility/ProgramExtension.OutputParser.cs#L83)
stderr is logged [here](https://github.com/dotnet/dotnet-monitor/blob/feature/egressExtension/src/Tools/dotnet-monitor/Extensibility/ProgramExtension.OutputParser.cs#L91)

src/Extensions/S3Storage/S3StorageEgressProviderOptions.cs
jander-msft: I'd rather preserve these for now. We need to be able to generate the schema.json file while including these options (TBD on how to do that with the extensions) and we need to validate the options within the extension (NYI).

PR #3755, opened 2023-02-22
Add Azure AD authentication support (aad-auth pt 3)

schmittjoseph:
###### Summary

This is the final PR in a series to add Azure Active Directory authentication support.

Notable changes:
- When using deferred authentication validate the used configuration before applying authentication. If validation errors are found, fail fast and prevent `dotnet monitor` from starting and inform the user why. We do this to strictly enforce the user's deferred authentication preferences as which auth mode is used cannot be adjusted at runtime.
  - This is **not** a breaking change as the API Key configuration behavior is unchanged.
  - deferred auth introduced in #3665
- Allow the user to interactively authenticate with `dotnet monitor` using our in-box Swagger UI with AzureAD.
- Introduce a new method (`ConfigureSwaggerUI`) on the `IAuthenticationConfigurator` interface.
- Change the recommended (but not default) authentication mode to Azure AD.

There are currently no end-to-end AzureAD authentication tests as that will require using actual Azure Resources. End-to-end tests will be added as a separate piece of work.

###### Release Notes Entry
Add Azure Active Directory authentication support. For more details see our [Azure AD documentation](https://github.com/dotnet/dotnet-monitor/blob/main/documentation/authentication.md#azure-active-directory-authentication).

### DIFF
--- a/documentation/authentication.md
+++ b/documentation/authentication.md
@@ -22,8 +22,6 @@ Azure Active Directory integration (referred to as Azure AD) is the recommended 
 - [Add a new app role](https://learn.microsoft.com/azure/active-directory/develop/howto-add-app-roles-in-azure-ad-apps#app-roles-ui) for general API access.
 - [Configure Azure AD in dotnet monitor](./configuration/azure-ad-authentication-configuration.md).
 
-> **Note**: Only organization accounts (non-MSA) are officially supported.
-
 > **Note**: Azure AD B2C is currently not supported.
 
 ### Authenticating with a Managed Identity


src/Microsoft.Diagnostics.Monitoring.Options/AzureAdOptions.cs
```diff
@@ -0,0 +1,48 @@
...
+    internal sealed partial class AzureAdOptions
+    {
+        private const string DefaultInstanceId = "https://login.microsoftonline.com";
+        private const string DefaultTenantId = "organizations";
+
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_AzureAdOptions_Instance))]
+        [DefaultValue(DefaultInstanceId)]
+        public string Instance { get; set; } = DefaultInstanceId;
```
jander-msft, line 19: Don't assign the property with a default value; this causes any test that is applying configuration via the object model to implicitly set these properties. Instead, add extension methods (e.g. `GetInstanceId(this AzureAdOptions)` and `GetTenantId(this AzureAdOptions)` to get the value of the property or fallback to the default value. Then use the methods instead of the properties in the rest of the code.

jander-msft: Can you specifically point to some other options class that is used in binding that has a default value set? All of the struct primitive types should be nullable and the strings should not have a default value. Note that I'll talking about the default value assignment, not the `[DefaultValue(...)]` attribute, which is necessary to describe the default value in the schema.

src/Tools/dotnet-monitor/Auth/AzureAd/AzureAdAuthConfigurator.cs
```diff
@@ -0,0 +1,132 @@
...
+                });
+            });
+        }
+
+        public void ConfigureSwaggerGenAuth(SwaggerGenOptions options)
+        {
+            const string OAuth2SecurityDefinitionName = "OAuth2";
+
+            // Only present an option to authenticate if a required scope is set.
+            // Otherwise a users cannot interactively authenticate.
```
kkeirstead, line 76:
```suggestion
            // Otherwise users cannot interactively authenticate.
```

src/Tools/dotnet-monitor/Auth/AzureAd/AzureAdAuthConfigurator.cs
```diff
@@ -0,0 +1,132 @@
...
+using Microsoft.AspNetCore.Authentication.JwtBearer;
+using Swashbuckle.AspNetCore.SwaggerGen;
+using Microsoft.OpenApi.Models;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Auth.AzureAd
+{
+    internal sealed class AzureAdAuthConfigurator : IAuthenticationConfigurator
+    {
+        private readonly AzureAdOptions _azureAdOptions;
+        private readonly string _fqRequiredScope;
```
kkeirstead, line 22: Pardon my ignorance - what is the `fq` in `fqRequiredScope`?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExpectedConfigurations/AuthenticationRedacted.json
```diff
@@ -1,6 +1,7 @@
 {
-  "MonitorApiKey": {
-    "Subject": "ae5473b6-8dad-498d-b915-ffffffffffff",
-    "PublicKey": ":REDACTED:"
-  }
+    "MonitorApiKey": {
+        "Subject": "ae5473b6-8dad-498d-b915-ffffffffffff",
+        "PublicKey": ":REDACTED:"
+    },
+    "AzureAd": ":NOTPRESENT:"
```
kkeirstead, line 6: nit: I'm assuming this is sufficient to get the tests to pass, but you could consider having a configuration where `AzureAd` is present (similar to what is done above for `MonitorApiKey`).

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Options/AzureAdOptionsTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            options.RequiredRole = null;
+
+            List<ValidationResult> results = new();
+
+            // Act
+            bool isValid = Validator.TryValidateObject(options, new(options), results, validateAllProperties: true);
+
+            // Assert
+            Assert.False(isValid);
+            Assert.Single(results);
```
kkeirstead, line 74: nit: Do we want to validate that the error message was the one we expected?

src/Tools/dotnet-monitor/Auth/DeferredAuthenticationValidationException.cs
```diff
@@ -0,0 +1,40 @@
...
+        public IEnumerable<string> FailureMessages { get; }
+
+        public DeferredAuthenticationValidationException(string configurationPath, IEnumerable<ValidationResult> results) : base()
+        {
+            ConfigurationPath = configurationPath;
+
+            List<string> failureMessages = new (results.Count());
+            foreach (ValidationResult result in results)
+            {
+                if (result.ErrorMessage != null)
```
kkeirstead, line 23: nit: not sure if applicable in this use case, but should this use `IsEmptyOrNull`?

jander-msft: Agreed, no point in adding a failure message that is empty.

jander-msft: Maybe even better to use `IsNullOrWhitespace`.

src/Tools/dotnet-monitor/Auth/AuthConfiguratorFactory.cs
```diff
@@ -29,12 +33,34 @@ public static IAuthenticationConfigurator Create(StartupAuthenticationMode start
...
                     throw new NotSupportedException();
             }
         }
+
+        private static void ValidateAuthConfigSection<T>(T options, string configurationPath)
+        {
+            List<ValidationResult> results = new();
+            if (!Validator.TryValidateObject(options, new ValidationContext(options), results, validateAllProperties: true))
+            {
+                throw new DeferredAuthenticationValidationException(configurationPath, results);
```
jander-msft, line 62: Because this is thrown during configuration, I think this is going to take down the process. This is not the existing behavior for the api key authentication; it keeps the process going but you have no way to access the api surface because auth is not configured (exception for maybe via Windows auth). This is a valid way to run because automation via collection rules will still work.

Since this only applies to a new feature, it is not technically a breaking change; however, it is a difference in behavior compared to the other authentication options.

jander-msft (approved):
Other than the crash-on-bad-configuration comment, everything else LGTM.

src/Tools/dotnet-monitor/Auth/AzureAd/AzureAdAuthConfigurator.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        public void ConfigureSwaggerGenAuth(SwaggerGenOptions options)
+        {
+            const string OAuth2SecurityDefinitionName = "OAuth2";
+            const string AzureAdBearerTokenSecurityDefinitionName = "AzureAd JWT";
+
+            // Only present an option to interactively authenticate if a swagger scope is set.
+            if (_fqSwaggerScope != null)
+            {
+                Uri baseEndpoint = new Uri(_azureAdOptions.GetInstance(), $"{_azureAdOptions.GetTenantId()}/oauth2/v2.0/");
```
wiktork, line 68: nit: Formattable.Invariant

src/Tools/dotnet-monitor/Auth/DeferredAuthenticationValidationException.cs
```diff
@@ -0,0 +1,40 @@
...
+            {
+                if (!string.IsNullOrWhiteSpace(result.ErrorMessage))
+                {
+                    failureMessages.Add(result.ErrorMessage);
+                }
+            }
+
+            FailureMessages = failureMessages;
+        }
+
```
wiktork, line 31: nit: It seems like we could have a shared implementation here and a mapping between results and messages

documentation/configuration/azure-ad-authentication-configuration.md
```diff
@@ -0,0 +1,56 @@
...
+# Azure Active Directory Authentication Configuration (8.0+)
+
+Azure Active Directory authentication must be configured before `dotnet monitor` starts, it does not support being configured or changed at runtime.
+
+## Configuration Options
+
+| Name | Type | Required | Description |
+|---|---|---|---|
+| ClientId | string | true | The unique application (client) id assigned to the app registration in Azure Active Directory. |
+| RequiredRole | string | true | The role required to be able to authenticate. |
```
wiktork, line 13: Is there a recommended default role for full access we could recommend here and perhaps use that as a default in our config?

documentation/authentication.md
```diff
@@ -3,11 +3,59 @@
...
+- [Add a new app role](https://learn.microsoft.com/azure/active-directory/develop/howto-add-app-roles-in-azure-ad-apps#app-roles-ui) for general API access.
+- [Configure Azure AD in dotnet monitor](./configuration/azure-ad-authentication-configuration.md).
+
+> **Note**: Azure AD B2C is currently not supported.
+
+### Authenticating with a Managed Identity
+
+`dotnet monitor` supports other applications calling its APIs using a [Managed Identity](https://learn.microsoft.com/azure/active-directory/managed-identities-azure-resources/) when Azure AD is configured. You will need to have added an app role to `dotnet monitor`'s App Registration as described above and then assign it to the Managed Identity. This can be done using [New-AzureADServiceAppRoleAssignment](https://learn.microsoft.com/powershell/module/azuread/new-azureadserviceapproleassignment?view=azureadps-2.0). Note that you will need to [configure Azure AD in dotnet monitor](./configuration/azure-ad-authentication-configuration.md) with the app role information.
```
wiktork, line 29: nit: consider the az equivalent since the token retrieval guidance is using az

documentation/schema.json
```diff
@@ -431,6 +442,58 @@
...
+          "format": "uri",
+          "default": "https://login.microsoftonline.com"
+        },
+        "TenantId": {
+          "type": [
+            "null",
+            "string"
+          ],
+          "description": "The tenant id of the Azure Active Directory tenant, or its tenant domain.",
+          "default": "organizations"
```
wiktork, line 468: I was under the impression that the tenantid is a guid. Does this get resolved by aad? Are there other meaningful values here?

src/Tools/dotnet-monitor/Auth/AzureAd/AzureAdAuthConfigurator.cs
```diff
@@ -0,0 +1,124 @@
...
+                    {
+                        options.Instance = _azureAdOptions.GetInstance().ToString();
+                        options.TenantId = _azureAdOptions.GetTenantId();
+                        options.ClientId = _azureAdOptions.ClientId;
+                    }
+                );
+
+            services.AddAuthorization(options =>
+            {
+                options.AddPolicy(AuthConstants.PolicyName, (builder) =>
```
wiktork, line 53: So to be clear, this will use the same [Authorize(Policy = AuthConstants.PolicyName)] attribute on our controllers correct?

wiktork (dismissed):

documentation/authentication.md
```diff
@@ -3,11 +3,59 @@
 
 # Authentication
 
-Authenticated requests to `dotnet monitor` help protect sensitive diagnostic artifacts from unauthorized users and lower privileged processes. `dotnet monitor` can be configured to use either [Windows Authentication](#windows-authentication) or via an [API Key](#api-key-authentication). It is possible, although strongly not recommended, to [disable authentication](#disabling-authentication).
+Authenticated requests to `dotnet monitor` help protect sensitive diagnostic artifacts from unauthorized users and lower privileged processes. `dotnet monitor` can be configured to use any one of the following authentication modes:
+- [API Key](#api-key-authentication)
+- [Azure Active Directory Authentication](#azure-active-directory-authentication) (8.0+)
```
wiktork, line 8: Would this work/ would you be able to test it with an MSA account rather than the organizational account?

src/Tools/dotnet-monitor/Auth/AuthConfiguratorFactory.cs
jander-msft: I'm okay with it as-is, however, if we do end up supporting both API key and AzureAD (and whatever future authN mechanism) side-by-side, then I think this behavior will need to be changed to allow one authN configuration to fail but keep the process running.

We've gone through great lengths to keep dotnet-monitor running (this is the primary reason why configuration of Kestrel is so complicated; the user specifying https without a configured certificate would normally bring down the process but we worked around that in dotnet-monitor; this was done so that a user can setup dotnet-monitor to collect metrics without having to configure the rest of it), even when misconfigured for some aspect. Not having the http API usable, while a big pain, is not catastrophic to dotnet-monitor working for other uses.

jander-msft (dismissed):

wiktork (approved):

PR #3759, opened 2023-02-23
Add substitution tokens for other process info

wiktork:

### DIFF
--- a/src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
+++ b/src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
@@ -151,11 +151,15 @@ namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules
                     }
                 }
             }
+            string commandLine = _ruleContext.EndpointInfo?.CommandLine ?? string.Empty;
 
             settings = _tokenParser.SubstituteOptionValues(settings, new TokenContext
             {
                 CloneOnSubstitution = ReferenceEquals(originalSettings, settings),
-                RuntimeId = _ruleContext.EndpointInfo?.RuntimeInstanceCookie ?? Guid.Empty
+                RuntimeId = _ruleContext.EndpointInfo?.RuntimeInstanceCookie ?? Guid.Empty,
+                ProcessId = _ruleContext.EndpointInfo?.ProcessId ?? 0,
+                CommandLine = commandLine,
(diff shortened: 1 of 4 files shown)


jander-msft (approved):

src/Tools/dotnet-monitor/CollectionRules/ActionOptionsDependencyAnalyzer.cs
```diff
@@ -151,11 +151,15 @@ public object SubstituteOptionValues(IDictionary<string, CollectionRuleActionRes
...
                 }
             }
+            string commandLine = _ruleContext.EndpointInfo?.CommandLine ?? string.Empty;
 
             settings = _tokenParser.SubstituteOptionValues(settings, new TokenContext
             {
                 CloneOnSubstitution = ReferenceEquals(originalSettings, settings),
-                RuntimeId = _ruleContext.EndpointInfo?.RuntimeInstanceCookie ?? Guid.Empty
+                RuntimeId = _ruleContext.EndpointInfo?.RuntimeInstanceCookie ?? Guid.Empty,
+                ProcessId = _ruleContext.EndpointInfo?.ProcessId ?? 0,
```
kkeirstead, line 160: Question - we previously set `ProcessId` to a default value of -1, but now we're falling back on 0 if there's no value. My understanding is that 0 can be a valid PID (whereas -1 cannot be...please correct me if this is inaccurate though) - is there a reason why we're using 0, as opposed to -1?

wiktork: In the unlikely case that we don't have a pid, I didn't want to embed signed values into names. I.e. artifact_-1. While you are correct, 0 is a reserved pid, I don't think we'll ever create artifacts against the 0 pid.

wiktork: As with -1, my concern here is that any parsing logic will have to handle these special cases. I.e. understanding coredump_<unsigned number> vs. coredump_<unsigned number|unknown>

jander-msft: Realistically, `_ruleContext.EndpointInfo?.ProcessId` will always have a value, so I'm not terribly concerned about it; I believe it was null only in unit test code, but this may have already been fixed as well. We'll definitely know when someone goes through the code and adds nullability checks :smile:

PR #3770, opened 2023-02-23
[release/7.x] Refactor authentication initialization to support additional modes (aad-auth pt 2)

schmittjoseph:
###### Summary

Manual backport of #3665 to release/7.x

**NOTE**:
This backport avoids the breaking change described in the PR.  See https://github.com/dotnet/dotnet-monitor/commit/37cfd6b390446b33cb7a2816ee831e93dbe214f5 for how this is done. The remainder of the backport has no functional differences.

**This PR is explicitly not being labeled with `backport` to avoid the breaking change release note entry from appearing in the `7.x` release notes.**

### DIFF
--- a/src/Tools/dotnet-monitor/Commands/ConfigShowCommandHandler.cs
+++ b/src/Tools/dotnet-monitor/Commands/ConfigShowCommandHandler.cs
@@ -3,10 +3,12 @@
 
 using Microsoft.Diagnostics.Monitoring.WebApi;
 using Microsoft.Diagnostics.Tools.Monitor.Auth;
+using Microsoft.Diagnostics.Tools.Monitor.Auth.ApiKey.Temporary;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using System;
+using System.Collections.Generic;
 using System.IO;
 
 namespace Microsoft.Diagnostics.Tools.Monitor.Commands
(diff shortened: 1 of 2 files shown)


src/Tools/dotnet-monitor/Commands/ConfigShowCommandHandler.cs
```diff
@@ -29,9 +32,32 @@ public static void Invoke(string[] urls, string[] metricUrls, bool metrics, stri
...
+            StartupAuthenticationMode startupAuthMode = HostBuilderHelper.GetStartupAuthenticationMode(noAuth, tempApiKey);
+            HostBuilderSettings settings = HostBuilderSettings.CreateMonitor(urls, metricUrls, metrics, diagnosticPort, startupAuthMode, configurationFilePath);
+            IHost host = HostBuilderHelper.CreateHostBuilder(settings)
+                .ConfigureAppConfiguration((HostBuilderContext context, IConfigurationBuilder builder) =>
+                {
+                    // HACK: generate a random jwt key and add it to the command line options.
+                    // Since we never fully startup it doesn't have to actually match what the auth configurator is using.
+                    //
+                    // Do this to avoid the breaking change described in https://github.com/dotnet/dotnet-monitor/pull/3665 in already shipping major versions.
+                    if (startupAuthMode == StartupAuthenticationMode.TemporaryKey)
```
jander-msft, line 44: If this were moved back to the `CreateHostBuilder` method as it was originally (but fixed so it's the last configuration source instead of second-to-last), then this has the downstream effect of:
1. Allowing `UserAuthorizationHandler` to not have to know a difference between a fixed subject and one that's provided via options. The way it current is does leak the notion of the fixed key vs configured key a little bit outside of the configuration layer.
2. Allowing `MonitorTempKeyAuthConfigurator` and `MonitorKeyAuthConfigurator` to be the same class with the caveat that it doesn't register and initialize the `MonitorApiKeyConfigurationObserver` for temp api key. Further optimizations could be made to the `ConfigureMonitorApiKeyAuthentication` method to no register change token sources when using temp api key.

This way, there is no difference between a configured key and a temp key to the rest of the system outside of configuration. Any future downstream consumers wouldn't have to understand the distinction. And it preserves the already shipped behavior (that the temp api key will be presented in `config show`, even though it's practically not useful). Finally, this can be applied to both 7.x and 8.x so we don't have merge conflicts if any other part of the configuration or config show aspects change and need to be backported.

PR #3771, opened 2023-02-23
Avoid breaking change with temp apikey configuration

schmittjoseph:
###### Summary

Per suggestion in #3770, this PR updates how `--temp-apikey` is initialized to avoid the breaking change in `config show --temp-apikey`.

### DIFF
--- a/src/Tools/dotnet-monitor/Auth/ApiKey/UserAuthorizationHandler.cs
+++ b/src/Tools/dotnet-monitor/Auth/ApiKey/UserAuthorizationHandler.cs
@@ -29,8 +29,8 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Auth.ApiKey
             if (context.User.Identity.AuthenticationType == AuthConstants.FederationAuthType)
             {
                 // If we get a FederationAuthType (Bearer from a Jwt Token) we need to check that the user has the specified subject claim.
-                string expectedSubjectClaim = _apiKeyConfig.CurrentValue.Subject;
-                if (context.User.HasClaim(ClaimTypes.NameIdentifier, expectedSubjectClaim))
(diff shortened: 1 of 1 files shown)


src/Tools/dotnet-monitor/Auth/ApiKey/MonitorApiKeyAuthConfigurator.cs
```diff
@@ -15,15 +16,19 @@
...
-        public AbstractMonitorKeyAuthConfigurator()
+
+        public MonitorApiKeyAuthConfigurator(GeneratedJwtKey pinnedJwtKey = null)
         {
+            _pinnedJwtKey = pinnedJwtKey;
+
             if (OperatingSystem.IsWindows())
             {
-                _enableNegotiation = true;
+                _enableNegotiation = false;
```
kkeirstead: I might just be missing something, but it looks like this value is always going to be false

src/Tools/dotnet-monitor/Auth/ApiKey/UserAuthorizationHandler.cs
```diff
@@ -36,7 +29,7 @@ protected override Task HandleRequirementAsync(AuthorizationHandlerContext conte
             if (context.User.Identity.AuthenticationType == AuthConstants.FederationAuthType)
             {
                 // If we get a FederationAuthType (Bearer from a Jwt Token) we need to check that the user has the specified subject claim.
-                string expectedSubjectClaim = _pinnedSubject ?? _apiKeyConfig.CurrentValue.Subject;
+                string expectedSubjectClaim = _apiKeyConfig.CurrentValue.Subject;
```
kkeirstead, line 32: nit: Is there any reason to keep this variable (as opposed to just using the value directly below in the conditional)?

jander-msft (approved):

PR #3790, opened 2023-02-24
[Learning Path] Using AKS To Test Dotnet Monitor Development

kkeirstead:
###### Summary

I don't have too much experience doing this, but this is the workflow I use when testing using AKS (originally based on Wiktor's script). Let me know if you'd prefer a different workflow or if there's anything missing/unnecessary.

### DIFF
--- a/documentation/learningPath/aks.md
+++ b/documentation/learningPath/aks.md
@@ -12,3 +12,28 @@ This workflow takes your local development copy of `dotnet-monitor`, patches it 
 ## Patching `dotnet monitor`
 
 >**Note**: If your changes do not involve the [.NET Core Diagnostics Repo](https://github.com/dotnet/diagnostics#net-core-diagnostics-repo), feel free to skip this step.
+
+1. Open `Powershell` and run the [generate-dev-sln script](https://github.com/dotnet/dotnet-monitor/blob/main/generate-dev-sln.ps1), providing a path to your local copy of the diagnostics repo.
+
+```bash
+cd C:\your-path\dotnet-monitor 
+.\generate-dev-sln.ps1 C:\your-path\diagnostics
+```
+
+2. Publish `dotnet monitor` to your desired (local) target location
+
+```bash
+dotnet publish .\src\Tools\dotnet-monitor -o $env:TEMP\dotnet-monitor -c Release -f net6.0
+```
+
+3. Add a Dockerfile to the `dotnet monitor` directory created in the previous step. Below is an example of the contents of the Dockerfile:
+
+```bash
+FROM mcr.microsoft.com/dotnet/nightly/aspnet:6.0-alpine 
+ENV COMPlus_EnableDiagnostics 0
+ENV ASPNETCORE_URLS=
+ENV DOTNET_MONITOR_VERSION=6
+WORKDIR /app
+COPY . .
+ENTRYPOINT ["dotnet", "dotnet-monitor.dll", "collect", "--no-auth"]
+```
--- a/documentation/learningPath/aks.md
+++ b/documentation/learningPath/aks.md
@@ -13,7 +13,7 @@ This workflow takes your local development copy of `dotnet-monitor`, patches it 
 
 >**Note**: If your changes do not involve the [.NET Core Diagnostics Repo](https://github.com/dotnet/diagnostics#net-core-diagnostics-repo), you don't need to complete this step.
 
-```bash
+```ps1
 cd C:\your-path\dotnet-monitor 
 .\generate-dev-sln.ps1 C:\your-path\diagnostics
 ```
@@ -26,7 +26,7 @@ dotnet publish .\src\Tools\dotnet-monitor -o $env:TEMP\dotnet-monitor -c Release
 
 3. Pull the latest copy of the base image to avoid using a cached version (this should be the same as the REPO used in the next step)
 
-```bash
+```ps1
 docker pull mcr.microsoft.com/dotnet/aspnet:8.0-alpine-amd64
 ```
 
@@ -68,7 +68,7 @@ CMD [ "collect", "--urls", "https://+:52323", "--metricUrls", "http://+:52325" ]
 
 5. Log in to your ACR
 
-```bash
+```ps1
 az account set -s <subscription_id>
 az aks get-credentials --resource-group <name_of_resource_group> --name <name_of_aks>
 az acr login --resource-group <name_of_resource_group> --name <name_of_acr>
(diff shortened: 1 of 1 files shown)


documentation/learningPath/aks.md
```diff
@@ -2,3 +2,58 @@
...
+
+2. Publish `dotnet monitor` to your desired (local) target location (TEMP is used throughout this example)
+
+```bash
+dotnet publish .\src\Tools\dotnet-monitor -o $env:TEMP\dotnet-monitor -c Release -f net6.0
+```
+
+3. Add a Dockerfile to the `dotnet monitor` directory created in the previous step. Below is an example of the contents of the Dockerfile:
+
+```bash
```
jander-msft, line 29: This gets .NET Monitor running in a container but it's not nearly as representative of what the actual .NET Monitor image is. It's missing many environment variables (behavior is going to be different), the entrypoint is different and lacks command (so those who are overriding it in their orchestrators are going to have mismatches), the profiler is missing, etc.

So either all of these caveats should be documented (so folks don't come us asking why it's not working as they expected), or the workflow should be updated to produce something more representative of the real image. For the latter, this would include:
1. Updating the publishing instructions to do a `.\build.cmd -ci -archive`.
1. Effectively copying https://github.com/dotnet/dotnet-docker/blob/main/src/monitor/7.0/alpine/amd64/Dockerfile and making minor adjustments for copying the built archive instead of downloading .NET Monitor.

This doesn't solve the missing profiler since that requires a Linux build (so folks doing this on Linux are in good shaped).

documentation/learningPath/aks.md
```diff
@@ -2,3 +2,58 @@
...
+az account set --subscription <subscription_id>
+az aks get-credentials --resource-group <name_of_resource_group> --name <name_of_aks>
+az acr login --resource-group <name_of_resource_group> --name <name_of_acr>
+```
+
+5. Build the Docker image locally
+
+```bash
+docker build $env:TEMP\dotnet-monitor -f $env:TEMP\dotnet-monitor\Dockerfile.localagent -t name_of_acr.azurecr.io/localagent
```
jander-msft, line 50:
```suggestion
docker build $env:TEMP\dotnet-monitor -f $env:TEMP\dotnet-monitor\Dockerfile.localagent -t <name_of_acr>.azurecr.io/localagent
```

documentation/learningPath/aks.md
```diff
@@ -2,3 +2,58 @@
...
+```
+
+6. Push the Docker image to your ACR
+
+```bash
+docker push name_of_acr.azurecr.io/localagent:latest
```
jander-msft, line 56:
```suggestion
docker push <name_of_acr>.azurecr.io/localagent:latest
```

documentation/learningPath/aks.md
jander-msft: After thinking about this more and discussing it, I think we should use your publish instructions because that will get most of .NET Monitor scenarios working without a lot of hassle. However, I think using a Dockerfile that is more representative of what is actually shipped would be beneficial in eliminating functional differences between the local image vs the official image.

documentation/learningPath/aks.md
```diff
@@ -2,3 +2,82 @@
...
+
+```bash
+dotnet publish .\src\Tools\dotnet-monitor -o $env:TEMP\dotnet-monitor -c Release -f net8.0
+```
+
+3. Add a Dockerfile to the `dotnet monitor` directory created in the previous step. Below is an example of the contents of the Dockerfile:
+
+```bash
+ARG REPO=mcr.microsoft.com/dotnet/aspnet
+FROM $REPO:7.0.3-alpine3.17-amd64
```
jander-msft, line 31:
```suggestion
FROM $REPO:8.0-alpine-amd64
```
Couple of points:
- The image target framework needs to match whichever target framework you publish for (above, you are publishing `net8.0`)
- Use the major.minor and versionless distro tag so that you get the latest supported image.
- Alpine images are not actually something we are offering for .NET 8, but it should still work.
- Add an instruction between #2 and #3 to pull the base image (e.g. `docker pull mcr.microsoft.com/dotnet/aspnet:8.0-alpine-amd64`) so you are not using a cached image that might be vulnerable.

jander-msft (dismissed):

documentation/learningPath/aks.md
```diff
@@ -2,3 +2,88 @@
...
+ENV \
+    # Unset ASPNETCORE_URLS from aspnet base image
+    ASPNETCORE_URLS= \
+    # Disable debugger and profiler diagnostics to avoid diagnosing self.
+    COMPlus_EnableDiagnostics=0 \
+    # Default Filter
+    DefaultProcess__Filters__0__Key=ProcessId \
+    DefaultProcess__Filters__0__Value=1 \
+    # Remove Unix Domain Socket before starting diagnostic port server
+    DiagnosticPort__DeleteEndpointOnStartup=true \
```
wiktork, line 50: DeleteEndpointOnStartup can be removed.

documentation/learningPath/aks.md
```diff
@@ -2,3 +2,88 @@
...
+    Logging__Console__FormatterOptions__TimestampFormat=yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffffff'Z' \
+    # Logging: Write timestamps using UTC offset (+0:00)
+    Logging__Console__FormatterOptions__UseUtcTimestamp=true \
+    # Add dotnet-monitor path to front of PATH for easier, prioritized execution
+    PATH="/app:${PATH}"
+
+COPY . .
+
+ENTRYPOINT [ "dotnet-monitor" ]
+CMD [ "collect", "--urls", "https://+:52323", "--metricUrls", "http://+:52325" ]
```
wiktork, line 65: If we want to keep this as simple as possible, I would suggest http://localhost:52323. For https, you need to deploy a certificate and certificate password.

documentation/learningPath/aks.md
jander-msft: Why?

wiktork (changes requested):
Looks good, but needs a few tweaks.

documentation/learningPath/aks.md
wiktork: I was suggesting it to keep the file simpler, but if we want to have more settings to match the real dockerfile we can keep it.

jander-msft: I'd rather keep the dockerfile as close as to the real one so that behaviors are not different, unless they raise the barrier for inner loop and debugging purposes too much.

documentation/learningPath/aks.md
```diff
@@ -2,3 +2,88 @@
...
+
+```dockerfile
+ARG REPO=mcr.microsoft.com/dotnet/aspnet
+FROM $REPO:8.0-alpine-amd64
+
+WORKDIR /app
+
+ENV \
+    # Unset ASPNETCORE_URLS from aspnet base image
+    ASPNETCORE_URLS= \
```
jander-msft, line 43: Because this is .NET 8 based, this environment variable should be `ASPNETCORE_HTTP_PORTS` because in .NET 8 the runtime-deps layer will set this instead of `ASPNETCORE_URLS`. See https://github.com/dotnet/dotnet-docker/blob/main/src/monitor/8.0/cbl-mariner-distroless/amd64/Dockerfile#L30

This may cause some confusion if folks switch between .NET 6/7/8, so maybe adding both (clears both values) would be able to cover any usage.

wiktork (dismissed):

wiktork (approved):

jander-msft (approved):

PR #3805, opened 2023-02-27
Basic exceptions store and route for default process

jander-msft:
###### Summary

- Adds a single instance of an exceptions store that contains the exception information for the default process. The exception store will quickly consume the information from the provider and queue it for processing shortly thereafter.
- Adds an exceptions service that will start the exceptions pipeline on the default process and insert the exception information into the store for the default process.
- Add the `/exceptions` route that only reports exception information from the default process. This route is authenticated and can produce `text/plain`, `application/x-ndjson`, and `application/json-seq` formats.

Example output of the `/exceptions` route:

Plain text:
```
First chance exception. System.InvalidOperationException: The current time is 2/27/2023 11:37:21 PM
First chance exception. System.InvalidOperationException: The current time is 2/27/2023 11:37:24 PM
First chance exception. System.InvalidOperationException: The current time is 2/27/2023 11:37:27 PM
First chance exception. System.InvalidOperationException: The current time is 2/27/2023 11:37:29 PM
First chance exception. System.InvalidOperationException: The current time is 2/27/2023 11:37:29 PM
```

Newline Delimited JSON:
```json
{"typeName":"System.InvalidOperationException","moduleName":"System.Private.CoreLib.dll","message":"The current time is 2/27/2023 11:37:21 PM"}
{"typeName":"System.InvalidOperationException","moduleName":"System.Private.CoreLib.dll","message":"The current time is 2/27/2023 11:37:24 PM"}
{"typeName":"System.InvalidOperationException","moduleName":"System.Private.CoreLib.dll","message":"The current time is 2/27/2023 11:37:27 PM"}
{"typeName":"System.InvalidOperationException","moduleName":"System.Private.CoreLib.dll","message":"The current time is 2/27/2023 11:37:29 PM"}
{"typeName":"System.InvalidOperationException","moduleName":"System.Private.CoreLib.dll","message":"The current time is 2/27/2023 11:37:29 PM"}
```

Caveats:

- Only the exception type and message are produced at this time. Future PRs will be able to add call stack, inner exception, and other possible ambient information such as activity IDs.
- There are no limits in place for the cache information on either the target application nor in the dotnet-monitor tool. All exception information is kept for the lifetime of each process. This will be addressed in future PRs where the caches will be cleared periodically or according to a prescribed lifetime policy (e.g. only keep exceptions for the last 15 minutes).
- The exceptions are the very beginning of the target application are not captured due to how the default process determination works. This will be addressed in a future PR where in "listen" mode, a store is created per-process and a session is initialized per-process during the diagnostic suspension point in order to capture all exception information.
- The functionality is provided via a startup hook in the target application. This startup hook is NOT automatically configured by .NET Monitor an must be manually set via the DOTNET_STARTUP_HOOKS environment variable.

###### Release Notes Entry
Add `/exceptions` route that reports captured exceptions from the default process.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
@@ -31,9 +31,19 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Pipeline.Steps
         {
             ArgumentNullException.ThrowIfNull(exception);
 
-            ulong identifier = _identifierCache.GetOrAdd(new ExceptionIdentifier(exception));
+            // Do not populate the cache or send via the EventSource until
+            // a listener is active; otherwise, the listener will not receive the identifiers
+            // for types/methods/modules/etc prior to listening to the event source. This
+            // means that exception at startup are likely to be dropped if the listener was
+            // not registered during the diagnostic startup suspension point.
+            // CONSIDER: Possible improvement is to cache the information and then send off
+            // the events once the listener is registered to effectively "catch up".
+            if (_eventSource.IsEnabled())
+            {
+                ulong identifier = _identifierCache.GetOrAdd(new ExceptionIdentifier(exception));
 
-            _eventSource.ExceptionInstance(identifier, exception.Message);
+                _eventSource.ExceptionInstance(identifier, exception.Message);
+            }
 
             _next(exception);
         }
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
@@ -34,7 +34,7 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Pipeline.Steps
             // Do not populate the cache or send via the EventSource until
             // a listener is active; otherwise, the listener will not receive the identifiers
             // for types/methods/modules/etc prior to listening to the event source. This
-            // means that exception at startup are likely to be dropped if the listener was
+            // means that exceptions at startup are likely to be dropped if the listener was
             // not registered during the diagnostic startup suspension point.
             // CONSIDER: Possible improvement is to cache the information and then send off
             // the events once the listener is registered to effectively "catch up".
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/ExceptionsController.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/ExceptionsController.cs
@@ -1,0 +1,96 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Diagnostics.Monitoring.WebApi.Exceptions;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Net.Http.Headers;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
+{
+    [Route("")]
+    [ApiController]
+    [HostRestriction]
+    [Authorize(Policy = AuthConstants.PolicyName)]
+    [ProducesErrorResponseType(typeof(ValidationProblemDetails))]
+    [ProducesResponseType(typeof(ValidationProblemDetails), StatusCodes.Status400BadRequest)]
+    public sealed class ExceptionsController : ControllerBase
+    {
+        private readonly IExceptionsStore _exceptionsStore;
+        private readonly IInProcessFeatures _inProcessFeatures;
+        private readonly IExceptionsOperationFactory _operationFactory;
+
+        public ExceptionsController(IServiceProvider serviceProvider)
+        {
+            // The exceptions store for the default process
+            _exceptionsStore = serviceProvider.GetRequiredService<IExceptionsStore>();
+            _inProcessFeatures = serviceProvider.GetRequiredService<IInProcessFeatures>();
+            _operationFactory = serviceProvider.GetRequiredService<IExceptionsOperationFactory>();
+        }
+
+        /// <summary>
+        /// Gets the exceptions from the default process.
+        /// </summary>
+        /// <returns></returns>
+        [HttpGet("exceptions", Name = nameof(GetExceptions))]
+        [ProducesWithProblemDetails(ContentTypes.ApplicationNdJson, ContentTypes.ApplicationJsonSequence, ContentTypes.TextPlain)]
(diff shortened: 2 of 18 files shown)


src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
```diff
@@ -31,9 +31,19 @@ public void Invoke(Exception exception)
         {
             ArgumentNullException.ThrowIfNull(exception);
 
-            ulong identifier = _identifierCache.GetOrAdd(new ExceptionIdentifier(exception));
+            // Do not populate the cache or send via the EventSource until
+            // a listener is active; otherwise, the listener will not receive the identifiers
+            // for types/methods/modules/etc prior to listening to the event source. This
+            // means that exception at startup are likely to be dropped if the listener was
```
kkeirstead, line 37:
```suggestion
            // means that exceptions at startup are likely to be dropped if the listener was
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/ExceptionsController.cs
```diff
@@ -0,0 +1,96 @@
...
+            // The exceptions store for the default process
+            _exceptionsStore = serviceProvider.GetRequiredService<IExceptionsStore>();
+            _inProcessFeatures = serviceProvider.GetRequiredService<IInProcessFeatures>();
+            _operationFactory = serviceProvider.GetRequiredService<IExceptionsOperationFactory>();
+        }
+
+        /// <summary>
+        /// Gets the exceptions from the default process.
+        /// </summary>
+        /// <returns></returns>
```
kkeirstead, line 39:
```suggestion
```

kkeirstead: I noticed we don't have this on our other ones and wasn't sure if it was intentionally included here, feel free to disregard.

src/Tools/dotnet-monitor/Exceptions/ExceptionsStore.cs
```diff
@@ -0,0 +1,125 @@
...
+            {
+                return new List<ExceptionInstance>(_instances).AsReadOnly();
+            }
+        }
+
+        private static Channel<ExceptionInstanceEntry> CreateChannel()
+        {
+            // TODO: Hook callback for when items are dropped an report appropriately.
+            return Channel.CreateBounded<ExceptionInstanceEntry>(
+                new BoundedChannelOptions(capacity: 1000)
```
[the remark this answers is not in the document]
jander-msft: No significance, just need some reasonable backlog count. We can adjust it or make it configurable depending on how much throughput the channel has.

src/Tools/dotnet-monitor/Exceptions/ExceptionsStore.cs
```diff
@@ -0,0 +1,131 @@
...
+            _channel.Writer.TryComplete();
+
+            await _processingTask.SafeAwait();
+
+            _disposalSource.SafeCancel();
+
+            _disposalSource.Dispose();
+        }
+
+        void IExceptionsStore.AddExceptionInstance(IExceptionsNameCache cache, ulong exceptionId, string message)
```
jander-msft, line 45: Undo explicit implementation

src/Tools/dotnet-monitor/Exceptions/ExceptionsService.cs
```diff
@@ -0,0 +1,77 @@
...
+                    // Monitor for exceptions
+                    await _pipeline.RunAsync(stoppingToken);
+                }
+                catch (Exception e) when (e is not OperationCanceledException || !stoppingToken.IsCancellationRequested)
+                {
+                    if (null != _pipeline)
+                    {
+                        await _pipeline.DisposeAsync();
+                    }
+                    await Task.Delay(5000, stoppingToken);
```
wiktork, line 63: nit: create const

src/Tools/dotnet-monitor/Exceptions/ExceptionsService.cs
```diff
@@ -0,0 +1,79 @@
...
+                    IProcessInfo pi = await _diagnosticServices.GetProcessAsync(processKey: null, stoppingToken);
+                    DiagnosticsClient client = new(pi.EndpointInfo.Endpoint);
+
+                    EventExceptionsPipelineSettings settings = new();
+                    _pipeline = new EventExceptionsPipeline(client, settings, _exceptionsStore);
+
+                    // Monitor for exceptions
+                    await _pipeline.RunAsync(stoppingToken);
+                }
+                catch (Exception e) when (e is not OperationCanceledException || !stoppingToken.IsCancellationRequested)
```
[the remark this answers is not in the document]
jander-msft: The token is only signaled when the application is being torn down. Background services can either exit gracefully or throw under this circumstance; see https://github.com/dotnet/runtime/blob/210a7a9feec64b7fac5147656cddb199ec90cf75/src/libraries/Microsoft.Extensions.Hosting/src/Internal/Host.cs#L102

This is the same behavior as the metrics service: https://github.com/dotnet/dotnet-monitor/blob/fdbfcaf49ccb104026806dc371ebb63c02d74d49/src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsService.cs#L67

I think this was done to avoid waiting 5 seconds on an exception when trying to shutdown. But we can fix both of them to catch but not wait under this circumstance (preferably in a separate PR).

src/Tools/dotnet-monitor/Exceptions/ExceptionsOperation.cs
```diff
@@ -0,0 +1,123 @@
...
+        private static async Task WriteTextInstance(Stream stream, IExceptionInstance instance, CancellationToken token)
+        {
+            // This format is similar of that which is written to the console when an unhandled exception occurs. Each
+            // exception will appear as:
+
+            // First chance exception. <TypeName>: <Message>
+
+            await using StreamWriter writer = new(stream, leaveOpen: true);
+
+            await writer.WriteAsync("First chance exception. ");
```
wiktork, line 116: Should this be in a resx or const?

src/Tools/dotnet-monitor/Exceptions/ExceptionsStore.cs
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        public void AddExceptionInstance(IExceptionsNameCache cache, ulong exceptionId, string message)
+        {
+            ExceptionInstanceEntry entry = new(cache, exceptionId, message);
+            // This should never fail to write because the behavior is to drop the oldest.
+            _channel.Writer.TryWrite(entry);
+        }
+
+        public IEnumerable<IExceptionInstance> GetSnapshot()
```
wiktork, line 54: nit: Consider returning IList instead of IEnumerable. IEnumerable signals deferred execution at the cost of random access. We really can't do deferred computation here so we might as well provide random access abilities.

src/Tools/dotnet-monitor/Exceptions/ExceptionsOperation.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            if (_format == ExceptionsFormat.JsonSequence)
+            {
+                await stream.WriteAsync(JsonSequenceRecordSeparator, token);
+            }
+
+            await using (Utf8JsonWriter writer = new(stream, new JsonWriterOptions() { Indented = false }))
+            {
+                writer.WriteStartObject();
+                writer.WriteString("typeName", instance.TypeName);
```
wiktork, line 90: Did you want to document some of the format considerations in the code?

wiktork (approved):

PR #3812, opened 2023-02-28
Adding `EventMeter` Trigger

kkeirstead:
###### Summary

This PR adds a trigger for `SystemDiagnosticsMetrics`. The options are similar to the `EventCounter` trigger, with two differences:
* Instead of `CounterName`, `InstrumentName` is used (there had previously been some discussion about unifying these names to reduce mental burden for users - I'm leaving it as is for now, but think it's worth discussing on this PR). 
* An additional property named `HistogramPercentile` that allows users to set which percentile they want to compare against for the `GreaterThan`/`LessThan` value. For now, this value would be either 50/95/99.

Docs and tests have been updated as part of this PR.

There is also a PR in the [diagnostics repo](https://github.com/dotnet/diagnostics/pull/3697) for this feature. 

###### Release Notes Entry

Added EventMeter Trigger for Collection Rules.


src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
```diff
@@ -811,6 +811,30 @@
...
+  </data>
+  <data name="DisplayAttributeDescription_SystemDiagnosticsMetricsOptions_HistogramPercentile" xml:space="preserve">
+    <value>When monitoring a histogram, this dictates which percentile to compare against using the value in GreaterThan/LessThan.</value>
+    <comment>The description provided for the HistogramPercentile parameter on SystemDiagnosticsMetricsOptions.</comment>
+  </data>
+  <data name="DisplayAttributeDescription_SystemDiagnosticsMetricsOptions_LessThan" xml:space="preserve">
+    <value>The threshold level the instrument must maintain (or lower) for the specified duration. Either GreaterThan or LessThan (or both) must be specified.</value>
+    <comment>The description provided for the LessThan parameter on SystemDiagnosticsMetricsOptions.</comment>
+  </data>
+  <data name="DisplayAttributeDescription_SystemDiagnosticsMetricsOptions_ProviderName" xml:space="preserve">
```
kkeirstead, line 830: Either have this as MeterName, or have InstrumentName be CounterName (not half and half)

kkeirstead: I've updated this to use MeterName and InstrumentName, while silently accepting ProviderName/CounterName for anyone who tries to follow the existing format

src/Tools/dotnet-monitor/CollectionRules/Options/Triggers/SystemDiagnosticsMetricsOptions.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.ComponentModel.DataAnnotations;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Triggers
+{
+    /// <summary>
+    /// Options for the SystemDiagnosticsMetrics trigger.
+    /// </summary>
+    internal sealed partial class SystemDiagnosticsMetricsOptions : ISlidingWindowDurationProperties
+    {
+        public string ProviderName { get; set; }
```
wiktork, line 16: What is the purpose of ProviderName and CounterName here?

kkeirstead: https://github.com/dotnet/dotnet-monitor/pull/3812#discussion_r1122272607

src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
wiktork: I think having duplicate options that do the same thing is more confusing than just changing the name; I think users who are familiar with System.Diagnostics.Metrics will find MeterName and InstrumentName very clear, even if not consistent with the previous trigger names.

kkeirstead: I think where this becomes interesting is that users are currently using the ProviderName/CounterName pattern to specify `SystemDiagnosticsMetrics` when listing custom providers (for `Metrics` in configuration) - which means in one part of config you'll call your meter a provider, and then you'll need to remember to call it a meter in the collection rule. That being said, it feels odd to make the official options incorrect (given that we know it's an `SDM` trigger, provider/counter aren't the correct terms, and I don't know if we want to establish a precedent of officially using the wrong name). This was intended to be a compromise that reduces the need for user education about the distinction, while keeping the provider/counter option undocumented so people don't think MeterName/ProviderName or InstrumentName/CounterName are different and both need to be configured with different values.

Basically, if a user read the documentation, they'd follow the "correct" naming scheme, but we're also not going to force users who don't read docs to understand the nuance since we know what they meant. I'm open to changing this if people feel strongly one way or another though.

src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
```diff
@@ -811,6 +811,30 @@
...
+  <data name="DisplayAttributeDescription_SystemDiagnosticsMetricsOptions_InstrumentName" xml:space="preserve">
+    <value>The name of the instrument to monitor.</value>
+    <comment>The description provided for the InstrumentName parameter on SystemDiagnosticsMetricsOptions.</comment>
+  </data>
+  <data name="DisplayAttributeDescription_SystemDiagnosticsMetricsOptions_GreaterThan" xml:space="preserve">
+    <value>The threshold level the instrument must maintain (or higher) for the specified duration. Either GreaterThan or LessThan (or both) must be specified.</value>
+    <comment>The description provided for the GreaterThan parameter on SystemDiagnosticsMetricsOptions.</comment>
+  </data>
+  <data name="DisplayAttributeDescription_SystemDiagnosticsMetricsOptions_HistogramPercentile" xml:space="preserve">
+    <value>When monitoring a histogram, this dictates which percentile to compare against using the value in GreaterThan/LessThan.</value>
```
wiktork, line 823: Maybe add the specific expected values like 50, 95, 99?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRuleOptionsTests.cs
```diff
@@ -254,6 +254,288 @@ public Task CollectionRuleOptions_EventCounterTrigger_GreaterThanLargerThanLessT
...
+                    Assert.Single(failures);
+                    VerifyRangeMessage<TimeSpan>(failures, 0, nameof(SystemDiagnosticsMetricsOptions.SlidingWindowDuration),
+                        TriggerOptionsConstants.SlidingWindowDuration_MinValue, TriggerOptionsConstants.SlidingWindowDuration_MaxValue);
+                });
+        }
+
+        [Fact]
+        public Task CollectionRuleOptions_SystemDiagnosticsMetricsTrigger_NoGreaterThanOrLessThan()
+        {
+            const string ExpectedMeterName = "Meter";
```
wiktork, line 380: nit: did you just want to make one set of const for all of these?

wiktork (commented):
Looks good, minus the additional properties for Provider/CounterName, which I think we should remove.

wiktork (approved):

documentation/collectionrules/collectionrules.md
```diff
@@ -46,6 +46,7 @@ The following are the currently available triggers:
 | [AspNetRequestDuration](../configuration/collection-rule-configuration.md#aspnetrequestduration-trigger) | Event Pipe | Satisfied when the number of HTTP requests have response times longer than the threshold duration. |
 | [AspNetResponseStatus](../configuration/collection-rule-configuration.md#aspnetresponsestatus-trigger) | Event Pipe | Satisfied when the number of HTTP responses that have status codes matching the pattern list is above the specified threshold. |
 | [EventCounter](../configuration/collection-rule-configuration.md#eventcounter-trigger) | Event Pipe | Satisfied when the value of a counter falls above, below, or between the described threshold. |
+| [SystemDiagnosticsMetrics](../configuration/collection-rule-configuration.md#systemdiagnosticsmetrics-trigger-71) | Event Pipe | Satisfied when the value of an instrument falls above, below, or between the described threshold. |
```
jander-msft, line 49: I think this should be renamed to `Meters` or `EventMeters` to be consistent with the rest of the tool. I don't think we make any reference to SDM or System.Diagnostics.Metrics anywhere, but we do call them meters.

documentation/collectionrules/collectionrules.md
```diff
@@ -46,6 +46,7 @@ The following are the currently available triggers:
 | [AspNetRequestDuration](../configuration/collection-rule-configuration.md#aspnetrequestduration-trigger) | Event Pipe | Satisfied when the number of HTTP requests have response times longer than the threshold duration. |
 | [AspNetResponseStatus](../configuration/collection-rule-configuration.md#aspnetresponsestatus-trigger) | Event Pipe | Satisfied when the number of HTTP responses that have status codes matching the pattern list is above the specified threshold. |
 | [EventCounter](../configuration/collection-rule-configuration.md#eventcounter-trigger) | Event Pipe | Satisfied when the value of a counter falls above, below, or between the described threshold. |
+| [EventMeter](../configuration/collection-rule-configuration.md#eventmeter-trigger-71) | Event Pipe | Satisfied when the value of an instrument falls above, below, or between the described threshold. |
```
jander-msft, line 49:
```suggestion
| [EventMeter](../configuration/collection-rule-configuration.md#eventmeter-trigger-80) | Event Pipe | Satisfied when the value of an instrument falls above, below, or between the described threshold. |
```

documentation/configuration/collection-rule-configuration.md
```diff
@@ -415,6 +416,111 @@ Usage that is satisfied when the CPU usage of the application is higher than 70%
   ```
 </details>
 
+### `EventMeter` Trigger (7.1+)
```
jander-msft, line 419:
```suggestion
### `EventMeter` Trigger (8.0+)
```

src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
```diff
@@ -858,4 +882,4 @@
   <data name="DisplayAttributeDescription_MetricsOptions_Meters" xml:space="preserve">
     <value>Names of meters to collect from the System.Diagnostics.Metrics provider.</value>
   </data>
-</root>
+</root>
```
jander-msft, line 885: nit: please add back the newline

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/CollectionRulePipelineTests.cs
```diff
@@ -160,6 +160,206 @@ public Task CollectionRulePipeline_EventCounterTriggerTest(TargetFrameworkMonike
                 });
         }
 
+        /// <summary>
+        /// Test that the pipeline works with the EventMeter trigger (gauge instrument).
+        /// </summary>
+        [Theory]
+        [InlineData(TargetFrameworkMoniker.Net70)]
```
jander-msft, line 167:
```suggestion
        [InlineData(TargetFrameworkMoniker.Net80)]
```

If we are only going to run these on one TFM, then it needs to be `Net8.0` so that PR will run them. However, this will prevent them from running in VS until we allow building the `net8.0` TFM in VS. This could be done with a `MemberData` attribute that calls a method that determines whether the test is running in VS based on an environment variable and return either `Net70` or `Net80`.

kkeirstead: I mimicked what the existing TFM logic does with `INCLUDE_NEXT_DOTNET` - let me know if you have something else in mind

jander-msft (approved):

PR #3817, opened 2023-03-01
Move Kestrel config to collect command and remove dynamic url config block

jander-msft:
###### Summary

Rework the `url` configuration clearing mechanism by moving Kestrel configuration to the `collect` command and simulate ASP.NET Core configuration in other usage. This provides a better separation of where Kestrel is actually used and removes the entire concept of the dynamic `url` configuration blocking mechanism.

The ConfigurationOrdering unit tests and the Kestrel functional tests show that this works as before the change. I manually validated that the `config show` command does respect the ASPNETCORE_* environment variables even though Kestrel is no longer configured for that command.

###### Release Notes Entry


src/Tools/dotnet-monitor/HostBuilder/HostBuilderHelper.cs
```diff
@@ -109,96 +109,6 @@ public static IHostBuilder CreateHostBuilder(HostBuilderSettings settings)
 
                         builder.AddCommandLine(arguments.ToArray());
                     }
-                })
-                //Note this is necessary for config only because Kestrel configuration
-                //is not added until WebHostDefaults are added.
-                .ConfigureWebHostDefaults(webBuilder =>
```
jander-msft, line 115: This mostly moved to the CollectCommandHandler and the dynamic configuration source for blocking the `urls` key has been completely removed.

PR #3839, opened 2023-03-03
Split meters from counters in the livemetrics route and action

jander-msft:

### DIFF
--- a/documentation/api/livemetrics-custom.md
+++ b/documentation/api/livemetrics-custom.md
@@ -126,10 +126,10 @@ Authorization: Bearer fffffffffffffffffffffffffffffffffffffffffff=
 
 {
     "includeDefaultProviders": false,
-    "providers": [
+    "meters": [
         {
-            "providerName": "CustomProvider",
-            "counterNames": [
+            "meterName": "CustomMeter",
+            "instrumentNames": [
                 "myHistogram"
             ]
         }
@@ -146,7 +146,7 @@ Location: localhost:52323/operations/67f07e40-5cca-4709-9062-26302c484f18
 
 {
     "timestamp": "2021-08-31T16:58:39.7514031+00:00",
-    "provider": "CustomProvider",
+    "provider": "CustomMeter",
     "name": "myHistogram",
     "displayName": "myHistogram",
     "unit": null,
@@ -156,7 +156,7 @@ Location: localhost:52323/operations/67f07e40-5cca-4709-9062-26302c484f18
 }
 {
     "timestamp": "2021-08-31T16:58:39.7514031+00:00",
-    "provider": "CustomProvider",
+    "provider": "CustomMeter",
     "name": "myHistogram",
     "displayName": "myHistogram",
     "unit": null,
@@ -166,7 +166,7 @@ Location: localhost:52323/operations/67f07e40-5cca-4709-9062-26302c484f18
 }
 {
     "timestamp": "2021-08-31T16:58:39.7514031+00:00",
-    "provider": "CustomProvider",
+    "provider": "CustomMeter",
     "name": "myHistogram",
     "displayName": "myHistogram",
     "unit": null,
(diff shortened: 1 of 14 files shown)


documentation/api/livemetrics-custom.md
```diff
@@ -146,7 +146,7 @@ Location: localhost:52323/operations/67f07e40-5cca-4709-9062-26302c484f18
 
 {
     "timestamp": "2021-08-31T16:58:39.7514031+00:00",
-    "provider": "CustomProvider",
+    "provider": "CustomMeter",
```
kkeirstead, line 149: I didn't think of this before, but worth considering before we lock ourselves into a schema - do we want to use `meter` instead of `provider` if we know we're using an SDM?

jander-msft:
I need to update CollectLiveMetrics with this change to. Will add to this PR.

documentation/api/livemetrics-custom.md
wiktork: This is likely to lead to a lot more unnecessary special casing. While I want to split the configuration, I'm not sure I want to split the data model so much.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/LiveMetricsTests.cs
```diff
@@ -467,12 +318,12 @@ public async Task TestSystemDiagnosticsMetrics_MaxTimeseries()
...
-                            Providers = new[]
+                            Meters = new[]
                             {
-                                new EventMetricsProvider
+                                new EventMetricsMeter
                                 {
-                                    ProviderName = p1.ProviderName,
-                                    CounterNames = counterNames
+                                    MeterName = p1.ProviderName,
+                                    InstrumentNames = counterNames
```
kkeirstead, line 326: nit: `counterNames` -> `instrumentNames` and `providers` -> `meters` (applies to all tests)

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/LiveMetricsTests.cs
```diff
@@ -286,89 +220,6 @@ public async Task TestSystemDiagnosticsMetrics()
                 });
         }
 
-        [Theory]
-        [InlineData(MetricProviderType.All, true)]
-        [InlineData(MetricProviderType.Meter, true)]
-        [InlineData(MetricProviderType.EventCounter, false)]
-        public async Task TestSystemDiagnosticsMetrics_MetricProviderType(MetricProviderType metricType, bool expectResults)
```
wiktork, line 293: Are these tests subsumed by other tests?

wiktork (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/LiveMetricsTests.cs
kkeirstead: These were just to test the MeterType mechanism (that we get results for Meter/All but not EventCounter) - if we aren't exposing that as configuration anymore I'm assuming we don't need them anymore

documentation/configuration.md
```diff
@@ -1613,6 +1613,7 @@ An action that collects live metrics for the process that the collection rule is
 |---|---|---|---|---|---|---|
 | `IncludeDefaultProviders` | bool | false | Determines if the default counter providers should be used. | `true` | | |
 | `Providers` | [EventMetricsProvider](api/definitions.md#eventmetricsprovider)[] | false | The array of providers for metrics to collect. | | | |
```
kkeirstead, line 1615: nit: we changed this for EventMetricsConfiguration to, "Array of **counter** providers for metrics to collect" - just if we're aiming for consistency

wiktork (dismissed):

kkeirstead (dismissed):

wiktork (dismissed):

wiktork (approved):

PR #3872, opened 2023-03-07
Startup hook validation for enabling exceptions capture

jander-msft:
###### Summary

These changes check that if the exceptions features are enabled then the default process needs to be configured correctly (the `DOTNET_STARTUP_HOOKS` environment variable needs to be set appropriately). The startup hook validator will log warning messages as to why the startup configuration of the target process is not sufficient. At this time, startup hook environment variable is the only way to configure target processes load the exceptions logging infrastructure.

Changes also include the ability to query for the managed libraries from the shared libraries path.

### DIFF
--- a/src/Tools/dotnet-monitor/Exceptions/ExceptionsService.cs
+++ b/src/Tools/dotnet-monitor/Exceptions/ExceptionsService.cs
@@ -1,11 +1,14 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using Microsoft.Diagnostics.Monitoring;
 using Microsoft.Diagnostics.Monitoring.WebApi;
 using Microsoft.Diagnostics.Monitoring.WebApi.Exceptions;
 using Microsoft.Diagnostics.NETCore.Client;
+using Microsoft.Diagnostics.Tools.Monitor.StartupHook;
 using Microsoft.Extensions.Hosting;
 using System;
+using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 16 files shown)


src/Tools/dotnet-monitor/Exceptions/ExceptionsService.cs
```diff
@@ -48,6 +55,22 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
                 {
                     // Get default process
                     IProcessInfo pi = await _diagnosticServices.GetProcessAsync(processKey: null, stoppingToken);
+
+                    // If previous checked configuration and it did not pass, do not check again or attempt
+                    // to start the event pipe session.
+                    UniqueProcessKey key = new(pi.EndpointInfo.ProcessId, pi.EndpointInfo.RuntimeInstanceCookie);
+                    if (_unconfiguredProcesses.Contains(key))
+                    {
+                        throw new MonitoringException(string.Empty);
```
wiktork, line 64: Exception message?

src/Tools/dotnet-monitor/Exceptions/ExceptionsService.cs
```diff
@@ -48,6 +55,22 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+                    }
+
+                    // Validate that the process is configured correctly for collecting exceptions.
+                    if (!await _startupHookValidator.CheckAsync(pi.EndpointInfo, stoppingToken))
+                    {
+                        _unconfiguredProcesses.Add(key);
+                        throw new MonitoringException(string.Empty);
```
wiktork, line 71: I'm guessing there's no message here since you already log from CheckAsync?

src/Tools/dotnet-monitor/StartupHook/StartupHookValidator.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.StartupHook
+{
+    internal sealed class StartupHookValidator
+    {
+        private const string StartupHookFileName = "Microsoft.Diagnostics.Monitoring.StartupHook.dll";
+        private const string StartupHookTargetFramework = "net6.0";
```
wiktork, line 20: So will we always be using v6 for the startup hook, since that reflects the minimum target process this feature would support?

wiktork (dismissed):

src/Tools/dotnet-monitor/StartupHook/StartupHookValidator.cs
jander-msft: That is the intent; I don't want to have to ship multiple TFMs for the startup hook unless we really need it for some technical reason.

src/Tools/dotnet-monitor/Exceptions/ExceptionsService.cs
jander-msft: I'll switch this to a non-MonitoringException since it is not intended to be observable by the user and add a comment to that effect.

src/Tools/dotnet-monitor/Exceptions/ExceptionsService.cs
jander-msft: Correct, and the exception is not user visible.

wiktork (approved):

PR #3883, opened 2023-03-08
Only publish platform neutral outputs on win-x64 release leg

jander-msft:
###### Summary

Currently, each build leg is publishing platform neutral outputs and uploading them to the same build artifact and sub path. This can cause a race condition between two build legs that are publishing at the same time. This change fixes this behavior by only publishing platform neutral outputs on the win-x64 release build leg. Additionally, the publish outputs for this build leg are also uploaded to the Build_Unified_Release artifact so that nupkg creation can consume it without publishing it (currently, the nupkg creation leg is invoking publish, but this will be fixed in a different change).

Example build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2130271&view=results

### DIFF
--- a/eng/pipelines/jobs/build-test.yml
+++ b/eng/pipelines/jobs/build-test.yml
@@ -31,6 +31,10 @@ jobs:
     architecture: ${{ parameters.architecture }}
 
     variables:
+    - _PublishAfterBuildArgs: '/p:PublishProjectsAfterBuild=true /p:SkipPlatformNeutralPublish=true'
+    - ${{ if and(eq(parameters.targetRid, 'win-x64'), eq(parameters.configuration, 'Release')) }}:
+      - _PublishAfterBuildArgs: '/p:PublishProjectsAfterBuild=true'
+
     - ${{ if and(ne(parameters.testGroup, 'None'), ne(parameters.architecture, 'arm64')) }}:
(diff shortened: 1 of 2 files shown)


eng/pipelines/jobs/build-test.yml
```diff
@@ -31,6 +31,10 @@ jobs:
     architecture: ${{ parameters.architecture }}
 
     variables:
+    - _PublishAfterBuildArgs: '/p:PublishProjectsAfterBuild=true /p:SkipPlatformNeutralPublish=true'
+    - ${{ if and(eq(parameters.targetRid, 'win-x64'), eq(parameters.configuration, 'Release')) }}:
+      - _PublishAfterBuildArgs: '/p:PublishProjectsAfterBuild=true'
```
kkeirstead, line 36: Just making sure I'm understanding this correctly, is this implicitly equivalent to `- _PublishAfterBuildArgs: '/p:PublishProjectsAfterBuild=true /p:SkipPlatformNeutralPublish=false'`

kkeirstead (approved):

eng/pipelines/jobs/build-test.yml
jander-msft: Sure, you can think of it that way. The `SkipPlatformNeutralPublish` usage is checking if it's exactly set to `true`, so any other value would cause it to not skip.

PR #3884, opened 2023-03-08
Allow skipping publishing of projects in archive and package legs

jander-msft:
###### Summary

Allow the archive and package legs to explicitly opt out of publishing projects and relying on the build legs to provide the published binaries. This change requires #3883 as seen from the example build where it fails to package due to missing published binaries.

Example build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2130279&view=results

### DIFF
--- a/src/Tools/dotnet-monitor/dotnet-monitor.csproj
+++ b/src/Tools/dotnet-monitor/dotnet-monitor.csproj
@@ -65,7 +65,7 @@
 
   <!-- Publish projects if they were not published after build -->
   <Target Name="PublishProjectsBeforePack"
-          Condition="'$(PublishProjectsAfterBuild)' != 'true'">
+          Condition="'$(SkipPublishProjects)' != 'true'">
     <CallTarget Targets="PublishProjects" />
   </Target>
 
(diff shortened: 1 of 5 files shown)


src/Tools/dotnet-monitor/dotnet-monitor.csproj
```diff
@@ -65,7 +65,7 @@
 
   <!-- Publish projects if they were not published after build -->
```
kkeirstead, line 66: Did you also want to update this comment (similar to the one in PublishedProjectPackage.targets), or is that not appropriate here?

kkeirstead (approved):

src/Tools/dotnet-monitor/dotnet-monitor.csproj
jander-msft, line 66:
```suggestion
  <!-- Publish projects unless skipped -->
```

jander-msft: Done

kkeirstead (approved):

PR #3906, opened 2023-03-08
[Egress] Support using shared framework host to execute extensions

jander-msft:
###### Summary

Extensions can specify `"UseSharedDotNetHost": true` in their manifest in order to instruct dotnet-monitor to execute the extension using the shared framework host (e.g. `dotnet.exe`) instead of the platform-specific apphost in the extension. This allows a platform-neutral extension to be used so that extensions do not have to have separate apphost executables for each supported platform.

### DIFF
--- a/src/Tools/dotnet-monitor/DotNetHost.Unix.cs
+++ b/src/Tools/dotnet-monitor/DotNetHost.Unix.cs
@@ -1,0 +1,58 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    partial class DotNetHost
+    {
+        private sealed class UnixDotNetHostHelper : IDotNetHostHelper
+        {
+            private const string InstallationFilePath = "/etc/dotnet/install_location";
+
+            private static readonly string CurrentArchInstallationFilePath =
+                FormattableString.Invariant($"{InstallationFilePath}_{RuntimeInformation.ProcessArchitecture.ToString("G").ToLowerInvariant()}");
+
+            public string ExecutableName => "dotnet";
+
+            public bool TryGetDefaultInstallationDirectory(out string dotnetRoot)
+            {
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+                {
+                    // CONSIDER: Account for emulation (emulated path would be under "./x64/")
+                    dotnetRoot = "/usr/local/share/dotnet";
+                }
+                else
+                {
+                    dotnetRoot = "/usr/share/dotnet";
+                }
+                return true;
+            }
+
+            public bool TryGetSelfRegisteredDirectory(out string dotnetRoot)
+            {
+                return TryReadFileFirstLine(CurrentArchInstallationFilePath, out dotnetRoot) ||
+                    TryReadFileFirstLine(InstallationFilePath, out dotnetRoot);
+            }
+
+            private static bool TryReadFileFirstLine(string filePath, out string content)
+            {
+                try
+                {
+                    using FileStream stream = new(filePath, FileMode.Open, FileAccess.Read, FileShare.Read);
+                    using StreamReader reader = new StreamReader(stream);
+                    content = reader.ReadLine();
+                    return !string.IsNullOrEmpty(content);
+                }
+                catch
+                {
+                    content = null;
+                    return false;
+                }
+            }
+        }
+    }
+}
--- a/src/Tools/dotnet-monitor/DotNetHost.Unix.cs
+++ b/src/Tools/dotnet-monitor/DotNetHost.Unix.cs
@@ -11,25 +11,31 @@ namespace Microsoft.Diagnostics.Tools.Monitor
     {
         private sealed class UnixDotNetHostHelper : IDotNetHostHelper
         {
-            private const string InstallationFilePath = "/etc/dotnet/install_location";
+            private const string DefaultPathLinux = $"/usr/share/{InstallationDirectoryName}";
+            private const string DefaultPathOSX = $"/usr/local/share/{InstallationDirectoryName}";
+            private const string InstallationFilePath = $"/etc/{InstallationDirectoryName}/install_location";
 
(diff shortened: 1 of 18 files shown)


src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
```diff
@@ -75,20 +76,42 @@ public async Task<EgressArtifactResult> EgressArtifact(ExtensionEgressPayload co
...
 
-            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            IFileInfo executableInfo;
+            if (Declaration.UseSharedDotNetHost)
             {
-                exeName += ".exe";
-            }
+                // .NET Monitor currently only runs with the shared framework, thus its ProcessPath
+                // will be the shared dotnet host (dotnet.exe).
+                executableInfo = new PhysicalFileInfo(new FileInfo(Environment.ProcessPath));
```
wiktork, line 93: For images, won't this result int the /app/dotnet-monitor path?

jander-msft: You're probably right. I'll have to think about how to differentiate between the apphost and the dotnet host from managed code.

jander-msft: The dotnet root resolution code is at https://github.com/dotnet/runtime/blob/7392fb7c4617ee831d160f0c61aaf43038a6aba8/src/native/corehost/fxr_resolver.cpp#L55

Order of looking this up is:
1. Check if hostfxr library is in the app directory, if so, assume self-contained.
1. Check for DOTNET_ROOT(_<arch>) environment variable
1. On Windows, check for installation in registry; on *nix, check `/etc/dotnet` for path file
1. Check default install location

We can probably get away with not doing this ourselves but instead doing:
1. If current process is named `dotnet`, just use that one.
1. Get location of currently loaded hostfxr library; if in app directory, then assume self-contained
1. If DOTNET_ROOT(_<arch>) is set, use that path.
1. Get dotnet relative to the hostfxr library path.

.NET Monitor doesn't ship self-contained at this time, so we don't need to handle that scenario just yet.

jander-msft: I mostly adopted the apphost's lookup algorithm (the first list).

jander-msft: I was able to test this on Windows and Ubuntu on WSL2. The Windows code path fell back to the default installation whereas the Linux code path used the self-registered location.

src/Tools/dotnet-monitor/DotNetHost.Unix.cs
```diff
@@ -0,0 +1,58 @@
...
+        private sealed class UnixDotNetHostHelper : IDotNetHostHelper
+        {
+            private const string InstallationFilePath = "/etc/dotnet/install_location";
+
+            private static readonly string CurrentArchInstallationFilePath =
+                FormattableString.Invariant($"{InstallationFilePath}_{RuntimeInformation.ProcessArchitecture.ToString("G").ToLowerInvariant()}");
+
+            public string ExecutableName => "dotnet";
+
+            public bool TryGetDefaultInstallationDirectory(out string dotnetRoot)
```
kkeirstead, line 21: Is there a reason this is a `Try...` if it only returns true?

src/Tools/dotnet-monitor/DotNetHost.Windows.cs
```diff
@@ -0,0 +1,29 @@
...
+    partial class DotNetHost
+    {
+        private sealed class WindowsDotNetHostHelper : IDotNetHostHelper
+        {
+            public string ExecutableName => "dotnet.exe";
+
+            public bool TryGetDefaultInstallationDirectory(out string dotnetRoot)
+            {
+                // CONSIDER: Account for emulation (emulated path would be under ".\x64\")
+                dotnetRoot = System.IO.Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles), "dotnet");
```
wiktork, line 17: will this work for aks + windows containers?

src/Tools/dotnet-monitor/DotNetHost.Windows.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using System;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    partial class DotNetHost
+    {
+        private sealed class WindowsDotNetHostHelper : IDotNetHostHelper
+        {
+            public string ExecutableName => "dotnet.exe";
```
wiktork, line 12: nit: deduplicate "dotnet"

src/Tools/dotnet-monitor/DotNetHost.Windows.cs
jander-msft: I'll check but I believe this should just work.

jander-msft: This should work for Windows containers. See the Windows dockerfile from dotnet/dotnet-docker where it is installed to `%ProgramFiles%\dotnet`: https://github.com/dotnet/dotnet-docker/blob/main/src/runtime/6.0/windowsservercore-ltsc2022/amd64/Dockerfile#L26

src/Tools/dotnet-monitor/DotNetHost.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Runtime.InteropServices;
+using System.Threading;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    internal static partial class DotNetHost
+    {
+        private const string DotNetName = "dotnet";
+        public const string ExecutableRootName = DotNetName;
+        public const string InstallationDirectoryName = DotNetName;
```
kkeirstead, line 15: Is there a reason why we have three constants with the same value, or is it just for readability?

kkeirstead (approved):

src/Tools/dotnet-monitor/DotNetHost.cs
jander-msft: I wanted to keep separate the installation folder name vs the executable name in case we need to change that and downstream consumers of the constants can be more deliberate about why they are using the constant that they chose.

PR #3921, opened 2023-03-10
Upload Logs Artifacts to S3 Egress

### LINKED ISSUE #3644: S3 Egress Provider Produces Empty Logs

kkeirstead:
###### Summary

This PR fixes our inability to use S3 egress for logs artifacts. Logging uses the `Write` method (intead of `WriteAsync`), which was previously not implemented - this change implements the `Write` method, temporarily pushing bytes into a list that are then consumed using the existing `WriteAsync` infrastructure.

Some exploration was done around using [Pipelines](https://learn.microsoft.com/en-us/dotnet/standard/io/pipelines); the code is currently not leveraging Pipelines, but this change could be made in the future.

Closes #3644


src/Tools/dotnet-monitor/Egress/S3/MultiPartUploadStream.cs
```diff
@@ -120,7 +167,18 @@ private async Task DoWriteAsync(bool allowPartialWrite, CancellationToken cancel
 
     public override void Write(byte[] buffer, int offset, int count)
     {
-        throw new NotSupportedException();
+        if (Closed)
```
jander-msft, line 170: Please move this implementation to an override of the `Write(ReadOnlySpan<byte>)` method and have this Write method call that one: `Write(buffer.AsMemory().Slice(offset, count));`

src/Tools/dotnet-monitor/Egress/S3/MultiPartUploadStream.cs
```diff
@@ -47,13 +59,43 @@ public override async Task FlushAsync(CancellationToken cancellationToken)
...
         if (Closed)
             throw new ObjectDisposedException(nameof(MultiPartUploadStream));
         if (_offset == 0)
             return;
         await DoWriteAsync(true, cancellationToken);
     }
 
+    public async Task WriteAsyncLoop(CancellationToken cancellationToken)
+    {
+        while (!_finalize || _syncTempBuffer.Count > 0)
```
jander-msft, line 76: If something doesn't immediately call Write on the stream after creating it, this loop is likely to exit very quickly because nothing is in the buffer. I don't think this loop should depend on data being in the buffer.

src/Tools/dotnet-monitor/Egress/S3/MultiPartUploadStream.cs
```diff
@@ -75,6 +117,11 @@ public override void SetLength(long value)
     }
 
     public override async ValueTask WriteAsync(ReadOnlyMemory<byte> buffer, CancellationToken cancellationToken)
+    {
+        await WriteAsync(buffer, false, cancellationToken);
+    }
+
+    public async ValueTask WriteAsync(ReadOnlyMemory<byte> buffer, bool allowPartialWrite, CancellationToken cancellationToken)
```
jander-msft, line 124: This `allowPartialWrite` parameter seems unnecessary; that's the purpose of the `FinalizeAsync` method.

src/Tools/dotnet-monitor/Egress/S3/MultiPartUploadStream.cs
```diff
@@ -47,13 +59,43 @@ public override async Task FlushAsync(CancellationToken cancellationToken)
...
+            if (Closed)
+                throw new ObjectDisposedException(nameof(MultiPartUploadStream));
+
+            await _semaphore.WaitAsync(cancellationToken);
+            try
+            {
+                if (_syncTempBuffer.Count > 0)
+                {
+                    await WriteAsync(_syncTempBuffer.ToArray(), _finalize, cancellationToken);
+                    _syncTempBuffer = new();
```
jander-msft, line 87: Clear the buffer instead of creating a new one to avoid extra allocations.

src/Tools/dotnet-monitor/Egress/S3/MultiPartUploadStream.cs
```diff
@@ -47,13 +59,43 @@ public override async Task FlushAsync(CancellationToken cancellationToken)
...
+            await _writeSynchronousArtifacts;
+        
         if (Closed)
             throw new ObjectDisposedException(nameof(MultiPartUploadStream));
         if (_offset == 0)
             return;
         await DoWriteAsync(true, cancellationToken);
     }
 
+    public async Task WriteAsyncLoop(CancellationToken cancellationToken)
```
jander-msft, line 74: This method appears to constantly loop (with the delay) until data is put into the buffer.

What it should do is wait on a semaphore that represents when data is available, await that, lock on a sync object to read the data out of the buffer into a temporary buffer, release the lock, then upload the temporary buffer data via the WriteAsync method.

The Write method should acquire the lock on the sync object, write the data to the buffer, release the lock, and release the data available semaphore.

By doing this, the WriteAsyncLoop method avoids spinning constantly and only does work when data is available.

jander-msft (changes requested):

src/Tools/dotnet-monitor/Egress/S3/MultiPartUploadStream.cs
kkeirstead: I had that initially but changed it to avoid situations where the `_syncTempBuffer` expands to a large size due to a brief spike in logs and then remains unnecessarily large after that point. I can change this back to clearing though if you'd prefer.

jander-msft: You can clear it and then set Capacity to some reasonable value so we avoid unnecessary allocations of the `List<byte>` itself and unnecessary reallocation of the array within it each time `AddRange` is called.

kkeirstead:
Closing without resolving since this should not be needed once the egress extensibility model is integrated.

PR #3952, opened 2023-03-14
Update SDK to final 8.0 Preview 2

jander-msft:

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/HttpApi/ApiClient.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/HttpApi/ApiClient.cs
@@ -871,6 +871,7 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests.HttpApi
         {
             JsonSerializerOptions options = CreateJsonDeserializeOptions();
 #if NET8_0_OR_GREATER
+            // Workaround for https://github.com/dotnet/aspnetcore/issues/47223
             options.PropertyNameCaseInsensitive = true;
 #endif
             return options;
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/HttpApi/ApiClient.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/HttpApi/ApiClient.cs
@@ -15,7 +15,6 @@ using System.Globalization;
 using System.IO;
 using System.Net;
 using System.Net.Http;
-using System.Reflection;
 using System.Text;
 using System.Text.Json;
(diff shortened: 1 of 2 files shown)


wiktork (dismissed):

jander-msft:
Functional tests are not able to validate HTTP 400 responses appropriately due to https://github.com/dotnet/aspnetcore/issues/47223

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/HttpApi/ApiClient.cs
```diff
@@ -743,8 +744,40 @@ private static async Task<Exception> CreateUnexpectedStatusCodeExceptionAsync(Ht
 
         private static async Task<ValidationProblemDetailsException> CreateValidationProblemDetailsExceptionAsync(HttpResponseMessage responseMessage)
         {
+            ValidationProblemDetails details = await ReadContentAsync<ValidationProblemDetails>(responseMessage).ConfigureAwait(false);
+
+#if NET8_0_OR_GREATER
+            // Workaround for https://github.com/dotnet/aspnetcore/issues/47223
+            if (null != details.Extensions)
```
wiktork, line 751: Do you think this might be a change in how they handle case sensitivity?

jander-msft: I did more testing around this. It does appear to be that the casing changed between .NET 7 (lowercase) and .NET 8 Preview 2 (titlecase). If I serialize and deserialize on the same .NET runtime, it works as expected.

jander-msft: Our .NET 8 version of .NET Monitor that runs on the 8.0.0-preview.3.23164.4 runtime and 8.0.0-preview.3.23164.1 aspnetcore in tests is returning `{"status":400,"detail":"HTTP egress is not enabled."}`. This is a result of the automatic serialization of effectively `new BadRequestObjectResult(new ValidationProblemDetails(...))`, so ASP.NET is causing it to serialize using lowercase as it had always done in the past. But the change in Preview 2 causes deserialization to expect title case.

This means that any .NET 8 app that deserializes ValidationProblemDetails from any prior .NET version (and it appears from ASP.NET 8 itself) will fail to deserialize it properly.

I think we'll need this workaround in place until either ASP.NET fixes it or gives us guidance on how to properly handle it.

wiktork: As a workaround, can you use DefaultJsonDeserializeOptions to ignore case as a workaround?

PR #3974, opened 2023-03-16
Update trace-until-event to use shared code from `dotnet/diagnostics`

schmittjoseph:
###### Summary

Now that the shared version of `EventMonitoringPassthroughStream` is available in public `dotnet/diagnostics` builds, switch over to using it. **NOTE:** The shared version splits `EventMonitoringPassthroughStream` into two separate classes (`EventMonitor` and `PassthroughStream`).


jander-msft (approved):

jander-msft:
Could you target release/7.x and then we'll merge back to main?

PR #3991, opened 2023-03-17
[Egress Extensibility] Disable Tools Folder Extension Discovery

kkeirstead:
###### Summary

@jander-msft I know we had discussed not supporting this scenario for now and leaving the infrastructure so we could easily re-integrate it in the future...from a maintainability standpoint, do you want it left in place (but unexercised), or should I rip it out here so we don't have stale unused code? This probably depends on whether we seriously plan to reintroduce this support in the future - thoughts?


jander-msft:
Let's completely remove it; we can go back in source code history and pull it back out if necessary.

PR #4031, opened 2023-03-23
Allow using extensions from build output

jander-msft:
###### Summary

Allow building and using extensions directly from debug build output without having to stage the egress extensions.

This is done by adding another ExtensionRepository implementation to the test hosting startup that understands how to locate extensions from the build output directory.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.TestHostingStartup/BuildOutput.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.TestHostingStartup/BuildOutput.cs
@@ -1,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.IO;
+using System.Reflection;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.TestHostingStartup
+{
+    internal static class BuildOutput
+    {
+        public const string ConfigurationName =
+#if DEBUG
+            "Debug";
+#else
+            "Release";
+#endif
+
+        // This is the binary output directory when built from the dotnet-monitor repo: <repoRoot>/artifacts/bin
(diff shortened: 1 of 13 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.TestHostingStartup/BuildOutput.cs
```diff
@@ -0,0 +1,22 @@
...
+        public const string ConfigurationName =
+#if DEBUG
+            "Debug";
+#else
+            "Release";
+#endif
+
+        // This is the binary output directory when built from the dotnet-monitor repo: <repoRoot>/artifacts/bin
+        public static readonly string RootPath =
+            Path.GetFullPath(Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "..", "..", ".."));
```
wiktork, line 20: We have a similar path resolution in shared library code. Did you want to unify this to one? Not a concern now, but Assembly.GetExecutingAssembly().Location will also need to be changed when using single file/trimmed executables.

wiktork: Disregard the first part, I made the comment before I finished reading the rest of the PR :-)

wiktork (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.TestHostingStartup/BuildOutput.cs
jander-msft: This is hosting startup that is loaded into the dotnet-monitor process. If dotnet-monitor is trimmed, it is very likely that startup hook code path will be trimmed, which means that the hosting startup assembly won't be loaded either. So I'm not too terribly concerned about this case. But when the time comes, the trimming analyzer will definitely catch this.

PR #4036, opened 2023-03-23
Simplified extension manifest

jander-msft:
###### Summary

Simplifies the extension manifest to only have three properties:
- Name: The name of the extension e.g. AzureBlobStorage or S3Storage
- ExecutableFileName: The name of the executable entry point if to be launched using its own executable.
- AssemblyFileName: The name of the entry point assembly if to be launched using the shared .NET host (e.g. dotnet.exe).

Everything else that was removed either was unused or would lead to future confusion if we add other extensibility points.

### DIFF
--- a/src/Extensions/AzureBlobStorage/extension.json
+++ b/src/Extensions/AzureBlobStorage/extension.json
@@ -1,10 +1,4 @@
 {
-    "Id": "Microsoft.Diagnostics.Monitoring.AzureStorage",
-    "Version": "1.0.0",
-    "Program": "dotnet-monitor-egress-azureblobstorage",
-    "UseSharedDotNetHost": true,
     "Name": "AzureBlobStorage",
-    "SupportedExtensionTypes": [
-        "Egress"
-    ]
+    "AssemblyFileName": "dotnet-monitor-egress-azureblobstorage"
 }
--- a/src/Tools/dotnet-monitor/Extensibility/ExtensionDeclaration.cs
+++ b/src/Tools/dotnet-monitor/Extensibility/ExtensionDeclaration.cs
@@ -30,8 +30,8 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
 
         public void Validate()
         {
-            bool hasAssemblyFileName = string.IsNullOrEmpty(AssemblyFileName);
-            bool hasExecutableFileName = string.IsNullOrEmpty(ExecutableFileName);
+            bool hasAssemblyFileName = !string.IsNullOrEmpty(AssemblyFileName);
+            bool hasExecutableFileName = !string.IsNullOrEmpty(ExecutableFileName);
 
             if (hasAssemblyFileName && hasExecutableFileName)
             {
--- a/src/Tools/dotnet-monitor/Extensibility/ExtensionDeclaration.cs
+++ b/src/Tools/dotnet-monitor/Extensibility/ExtensionDeclaration.cs
@@ -1,42 +1,57 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
 
-using System;
+using System.Globalization;
 
 namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
 {
     internal class ExtensionDeclaration
     {
         /// <summary>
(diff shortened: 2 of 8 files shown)


src/Tools/dotnet-monitor/Extensibility/ExtensionDeclaration.cs
```diff
@@ -1,42 +1,57 @@
...
+        /// </remarks>
+        public string AssemblyFileName { get; set; }
 
-        /// <summary>
-        /// This is the name that users specify in configuration to refer to the extension.
-        /// </summary>
-        public string Name { get; set; }
+        public void Validate()
+        {
+            bool hasAssemblyFileName = string.IsNullOrEmpty(AssemblyFileName);
```
kkeirstead, line 33: Should these be inverted?

jander-msft: That's true, however the validation logic still works 😅

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
```diff
@@ -203,7 +205,7 @@ private void ValidateExecutable(IFileInfo entrypointInfo)
         {
             if (!entrypointInfo.Exists || entrypointInfo.IsDirectory || entrypointInfo.PhysicalPath == null)
             {
-                _logger.ExtensionProgramMissing(_extensionName, Path.Combine(_targetFolder, _exePath), Declaration.Program);
+                _logger.ExtensionProgramMissing(_extensionName, Path.Combine(_targetFolder, _exePath), Declaration.ExecutableFileName);
```
kkeirstead, line 208: In theory, this isn't specified when we validate the AssemblyFileName, correct? Do we just want to pass the file name as a parameter so we don't have to do the if/else check again to determine which one to use?

src/Tools/dotnet-monitor/Extensibility/ExtensionNotFoundException.cs
```diff
@@ -17,21 +15,20 @@ private ExtensionException(string message)
...
         {
-            throw new ExtensionException(string.Format(CultureInfo.CurrentCulture, Strings.ErrorMessage_ExtensionLaunchFailed, extensionName));
+            throw new ExtensionException(string.Format(CultureInfo.CurrentCulture, Strings.ErrorMessage_ExtensionManifestInvalid, reason));
         }
 
         [DoesNotReturn]
-        public static ExtensionException ThrowWrongType(string extensionName, string extensionPath, Type requiredType)
+        public static void ThrowLaunchFailure(string extensionName)
         {
-            throw new ExtensionException(string.Format(CultureInfo.CurrentCulture, Strings.LogFormatString_ExtensionNotOfType, extensionName, extensionPath, requiredType.Name));
```
kkeirstead, line 34: Assuming this string isn't used anywhere else, can we remove it from the resx?

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.cs
jander-msft: You're absolutely right... but if you are okay with it, I'd rather defer fixing this to a separate change to cleanup the entire ProgramExtension class. It's kind of a mess right now and several of the fields don't represent what their names suggest. I'd like to take a separate pass over it after this PR is merged in.

src/Extensions/AzureBlobStorage/extension.json
```diff
@@ -1,10 +1,4 @@
 {
-    "Id": "Microsoft.Diagnostics.Monitoring.AzureStorage",
```
wiktork, line 2: Did we want to leave version? For example, if there are multiple releases of a private egress extension, it might be worthwhile to know the version for troubleshooting purposes.

We could also add tags (similar to nuget package tags, but perhaps kv pairs) for arbitrary data that could be added to the extension manifest.

wiktork (approved):

src/Extensions/AzureBlobStorage/extension.json
jander-msft: I see the version as being fairly superfluous at the moment. Nothing uses it and the extension version is the same exact version as dotnet-monitor for the time being. We can add it back if we use it for something (logging, extension enumeration for debugging purposes through some configuration validation or a dedicated `dotnet-monitor extensions list` command, etc). I'll add this to the backlog though.

PR #4045, opened 2023-03-24
Refactor ExtensionManifest creation out of ProgramExtension class.

jander-msft:
###### Summary

This change centralizes the parsing and validation of the extension manifest in the `ExtensionManifest` class so that it can be separately constructed outside of the `ProgramExtension` implementation. Also added unit tests to verify parsing and validation logic.


src/Tools/dotnet-monitor/Extensibility/ExtensionManifest.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            try
+            {
+                return JsonSerializer.Deserialize<ExtensionManifest>(stream);
+            }
+            catch (JsonException ex)
+            {
+                ExtensionException.ThrowInvalidManifest(ex);
+
+                throw;
```
wiktork, line 56: Remove second throw?

src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
```diff
@@ -28,12 +29,23 @@ public override bool TryFindExtension(string extensionName, out IExtension exten
...
+                    {
+                        manifest = ExtensionManifest.FromPath(manifestPath);
+                    }
+                    catch (Exception ex) when (LogManifestParseError(ex, manifestPath))
                     {
-                        extension = currExtension;
+                        continue;
+                    }
+
+                    if (extensionName == manifest.Name)
```
wiktork, line 46: Do we have any case-sensitivity or restrictions on extension names that should be reflected here?

wiktork (approved):

src/Tools/dotnet-monitor/Extensibility/ExtensionManifest.cs
jander-msft: Control flow analysis doesn't understand that this will never be reached, so we have to "throw" here to make the compiler happy.

src/Tools/dotnet-monitor/Extensibility/FolderExtensionRepository.cs
jander-msft: I don't think that that was explicitly decided, but I think it's correct to enforce case sensitivity here. This name is not used in the file system anymore (e.g. you can name the extension's folder `xyz` but the manifest has `abc`; the `abc` is what will be expected in configuration and the `xyz` folder name has no bearing on the extension).

PR #4061, opened 2023-03-27
[Egress Extensibility] Removed `generatePreSignedUrl` Property

kkeirstead:
###### Summary

The `generatePreSignedUrl` property seems redundant - users can just set/omit `preSignedUrlExpiry` when they want a pre-signed url (this helps reduce configuration errors and users needing to remember to set both properties). The pre-signed url also comes back with escaped characters (specifically, the &) that mean you can't click to follow the link provided in the egress payload; this PR now unescapes messages from the extension before printing them.

### DIFF
--- a/src/Tools/dotnet-monitor/Extensibility/ProgramExtension.OutputParser.cs
+++ b/src/Tools/dotnet-monitor/Extensibility/ProgramExtension.OutputParser.cs
@@ -6,7 +6,6 @@ using Microsoft.Extensions.Logging;
 using System;
 using System.Diagnostics;
 using System.Text.Json;
-using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -81,7 +80,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
                     {
                         // Expected that some things won't parse correctly
                     }
-                    _logger.ExtensionOutputMessage(_processId, Regex.Unescape(eventArgs.Data));
+                    _logger.ExtensionOutputMessage(_processId, eventArgs.Data);
                 }
             }
 
--- a/src/Tools/dotnet-monitor/Extensibility/ProgramExtension.OutputParser.cs
+++ b/src/Tools/dotnet-monitor/Extensibility/ProgramExtension.OutputParser.cs
@@ -6,6 +6,7 @@ using Microsoft.Extensions.Logging;
 using System;
 using System.Diagnostics;
 using System.Text.Json;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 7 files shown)


src/Tools/dotnet-monitor/Extensibility/ProgramExtension.OutputParser.cs
```diff
@@ -80,7 +81,7 @@ private void ParseStdOut(object sender, DataReceivedEventArgs eventArgs)
                     {
                         // Expected that some things won't parse correctly
                     }
-                    _logger.ExtensionOutputMessage(_processId, eventArgs.Data);
+                    _logger.ExtensionOutputMessage(_processId, Regex.Unescape(eventArgs.Data));
```
jander-msft, line 84: The individual extensions should unescape their results, not dotnet-monitor.

jander-msft (changes requested):

src/Tools/dotnet-monitor/Extensibility/ProgramExtension.OutputParser.cs
kkeirstead: Handling this on the extension side, my understanding is I need to do this as an option for the JSON Serialization - the latest update appears to work by doing this, but I'm not sure if it's considered good practice. Let me know if you had something else in mind.

jander-msft: I think this is still not correct. The JSON serialization and deserialization shouldn't have to know about this issue; the data that is fed to the serializer needs to be unescaped before being serialized. So if something needs to be unescaped, it should be done at https://github.com/dotnet/dotnet-monitor/blob/feature/egressExtension/src/Extensions/AzureBlobStorage/AzureBlobEgressProvider.cs#L82 and https://github.com/dotnet/dotnet-monitor/blob/feature/egressExtension/src/Extensions/S3Storage/S3StorageEgressProvider.cs#L67

kkeirstead: Here's my understanding of what's going on (prior to my changes):

result.ArtifactPath (on the extension side) contains an & symbol.
JsonSerializer.Serialize converts that to \u0026
ParseStdOut prints eventArgs.Data for our EgressArtifactResult, which hasn't been deserialized (and thus contains \u0026)
EgressOperation receives the deserialized artifact path, and prints 'Egressed artifact to: ...", with the correct &

Basically, the issue is that we print the raw eventArgs.Data, not the deserialized result. What I did gets around that by overriding the default serialization behavior, so JsonSerializer.Serialize doesn't convert & to \u0026, but alternatively the logic in ParseStdOut on the DM side could just print the deserialized result instead. Does that make sense?

jander-msft: When I take what's in the egress extension branch, modify the Azure blob extension to return "something&otherthing", the `eventArgs.Data` property value is `{"Succeeded":true,"FailureMessage":null,"ArtifactPath":"something\u0026otherthing"}`; after deserialization, the `result.ArtifactPath` property value is `something&otherthing`. This is completely what I would expect to happen.

If the resource ID that is returned by the S3 libraries is returning something that is escaped, then you need to unescape it before returning it from the `EgressAsync` method. It should not be the responsibility of the transport or the serialization mechanism between the process and dotnet-monitor to handle that.

jander-msft (approved):

PR #4082, opened 2023-03-28
Convert FileSystem egress to built-in egress extension

jander-msft:
###### Summary

Refactor the `FileSystem` egress provider to be an `IEgressExtension` implementation. This will allow the egress architecture to be greatly simplified (in a future PR) since all egress providers would be implemented as either built-in and external egress extensions.


src/Microsoft.Diagnostics.Monitoring.Options/EgressOptions.cs
```diff
@@ -11,13 +10,11 @@ namespace Microsoft.Diagnostics.Tools.Monitor
     internal sealed class EgressOptions
     {
         [Display(
-            ResourceType = typeof(OptionsDisplayStrings),
-            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_EgressOptions_FileSystem))]
+            Description = "Mapping of file system egress provider names to their options.")]
```
kkeirstead, line 13: Was there a reason why you in-lined these? Is it just to be consistent with the other extensions not reaching into the main OptionsDisplayStrings resx?

src/Tools/dotnet-monitor/Egress/FileSystem/FileSystemEgressExtension.cs
```diff
@@ -163,5 +168,33 @@ private static string WrapMessage(string innerMessage)
...
+        {
+            if (configuration.TryGetValue(nameof(FileSystemEgressProviderOptions.CopyBufferSize), out string copyBufferSizeString) &&
+                !string.IsNullOrEmpty(copyBufferSizeString))
+            {
+                try
+                {
+                    options.CopyBufferSize = (int)TypeDescriptor.GetConverter(typeof(int)).ConvertFromInvariantString(copyBufferSizeString);
+                }
+                catch
+                {
```
kkeirstead, line 185: Do we want this to fail to bind silently?

src/Tools/dotnet-monitor/Egress/Configuration/OptionsTypeToProviderTypesMapper.cs
```diff
@@ -53,14 +44,13 @@ public IEnumerable<IConfigurationSection> GetProviderSections(Type optionsType)
...
-        /// Determines if the given section is a built in type property/field of <see cref="EgressOptions"/>.
+        /// Determines if the given section is a built in type property/field of the Egress configuration section.
         /// </summary>
         /// <param name="s">The <see cref="IConfigurationSection"/> to evaluate, this should be one of the children of the "Egress" section.</param>
         /// <returns><see langword="true"/> if and only if the given section is one of of the pre-defined</returns>
         private static bool IsBuiltInSection(IConfigurationSection s)
         {
-            Type egressType = typeof(EgressOptions);
-            return egressType.GetProperty(s.Key) != null;
+            return ConfigurationKeys.Egress_Properties == s.Key;
```
kkeirstead, line 53: If I'm reading this correctly, we're just checking if the key is "Properties" since all of the providers are now out of the box - if that's the case, perhaps either the commenting/naming of this method could be updated to make that more transparent.

src/Tools/dotnet-monitor/Egress/Extension/WellKnownEgressExtensionRepository.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Egress.Extension
+{
+    internal sealed class WellKnownEgressExtensionRepository :
+        ExtensionRepository
+    {
+        private readonly ILogger<FileSystemEgressExtension> _logger;
+
+        public WellKnownEgressExtensionRepository(ILogger<FileSystemEgressExtension> logger)
```
kkeirstead, line 15: Just confirming my understanding - a "Well Known" egress extension is just one that's built into dotnet-monitor - which in this case is just `FileSystem`? The only thing that feels a bit odd here is that the logger is specifically for `FileSystemEgressExtension`, even though the method is presumably meant to be generic, though for now that doesn't really matter since we're only looking for `FileSystem`.

src/Microsoft.Diagnostics.Monitoring.Options/EgressOptions.cs
jander-msft: I thought that I'd move EgressOptions out of the Microsoft.Diagnostics.Monitoring.Options assembly and since it no longer is used by product code, it doesn't need localized descriptions any more. I'll take another look at actually moving the class if possible.

src/Tools/dotnet-monitor/Egress/FileSystem/FileSystemEgressExtension.cs
jander-msft: Yes, configuration binding fails silently for property binding. Note the comment above the method about this being a temporary stop-gap solution until more refactoring is done in future PRs.

src/Tools/dotnet-monitor/Egress/Extension/WellKnownEgressExtensionRepository.cs
jander-msft: Yes, it's the only statically known egress extension because it's "in the box". Removing the `ILogger` injection would mean that we have some other implementation class that gets injected that would create the `FileSystemEgressExtension` instance, which is basically the same problem.

I'll play around with is some more to see if I can remove the dependency on the concrete types.

src/Microsoft.Diagnostics.Monitoring.Options/EgressOptions.cs
jander-msft: I'll revert the display string removal and leave moving the class out to another PR.

kkeirstead (approved):

PR #4086, opened 2023-03-28
[Egress] Clean-Up For CopyBufferSize

kkeirstead:
###### Summary

Cleaned up our usage of `CopyBufferSize` for Azure & S3. I'll leave comments around certain lines of code down below that need further context.

### DIFF
--- a/src/Extensions/AzureBlobStorage/AzureBlobEgressProvider.cs
+++ b/src/Extensions/AzureBlobStorage/AzureBlobEgressProvider.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Diagnostics.Monitoring.AzureBlobStorage
     /// </remarks>
     internal partial class AzureBlobEgressProvider : EgressProvider<AzureBlobEgressProviderOptions>
     {
-        private int BlobStorageBufferSize = 4 * 1024 * 1024;
+        private int DefaultBlobStorageBufferSize = 4 * 1024 * 1024;
 
         public AzureBlobEgressProvider(ILogger logger) : base(logger)
         {
@@ -49,10 +49,10 @@ namespace Microsoft.Diagnostics.Monitoring.AzureBlobStorage
 
                 var bloboptions = new BlockBlobOpenWriteOptions
                 {
-                    BufferSize = BlobStorageBufferSize,
+                    BufferSize = options.CopyBufferSize.GetValueOrDefault(DefaultBlobStorageBufferSize),
                 };
                 using (Stream blobStream = await blobClient.OpenWriteAsync(overwrite: true, options: bloboptions, cancellationToken: token))
-                using (AutoFlushStream flushStream = new AutoFlushStream(blobStream, BlobStorageBufferSize))
+                using (AutoFlushStream flushStream = new AutoFlushStream(blobStream, bloboptions.BufferSize.Value))
                 {
                     //Azure's stream from OpenWriteAsync will do the following
                     //1. Write the data to a local buffer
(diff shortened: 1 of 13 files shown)


src/Extensions/AzureBlobStorage/AzureBlobEgressProviderOptions.cs
```diff
@@ -55,6 +55,7 @@ internal sealed partial class AzureBlobEgressProviderOptions
         [Display(
             ResourceType = typeof(OptionsDisplayStrings),
             Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_CommonEgressProviderOptions_CopyBufferSize))]
+        [Range(1, int.MaxValue)]
```
kkeirstead, line 58: I added range validation for this to keep it aligned with S3 - but it also means we fail the egress operation if this isn't set correctly (as opposed to just disregarding it). I would be fine with removing the Range check and just internally enforcing that CopyBufferSize > 0 if people prefer.

src/Tools/dotnet-monitor/Egress/EgressProvider.cs
```diff
@@ -44,47 +44,6 @@ protected EgressProvider(ILogger logger)
...
-            TOptions options,
-            Func<CancellationToken, Task<Stream>> action,
-            EgressArtifactSettings artifactSettings,
-            CancellationToken token)
-        {
-            Func<Stream, CancellationToken, Task> wrappingAction = async (targetStream, token) =>
-            {
-                using var sourceStream = await action(token);
-
-                int copyBufferSize = options.CopyBufferSize.GetValueOrDefault(0x100000);
```
kkeirstead, line 69: I noticed that `CopyBufferSize` was being used here, but then realized that all the deleted code appears to be an unused code path. If there's a reason to leave this in I can revert it, but I believe it's just stale code.

src/Extensions/AzureBlobStorage/AzureBlobEgressProvider.cs
```diff
@@ -49,10 +49,10 @@ public AzureBlobEgressProvider(ILogger logger) : base(logger)
 
                 var bloboptions = new BlockBlobOpenWriteOptions
                 {
-                    BufferSize = BlobStorageBufferSize,
+                    BufferSize = options.CopyBufferSize.GetValueOrDefault(DefaultBlobStorageBufferSize),
```
kkeirstead, line 52: Assuming that `CopyBufferSize` is > 0, we'll use it here...while we don't have the concept of a mandatory 5 Mb minimum like on S3, would it make sense to have a reasonable minimum here? I don't expect that it'll matter much since most people won't be setting this, but I don't know if it's worth protecting people from using really low values (maybe if they mistakenly assume the units are in Mb?)

src/Extensions/S3Storage/S3StorageEgressProvider.cs
```diff
@@ -43,8 +43,7 @@ public S3StorageEgressProvider(ILogger logger) : base(logger)
             {
                 client = await ClientFactory.CreateAsync(options, artifactSettings, token);
                 uploadId = await client.InitMultiPartUploadAsync(artifactSettings.Metadata, token);
-                int copyBufferSize = options.CopyBufferSize.GetValueOrDefault(0x100000);
```
kkeirstead, line 46: I believe the `0x100000` value was originally a copy-paste from our existing usage in `EgressProvider.cs` - it wasn't actually getting used, because in `MultiPartUploadStream` that would be overridden by the mimimum value anyway.

jander-msft (approved):
LGTM! Good job on the removal of the EgressAsync overload too!

src/Extensions/AzureBlobStorage/AzureBlobEgressProviderOptions.cs
wiktork: I think as long as we have a reasonable default here, we should leave the range check

PR #4087, opened 2023-03-28
Move egress configuration parsing into EgressExtension

jander-msft:
###### Summary

This is a "separation of concerns" fixup PR.

Move the parsing of the egress provider configuration section from `ExtensionEgressProvider` into `EgressExtension`. This allows the `FileSystemEgressExtension` to not manually look up its options from a dictionary but instead using options binding. Move the creation o the `EgressExtension` into a factory so that other callers do not have to provide the myriad of services it requires.

### DIFF
--- a/src/Tools/dotnet-monitor/Egress/Extension/EgressExtension.cs
+++ b/src/Tools/dotnet-monitor/Egress/Extension/EgressExtension.cs
@@ -1,12 +1,16 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using Microsoft.Diagnostics.Tools.Monitor.Egress.Configuration;
 using Microsoft.Diagnostics.Tools.Monitor.Extensibility;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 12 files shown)


src/Tools/dotnet-monitor/Egress/Extension/EgressExtension.cs
```diff
@@ -158,5 +191,34 @@ private void ValidateFileExists(FileInfo fileInfo)
                 ExtensionException.ThrowFileNotFound(_manifest.Name, fileInfo.FullName);
             }
         }
+
+        private Dictionary<string, string> GetConfigurationSection(string providerName, string providerType)
+        {
+            IConfigurationSection providerTypeSection = _configurationProvider.GetProviderTypeConfigurationSection(providerType);
+            IConfigurationSection providerNameSection = providerTypeSection.GetSection(providerName);
```
kkeirstead, line 198: I'm assuming this was just a 1:1 copy-paste of the existing implementation (with the refactored method name for `GetProvider**Type**ConfigurationSection`), but isn't this effectively recreating what `GetProviderConfigurationSection` does?

jander-msft: Will fix this in next iteration

kkeirstead (approved):

PR #4121, opened 2023-03-31
Validate configured egress providers at startup

jander-msft:
###### Summary

- Moved enumeration and caching of egress providers into separate class named `EgressProviderSource`.
- Add startup logger that initializes the `EgressProviderSource`, which causes it to log any missing egress providers.

### DIFF
--- a/src/Tools/dotnet-monitor/Egress/EgressProviderSource.cs
+++ b/src/Tools/dotnet-monitor/Egress/EgressProviderSource.cs
@@ -1,0 +1,97 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Diagnostics.Tools.Monitor.Egress.Configuration;
+using Microsoft.Diagnostics.Tools.Monitor.Extensibility;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Primitives;
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Threading;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Egress
+{
+    internal sealed class EgressProviderSource : IDisposable
+    {
+        private readonly IEgressConfigurationProvider _configurationProvider;
+        private readonly ExtensionDiscoverer _extensionDiscoverer;
+        private readonly ILogger _logger;
+        private readonly IDictionary<string, string> _providerNameToTypeMap;
+
+        private IDisposable _changeRegistration;
+        private long _initialized;
+
+        public EgressProviderSource(
+            IEgressConfigurationProvider configurationProvider,
+            ExtensionDiscoverer extensionDiscoverer,
+            ILogger<EgressProviderSource> logger)
+        {
+            _configurationProvider = configurationProvider;
+            _extensionDiscoverer = extensionDiscoverer;
+            _logger = logger;
+            _providerNameToTypeMap = new ConcurrentDictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+        }
+
+        public void Dispose()
+        {
+            _changeRegistration.Dispose();
+        }
+
+        public void Initialize()
+        {
+            if (0 != Interlocked.CompareExchange(ref _initialized, 1, 0))
+                return;
+
+            _changeRegistration = ChangeToken.OnChange(
+                _configurationProvider.GetReloadToken,
+                Reload);
+
+            Reload();
+        }
+
+        public IEgressExtension GetEgressProvider(string name)
+        {
+            if (!_providerNameToTypeMap.TryGetValue(name, out string providerType))
+            {
+                throw new EgressException(string.Format(CultureInfo.CurrentCulture, Strings.ErrorMessage_EgressProviderDoesNotExist, name));
+            }
+
+            return _extensionDiscoverer.FindExtension<IEgressExtension>(providerType);
+        }
+
+        private void Reload()
+        {
+            _providerNameToTypeMap.Clear();
+
+            foreach (string providerType in _configurationProvider.ProviderTypes)
+            {
+                try
+                {
+                    IEgressExtension _ = _extensionDiscoverer.FindExtension<IEgressExtension>(providerType);
+                }
+                catch (ExtensionException)
+                {
+                    _logger.EgressProviderTypeNotExist(providerType);
+                }
+
+                IConfigurationSection typeSection = _configurationProvider.GetProviderTypeConfigurationSection(providerType);
+
+                foreach (IConfigurationSection optionsSection in typeSection.GetChildren())
+                {
+                    string providerName = optionsSection.Key;
+                    if (_providerNameToTypeMap.TryGetValue(providerName, out string existingProviderType))
+                    {
+                        _logger.DuplicateEgressProviderIgnored(providerName, providerType, existingProviderType);
+                    }
+                    else
+                    {
+                        _providerNameToTypeMap.Add(providerName, providerType);
+                    }
+                }
+            }
+        }
+    }
+}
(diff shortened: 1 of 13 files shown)


kkeirstead:
I pulled this locally to play around with it - these improvements can come as part of a later PR if you'd rather not include them here, but I think we should be effectively exercising the extension end-to-end.

Right now, there are a few ways that I'll pass the initial validation check that I'd like to cause failures. For a well-configured extension that we supply, these might not be common occurrences, but for 3rd party ones these could be helpful:

1. If I have a folder in my `Extensions` directory with the name of my type (e.g. AzureBlobStorage), we only check that the `extension.json` file is there - we don't validate that there's an executable/assembly name, and we also don't verify that there's an executable/assembly actually provided in the folder. This causes a failure when we try to use the extension, because there's nothing to execute.
2. If an empty manifest is provided (`extension.json` just contains the basic `{}` for being a JSON file), we will fail immediately, but it's generic. It'll be more helpful to the user if we can tell them the required fields that are missing.

Can we "test-launch" each extension at start-up - basically go through the entire flow, right before we would actually egress an artifact? We already have warnings when an extension is launched that address some of these issues, but users won't hit them until they actually need the extension (instead of at startup).

src/Tools/dotnet-monitor/Egress/EgressProviderSource.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            foreach (string providerType in _configurationProvider.ProviderTypes)
+            {
+                try
+                {
+                    IEgressExtension _ = _extensionDiscoverer.FindExtension<IEgressExtension>(providerType);
+                }
+                catch (ExtensionException)
+                {
+                    _logger.EgressProviderTypeNotExist(providerType);
```
kkeirstead, line 83: I effectively get two of the same error message right now:
`08:00:43 fail: Microsoft.Diagnostics.Tools.Monitor.Extensibility.ExtensionDiscoverer[86]
      Failed to locate extension 'AzureBlobStorage_dotnet_tools'.
08:00:43 warn: Microsoft.Diagnostics.Tools.Monitor.Egress.EgressProviderSource[96]
      Egress provider type 'AzureBlobStorage_dotnet_tools' does not exist.`

I suppose there isn't any harm in this - it's just redundant for these to both print out without providing any unique information.

jander-msft:
> I pulled this locally to play around with it - these improvements can come as part of a later PR if you'd rather not include them here, but I think we should be effectively exercising the extension end-to-end.
> 
> Right now, there are a few ways that I'll pass the initial validation check that I'd like to cause failures. For a well-configured extension that we supply, these might not be common occurrences, but for 3rd party ones these could be helpful:
> 
> 1. If I have a folder in my `Extensions` directory with the name of my type (e.g. AzureBlobStorage), we only check that the `extension.json` file is there - we don't validate that there's an executable/assembly name, and we also don't verify that there's an executable/assembly actually provided in the folder. This causes a failure when we try to use the extension, because there's nothing to execute.
> 2. If an empty manifest is provided (`extension.json` just contains the basic `{}` for being a JSON file), we will fail immediately, but it's generic. It'll be more helpful to the user if we can tell them the required fields that are missing.
> 
> Can we "test-launch" each extension at start-up - basically go through the entire flow, right before we would actually egress an artifact? We already have warnings when an extension is launched that address some of these issues, but users won't hit them until they actually need the extension (instead of at startup).

Both of those concerns are about extension authoring not about extension usage. I would argue that an extension author should do their due diligence to test that their extension actually is launchable and egresses before those extensions are published for others to use. The consumer of the extension has no ability to fix any of the potential issues that you've highlighted; their only responsibility is to install the extension (which the code verifies it will find an extension that matches) and to configure it correctly (which cannot be verified until execution time).

src/Tools/dotnet-monitor/Egress/EgressProviderSource.cs
jander-msft: Actually, I'd like to defer this to a different PR if actually necessary. I think the logging for extension discovery should be simplified or replaced with exceptions but don't want to complicate this PR further.

kkeirstead:
> > I pulled this locally to play around with it - these improvements can come as part of a later PR if you'd rather not include them here, but I think we should be effectively exercising the extension end-to-end.
> > Right now, there are a few ways that I'll pass the initial validation check that I'd like to cause failures. For a well-configured extension that we supply, these might not be common occurrences, but for 3rd party ones these could be helpful:
> > 
> > 1. If I have a folder in my `Extensions` directory with the name of my type (e.g. AzureBlobStorage), we only check that the `extension.json` file is there - we don't validate that there's an executable/assembly name, and we also don't verify that there's an executable/assembly actually provided in the folder. This causes a failure when we try to use the extension, because there's nothing to execute.
> > 2. If an empty manifest is provided (`extension.json` just contains the basic `{}` for being a JSON file), we will fail immediately, but it's generic. It'll be more helpful to the user if we can tell them the required fields that are missing.
> > 
> > Can we "test-launch" each extension at start-up - basically go through the entire flow, right before we would actually egress an artifact? We already have warnings when an extension is launched that address some of these issues, but users won't hit them until they actually need the extension (instead of at startup).
> 
> Both of those concerns are about extension authoring not about extension usage. I would argue that an extension author should do their due diligence to test that their extension actually is launchable and egresses before those extensions are published for others to use. The consumer of the extension has no ability to fix any of the potential issues that you've highlighted; their only responsibility is to install the extension (which the code verifies it will find an extension that matches) and to configure it correctly (which cannot be verified until execution time).

You bring up a good point about the fact that consumers don't really have the ability to fix any issues. I agree that authors _should_ be doing their own testing, but I think of this as more of a safeguard for users - if they choose to use 3rd-party extensions that aren't well-maintained/tested and something breaks with an updated version, dotnet-monitor can't fix the issue - but we can proactively make them aware of it. It helps prevent users from missing important artifacts at collection-time due to an issue - if users know about an issue when they launch dotnet-monitor, they can go back to a prior version or choose not to use that extension.

All of that being said - this really comes down to quality of life and ease-of-use, not core functionality. If you don't want to incorporate this, I won't block on it.

kkeirstead (approved):
Approving, but left a comment in discussion about what level of validation I'd like to see (if not in this PR)

jander-msft:
> > > I pulled this locally to play around with it - these improvements can come as part of a later PR if you'd rather not include them here, but I think we should be effectively exercising the extension end-to-end.
> > > Right now, there are a few ways that I'll pass the initial validation check that I'd like to cause failures. For a well-configured extension that we supply, these might not be common occurrences, but for 3rd party ones these could be helpful:
> > > 
> > > 1. If I have a folder in my `Extensions` directory with the name of my type (e.g. AzureBlobStorage), we only check that the `extension.json` file is there - we don't validate that there's an executable/assembly name, and we also don't verify that there's an executable/assembly actually provided in the folder. This causes a failure when we try to use the extension, because there's nothing to execute.
> > > 2. If an empty manifest is provided (`extension.json` just contains the basic `{}` for being a JSON file), we will fail immediately, but it's generic. It'll be more helpful to the user if we can tell them the required fields that are missing.
> > > 
> > > Can we "test-launch" each extension at start-up - basically go through the entire flow, right before we would actually egress an artifact? We already have warnings when an extension is launched that address some of these issues, but users won't hit them until they actually need the extension (instead of at startup).
> > 
> > 
> > Both of those concerns are about extension authoring not about extension usage. I would argue that an extension author should do their due diligence to test that their extension actually is launchable and egresses before those extensions are published for others to use. The consumer of the extension has no ability to fix any of the potential issues that you've highlighted; their only responsibility is to install the extension (which the code verifies it will find an extension that matches) and to configure it correctly (which cannot be verified until execution time).
> 
> You bring up a good point about the fact that consumers don't really have the ability to fix any issues. I agree that authors _should_ be doing their own testing, but I think of this as more of a safeguard for users - if they choose to use 3rd-party extensions that aren't well-maintained/tested and something breaks with an updated version, dotnet-monitor can't fix the issue - but we can proactively make them aware of it. It helps prevent users from missing important artifacts at collection-time due to an issue - if users know about an issue when they launch dotnet-monitor, they can go back to a prior version or choose not to use that extension.
> 
> All of that being said - this really comes down to quality of life and ease-of-use, not core functionality. If you don't want to incorporate this, I won't block on it.

Even if we did add validation upfront about the details of the extension that the extension author should have done themselves, in practice, I don't think it's really going to help. My gut says that folks only look at the logs (honestly, if they look at logs at all) if a failure in operation occurs. If we are only logging warnings because of a bad extension, they are not going to notice until they go to use it.

PR #4133, opened 2023-04-03
Add egress extensibility, refactor built-in egress providers, and extension packaging

jander-msft:
###### Summary

Refactor the AzureBlobStorage and S3Storage egress providers into separate executables as egress extensions using an extensibility surface that allows others to write their own egress providers that can be configured for use with dotnet-monitor. The .NET Tool package remains functionally unchanged (it includes both egress extensions) whereas the product archive no longer contains the egress extension binaries; consumers will have to layer the egress extension archives on top of the product archive.

###### Release Notes Entry
Refactor AzureBlobStorage and S3Storage egress into extensions


src/Tools/dotnet-monitor/DotNetHost.Unix.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    partial class DotNetHost
+    {
+        private sealed class UnixDotNetHostHelper : IDotNetHostHelper
+        {
+            private const string DefaultPathLinux = $"/usr/share/{InstallationDirectoryName}";
```
wiktork, line 14: Is this unniversal across all the dotnet supported distros?

jander-msft: Yes, it's the final fallback; see https://github.com/dotnet/runtime/blob/main/src/native/corehost/hostmisc/pal.unix.cpp#L596

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -0,0 +1,130 @@
...
+
+            return egressCmd;
+        }
+
+        private static async Task<int> Egress<TOptions>(EgressProvider<TOptions> provider, CancellationToken token, Action<ExtensionEgressPayload, TOptions, ILogger> configureOptions = null) where TOptions : class, new()
+        {
+            EgressArtifactResult result = new();
+            try
+            {
+                string jsonConfig = Console.ReadLine();
```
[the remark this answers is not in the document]
jander-msft: I think there's a chance that the `StreamReader` instance for `Console.In` overreads given that `StreamReader` fills its internal buffer with 4k of data. However, we haven't seen an issue in practice with it.

I'll take a look at refactoring this to see if we can read exactly the amount that we need or preserve the reader for use by the egress implementation. I'd like to do this after this PR is merged as a follow up item.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressException.cs
```diff
@@ -0,0 +1,17 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace Microsoft.Diagnostics.Monitoring.Extension.Common
+{
+    /// <summary>
+    /// Exception that egress providers can throw when an operational error occurs (e.g. failed to write the stream data).
+    /// </summary>
+    internal sealed class EgressException : MonitoringException
```
wiktork, line 11: Do we want to have a separate exception hierarchy here? It seems unnecessary to share a base exception type since we can't catch out of proc exceptions anyway.

wiktork (approved):

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressException.cs
jander-msft: I think that it would still be good for us to have a single exception type that can be thrown that indicates that we know something is wrong (e.g. validation of the options) as opposed to some logical error (e.g. NRE). But this can definitely be flattened to one type.

PR #4140, opened 2023-04-03
Derive release notes file name from only `eng/Versions.props`

schmittjoseph:
Re-opening this PR as it's applicable again. With the recent problems calculating build numbers for internal builds, I'm in favor of simplifying our release notes file name to not depend on the build version. This would fix the above scenario, as well as allowing this action to be ran before the final build is completed. 

---

###### Summary

As mentioned in https://github.com/dotnet/dotnet-monitor/pull/4139, nupkgs are no longer published to blob storage so the aka.ms links that release notes generation relied on are no longer updated.

Instead of using the aka.ms links to generate a file name, use only the information from `eng/Versions.props`.  As a result the `Test` option is no longer needed either.

**NOTE**: Before this PR is merged in, the release pipeline will need to be updated to also use the new file names.

Example PRs (with intentionally blank changelogs):
- Preview release: https://github.com/schmittjoseph/dotnet-monitor/pull/302
- RTM/Servicing release: https://github.com/schmittjoseph/dotnet-monitor/pull/304


jander-msft:
FYI, this script looks up the release version in the release pipeline: https://github.com/dotnet/dotnet-monitor/blob/main/eng/release/Scripts/GetReleaseVersion.ps1

.github/workflows/generate-release-notes.yml
```diff
@@ -39,27 +34,14 @@ jobs:
           versionFile="./eng/Versions.props"
           release_version=$(perl -ne '/<VersionPrefix>([^<]*)/ && print $1' $versionFile)
           release_version_label=$(perl -ne '/<PreReleaseVersionLabel>([^<]*)/ && print $1' $versionFile)
-          major_minor_version=${release_version%.*}
-
-          if [[ "$TEST_RUN" == "true" ]]; then
-            qualified_release_version="test-$release_version"
-          else
-            version_url="https://aka.ms/dotnet/diagnostics/monitor${major_minor_version}/release/dotnet-monitor.nupkg.version"
```
jander-msft, line 47: Can we use `aka.ms/dotnet/diagnostics/monitor${major_minor_version}/release/dotnet-monitor-win-x64.zip.version` rather than changing how all of this works?

jander-msft:
> Re-opening this PR as it's applicable again. With the recent problems calculating build numbers for internal builds, I'm in favor of simplifying our release notes file name to not depend on the build version. This would fix the above scenario, as well as allowing this action to be ran before the final build is completed.

This does not fix the ability to generate release notes for internal release builds. Internal releases have commits that are not publicly available at the time of release. Release note generation will not know of the commit nor have access to the version file (since it's in an authenticated blob storage account).

The only way I see the current generation scheme to work for internal builds is to actually generate the release notes on the most-nearly corresponding public release branch. I suppose this is why you want to generate based on the version.props file rather than the actual version. This would generate the same content, except for any release-note-worthy fixes that were made internally.

jander-msft (approved):

PR #4176, opened 2023-04-07
Validate FileSystem egress options

jander-msft:
###### Summary

FileSystem egress options validation was accidentally removed in #4082. This change adds it back to the extension implementation and adds tests to check that the validation exists and is correct.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/FileSystemEgressExtensionTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/FileSystemEgressExtensionTests.cs
@@ -1,0 +1,135 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Diagnostics.Monitoring.TestCommon;
+using Microsoft.Diagnostics.Tools.Monitor;
+using Microsoft.Diagnostics.Tools.Monitor.Egress;
+using Microsoft.Diagnostics.Tools.Monitor.Egress.Configuration;
+using Microsoft.Diagnostics.Tools.Monitor.Egress.FileSystem;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Moq;
+using System.Collections.Generic;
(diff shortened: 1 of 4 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/FileSystemEgressExtensionTests.cs
```diff
@@ -0,0 +1,135 @@
...
+
+            // Act & Assert
+            OptionsValidationException exception = await Assert.ThrowsAsync<OptionsValidationException>(
+                () => extension.EgressArtifact(ProviderName, null, (stream, token) => Task.CompletedTask, CancellationToken.None));
+            string errorMessage = Assert.Single(exception.Failures);
+            Assert.Equal(CreateRequiredMessage(nameof(FileSystemEgressProviderOptions.DirectoryPath)), errorMessage);
+        }
+
+        [Fact]
+        public async Task FileSystemEgressExtension_CopyBufferSizeZero_ThrowsException()
```
kkeirstead, line 59: nit: Is it worth having a test where the `CopyBufferSize` **is** valid, like for `DirectoryPath`?

kkeirstead (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/FileSystemEgressExtensionTests.cs
```diff
@@ -0,0 +1,203 @@
...
+
+        [Fact]
+        public async Task FileSystemEgressExtension_EmptyConfiguration_ThrowsException()
+        {
+            // Arrange
+            IEgressExtension extension = CreateExtension();
+
+            // Act & Assert
+            OptionsValidationException exception = await Assert.ThrowsAsync<OptionsValidationException>(
+                () => extension.EgressArtifact(ProviderName, CreateSettings(), WriteFileContent, CancellationToken.None));
```
kkeirstead, line 51: Just double checking that we want to be using `CancellationToken.None` for all of these.

jander-msft: In general, I would say it's necessary to pass a real token, but there is no code path that realistically require an async completion.

PR #4179, opened 2023-04-07
[Egress] Update StdIn Protocol

kkeirstead:
###### Summary

The existing StdIn protocol relied on a newline to separate the `ExtensionEgressPayload` from the artifact in the stream; though this worked in most cases, this resulted in empty artifacts being egressed in some edge cases (the belief is that this was a result of Console.ReadLine's buffer). This PR moves away from separating the payload from the artifact via newline, instead specifying the size of the payload at the beginning of the StdIn stream, then reading the exact number of bytes required for the payload (before copying the artifact to the outputStream).


src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -35,8 +35,15 @@ private static async Task<int> Egress<TOptions>(EgressProvider<TOptions> provide
...
             try
             {
-                string jsonConfig = Console.ReadLine();
-                ExtensionEgressPayload configPayload = JsonSerializer.Deserialize<ExtensionEgressPayload>(jsonConfig);
+                StdInStream = Console.OpenStandardInput();
+
+                byte[] payloadLengthBuffer = new byte[sizeof(long)]; // Stream's Position is a long
+                _ = StdInStream.Read(payloadLengthBuffer);
+
+                byte[] payloadBuffer = new byte[BitConverter.ToInt64(payloadLengthBuffer)];
```
wiktork, line 43: Consider an upperbound or sanity check here on the value you receive.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -35,8 +35,15 @@ private static async Task<int> Egress<TOptions>(EgressProvider<TOptions> provide
             EgressArtifactResult result = new();
             try
             {
-                string jsonConfig = Console.ReadLine();
-                ExtensionEgressPayload configPayload = JsonSerializer.Deserialize<ExtensionEgressPayload>(jsonConfig);
+                StdInStream = Console.OpenStandardInput();
+
+                byte[] payloadLengthBuffer = new byte[sizeof(long)]; // Stream's Position is a long
```
wiktork, line 40: We should talk about whether we want to add a more elaborate header here.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -35,8 +35,15 @@ private static async Task<int> Egress<TOptions>(EgressProvider<TOptions> provide
...
+                _ = StdInStream.Read(payloadLengthBuffer);
```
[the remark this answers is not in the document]
jander-msft: See https://github.com/dotnet/runtime/blob/f9ebd42d05ca88f086d472832487fbccfc7fe4ac/src/libraries/System.Private.CoreLib/src/System/IO/Stream.cs#L428 for how to write the read loop (doens't have to be exactly this, but gives you the idea on how to check whether more needs to be read).

For .NET 7 and above, you can use [Stream.ReadExactly](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.readexactly?view=net-8.0) instead of writing your own read loop.

src/Tools/dotnet-monitor/Egress/Extension/EgressExtension.cs
```diff
@@ -148,8 +145,13 @@ public void AddEnvironmentVariable(string key, string value)
 
             parser.BeginReading();
 
-            await JsonSerializer.SerializeAsync<ExtensionEgressPayload>(p.StandardInput.BaseStream, payload, options: null, token);
-            await p.StandardInput.WriteAsync(NewLine, token);
+            Stream intermediateStream = new MemoryStream();
+            await JsonSerializer.SerializeAsync(intermediateStream, payload, options: null, token);
```
wiktork, line 149: What is the default pretty printing for serialization? We can save size by making sure the json is as compact as possible.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
jander-msft: Probably at most a version field and then the payload length. The payload can be differentiated by version if necessary.

src/Tools/dotnet-monitor/Egress/Extension/EgressExtension.cs
kkeirstead: It appears that it's compacted by default (there's no spacing) - I believe there's an option to make it indented, but by default it isn't.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
kkeirstead: Was there other info you wanted to include in it? The primary purpose of the header is a workaround for the buffer issue with `Console.ReadLine` as opposed to passing data to the extension, but I'm curious what you had in mind.

jander-msft: Let's write out:
- Version (4-byte integer); its value is fixed to 1
- Payload length in bytes (8-byte integer)
- Payload as bytes

The version will indicate the format of the remainder of the data stream. Egress extensions should validate the version is what is expected and otherwise fail (throw exception).

We likely won't change the remainder of the stream in any significant manner after 8.0 is released, but this allows us to have that ability in the future if necessary.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -35,8 +37,27 @@ private static async Task<int> Egress<TOptions>(EgressProvider<TOptions> provide
...
+                if (payloadLengthBufferSize != await StdInStream.ReadAsync(payloadLengthBuffer))
+                {
+                    throw new EgressException(GenericEgressFailureMessage);
+                }
+
+                long payloadBufferSize = BitConverter.ToInt64(payloadLengthBuffer);
+
+                byte[] payloadBuffer = new byte[payloadBufferSize];
+
+                if (payloadBufferSize != await StdInStream.ReadAsync(payloadBuffer))
```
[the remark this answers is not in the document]
kkeirstead: Yep, GitHub is acting up today for some reason...I didn't see any of his comments until _after_ I put up the first set of changes unfortunately. This will be in the next version.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -113,9 +128,39 @@ private static async Task GetStream(Stream outputStream, CancellationToken cance
...
+
+        private static async Task<byte[]> ReadFromStream(long bufferSize, CancellationToken token)
+        {
+            Memory<byte> buffer = new(new byte[bufferSize]);
+            if (bufferSize != await ReadHelper(buffer, bufferSize, token))
+            {
+                throw new EgressException(GenericEgressFailureMessage);
+            }
+
+            return buffer.ToArray();
```
kkeirstead, line 162: My understanding is that this makes a copy...is there a more efficient way I should be doing this, or for the amount of data we're working with is this not a concern?

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -20,6 +21,9 @@ internal sealed class EgressHelper
     {
         private static Stream StdInStream;
         private static CancellationTokenSource CancelSource = new CancellationTokenSource();
+        private const int PayloadVersion = 1;
+
+        private const string GenericEgressFailureMessage = "The egress operation failed due to an internal error.";
```
kkeirstead, line 26: We currently don't have a Resx, and I wasn't sure what we want to expose to users for what's effectively an internal protocol error - I put this here temporarily, but we can do something else before this gets checked in.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -35,8 +39,19 @@ private static async Task<int> Egress<TOptions>(EgressProvider<TOptions> provide
             EgressArtifactResult result = new();
             try
             {
-                string jsonConfig = Console.ReadLine();
-                ExtensionEgressPayload configPayload = JsonSerializer.Deserialize<ExtensionEgressPayload>(jsonConfig);
+                StdInStream = Console.OpenStandardInput();
+
+                var payloadVersionBuffer = await ReadFromStream(sizeof(int), token); // Payload Version
```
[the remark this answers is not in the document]
kkeirstead: I'll let @jander-msft weigh in on this, but a few thoughts:

1. For the sake of not making the manifest overly verbose, we had discussed keeping it more minimal to start with (and expanding it as necessary over time). For fields that weren't initially part of the manifest, from a backwards-compatibility standpoint we would just assume the omission of the field indicates a default (in this case, no version would be interpreted as v1).
2. This might be covered by the startup validation feature I'm working on right now (at least, for extension authors who include the ability to do startup validation). For our extension, when dotnet-monitor starts up we would send a test payload to the extension, and if the version numbers weren't the same (or any other issue) the user would be notified of the failure.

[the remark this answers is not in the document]
jander-msft: I agree with Kyle. The lack of a specified version in the manifest can be interpreted as version 1. Keeping it optional for version 1 protocol with keep things simple in the manifest for now.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -113,9 +128,39 @@ private static async Task GetStream(Stream outputStream, CancellationToken cance
         {
             const int DefaultBufferSize = 0x10000;
 
-            StdInStream = Console.OpenStandardInput();
             await StdInStream.CopyToAsync(outputStream, DefaultBufferSize, cancellationToken);
         }
+
+        // TODO: For .NET 7 and above, can use Stream.ReadExactly
+        private static async Task<int> ReadHelper(Memory<byte> buffer, long bufferSize, CancellationToken token)
```
jander-msft, line 135: Let's try this:
- Make this method `Task ReadExactlyAsync(Memory<byte> buffer, CancellationToken)`; read exactly the number of bytes from the stream as there is room for in the buffer. Throw `EndOfStreamExecption` if cannot fill the buffer. Use `#if NET7_0_OR_GREATER` to defer to the `Stream.ReadExactlyAsync` method and fallback to your implementation for .NET 6 and lower.
- Remove the `ReadFromStream` method and have the former callers (1) `stackalloc` their `byte[]` and (2) pass it into `ReadExactlyAsync`.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -35,8 +37,24 @@ private static async Task<int> Egress<TOptions>(EgressProvider<TOptions> provide
             EgressArtifactResult result = new();
             try
             {
-                string jsonConfig = Console.ReadLine();
-                ExtensionEgressPayload configPayload = JsonSerializer.Deserialize<ExtensionEgressPayload>(jsonConfig);
+                StdInStream = Console.OpenStandardInput();
+
+                var payloadVersionBuffer = stackalloc byte[sizeof(int)].ToArray();
```
jander-msft, line 42:
```suggestion
                Span<byte> payloadVersionSpan = stackalloc byte[sizeof(int)];
```

Using `ToArray` defeats the purpose of using `stackalloc`.

kkeirstead: That's what I tried initially, but I got "Parameters or locals of type 'Span<byte>' cannot be declared in async methods or async lambda expressions."

jander-msft: Totally forgot this is in an async method; arrays allocated with `stackalloc` cannot be boxed since they are purely on the stack and the async state machine hoists its locals to properties. Just allocate the arrays without using `stackalloc`.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -35,8 +37,24 @@ private static async Task<int> Egress<TOptions>(EgressProvider<TOptions> provide
...
+                var payloadVersionBuffer = new byte[sizeof(int)];
+                await ReadExactlyAsync(payloadVersionBuffer, token);
+
+                int dotnetMonitorPayloadProtocolVersion = BitConverter.ToInt32(payloadVersionBuffer);
+                if (dotnetMonitorPayloadProtocolVersion != ExpectedPayloadProtocolVersion)
+                {
+                    throw new EgressException(string.Format(CultureInfo.CurrentCulture, Strings.ErrorMessage_IncorrectPayloadVersion, dotnetMonitorPayloadProtocolVersion, ExpectedPayloadProtocolVersion));
+                }
+
+                var payloadLengthBuffer = new byte[sizeof(long)].ToArray();
```
jander-msft, line 51:
```suggestion
                var payloadLengthBuffer = new byte[sizeof(long)];
```

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -35,8 +37,24 @@ private static async Task<int> Egress<TOptions>(EgressProvider<TOptions> provide
             EgressArtifactResult result = new();
             try
             {
-                string jsonConfig = Console.ReadLine();
-                ExtensionEgressPayload configPayload = JsonSerializer.Deserialize<ExtensionEgressPayload>(jsonConfig);
+                StdInStream = Console.OpenStandardInput();
+
+                var payloadVersionBuffer = new byte[sizeof(int)];
+                await ReadExactlyAsync(payloadVersionBuffer, token);
```
jander-msft, line 43: Try taking a look at using [BinaryReader](https://learn.microsoft.com/en-us/dotnet/api/system.io.binaryreader?view=net-8.0).
- It seems to only read exactly what is asked for from the stream.
- It establishes that values are written in little endian form (whereas right now we leave that up to the runtime).
- We avoid having to allocate buffers for each piece of data.

This will require you to use [BinaryWriter](https://learn.microsoft.com/en-us/dotnet/api/system.io.binarywriter?view=net-8.0) on the dotnet-monitor side.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -35,8 +38,28 @@ private static async Task<int> Egress<TOptions>(EgressProvider<TOptions> provide
...
+                {
+                    dotnetMonitorPayloadProtocolVersion = reader.ReadInt32();
+                    payloadLengthBuffer = reader.ReadInt64();
+
+                    payloadBuffer = new byte[payloadLengthBuffer];
+
+                    reader.Read(payloadBuffer, 0, (int)payloadLengthBuffer); // this should be safe, given we know the payload is small
+                }
+
+                if (dotnetMonitorPayloadProtocolVersion != ExpectedPayloadProtocolVersion)
```
jander-msft, line 57: This check needs to be moved to immediately after reading the protocol version. There's no point in attempting to read anything else until it verifies the expected protocol.

src/Tools/dotnet-monitor/Egress/Extension/EgressExtension.cs
```diff
@@ -148,9 +147,24 @@ public void AddEnvironmentVariable(string key, string value)
...
-            await JsonSerializer.SerializeAsync<ExtensionEgressPayload>(p.StandardInput.BaseStream, payload, options: null, token);
-            await p.StandardInput.WriteAsync(NewLine, token);
-            await p.StandardInput.BaseStream.FlushAsync(token);
+            // p.StandardInput.BaseStream Format: Version (int), Payload Length (long), Payload, Artifact
+            using Stream intermediateStream = new MemoryStream();
+            await JsonSerializer.SerializeAsync(intermediateStream, payload, options: null, token);
+
+            using (var writer = new BinaryWriter(p.StandardInput.BaseStream, Encoding.UTF8, true))
+            {
+                writer.Write(BitConverter.GetBytes(PayloadProtocolVersion));
```
jander-msft, line 156:
```suggestion
                writer.Write(PayloadProtocolVersion);
```

Since you are not using `BitConverter` on the extension side, then don't use it here. It is fact unnecessary since `BinaryWriter` will write the bytes in a specific endianness and correctly decode it with `BinaryReader`.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -35,8 +38,28 @@ private static async Task<int> Egress<TOptions>(EgressProvider<TOptions> provide
...
             {
-                string jsonConfig = Console.ReadLine();
-                ExtensionEgressPayload configPayload = JsonSerializer.Deserialize<ExtensionEgressPayload>(jsonConfig);
+                StdInStream = Console.OpenStandardInput();
+
+                int dotnetMonitorPayloadProtocolVersion;
+                long payloadLengthBuffer;
+                byte[] payloadBuffer;
+
+                using (var reader = new BinaryReader(StdInStream, Encoding.UTF8, true))
```
jander-msft, line 47:
```suggestion
                using (BinaryReader reader = new(StdInStream, Encoding.UTF8, leaveOpen: true))
```

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -35,8 +38,28 @@ private static async Task<int> Egress<TOptions>(EgressProvider<TOptions> provide
...
+                {
+                    dotnetMonitorPayloadProtocolVersion = reader.ReadInt32();
+                    payloadLengthBuffer = reader.ReadInt64();
+
+                    payloadBuffer = new byte[payloadLengthBuffer];
+
+                    reader.Read(payloadBuffer, 0, (int)payloadLengthBuffer); // this should be safe, given we know the payload is small
```
jander-msft, line 54: Let's not use the `BinaryReader` to read out the payload since we aren't using the `BinaryWriter` to write the payload. Basically, the same methodology that we use to write is the same methodology that we should use to read.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -35,8 +38,28 @@ private static async Task<int> Egress<TOptions>(EgressProvider<TOptions> provide
...
+                StdInStream = Console.OpenStandardInput();
+
+                int dotnetMonitorPayloadProtocolVersion;
+                long payloadLengthBuffer;
+                byte[] payloadBuffer;
+
+                using (var reader = new BinaryReader(StdInStream, Encoding.UTF8, true))
+                {
+                    dotnetMonitorPayloadProtocolVersion = reader.ReadInt32();
+                    payloadLengthBuffer = reader.ReadInt64();
```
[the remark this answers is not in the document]
kkeirstead: I believe we only need to do this downcast if we're using the BinaryReader; if we're switching back to ReadExactlyAsync for the payload, I think we'll be able to keep it as a long

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
kkeirstead: I re-integrated `ReadExactlyAsync` for this - given that we still need it, what's the advantage of using Binary*er? My initial thought was it would enable us to get away from the manual `ReadExactlyAsync` method, but it seems like it's only solving part of the problem.

jander-msft: `BinaryWriter`/`BinaryReader` are good for writing protocols for exact serialization due to endianness concerns. `BinaryWriter`/`BinaryReader` will always write/read using little endian. Thus we can say "the first 4 bytes are the protocol version and its in little endian form".

I think that Windows is always little endian whereas Linux and Mac can be either depending on the processor its running on. While this is probably okay for .NET code to not care about it (because the .NET runtime will do the same thing on both sides of the communication so long as we use the same serialization methodology), native code extensions will not be as tolerant.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -35,8 +38,27 @@ private static async Task<int> Egress<TOptions>(EgressProvider<TOptions> provide
...
+                    dotnetMonitorPayloadProtocolVersion = reader.ReadInt32();
+                    if (dotnetMonitorPayloadProtocolVersion != ExpectedPayloadProtocolVersion)
+                    {
+                        throw new EgressException(string.Format(CultureInfo.CurrentCulture, Strings.ErrorMessage_IncorrectPayloadVersion, dotnetMonitorPayloadProtocolVersion, ExpectedPayloadProtocolVersion));
+                    }
+
+                    payloadLengthBuffer = reader.ReadInt64();
+                }
+
+                payloadBuffer = new byte[payloadLengthBuffer];
```
jander-msft, line 58: Can you validate that the length is not less than zero, as in, throw an ArgumentOutOfRangeException if it is less than zero?

jander-msft (approved):

jander-msft (approved):

PR #4207, opened 2023-04-11
[Egress] Validate Egress Extensions At Startup

kkeirstead:
###### Summary

For extensions with the `Validate` mode listed in their manifest, this change now validates that extensions will start up correctly (and data annotation validation) on startup. This gives users an opportunity to fix their configuration immediately, instead of having it fail the first time it's used.


src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -58,13 +66,20 @@ private static async Task<int> Egress<TOptions>(EgressProvider<TOptions> provide
                     }
                 }
 
-                Console.CancelKeyPress += Console_CancelKeyPress;
+                if (mode == ExtensionModes.Execute)
```
jander-msft, line 69: Rather than making this method serve a dual purpose, let's split the validation into a separate method.

I did a decent refactor in #4247. If we take those changes, you can easily create the service collection, pull out the `IOptionsSnapshot<T>` from it, and call `Get`. That will then throw an `OptionsValidationException` that you can report the results.

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -83,13 +71,69 @@ internal sealed class EgressHelper
...
+        private static int ProcessEgressResult(EgressArtifactResult result)
+        {
             string jsonBlob = JsonSerializer.Serialize(result);
             Console.Write(jsonBlob);
-
             // return non-zero exit code when failed
             return result.Succeeded ? 0 : 1;
         }
 
+        private static async Task<(TOptions, EgressProvider<TOptions>)> GetProviderAndOptions<TProvider, TOptions>(Action<IServiceCollection> configureServices, ExtensionEgressPayload configPayload)
```
jander-msft, line 107: I'm not certain that this refactor makes sense, given that the validation path doesn't involve the egress provider implementation.

src/Microsoft.Diagnostics.Monitoring.WebApi/IEgressService.cs
```diff
@@ -20,5 +20,9 @@ internal interface IEgressService
             IEndpointInfo source,
             CollectionRuleMetadata collectionRuleMetadata,
             CancellationToken token);
+
+        Task<EgressResult> ValidateProviderAsync(
```
jander-msft, line 24: Given that `ValidateProvider` and `ValidateProviderAsync` only differ from an interface perspective in that one is async and the other isn't, I think we whouls rename `ValidateProvider` to `ValidateProviderExists` and `ValidateProviderAsync` to `ValidateProviderOptionsAsync` to better convey the difference between their behaviors.

src/Tools/dotnet-monitor/Egress/EgressService.cs
```diff
@@ -51,6 +51,24 @@ public async Task<EgressResult> EgressAsync(string providerName, Func<Stream, Ca
             return new EgressResult(result.ArtifactPath);
         }
 
+        public async Task<EgressResult> ValidateProviderAsync(string providerName, CancellationToken token)
```
jander-msft, line 54:
```suggestion
        public async Task ValidateProviderAsync(string providerName, CancellationToken token)
```

Callers shouldn't care about the result and the result is fairly uninformative in this case since it would always be successful with an empty artifact path. Just throw on failure and return void on success.

src/Tools/dotnet-monitor/Egress/Extension/EgressExtension.cs
```diff
@@ -137,7 +168,7 @@ public void AddEnvironmentVariable(string key, string value)
                 StartInfo = pStart,
             };
 
-            using OutputParser<EgressArtifactResult> parser = new(p, _logger);
+            using OutputParser<EgressArtifactResult> parser = new(p, _logger, mode);
```
jander-msft, line 171: Rather than updating `OutputParser` to understand modes and conditional logging, please try to pass `NullLogger<EgressExtension>.Instance` instead.

src/Tools/dotnet-monitor/Egress/Extension/EgressExtension.cs
```diff
@@ -70,10 +70,40 @@ public void AddEnvironmentVariable(string key, string value)
...
+
+            return result;
+        }
+
         public async Task<EgressArtifactResult> EgressArtifact(
             ExtensionEgressPayload payload,
             Func<Stream, CancellationToken, Task> action,
-            CancellationToken token)
+            CancellationToken token,
+            ExtensionModes mode = ExtensionModes.Execute)
```
jander-msft, line 106: nit: Remove the default value and have the caller be explicit about which mode to use.

src/Tools/dotnet-monitor/Egress/FileSystem/FileSystemEgressExtension.cs
```diff
@@ -38,6 +38,16 @@ public FileSystemEgressExtension(IEgressConfigurationProvider configurationProvi
             EgressArtifactSettings settings,
             Func<Stream, CancellationToken, Task> action,
             CancellationToken token)
+        {
+            return await EgressArtifact(providerName, settings, action, false, token);
+        }
+
+        public async Task<EgressArtifactResult> EgressArtifact(
```
jander-msft, line 45: Undo everything related to this "new" method since validation is handled separately.

jander-msft: Try to remove the validation code from here and create a separate method for validation that is sharable among `EgressArtifact` and `ValidateProviderAsync`. It can even use https://github.com/dotnet/dotnet-monitor/blob/main/src/Tools/dotnet-monitor/DataAnnotationValidateOptions.cs if you want to deduplicate code.

src/Tools/dotnet-monitor/Egress/EgressProviderSource.cs
```diff
@@ -98,6 +104,13 @@ private void Reload()
                     }
                 }
             }
+
+            CancellationTokenSource source = new(ValidationTimeoutMilliseconds);
+
+            foreach (var providerName in _providerNameToTypeMap.Keys)
```
jander-msft, line 110: This isn't particularly safe to do from a non-async context and creates unobservable Tasks. Let's move this to a new implementation that derives from BackgroundService. Then add a Reloaded event to the EgressProviderSource that the background services can listen to. Each time the event is raised, the background service can then async await the validation of each provider. If validation is already in flight, then those need to be canceled before validating the reload.

src/Tools/dotnet-monitor/Egress/Extension/EgressExtension.OutputParser.cs
```diff
@@ -73,22 +73,22 @@ private void ParseStdOut(object sender, DataReceivedEventArgs eventArgs)
                         }
                         else
                         {
-                            _logger.ExtensionMalformedOutput(_processId, eventArgs.Data, typeof(TResult));
+                            _logger?.ExtensionMalformedOutput(_processId, eventArgs.Data, typeof(TResult));
```
jander-msft, line 76: Please revert remaining changes in this file

src/Tools/dotnet-monitor/Egress/Extension/EgressArtifactResult.cs
```diff
@@ -17,8 +17,8 @@ public bool IsValid()
         {
             if (Succeeded)
             {
-                // If Success, we must have an artifact path, and no failure message
-                return !string.IsNullOrEmpty(ArtifactPath) && string.IsNullOrEmpty(FailureMessage);
+                // If Success, we must have no failure message
+                return string.IsNullOrEmpty(FailureMessage);
```
kkeirstead, line 21: When we weren't keeping track of tasks this slipped through the cracks - the artifact path wasn't being set, so the result was never valid. This was causing `_resultCompletionSource` to never set the result in the output parser, which made the `WhenAll` for the validation tasks hang. This is one way of addressing the issue (you could also mandate that validation includes an arbitrary `ArtifactPath`, but I don't think that's a good idea).

src/Tools/dotnet-monitor/Egress/EgressValidationService.cs
```diff
@@ -0,0 +1,99 @@
...
+            _serviceProvider = serviceProvider;
+        }
+
+        protected override async Task ExecuteAsync(CancellationToken token)
+        {
+            // Indicates that rules changed while validating a previous change.
+            // Used to indicate that the next iteration should not wait for
+            // another configuration change since a change was already detected.
+            bool configurationChanged = false;
+            bool initialLoad = true;
```
kkeirstead, line 31: @jander-msft I know we discussed this yesterday, but after playing around with initializing `EgressProviderSource` my guess is that I didn't explain my issue well enough (that or I'm doing something incorrectly here). `EgressProviderSource` is Reloaded **before** this background service starts, so when `ConfigurationChanged` is invoked there, nothing happens here. I'm currently working around this by doing a one-off initial check, but let me know if there's a more correct solution I should be using.

jander-msft: Try something like this in ExecuteAsync:
- Unconditionally call Initiailize on the EgressProviderSource.
- Create a SemaphoreSlim (initial count 0, max count 1) and an event handler for the EgressProviderSource change event that will call Release on the SemaphoreSlim. Catch SemaphoreFullException since multiple reloads can overflow it.
- In a while true loop, validate the current list of egress providers and then await the SemaphoreSlim. This allows the first validation pass before waiting for the event to be raised.

There are other things to fill in like cancelling the current validation pass when the event is raised again, but that can be done with a combination of a second event handler and a linked cancellation token.

It's okay that EgressProviderSource loads before ExecutAsync starts; it shouldn't care how many times it reloaded before it does the initial validation pass. And putting a upper bound on the SemaphoreSlim ensures that the loop doesn't continuously validate the providers due to quick successive changes in configuration.

src/Tools/dotnet-monitor/Egress/EgressProviderSource.cs
```diff
@@ -21,6 +21,8 @@ internal sealed class EgressProviderSource : IDisposable
         private readonly ExtensionDiscoverer _extensionDiscoverer;
         private readonly ILogger _logger;
         private readonly IDictionary<string, string> _providerNameToTypeMap;
+        public IEnumerable<string> ProviderNames { get; set; } = new List<string>();
+        public event EventHandler ConfigurationChanged;
```
jander-msft, line 25: ProvidersChanged

src/Tools/dotnet-monitor/Egress/EgressProviderSource.cs
```diff
@@ -98,6 +100,13 @@ private void Reload()
                     }
                 }
             }
+
+            lock(ProviderNames)
```
jander-msft, line 104: The lock on the ProviderNames property is not necessary, and may actually cause confusion; I would remove it.

If a caller is getting the provider list, they are either getting the current list or they are getting the list that is about to be outdated but hasn't be updated yet because the reload is in progress. That's okay because the caller should be listening to the change event and will get notified imminently when that list changes.

Additionally, reads and writes of fields/locals/etc of types that are 4-byte or smaller and references types are guaranteed to be atomic.

src/Tools/dotnet-monitor/Egress/EgressValidationService.cs
kkeirstead: I did something similar to what you suggested, though I didn't use a second event handler (I believe the way I'm doing it now shouldn't require it, but you can let me know if there's an issue with that approach).

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperation.cs
```diff
@@ -76,7 +74,22 @@ public void Validate(IServiceProvider serviceProvider)
         {
             serviceProvider
                 .GetRequiredService<IEgressService>()
-                .ValidateProvider(EgressProviderName);
+                .ValidateProviderExists(EgressProviderName);
+        }
+
+        public static async Task<ExecutionResult<EgressResult>> ValidateAsync(IServiceProvider serviceProvider, string endpointName, CancellationToken token)
```
jander-msft, line 80:
```suggestion
        public static async Task ValidateAsync(IServiceProvider serviceProvider, string endpointName, CancellationToken token)
```

Caller shouldn't have to inspect the result so don't provide it.

src/Tools/dotnet-monitor/Egress/EgressValidationService.cs
```diff
@@ -0,0 +1,79 @@
...
+                    {
+                        // Ignore since multiple reloads can overflow it
+                    }
+                };
+
+                try
+                {
+                    _egressProviderSource.ProvidersChanged += configChangedHandler;
+                    List<Task> validationTasks = new();
+                    lock (_egressProviderSource.ProviderNames)
```
jander-msft, line 57: Remove this lock

src/Tools/dotnet-monitor/Egress/EgressValidationService.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            _egressProviderSource = source;
+            _serviceProvider = serviceProvider;
+        }
+
+        protected override async Task ExecuteAsync(CancellationToken token)
+        {
+            _egressProviderSource.Initialize();
+
+            SemaphoreSlim semaphore = new(0, 1);
```
jander-msft, line 28: Please name the arguments so it's easier to read how the arguments are used.

src/Tools/dotnet-monitor/Egress/FileSystem/FileSystemEgressExtension.cs
```diff
@@ -195,5 +173,42 @@ private static string WrapMessage(string innerMessage)
...
+            EgressArtifactSettings settings,
+            CancellationToken token)
+        {
+            EgressArtifactResult result = new();
+            try
+            {
+                FileSystemEgressProviderOptions options = ValidateOptions(providerName);
+                result.Succeeded = true;
+            }
+            catch (Exception ex)
```
jander-msft, line 187: Only catch OptionsValidationException; all other exceptions should flow through to help indicate some error in the system.

src/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -83,13 +74,68 @@ internal sealed class EgressHelper
...
+            {
+                ExtensionEgressPayload configPayload = await GetPayload(token);
+                await using var serviceProvider = BuildServiceProvider<TProvider, TOptions>(configureServices, configPayload);
+
+                TOptions options = serviceProvider.GetRequiredService<IOptionsSnapshot<TOptions>>().Get(configPayload.ProviderName);
+
+                result.ArtifactPath = string.Empty;
+                result.Succeeded = true;
+            }
+            catch (Exception ex)
```
jander-msft, line 95: Only catch OptionsValidationException; all other exceptions should flow through to help indicate some error in the system.

jander-msft (commented):
I think this is close to finished. Leaving a few more comments.

src/Tools/dotnet-monitor/Egress/FileSystem/FileSystemEgressExtension.cs
```diff
@@ -195,5 +173,42 @@ private static string WrapMessage(string innerMessage)
...
+        }
+
+        private FileSystemEgressProviderOptions ValidateOptions(string providerName)
+        {
+            IConfigurationSection configuration = _configurationProvider.GetProviderConfigurationSection(EgressProviderTypes.FileSystem, providerName);
+
+            FileSystemEgressProviderOptions options = new();
+            configuration.Bind(options);
+
+            DataAnnotationValidateOptions<FileSystemEgressProviderOptions> validateOptions = new(null);
```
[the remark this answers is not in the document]
jander-msft: I don't understand the concern. Is it that you are not sure how to provide an IServiceProvider instance to the FileSystemEgressExtension in unit tests? Are you not sure how to verify that the IServiceProvider is accessible from the validation context during testing?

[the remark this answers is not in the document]
kkeirstead: The former - I wasn't sure what the precedent is for doing this with UnitTests (as opposed to our functional tests, where we use ServiceProviderFixture).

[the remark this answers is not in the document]
jander-msft: Use `Mock<IServiceProvider>` in the unit tests. We use the `ServiceProviderFixture` to provision only a few `HttpClient`s across all of the test instances.

src/Tools/dotnet-monitor/ServiceCollectionExtensions.cs
```diff
@@ -239,6 +239,7 @@ public static IServiceCollection ConfigureExtensions(this IServiceCollection ser
             // Well-known extensions
             services.AddSingleton<ExtensionRepository, WellKnownExtensionRepository>();
             services.AddSingleton<IWellKnownExtensionFactory, FileSystemEgressExtensionFactory>();
+            services.AddHostedService<EgressValidationService>();
```
jander-msft, line 242: Move this to `ConfigureEgress`

src/Tools/dotnet-monitor/Egress/EgressValidationService.cs
```diff
@@ -0,0 +1,76 @@
...
+                    {
+                        // Ignore since multiple reloads can overflow it
+                    }
+                };
+
+                try
+                {
+                    _egressProviderSource.ProvidersChanged += configChangedHandler;
+                    List<Task> validationTasks = new();
+                    foreach (var providerName in _egressProviderSource.ProviderNames)
```
jander-msft, line 57:
```suggestion
                    IReadOnlyCollection providerNames = _egressProviderSource.ProviderNames;
                    List<Task> validationTasks = new(providerNames.Count);
                    foreach (var providerName in providerNames)
```

When the count of a list/collection/dictionary/etc are known ahead of time, we should set their initial capacity to that count to avoid reallocation when adding items to it.

src/Tools/dotnet-monitor/Egress/EgressProviderSource.cs
```diff
@@ -21,6 +21,8 @@ internal sealed class EgressProviderSource : IDisposable
         private readonly ExtensionDiscoverer _extensionDiscoverer;
         private readonly ILogger _logger;
         private readonly IDictionary<string, string> _providerNameToTypeMap;
+        public IEnumerable<string> ProviderNames { get; set; } = new List<string>();
```
jander-msft, line 24:
```suggestion
        public IReadOnlyCollection<string> ProviderNames { get; set; } = new List<string>();
```

src/Tools/dotnet-monitor/Egress/EgressProviderSource.cs
```diff
@@ -98,6 +100,10 @@ private void Reload()
                     }
                 }
             }
+
+            ProviderNames = new List<string>(_providerNameToTypeMap.Keys);
```
jander-msft, line 104:
```suggestion
            ProviderNames = new List<string>(_providerNameToTypeMap.Keys).AsReadOnly();
```

src/Tools/dotnet-monitor/Extensibility/ExtensionMode.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Extensibility
+{
+    [JsonConverter(typeof(JsonStringEnumConverter))]
```
jander-msft, line 8: Could you see if moving this attribute to the `ExtensionManifest.Modes` property works? This way, there isn't an indication or expectation from the extension implementation that this is involved in JSON serialization (even though it is for this one specific usage).

jander-msft (commented):
I think these are the last set of changes that I'll ask for. Thanks for the great work!

src/Tools/dotnet-monitor/Extensibility/ExtensionManifest.cs
```diff
@@ -47,7 +53,11 @@ public static ExtensionManifest FromPath(string path)
 
             try
             {
-                return JsonSerializer.Deserialize<ExtensionManifest>(stream);
+                JsonSerializerOptions serializerOptions = new()
```
jander-msft, line 56: nit: Can you make this a `static readonly` on the class so it only has to be create once?

jander-msft (approved):

jander-msft (approved):

PR #4247, opened 2023-04-13
Separate serialization from egress implementations and use dependency injection

jander-msft:
###### Summary

This change allows removing the need from the common egress extension infrastructure of how to configure additional egress options, how to validate them, how to pass additional services to the egress provider, etc. It additionally separates the serialization model (namely the `ExtensionEgressPayload`) from egress implementations so that the model and protocol can be changed independently from the egress providers.

This should overall simplify the need for egress implementations to construct and pass instances of services that may not be necessary or used and allow the callers to provide their own services that participate in dependency injection.


src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressProvider.cs
```diff
@@ -11,12 +10,10 @@ namespace Microsoft.Diagnostics.Monitoring.Extension.Common
...
     internal abstract class EgressProvider<TOptions> where TOptions : class
     {
-        protected ILogger Logger { get; set; }
-
-        public EgressProvider() { }
+        protected EgressProvider() { }
 
         public abstract Task<string> EgressAsync(
-            ILogger logger,
+            string providerName,
```
kkeirstead, line 16: Is there a reason why we're passing this here? I couldn't find how S3/Azure were using it

src/Extensions/AzureBlobStorage/PostConfigureAzureBlobEgressProviderOptions.cs
```diff
@@ -0,0 +1,67 @@
...
+        public PostConfigureAzureBlobEgressProviderOptions(ILogger<AzureBlobEgressProvider> logger, EgressProperties properties)
+        {
+            _properties = properties;
+            _logger = logger;
+        }
+
+        public void PostConfigure(string name, AzureBlobEgressProviderOptions options)
+        {
+            // If account key was not provided but the name was provided,
+            // lookup the account key property value from EgressOptions.Properties
```
kkeirstead, line 25: nit: I realize this comment was already out of date, but let's fix the reference to `EgressOptions.Properties` for all three of these.

src/Tools/dotnet-monitor/DataAnnotationValidateOptions.cs
```diff
@@ -7,7 +7,11 @@
 using System.Collections.ObjectModel;
 using System.ComponentModel.DataAnnotations;
 
+#if EXTENSION
+namespace Microsoft.Diagnostics.Monitoring.Extension.Common
```
kkeirstead, line 11: I'm assuming this is temporary, given our future plans to physically separate dotnet monitor from the extensions?

src/Extensions/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressProvider.cs
jander-msft: These two extensions don't need it but it's there for if we add other extensions that need to look up configuration external from dotnet-monitor. I suppose the argument would be that all of the extensions that we maintain should get their configuration from dotnet-monitor so this wouldn't be necessary. I can remove it for now.

src/Tools/dotnet-monitor/DataAnnotationValidateOptions.cs
jander-msft: Yes, linking in files is good for now to avoid code duplication, especially if we don't intend to ship libraries. If there is a bifurcation, then whoever does it needs to just copy the file as-is.

kkeirstead (approved):

kkeirstead (approved):

PR #4266, opened 2023-04-14
Add stack frame information to exceptions event source

jander-msft:
###### Summary

- Report stack frame information (list of method and IL offset pairs) from exceptions
- Cache stack frame information so that it is not repeated through the event source
- New stack frame information is reported to the event source.
- Added tests to validate caching and event source information.
- Fix call stack gathering in order to get full call stack of thrown exception.
- Fixed TargetFrameworkMonikerTrait attributes on tests to have them correctly run when filtered by TFM (e.g. PR builds only run tests for the latest TFM)

Note: The dotnet-monitor tool does nothing with the stack frame information. This is will be implemented in a follow up PR in order to keep this one simple.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Identification/ExceptionIdentifierCache.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Identification/ExceptionIdentifierCache.cs
@@ -16,14 +16,14 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Identification
     /// </summary>
     internal sealed class ExceptionIdentifierCache
     {
-        private const ulong InvalidId = 0;
-
         // List of callbacks to invoke for newly provided metadata.
         private readonly IEnumerable<ExceptionIdentifierCacheCallback> _callbacks;
 
         // Mapping of ExceptionIdentifier to a unique ID; ExceptionIdentifier itself does not have a way
         // of uniquely identifying itself using a primitive type.
         private readonly ConcurrentDictionary<ExceptionIdentifier, ulong> _exceptionIds = new();
+
+        private readonly ConcurrentDictionary<MethodBase, ulong> _methodIds = new();
 
         // Mapping of Module to a unique ID; Module itself does not have a way
         // of uniquely identifying itself using a primitive type.
@@ -35,6 +35,7 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Identification
         private readonly NameCache _nameCache = new();
 
         private ulong _nextRegistrationId = 1;
+        private ulong _nextMethodId = 1;
         private ulong _nextModuleId = 1;
         private ulong _nextStackFrameId = 1;
 
@@ -85,23 +86,37 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Identification
         /// </returns>
         public ulong GetOrAdd(MethodBase method)
         {
-            // Dynamic methods do not have method handles and can be detected by checking if their declaring type is null
-            // Do not attempt to create an identifier for these and skip them.
-            // CONSIDER: Can an artificial identifier be created that does not collide with the algorithm used in GetId(MethodBase)?
-            if (null == method.DeclaringType)
-            {
-                return InvalidId;
-            }
+            if (_methodIds.TryGetValue(method, out ulong methodId))
+                return methodId;
 
-            ulong methodId = GetId(method);
-            if (!_nameCache.FunctionData.ContainsKey(methodId))
+            methodId = Interlocked.Increment(ref _nextMethodId);
+
+            ulong actualMethodId = _methodIds.GetOrAdd(method, methodId);
+
+            if (methodId == actualMethodId)
             {
+                // Dynamic methods do not have metadata tokens
+                uint metadataToken = 0;
+                try
+                {
+                    metadataToken = Convert.ToUInt32(method.MetadataToken);
+                }
+                catch (Exception) { }
+
+                // RTDynamicMethod does not implement GetGenericArguments.
+                Type[] genericArguments = Array.Empty<Type>();
+                try
+                {
+                    genericArguments = method.GetGenericArguments();
+                }
+                catch (Exception) { }
+
                 FunctionData data = new(
                     method.Name,
                     AddOrDefault(method.DeclaringType),
-                    Convert.ToUInt32(method.MetadataToken),
+                    metadataToken,
                     GetOrAdd(method.Module),
-                    GetOrAdd(method.GetGenericArguments())
+                    GetOrAdd(genericArguments)
                     );
 
                 if (_nameCache.FunctionData.TryAdd(methodId, data))
@@ -276,11 +291,6 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Identification
                 return default;
 
             return GetOrAdd(type);
-        }
-
-        private static ulong GetId(MethodBase method)
-        {
-            return Convert.ToUInt64(method.MethodHandle.Value.ToInt64());
         }
 
         private ulong GetId(Module module)
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Identification/ExceptionIdentifierCache.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Identification/ExceptionIdentifierCache.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Reflection;
 using System.Threading;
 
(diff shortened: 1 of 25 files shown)


src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Identification/ExceptionIdentifierCache.cs
```diff
@@ -121,6 +125,57 @@ public ulong GetOrAdd(Module module)
             return moduleId;
         }
 
+        /// <summary>
+        /// Gets the identifier of the <see cref="StackFrame"/> instance and
+        /// caches its data if it has not be encountered yet.
```
kkeirstead, line 130:
```suggestion
        /// caches its data if it has not been encountered yet.
```

src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/Exceptions/Eventing/ExceptionsEventSourceTests.cs
```diff
@@ -99,11 +107,28 @@ public void ExceptionsEventSource_WriteException_NotEnabled()
...
 
-            source.ExceptionInstance(9, OperationCancelledExceptionMessage);
+            source.ExceptionInstance(9, OperationCancelledExceptionMessage, Array.Empty<ulong>());
 
             Assert.Empty(listener.Exceptions);
         }
 
+        [Theory]
+        [InlineData(0, 0, 0)]
+        [InlineData(1, 42, 7)]
```
kkeirstead, line 117: Any significance to these values?

jander-msft: No, they are arbitrary. Want to test with _something_ but not random numbers.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
```diff
@@ -40,12 +41,64 @@ public void Invoke(Exception exception)
...
+                _eventSource.ExceptionInstance(
+                    identifier,
+                    exception.Message,
+                    frameIds);
             }
 
             _next(exception);
         }
+
+        private static ReadOnlySpan<StackFrame> ComputeEffectiveCallStack(Exception exception)
```
jander-msft, line 59: Let's validate this works for reverse p/invoke scenarios.

jander-msft, line 59: Let's check what happens for first chance exceptions that will be fatal.

jander-msft: When throwing an exception from a reverse p/invoke (native code calling managed code), the call stack only contains the managed frames; so it works, but there's no indication that manged-to-native or native-to-managed transitions have occurred. We might be able to inspect each method and see if there are a pair of matching extern methods:
reverse
Because the managed callstack is intact, I'll save transition indication for a separate work item.

jander-msft: We are able to observe unhandled exceptions from the UnhandledException event in the app domain. However, the call stack computation here doesn't work for an exception that was already thrown from within the context of an async method execution. The call stack is preserved on the exception and doesn't match the current thread's call stack because of the async state machine behavior.

This isn't really an issue because we ignore rethrown exceptions.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Identification/ExceptionIdentifierCache.cs
```diff
@@ -79,6 +85,14 @@ public ulong GetOrAdd(ExceptionIdentifier exceptionId)
         /// </returns>
         public ulong GetOrAdd(MethodBase method)
         {
+            // Dynamic methods do not have method handles and can be detected by checking if their declaring type is null
+            // Do not attempt to create an identifier for these and skip them.
+            // CONSIDER: Can an artificial identifier be created that does not collide with the algorithm used in GetId(MethodBase)?
+            if (null == method.DeclaringType)
```
wiktork, line 91: Are there situations where declaring type is null but method id is not null? In the profiler this is possible (I.e. FunctionId is not null, but the method is not associated with a type)

wiktork: Another interesting scenario is arrays. I.e. throw an exception from one of the methods on array instance. Does declaring type resolve to Array or somethign else.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Identification/ExceptionIdentifierCache.cs
```diff
@@ -121,6 +135,57 @@ public ulong GetOrAdd(Module module)
...
+                frame.GetILOffset());
+
+            if (_stackFrameIds.TryGetValue(identifier, out ulong frameId))
+                return frameId;
+
+            frameId = Interlocked.Increment(ref _nextStackFrameId);
+
+            ulong actualFrameId = _stackFrameIds.GetOrAdd(identifier, frameId);
+
+            if (actualFrameId == frameId)
```
wiktork, line 158: Can you add some comments around this scenario? I'm assuming this is to handle concurrency.

wiktork (dismissed):

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Identification/ExceptionIdentifierCache.cs
jander-msft: I reworked this part to allow handling of dynamic methods and catch exceptions appropriately for API calls that will fail for them; ideally I would be able to easily identify a dynamic method but the only way I found to do that was to check the runtime type of `method` and see if it's a RTDynamicMethod (this type is internal to System.Private.CoreLib), but I decided against doing that type of check since we don't care that it is dynamic and just want to get as best information out of it as possible. So now we only using `DeclaringType` to get the owner type of the method.

jander-msft: Arrays work fine; the `DeclaringType` is `System.Array`.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Identification/ExceptionIdentifierCache.cs
jander-msft: I ended up refactoring all of the usages of this pattern into one method that gets called by each concept. The comments have been added to that method to explain in great detail how it's handling concurrent inserts into the dictionary.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Identification/ExceptionIdentifierCache.cs
```diff
@@ -121,6 +135,57 @@ public ulong GetOrAdd(Module module)
...
+        public ulong GetOrAdd(StackFrame frame)
+        {
+            StackFrameIdentifier identifier = new(
+                AddOrDefault(frame.GetMethod()),
+                frame.GetILOffset());
+
+            if (_stackFrameIds.TryGetValue(identifier, out ulong frameId))
+                return frameId;
+
+            frameId = Interlocked.Increment(ref _nextStackFrameId);
```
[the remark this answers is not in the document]
jander-msft: I'm going to keep this as `ulong` in this change since it's what we use everywhere at this time. I can do a separate PR to reduce the field sizes so as to not complicate this PR further.

jander-msft:
I've added additional test stubs for reverse p/invokes, dynamic methods, and array methods. These will be better filled out in the next PR where I plumb through the call stack information on the dotnet-monitor side.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Identification/ExceptionIdentifierCache.cs
```diff
@@ -79,22 +79,38 @@ public ulong GetOrAdd(ExceptionIdentifier exceptionId)
...
-                    Convert.ToUInt32(method.MetadataToken),
-                    GetOrAdd(method.Module),
-                    GetOrAdd(method.GetGenericArguments())
-                    );
-
-                if (_nameCache.FunctionData.TryAdd(methodId, data))
-                {
-                    InvokeFunctionDataCallbacks(methodId, data);
-                }
+                metadataToken = Convert.ToUInt32(method.MetadataToken);
```
wiktork, line 89: nit: Consider a more specialized exception catch

wiktork (approved):

PR #4316, opened 2023-04-26
Use Ubuntu 18.04 for Linux x64 builds

jander-msft:
###### Summary

Recent changes in the minimum requirement for the glibc version for .NET 8 prevent it from being installed on CentOS 7 (has a glibc version of 2.17). The new minimum is 2.23 (see https://github.com/dotnet/core/pull/8401). To resolve this for the medium term, switch over to using Ubuntu 18.04, which has a glibc version of 2.27. The longer term fix to use glibc version 2.23 will require much more churn in the build process, but it should be addressed prior to shipping .NET Monitor 8.

### DIFF
--- a/eng/pipelines/jobs/build.yml
+++ b/eng/pipelines/jobs/build.yml
@@ -73,7 +73,7 @@ jobs:
       ${{ if eq(parameters.architecture, 'arm64') }}:
         container: mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-cross-arm64
       ${{ else }}:
-        container: mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7
+        container: mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04
 
     ${{ if eq(parameters.osGroup, 'Linux_Musl') }}:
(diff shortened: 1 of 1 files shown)


eng/pipelines/jobs/build.yml
```diff
@@ -73,7 +73,7 @@ jobs:
       ${{ if eq(parameters.architecture, 'arm64') }}:
         container: mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-cross-arm64
       ${{ else }}:
-        container: mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7
+        container: mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04
```
wiktork, line 76: Does 2.27 glibc have any incompatibilities with 2.23?

wiktork (approved):

eng/pipelines/jobs/build.yml
jander-msft: In the future, we're likely to split test phase into its own set of jobs where we run all of the tests on different distros to validate the backward compatibility of newer distros with old glibc reference in the profiler.

PR #4330, opened 2023-04-28
Split test phase into separate jobs

jander-msft:
###### Summary

This change splits the test phase into its own set of jobs separately from the build phase. This will allow a future change where the build phase will use different containers from cross-arch and cross-C-runtime compilation. An added benefit is that tests can be rerun without having to rebuild the binaries. This does add some time to the build because of agent acquisition wait time and having to upload/download the binaries.

The only issue that occurred is that the extension tests were starting to fail due to the inability to run the executable (permission denied). This was fixed by `chmod`ing the executable after copying it to the temporary directory and then performing the remainder of the test.

Example build (pre-squash): https://dev.azure.com/dnceng/internal/_build/results?buildId=2169914&view=results


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EgressExtensibilityTests.cs
```diff
@@ -132,6 +133,32 @@ private IEgressExtension FindEgressExtension(ConfigDirectory configDirectory, Ho
...
+                {
+                    FileName = "chmod",
+                    Arguments = "+x " + executablePath,
+                    UseShellExecute = true
+                };
+
+                using Process proc = Process.Start(startInfo);
+                if (!proc.WaitForExit(60_000)) // 1 minute
+                {
+                    throw new InvalidOperationException("Unable to make extension executable: Timeout out.");
```
kkeirstead, line 153:
```suggestion
                    throw new InvalidOperationException("Unable to make extension executable: Timed out.");
```

eng/pipelines/jobs/build-binaries.yml
```diff
@@ -0,0 +1,90 @@
+# Builds and tests dotnet-monitor for a specific platform and configuration
```
wiktork, line 1: Update comment

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EgressExtensibilityTests.cs
```diff
@@ -132,6 +133,33 @@ private IEgressExtension FindEgressExtension(ConfigDirectory configDirectory, Ho
 
             CopyExtensionFiles(extensionDirPath);
 
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
wiktork, line 136: Would it be possible to shift this into the test pipeline itself instead of doing it in the code?

wiktork (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EgressExtensibilityTests.cs
jander-msft: I don't think so. Each invocation of the test creates a new temporary directory to which this is copied. The pipeline won't know what that directory is.

We can take a look at running the extension as-is from where it was built/downloaded rather than copying it to a new directory, unless the point of the test is to test that it can run elsewhere.

eng/pipelines/jobs/build-binaries.yml
jander-msft: Will do in next PR

PR #4349, opened 2023-05-03
Use CBL Mariner build images

jander-msft:
###### Summary

This change switch all of the build jobs to use CBL Mariner with cross build support for clib and architecture variants.

Overview of changes:
- Switch to using CBL Mariner as the default container for Linux builds; allow test jobs to override the container to run the tests on distros that have the target clib and architecture types.
- Move the cross-build notion up into the build-binaries job since it's only necessary when building binaries; base codeql job on this job instead of base build job.
- Disable tests the execute the test egress extension on Alpine images for now. The apphost for the extension is glibc x64, so it won't execute on Alpine. Will attempt a solution for this after this PR.

Example build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2173397&view=results

### DIFF
--- a/eng/pipelines/jobs/test-binaries.yml
+++ b/eng/pipelines/jobs/test-binaries.yml
@@ -32,6 +32,12 @@ jobs:
     disableComponentGovernance: true
     disableSbom: true
 
+    # Override containers for Linux to test on real clib and architecture variants
+    ${{ if eq(parameters.osGroup, 'Linux') }}:
+      container: mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04
+    ${{ if eq(parameters.osGroup, 'Linux_Musl') }}:
+      container: mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.14-WithNode
+
     variables:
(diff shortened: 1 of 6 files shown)


eng/pipelines/jobs/test-binaries.yml
```diff
@@ -32,6 +32,12 @@ jobs:
     disableComponentGovernance: true
     disableSbom: true
 
+    # Override containers for Linux to test on real clib and architecture variants
+    ${{ if eq(parameters.osGroup, 'Linux') }}:
+      container: mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04
+    ${{ if eq(parameters.osGroup, 'Linux_Musl') }}:
+      container: mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.14-WithNode
```
wiktork, line 39: I know this is the image we used before, but why do we need node?

wiktork (dismissed):

eng/pipelines/jobs/test-binaries.yml
jander-msft: Azure Pipelines requires containers to have node on them in order to execute steps within the image.

PR #4389, opened 2023-05-10
[Parameter Capturing] Import external code for il rewriting

schmittjoseph:
###### Summary

Note: This is targeting a feature branch.

This PR imports code from the runtime repo relating to IL rewriting.  Overview of commits:
- [The first commit](https://github.com/dotnet/dotnet-monitor/pull/4389/commits/e14ab31e6dd781afb1a8d4cf28a4331d34d5c0e4) only contains a direct import of external code with no modifications.
- [The second commit](https://github.com/dotnet/dotnet-monitor/pull/4389/commits/55bcf6b8889997c2a821105ded8df44daef20bbd) contains changes I've made to properly compile the new code and remove unnecessary code from `ILRewriter.{cpp,.h}` that won't be used.
- Everything afterwards are changes made during the PR process.

### DIFF
--- a/src/MonitorProfiler/Utilities/ILRewriter.h
+++ b/src/MonitorProfiler/Utilities/ILRewriter.h
@@ -4,7 +4,6 @@
 #pragma once
 
 #include <assert.h>
-#include <cstdio>
 #include "cor.h"
 #include "corprof.h"
 #include "corhlpr.h"


src/MonitorProfiler/Utilities/ILRewriter.cpp
```diff
@@ -600,195 +594,4 @@ void ILRewriter::DeallocateILMemory(LPBYTE pBody)
...
-
-
-/////////////////////////////////////////////////////////////////////////////////////////////////
-//
-// R E W R I T E
-//
-////////////////////////////////////////////////////////////////////////////////////////////////
-
-UINT ILRewriter::AddNewInt32Local()
-{
```
wiktork: Even if we are not using these methods, keeping them in is likely going to make future merges easier. Or did you need to take these out because they break compilation?

wiktork (dismissed):

wiktork (approved):

PR #4391, opened 2023-05-10
[Parameter Capturing] Dynamically configure if the profiler is notification-only

schmittjoseph:
###### Summary

Note: This is targeting a feature branch.

This PR introduces a new environment variable (`DotnetMonitor_Profiler_IsMutatingProfiler`) that can be set to cause our profiler to be capable of mutating the target process (not-notification-only).  This PR only includes the necessary changes to the profiler and not the supporting logic in our managed code.

Our `LoadAsNotificationOnly` callback is invoked before profiler initialization has fully occurred so we're unable to leverage `ICorProfiler` to get environment variables. Instead introduce a new helper method that uses platform-specific APIs.

### DIFF
--- a/src/MonitorProfiler/Environment/StartupEnvironmentUtilities.h
+++ b/src/MonitorProfiler/Environment/StartupEnvironmentUtilities.h
@@ -3,6 +3,8 @@
 
 #pragma once
 
+#include "corhlpr.h"
+
 #if TARGET_WINDOWS
 #include <processenv.h>
 #include "tstring.h"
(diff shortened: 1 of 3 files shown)


src/MonitorProfiler/Environment/StartupEnvironmentUtilities.h
```diff
@@ -0,0 +1,67 @@
...
+
+            isSet = (buffer[0] == EnabledValue);
+#else
+            const char EnabledValue = '1';
+
+            // Subsequent calls to getenv may invalidate the pointer returned by previous calls.
+            std::lock_guard<std::mutex> lock(s_getEnvMutex);
+
+            char *ret = std::getenv(name);
+            isSet = (ret != nullptr && ret[0] == EnabledValue && ret[1] == '\0');
```
kkeirstead, line 62: Just to confirm my understanding: `ret` is guaranteed to be null-terminated, and thus if `ret` was just `\0` we would short circuit and never evaluate `ret[1]`?

PR #4392, opened 2023-05-10
Add timestamp information to exception history

jander-msft:
###### Summary

These changes capture the time at which the exception is thrown, propagates it as UTC, and reports it in the exception history store. For JSON, it is formatted using ISO 8601 for standard interchange. Example output:

```
{"timestamp":"2023-05-10T23:04:35.572644Z","typeName":"System.InvalidOperationException","moduleName":"System.Private.CoreLib.dll","message":"I don\u0027t like the count!"}
{"timestamp":"2023-05-10T23:04:42.4341469Z","typeName":"System.InvalidOperationException","moduleName":"System.Private.CoreLib.dll","message":"I don\u0027t like the count!"}
```

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/ExceptionPipelineExceptionContext.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/ExceptionPipelineExceptionContext.cs
@@ -1,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Pipeline
+{
+    internal sealed class ExceptionPipelineExceptionContext
+    {
+        public ExceptionPipelineExceptionContext(DateTime timestamp)
+        {
+            Timestamp = timestamp;
+        }
+
+        public DateTime Timestamp { get; }
+    }
+}
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/Exceptions/Eventing/ExceptionsEventSourceTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/Exceptions/Eventing/ExceptionsEventSourceTests.cs
@@ -77,7 +77,9 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Eventing
             listener.EnableEvents(source, EventLevel.Informational);
 
             ulong[] frameIds = frameIdsString.Split(',', StringSplitOptions.RemoveEmptyEntries).Select(ulong.Parse).ToArray();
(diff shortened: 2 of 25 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/Exceptions/Eventing/ExceptionsEventSourceTests.cs
```diff
@@ -77,14 +77,17 @@ public void ExceptionsEventSource_WriteException_Event(ulong id, string message,
             listener.EnableEvents(source, EventLevel.Informational);
 
             ulong[] frameIds = frameIdsString.Split(',', StringSplitOptions.RemoveEmptyEntries).Select(ulong.Parse).ToArray();
-            source.ExceptionInstance(id, message, frameIds);
+            DateTime timestamp = DateTime.UtcNow;
```
kkeirstead, line 80: I think using the actual time makes sense for now - eventually if we only display the timestamps in the past X time window, we might want to consider using fixed values for some tests to check that old exceptions get weeded out (can't remember if this was something we discussed as being on the to-do list, or something like limiting by only keeping X most recent exceptions)

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/ExceptionPipelineExceptionContext.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Pipeline
+{
+    internal sealed class ExceptionPipelineExceptionContext
```
kkeirstead, line 8: From a design perspective, what's the purpose of this class as an abstraction around `DateTime`?

src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/Exceptions/Eventing/ExceptionsEventSourceTests.cs
jander-msft: That's NYI (not-yet-implemented). And the event source wouldn't understand any of that so it wouldn't be part of this test.

Clearing the cache from the startup hook exception pipeline step wouldn't be something observable by the dotnet-monitor side of things other than some type of reset event being emitted; it would only really be seen as the IDs being "reused".

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/ExceptionPipelineExceptionContext.cs
jander-msft: For now, it's just the timestamp. But I made it a separate class so that way if we add more contextual information in the future (e.g. first chance vs unhandled) then we don't need to refactor a bunch of the code to add yet another property.

PR #4428, opened 2023-05-16
[Parameter Capturing] Add probe instrumentation APIs to profiler

schmittjoseph:
###### Summary

This PR creates an API surface exported by our profiler dll for managed code to request function probe installation and removal on-demand.  The API surface includes:
1. `RegisterFunctionProbe` - This registers an existing managed function as a function probe. This can only be called once during an application's lifetime.
2. `RequestFunctionProbeInstallation` - This requests that the previously registered probe is installed into one or more functions next time they are invoked. Note: Only one set of functions can have probes installed at a given time. If a new set of functions need probes, `RequestFunctionProbeShutdown` must be called first.
3. `RequestFunctionProbeShutdown` - This requests that any functions that currently have probes installed have their probes uninstalled.

Note: This PR only contains the profiler changes and not the managed code layer.

#### Overview of Code
`src/MonitorProfiler/Mainprofiler/ProbeInstrumentation` is divided into 3 distinct layers:
1. `ProbeInstrumentation`: This is the top-most layer that provides the API surface, handling requests rejits+reverts.
4. `AssemblyProbePrep`: Once `ProbeInstrumentation` receives a request to install a probe into a given function, work needs to be done to first prepare the function's assembly for a probe to be installed into it. This includes emitting necessary types, assembly references, etc.
5. `ProbeInjector`: This handles the actual IL rewriting to install the function probe into a given method.

### DIFF
--- a/src/MonitorProfiler/ProbeInstrumentation/ProbeInjector.cpp
+++ b/src/MonitorProfiler/ProbeInstrumentation/ProbeInjector.cpp
@@ -13,7 +13,7 @@
 // Instrumentation requests provide special boxing instructions by using SpecialCaseBoxingTypeFlag
 // as the "token type" and the SpecialCaseBoxingTypes enum as the RID.
 //
-constexpr ULONG32 SpecialCaseBoxingTypeFlag = 0xff000000;
+constexpr ULONG32 SpecialCaseBoxingTypeFlag = 0x7f000000;
 enum class SpecialCaseBoxingTypes : ULONG32
 {
     TYPE_UNKNOWN = 0,


src/MonitorProfiler/ProbeInstrumentation/ProbeInstrumentation.cpp
```diff
@@ -0,0 +1,300 @@
...
+{
+}
+
+HRESULT ProbeInstrumentation::RegisterFunctionProbe(FunctionID enterProbeId)
+{
+    lock_guard<mutex> lock(m_probePinningMutex);
+
+    if (HasPinnedProbe())
+    {
+        // Probes have already been pinned.
```
kkeirstead, line 25: Not sure whether it makes sense in the current context, but given that we log down below when we _do_ receive probes, would we want to log when probes were already pinned? Or would you just note the absence of that message?

src/MonitorProfiler/ProbeInstrumentation/ProbeInstrumentation.cpp
```diff
@@ -0,0 +1,300 @@
...
+        case ProbeWorkerInstruction::UNINSTALL_PROBES:
+            hr = UninstallProbes();
+            if (hr != S_OK)
+            {
+                m_pLogger->Log(LogLevel::Error, _LS("Failed to uninstall probes: 0x%08x"), hr);
+            }
+            break;
+
+        default:
+            m_pLogger->Log(LogLevel::Error, _LS("Unknown message"));
```
kkeirstead, line 83: Would it be helpful to have a more descriptive message here (potentially providing the value `payload.instruction`)? Or does that not make sense for what this is guarding against?

src/MonitorProfiler/ProbeInstrumentation/ProbeInstrumentation.cpp
```diff
@@ -0,0 +1,300 @@
...
+
+HRESULT ProbeInstrumentation::RequestFunctionProbeInstallation(
+    ULONG64 functionIds[],
+    ULONG32 count,
+    ULONG32 argumentBoxingTypes[],
+    ULONG32 argumentCounts[])
+{
+    m_pLogger->Log(LogLevel::Debug, _LS("Probe installation requested"));
+
+    if (!HasPinnedProbe())
```
kkeirstead, line 103: I'm a bit confused here - if we're requesting an installation, why do we fail if there isn't a pinned probe (as opposed to the other way around like in `RegisterFunctionProbe`)? I also might be misunderstanding what this is supposed to be doing

src/MonitorProfiler/ProbeInstrumentation/ProbeInstrumentation.cpp
```diff
@@ -0,0 +1,300 @@
...
+    ULONG32 offset = 0;
+    for (ULONG32 i = 0; i < count; i++)
+    {
+        vector<ULONG32> tokens;
+        tokens.reserve(argumentCounts[i]);
+        ULONG32 j;
+
+        for (j = 0; j < argumentCounts[i]; j++)
+        {
+            tokens.push_back(argumentBoxingTypes[offset+j]);
```
kkeirstead, line 120: Should the `E_INVALIDARG` check be above this? Meaning, if `offset + j` really is greater than UINT32_MAX, we wouldn't want to try indexing into `argumentBoxingTypes` with it?

src/MonitorProfiler/ProbeInstrumentation/ProbeInstrumentation.cpp
```diff
@@ -0,0 +1,300 @@
...
+    vector<ModuleID> requestedModuleIds;
+    vector<mdMethodDef> requestedMethodDefs;
+
+    // JSFIX: Handle OOM scenarios.
+    requestedModuleIds.reserve(requests.size());
+    requestedMethodDefs.reserve(requests.size());
+
+    for (auto const& req : requests)
+    {
+        INSTRUMENTATION_REQUEST request;
```
kkeirstead, line 185: Having both `req` and `request` might be confusing from a naming perspective

src/MonitorProfiler/Utilities/TypeNameUtilities.cpp
```diff
@@ -8,6 +8,17 @@ TypeNameUtilities::TypeNameUtilities(ICorProfilerInfo12* profilerInfo) : _profil
 {
 }
 
+HRESULT TypeNameUtilities::CacheModuleNames(NameCache& nameCache, ModuleID moduleId)
+{
+    std::shared_ptr<ModuleData> moduleData;
+    if (!nameCache.TryGetModuleData(moduleId, moduleData))
```
kkeirstead, line 14: nit: is this redundant? Looking at the implementation of `TryGetModuleData`, it looks like one of its first checks is the inverse of this (in which case it just returns `S_OK`)

src/MonitorProfiler/ProbeInstrumentation/AssemblyProbePrep.h
```diff
@@ -0,0 +1,120 @@
...
+class AssemblyProbePrepData
+{
+public:
+    AssemblyProbePrepData(mdMemberRef probeMemberRef, COR_LIB_TYPE_TOKENS corLibTypeTokens) :
+        m_probeMemberRef(probeMemberRef), m_corLibTypeTokens(corLibTypeTokens)
+    {
+    }
+
+    const mdMemberRef GetProbeMemberRef() const { return m_probeMemberRef; }
+    const COR_LIB_TYPE_TOKENS GetCorLibTypeTokens() const { return m_corLibTypeTokens; }
```
wiktork, line 45: nit: return const reference

src/MonitorProfiler/ProbeInstrumentation/AssemblyProbePrep.cpp
```diff
@@ -0,0 +1,489 @@
...
+    HRESULT hr;
+    corlibAssemblyRef = mdAssemblyRefNil;
+
+    IfFailRet(HydrateResolvedCorLib());
+
+    ComPtr<IMetaDataAssemblyImport> pMetadataAssemblyImport;
+    IfFailRet(pMetadataImport->QueryInterface(IID_IMetaDataAssemblyImport, reinterpret_cast<void **>(&pMetadataAssemblyImport)));
+
+    // JSFIX: Consider RAII for scope guarding instead of closing this enum in all needed cases.
+    HCORENUM corEnum = 0;
```
wiktork, line 219: We have some utility classes for HCORENUM that can help with this.

src/MonitorProfiler/ProbeInstrumentation/AssemblyProbePrep.cpp
```diff
@@ -0,0 +1,489 @@
...
+                nullptr,
+                assemblyName.get(),
+                expectedLength,
+                &nameLength,
+                nullptr,
+                nullptr,
+                nullptr,
+                nullptr);
+
+            if (hr == CLDB_S_TRUNCATION)
```
wiktork, line 246: The typical approach here is to call GetAssemblyRefProps with just nameLength, and then call it again with a buffer that matches the right size. Alternatively, if you get the error code, reallocate with the right size and try again.

src/MonitorProfiler/ProbeInstrumentation/AssemblyProbePrep.cpp
```diff
@@ -0,0 +1,489 @@
...
+        return S_OK;
+    }
+
+    HRESULT hr;
+    ModuleID corLibId = 0;
+    ComPtr<ICorProfilerModuleEnum> pEnum;
+    ModuleID curModuleId;
+    mdTypeDef sysObjectTypeDef = mdTypeDefNil;
+
+    IfFailRet(m_pCorProfilerInfo->EnumModules(&pEnum));
```
wiktork, line 324: You might have to validate if this is still true, but it used to be the case that the corelib would always load first in the ModuleLoaded callback. Rather than attempting to resolve corlib by enumerating all the modules, you can simply propagate it from the load event.

src/MonitorProfiler/ProbeInstrumentation/AssemblyProbePrep.cpp
```diff
@@ -0,0 +1,489 @@
...
+    {
+        return E_UNEXPECTED;
+    }
+
+    corLibName = moduleData->GetName();
+
+    // Trim the .dll file extension
+    const tstring dllExtension = _T(".dll");
+    if (corLibName.length() > dllExtension.length() &&
+        corLibName.compare(corLibName.length() - dllExtension.length(), dllExtension.length(), dllExtension))
```
wiktork, line 396: will this break for different casing?

src/MonitorProfiler/ProbeInstrumentation/AssemblyProbePrep.cpp
```diff
@@ -0,0 +1,480 @@
...
+        return S_OK;
+    }
+
+    COR_LIB_TYPE_TOKENS corLibTypeTokens = {};
+    IfFailRet(EmitNecessaryCorLibTypeTokens(moduleId, corLibTypeTokens));
+
+    mdMemberRef probeMemberRef;
+    IfFailRet(EmitProbeReference(moduleId, probeMemberRef));
+
+    shared_ptr<AssemblyProbePrepData> data(new (nothrow) AssemblyProbePrepData(probeMemberRef, corLibTypeTokens));
```
wiktork, line 52: We handle OOM scenarios here but we will throw bad_alloc on stl unordered map operations if there isn't enough memory for allocations. One option here:
- Use throwing new for single allocations
- Use no throw for array like operations, since these could fail if you are requesting larger blocks

src/MonitorProfiler/ProbeInstrumentation/AssemblyProbePrep.cpp
```diff
@@ -0,0 +1,480 @@
...
+    HRESULT hr;
+
+    auto const& it = m_assemblyProbeCache.find(moduleId);
+    if (it != m_assemblyProbeCache.end())
+    {
+        return S_OK;
+    }
+
+    COR_LIB_TYPE_TOKENS corLibTypeTokens = {};
+    IfFailRet(EmitNecessaryCorLibTypeTokens(moduleId, corLibTypeTokens));
```
wiktork, line 47: If you keep the nothrow check, you should heap allocate first, then emit the tokens.

src/MonitorProfiler/ProbeInstrumentation/AssemblyProbePrep.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+    typeToken = mdTokenNil;
+
+    mdTypeRef typeRefToken;
+    hr = pMetadataImport->FindTypeRef(
+        resolutionScope,
+        name.c_str(),
+        &typeRefToken);
+
+    if (FAILED(hr))
```
wiktork, line 189: Do we actually need to find/Define or would Define just give back an existing token?

src/MonitorProfiler/ProbeInstrumentation/AssemblyProbePrep.h
```diff
@@ -0,0 +1,117 @@
...
+
+private:
+    mdMemberRef m_probeMemberRef;
+    COR_LIB_TYPE_TOKENS m_corLibTypeTokens;
+};
+
+typedef struct _PROBE_INFO_CACHE
+{
+    tstring assemblyName;
+    std::unique_ptr<BYTE[]> signature;
```
wiktork, line 52: nit: vector<BYTE> would probably work too and might be more convenient to use.

src/MonitorProfiler/ProbeInstrumentation/ProbeInjector.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+HRESULT ProbeInjector::InstallProbe(
+    ICorProfilerInfo* pICorProfilerInfo,
+    ICorProfilerFunctionControl* pICorProfilerFunctionControl,
+    const INSTRUMENTATION_REQUEST& request)
+{
+    IfNullRet(pICorProfilerInfo);
+    IfNullRet(pICorProfilerFunctionControl);
+
+    if (request.boxingTypes.size() > UINT32_MAX)
```
wiktork, line 43: Some other possible options here:
- Only process UINT32_MAX entries, ignore the rest
- resize the vector down to UINT32_MAX
- Ignore this condition
- Inject a log entry with a celebratory message

src/MonitorProfiler/ProbeInstrumentation/ProbeInjector.cpp
```diff
@@ -0,0 +1,213 @@
...
+            }
+        }
+
+        // Replace the i'th element in our new array with what we just pushed on the stack
+        pNewInstr = rewriter.NewILInstr();
+        pNewInstr->m_opcode = CEE_STELEM_REF;
+        rewriter.InsertBefore(pInsertProbeBeforeThisInstr, pNewInstr);
+    }
+
+    pNewInstr = rewriter.NewILInstr();
```
wiktork, line 136: Can you test instrumentation against a method that has a protected exception region starting at il offset 0?

src/MonitorProfiler/ProbeInstrumentation/ProbeInstrumentation.cpp
```diff
@@ -0,0 +1,313 @@
...
+            req.functionId,
+            NULL,
+            nullptr,
+            &processedRequest.moduleId,
+            &processedRequest.methodDef,
+            0,
+            nullptr,
+            nullptr));
+
+        IfFailLogRet(m_pAssemblyProbePrep->PrepareAssemblyForProbes(processedRequest.moduleId));
```
wiktork, line 215: Not something for this review, but it would probably be worth testing this system with AssemblyLoadContext scenarios

wiktork (dismissed):

wiktork (approved):

Issue #4427, opened 2023-05-16
Capture dump from root container using non-root .NET Monitor doesn't work


jander-msft:
Use the following example where the target application is running as root and the .NET Monitor container is running as non-root:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: deploy-exampleapp
spec:
  replicas: 1
  selector:
    matchLabels:
      app: exampleapp
  template:
    metadata:
      labels:
        app: exampleapp
    spec:
      restartPolicy: Always
      containers:
      - name: app
        image: mcr.microsoft.com/dotnet/samples:aspnetapp
        imagePullPolicy: Always
        env:
        - name: ASPNETCORE_URLS
          value: http://+:80
        - name: DOTNET_DiagnosticPorts
          value: /diag/dotnet-monitor.sock
        volumeMounts:
        - mountPath: /diag
          name: diagvol
      - name: monitor
        image: mcr.microsoft.com/dotnet/nightly/monitor:8.0.0-preview.4
        args: [ "collect", "--urls", "http://localhost:52323", "--metricUrls", "http://+:52325", "--no-auth" ]
        imagePullPolicy: Always
        env:
        - name: DOTNETMONITOR_DiagnosticPort__ConnectionMode
          value: Listen
        - name: DOTNETMONITOR_Storage__DefaultSharedPath
          value: /diag
        volumeMounts:
        - mountPath: /diag
          name: diagvol
      volumes:
      - name: diagvol
        emptyDir: {}
```

Attempting to collect a dump from the target application will fail with an error message such as:

```
Access to the path '/diag/dumps/510aa00b-cfe1-4207-b965-90805b922b59_1' is denied.
```

The issue is that the dump is written by the runtime with root read-write permissions but .NET Monitor cannot read it because the 8.0 images run as non-root:

```bash
ls -la /diag/dumps/

drwxr-xr-x    2 64198    64198         4096 May 16 17:51 .
drwxrwxrwx    3 root     root          4096 May 16 17:43 ..
-rw-------    1 root     root     127311872 May 16 17:51 510aa00b-cfe1-4207-b965-90805b922b59_1
```

The only current workaround is to have the .NET Monitor container run as root:

```diff
      - name: monitor
        image: mcr.microsoft.com/dotnet/monitor:8.0.0-preview.4
        args: [ "collect", "--urls", "http://localhost:52323", "--metricUrls", "http://+:52325", "--no-auth" ]
[...]

PR #4436, opened 2023-05-17
Egress Extensibility - Updates to User-Facing Docs

kkeirstead:
###### Summary

These may be expanded in the future, but I wanted to get some initial coverage in our docs regarding the new model. Note that there are references to the learning path in this doc (which has not yet been merged) - if we don't want to link users there, I can also duplicate some of that information here, or have the learning path point to this documentation (instead of vice versa).

### DIFF
--- a/documentation/egress.md
+++ b/documentation/egress.md
@@ -25,16 +25,16 @@ In addition to the current `dotnet monitor` offerings, a `monitor-base` image is
 
 Users using the `monitor-base` image can manually install supported extensions via Multi-Stage Docker Builds, creating their own image that includes any desired egress providers.
 
-For an example of using Multi-Stage Docker Builds, see the [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/monitor/8.0/ubuntu-chiseled/amd64/Dockerfile) that the `dotnet monitor` team used to construct the `monitor` image.
+For an example of using Multi-Stage Docker Builds, see the [Dockerfile](https://github.com/dotnet/dotnet-docker/blob/nightly/src/monitor/8.0/ubuntu-chiseled/amd64/Dockerfile) that the `dotnet monitor` team uses to construct the `amd64` `monitor` image.
 
-To directly access archives for one of `dotnet monitor`'s supported extensions, these are available using the following link (this example is specifically for the `linux-x64` archive): `https://dotnetbuilds.azureedge.net/public/diagnostics/monitor/$dotnet_monitor_extension_version/dotnet-monitor-egress-azureblobstorage-$dotnet_monitor_extension_version-linux-x64.tar.gz`. Note: the versions in this link will change in response to new `dotnet monitor` releases; as a result, the link will need to be updated to reflect the most recent version when updating your extensions.
+To directly access archives for one of `dotnet monitor`'s supported extensions, these are available using the following link (this example is specifically for the `linux-x64` archive): `https://dotnetbuilds.azureedge.net/public/diagnostics/monitor/$dotnet_monitor_extension_version/dotnet-monitor-egress-azureblobstorage-$dotnet_monitor_extension_version-linux-x64.tar.gz`. Note: the versions in this link will change in response to new `dotnet monitor` releases; as a result, the link will need to be changed to reflect the most recent version when updating your extensions.
 
 #### Example Of Manually Installing AzureBlobStorage Extension Locally On Windows (Version Numbers May Vary)
 
 1. Download the archive: `curl -fSL --output dotnet-monitor-egress-azureblobstorage.tar.gz https://dotnetbuilds.azureedge.net/public/diagnostics/monitor/8.0.0-preview.4.23260.4/dotnet-monitor-egress-azureblobstorage-8.0.0-preview.4.23260.4-linux-x64.tar.gz`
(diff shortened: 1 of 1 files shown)


documentation/egress.md
```diff
@@ -15,6 +15,36 @@ Egress providers use [operations](./api/operations.md) to provide status.
 
 The `--no-http-egress` flag requires users to specify an egress provider by preventing the default HTTP response for logs, traces, dumps, gcdumps, and live metrics.
 
+## Egress Extensibility (8.0+)
+
+The `dotnet monitor` tool has transitioned to an extensible egress model. **This should not be a breaking change - by default, users migrating from `dotnet monitor` 6/7 should see no difference in `dotnet monitor`'s behavior**.
```
kkeirstead, line 20: Do we want to call out the known bug https://github.com/dotnet/dotnet-monitor/issues/4427? My only concern with doing that is forgetting to update the docs once that's resolved, but technically this statement is misleading at the moment in certain situations.

documentation/egress.md
```diff
@@ -15,6 +15,36 @@ Egress providers use [operations](./api/operations.md) to provide status.
...
+
+To directly access archives for one of `dotnet monitor`'s supported extensions, these are available using the following link (this example is specifically for the `linux-x64` archive): `https://dotnetbuilds.azureedge.net/public/diagnostics/monitor/$dotnet_monitor_extension_version/dotnet-monitor-egress-azureblobstorage-$dotnet_monitor_extension_version-linux-x64.tar.gz`. Note: the versions in this link will change in response to new `dotnet monitor` releases; as a result, the link will need to be changed to reflect the most recent version when updating your extensions.
+
+#### Example Of Manually Installing AzureBlobStorage Extension Locally On Windows (Version Numbers May Vary)
+
+1. Download the archive: `curl -fSL --output dotnet-monitor-egress-azureblobstorage.tar.gz https://dotnetbuilds.azureedge.net/public/diagnostics/monitor/8.0.0-preview.4.23260.4/dotnet-monitor-egress-azureblobstorage-8.0.0-preview.4.23260.4-linux-x64.tar.gz`
```
kkeirstead, line 34: If there's a different flow we'd like to represent here, I can update this

documentation/egress.md
wiktork: Egress extensibility didn't introduce this issue though. I don't think it should be mentioned here.

documentation/egress.md
wiktork: The install locally on windows part of this is very confusing, given the linux payload and tar.gz files. Perhaps we should demonstrate this fully on linux or wsl2?

documentation/egress.md
```diff
@@ -15,6 +15,37 @@ Egress providers use [operations](./api/operations.md) to provide status.
 
 The `--no-http-egress` flag requires users to specify an egress provider by preventing the default HTTP response for logs, traces, dumps, gcdumps, and live metrics.
 
+## Egress Extensibility (8.0+)
```
jander-msft, line 18: I think this should go to the bottom of the page, below the `Examples of Egressing a dump to blob storage` section; concerns about egress extensibility will be a very small slice of .NET Monitor users and we shouldn't force everyone to attempt to understand it and then realize they generally don't need to know about it.

documentation/egress.md
```diff
@@ -15,6 +15,37 @@ Egress providers use [operations](./api/operations.md) to provide status.
...
+
+The `dotnet monitor` tool has transitioned to an extensible egress model. **This should not be a breaking change - by default, users migrating from `dotnet monitor` 6/7 should see no difference in `dotnet monitor`'s behavior**.
```
jander-msft, line 20: > This should not be a breaking change - by default, users migrating from `dotnet monitor` 6/7 should see no difference in `dotnet monitor`'s behavior

This feels too broad of a statement. If you are using a docker image, then migrating from `mcr.microsoft.com/dotnet/monitor:7` to `mcr.microsoft.com/dotnet/monitor:8` can be seen as a behavior difference because the tool will be running as a non-root user whereas it ran as root previously.

Also, the statement that there is an extensibility model doesn't imply anything about whether the prior egress providers exist or are missing from certain packages.

Maybe state something more like:

```
Starting with `dotnet monitor` 8, the tool includes an egress extensibility model that allows additional egress providers to be discovered and usable by a `dotnet monitor` installation. The existing `AzureBlobStorage` egress provider has been moved to this model and remains as an available egress provider in the .NET Tool and `mcr.microsoft.com/dotnet/monitor` image offerings.
```

kkeirstead: In reference to the first part: https://github.com/dotnet/dotnet-monitor/pull/4436#discussion_r1196559127 - I think that will need to be called out, I think it's just a question of where (which may very well be here).

jander-msft: > In reference to the first part: [#4436 (comment)](https://github.com/dotnet/dotnet-monitor/pull/4436#discussion_r1196559127) - I think that will need to be called out, I think it's just a question of where (which may very well be here).

The behavior differences between the 6/7 vs 8 images have nothing to do with egress extensibility. Calling it out here is muddying the separation of concerns. It should be called out in a breaking change document (that has been created yet for 8.0).

kkeirstead: In which case for this PR - do you want me to omit the statement about breaking changes, or make a breaking change doc and add a reference to it?

jander-msft (approved):

PR #4471, opened 2023-05-23
[Exceptions] Report Stack Frame Information

kkeirstead:
###### Summary

Builds on top of @jander-msft 's Exceptions work to display stack frame information in JSON/plaintext. Includes testing for top stack frame.


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsPipelineTests.cs
```diff
@@ -256,6 +289,15 @@ public Task EventExceptionsPipeline_ArrayException()
             });
         }
 
+        private static void ValidateStack(TestExceptionsStore.ExceptionInstance instance, ulong expectedFunctionId, ulong expectedOffset)
+        {
+            CallStack stack = Assert.Single(instance.CallStackResult.Stacks);
+            Assert.NotEmpty(stack.Frames);
+            Assert.True(0 < stack.ThreadId);
+            Assert.Equal(expectedFunctionId, stack.Frames[0].FunctionId);
+            Assert.Equal(expectedOffset, stack.Frames[0].Offset);
```
kkeirstead, line 298: Looks like offsets aren't consistent - may need to remove this check.

src/Tools/dotnet-monitor/Exceptions/ExceptionsStore.cs
```diff
@@ -104,24 +104,74 @@ private async Task ProcessEntriesAsync(CancellationToken token)
...
+                        if (cache.TryGetModuleId(functionData.ModuleId, out ModuleData moduleData))
+                        {
+                            callStackResult.NameCache.ModuleData.TryAdd(functionData.ModuleId, moduleData);
+                        }
+                    }
+
+                    CallStackFrame frame = new()
+                    {
+                        FunctionId = methodId,
+                        Offset = (ulong)ilOffset // is this safe?
```
wiktork, line 153: IIRC iloffset as reported by the stacks api is an absolute memory location (hence the ulong). I believe you are reporting a relative to function iloffset here correct?

jander-msft: Correct, this is the relative managed ILOffset within the method body; I don't think we have access to the native IL Offset.

src/Microsoft.Diagnostics.Monitoring.WebApi/Exceptions/IExceptionInstance.cs
```diff
@@ -14,5 +15,7 @@ internal interface IExceptionInstance
         string TypeName { get; }
 
         DateTime Timestamp { get; }
+
+        CallStackResult CallStackResult { get; }
```
jander-msft, line 19: This should be the `CallStackResult` from `Microsoft.Diagnostics.Monitoring.WebApi.Models` not from `Microsoft.Diagnostics.Monitoring.WebApi.Stacks`. The `IExceptionInstance` needs to be reporting realized information instead of numeric identifiers e.g. the stack frame should have the function name already resolved instead of providing a `FunctionId` property.

src/Tools/dotnet-monitor/Exceptions/EventExceptionsPipelineNameCache.cs
```diff
@@ -63,8 +63,70 @@ public bool TryGetExceptionId(ulong exceptionId, out ulong exceptionClassId, out
             return true;
         }
 
+        public bool TryGetClassId(ulong classId, out ClassData data)
```
jander-msft, line 66: All of these new methods don't really provide much value because you can already do the same thing from the already publicly available `NameCache` property.

src/Tools/dotnet-monitor/Exceptions/ExceptionsOperation.cs
```diff
@@ -126,6 +138,25 @@ private static async Task WriteTextInstance(Stream stream, IExceptionInstance in
...
+
+                    await writer.WriteLineAsync(
+                        string.Format(
+                            CultureInfo.InvariantCulture,
+                            Strings.OutputFormatString_FirstChanceExceptionStackFrame,
+                            frameModel.ClassName,
+                            frameModel.MethodName));
+                }
+
+                writer.WriteLine();
```
jander-msft, line 157: If this is to provide an empty newline between exception instances, then this should be outside of the if condition.

src/Microsoft.Diagnostics.Monitoring.WebApi/Exceptions/IExceptionInstance.cs
kkeirstead: I can do this, but doesn't the /stacks endpoint only do this translation when it's being formatted? I reused the logic being used there by doing the translation in `FormatStack`. Or do we want this to behave differently than that?

src/Tools/dotnet-monitor/Exceptions/ExceptionsStore.cs
```diff
@@ -104,24 +105,85 @@ private async Task ProcessEntriesAsync(CancellationToken token)
...
             }
         }
 
+        internal static CallStackResultModel GenerateCallStackResult(ulong[] stackFrameIds, IExceptionsNameCache cache, int threadId)
+        {
+            CallStackResult callStackResult = new CallStackResult();
+            CallStack callStack = new();
+            callStack.ThreadId = (uint)threadId;
+
+            foreach (var stackFrameId in stackFrameIds)
```
jander-msft, line 126: Rather than iterating the name cache and inserting the data into another name cache, could you add another constructor to [CallStackData](https://github.com/dotnet/dotnet-monitor/blob/1e1bd44c6a049380b5c60cdcdca5a825858c1176/src/Microsoft.Diagnostics.Monitoring.WebApi/Stacks/CallStackData.cs#L12) that takes an existing NameCache? We're going to throw away this CallStackResult immediately after converting it to the model, so there shouldn't be any lifetime issues.

src/Tools/dotnet-monitor/Exceptions/ExceptionsStore.cs
```diff
@@ -104,24 +105,85 @@ private async Task ProcessEntriesAsync(CancellationToken token)
...
-                        _instances.Add(new ExceptionInstance(exceptionTypeName, moduleName, entry.Message, entry.Timestamp));
+                        _instances.Add(new ExceptionInstance(exceptionTypeName, moduleName, entry.Message, entry.Timestamp, callStackResult));
                     }
                 }
 
                 shouldReadEntry = await _channel.Reader.WaitToReadAsync(token);
             }
         }
 
+        internal static CallStackResultModel GenerateCallStackResult(ulong[] stackFrameIds, IExceptionsNameCache cache, int threadId)
```
jander-msft, line 120: Why not return [CallStack](https://github.com/dotnet/dotnet-monitor/blob/main/src/Microsoft.Diagnostics.Monitoring.WebApi/Models/CallStackResults.cs#L25) instead of CallStackResult? A thrown exception will only have a single call stack.

src/Microsoft.Diagnostics.Monitoring.WebApi/Exceptions/IExceptionInstance.cs
jander-msft, line 19: I think it makes more sense to have this be [CallStack](https://github.com/dotnet/dotnet-monitor/blob/main/src/Microsoft.Diagnostics.Monitoring.WebApi/Models/CallStackResults.cs#L25) instead of CallStackResult since there is only ever one call stack.

src/Tools/dotnet-monitor/Exceptions/ExceptionsOperation.cs
```diff
@@ -96,6 +98,14 @@ await using (Utf8JsonWriter writer = new(stream, new JsonWriterOptions() { Inden
                 writer.WriteString("typeName", instance.TypeName);
                 writer.WriteString("moduleName", instance.ModuleName);
                 writer.WriteString("message", instance.Message);
+
+                MemoryStream outputStream = new();
+                await JsonSerializer.SerializeAsync(outputStream, instance.CallStackResult, cancellationToken: token);
+                outputStream.Position = 0;
+
+                writer.WritePropertyName("callStack");
+                writer.WriteRawValue(new StreamReader(outputStream).ReadToEnd());
```
jander-msft, line 107: Won't this write the entire call stack payload on a single line without white space? That's not desirable here.

I understand that we use `JsonSerializer` [here](https://github.com/dotnet/dotnet-monitor/blob/main/src/Microsoft.Diagnostics.Monitoring.WebApi/Stacks/JsonStacksFormatter.cs#L69) but we should probably use the existing `Utf8JsonWriter` instead.

You can at least share the property names since those are declared as constants on the model attributes, however it might be best to change the stacks feature to use `Utf8JsonWriter` instead of `JsonSerializer`, create a common writing method, and use the method in both features.

src/Microsoft.Diagnostics.Monitoring.WebApi/Stacks/CallStackData.cs
```diff
@@ -11,6 +11,13 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Stacks
     /// </summary>
     internal sealed class CallStackResult
     {
+        public CallStackResult(NameCache nameCache)
+        {
+            NameCache = nameCache;
+        }
+
+        public CallStackResult() { }
```
jander-msft, line 19:
```suggestion
        public CallStackResult() : this(new NameCache()) { }
```

and remove the property value initializer for `NameCache`.

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/StackUtilities.cs
```diff
@@ -19,6 +20,66 @@ internal enum StackFormat
 
     internal static class StackUtilities
     {
+        public static Models.CallStackResult TranslateCallStackResultToModel(CallStackResult result)
```
jander-msft, line 23: Change this to be `Models.CallStack TranslateCallStackToModel(CallStack stack)` and update `JsonStacksFormatter.FormatStack` to use this method too.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsPipelineTests.cs
```diff
@@ -285,17 +312,20 @@ public TestExceptionsStore(int instanceThreshold = 1)
...
                 StringBuilder typeBuilder = new();
-                FunctionData throwingMethodData;
+                WebApi.Stacks.FunctionData throwingMethodData;
+                CallStack callStack;
                 try
                 {
                     Assert.True(cache.TryGetExceptionGroup(groupId, out ulong exceptionClassId, out ulong throwingMethodId, out _));
 
-                    NameFormatter.BuildClassName(typeBuilder, cache.NameCache, exceptionClassId);
+                    WebApi.Stacks.NameFormatter.BuildClassName(typeBuilder, cache.NameCache, exceptionClassId);
```
jander-msft, line 324: Is the namespace qualification necessary here?

src/Tools/dotnet-monitor/Exceptions/ExceptionsOperation.cs
```diff
@@ -96,7 +97,27 @@ await using (Utf8JsonWriter writer = new(stream, new JsonWriterOptions() { Inden
...
                 writer.WriteString("message", instance.Message);
-                writer.WriteEndObject();
+
+                writer.WriteStartObject("callStack");
+                writer.WriteNumber("threadId", instance.CallStack.ThreadId);
+                writer.WriteString("threadName", instance.CallStack.ThreadName);
+
+                writer.WriteStartArray("frames");
+
+                foreach (var frame in  instance.CallStack.Frames)
```
jander-msft, line 107:
```suggestion
                foreach (var frame in instance.CallStack.Frames)
```

kkeirstead:
It appears that something broke when I manually merged the latest changes from release/7.x - I haven't had a chance to debug into it yet, but I'll plan to push another update with that fix tomorrow.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsPipelineTests.cs
kkeirstead: I believe so, due to the ambiguity between `WebApi.Stacks` and `WebApi.Models` for some things (like `CallStack`).

jander-msft: But there's only one `NameFormatter` class [here](https://github.com/dotnet/dotnet-monitor/blob/main/src/Microsoft.Diagnostics.Monitoring.WebApi/Stacks/NameFormatter.cs#L15); the namespace only changes if it is code in the startup hook assembly, which this is not.

kkeirstead: I just meant that we can't include `WebApi.Stacks` and `WebApi.Models` without needing to qualify something because of the ambiguity around `CallStack` - I can add a `using CallStack = Microsoft.Diagnostics.Monitoring.WebApi.Models.CallStack` to address that though.

kkeirstead:
> It appears that something broke when I manually merged the latest changes from release/7.x - I haven't had a chance to debug into it yet, but I'll plan to push another update with that fix tomorrow.

Unrelated to these changes - resolved by https://github.com/dotnet/dotnet-monitor/pull/4648

src/Microsoft.Diagnostics.Monitoring.WebApi/Stacks/CallStackData.cs
kkeirstead: I ended up just reverting this, since we no longer need the constructor if we take https://github.com/dotnet/dotnet-monitor/pull/4471#discussion_r1228775857

PR #4515, opened 2023-06-02
[Parameter Capturing] Add functional tests for function probes

schmittjoseph:
###### Summary

Add functional tests for the function probes infrastructure. The log emitting probes and `ParameterCapturingService` will be tested separately.

Overview of changes:
1. Refactor function probes management into its own class (`FunctionProbesManager`) so that test code can easily leverage it without the full `ParameterCapturingService`. The **only** functional difference is that the `InstrumentedMethodCache` instance is now static and belongs to `FunctionProbesStub`.
2. Add a new test scenario for function probes. Each command represents an end-to-end test case.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/FunctionProbesTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/FunctionProbesTests.cs
@@ -66,10 +66,7 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests
                 _httpClientFactory,
                 DiagnosticPortConnectionMode.Listen,
                 TestAppScenarios.FunctionProbes.Name,
-                appValidate: async (runner, client) =>
-                {
-                    await AppRunnerExtensions.SendCommandAsync(runner, TestAppScenarios.FunctionProbes.Commands.RunTest);
-                },
+                appValidate: (runner, client) => { return Task.CompletedTask; },
                 configureApp: runner =>
                 {
                     runner.Architecture = targetArchitecture;
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/FunctionProbes/FunctionProbesScenario.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/FunctionProbes/FunctionProbesScenario.cs
@@ -53,8 +53,6 @@ namespace Microsoft.Diagnostics.Monitoring.UnitTestApp.Scenarios.FunctionProbes
                 {
                     return ScenarioHelpers.RunScenarioAsync(async logger =>
                     {
-                        await ScenarioHelpers.WaitForCommandAsync(TestAppScenarios.FunctionProbes.Commands.RunTest, logger);
-
                         PerFunctionProbeProxy probeProxy = new PerFunctionProbeProxy();
                         using FunctionProbesManager probeManager = new(probeProxy);
 
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/TestAppScenarios.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/TestAppScenarios.cs
@@ -88,11 +88,6 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon
                 public const string GenericMethods = nameof(GenericMethods);
                 public const string ExceptionRegionAtBeginningOfMethod = nameof(ExceptionRegionAtBeginningOfMethod);
             }
-
-            public static class Commands
-            {
-                public const string RunTest = nameof(RunTest);
-            }
         }
 
         public static class Stacks


src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/FunctionProbes/FunctionProbesScenario.cs
```diff
@@ -0,0 +1,326 @@
...
+        public static CliCommand Command()
+        {
+            CliCommand command = new(TestAppScenarios.FunctionProbes.Name);
+            command.SetAction(ExecuteAsync);
+            return command;
+        }
+
+        public static Task<int> ExecuteAsync(ParseResult result, CancellationToken token)
+        {
+            Dictionary<string, TestCaseAsync> testCases = new()
```
jander-msft, line 32: Rather than treating these as commands, I think it makes more sense to make them subscenarios e.g. https://github.com/dotnet/dotnet-monitor/blob/main/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/ExceptionsScenario.cs#L15

These don't seem like they are for orchestration or used multiple times per test app process but represent distinct scenarios that you are trying to test.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/FunctionProbes/FunctionProbesScenario.cs
```diff
@@ -0,0 +1,326 @@
...
+
+            probeProxy.RegisterPerFunctionProbe(installationTestMethod, (object[] args) => { });
+
+            List<MethodInfo> methodsToCapture = new(methods.Count + 2)
+            {
+                installationTestMethod,
+                uninstallationTestMethod
+            };
+            methodsToCapture.AddRange(methods);
+            probeManager.StartCapturing(methodsToCapture);
```
jander-msft, line 303: Doesn't have to be done in this PR...

I'd like to see a callback/synchronization mechanism here so that way we can definitively tell when the probes have been registered. As it is, this effective tosses the methods over a fence and the caller never gets to know "were all of the methods processed?", "did some failure occur?", etc. Adding this mechanism allows the caller to choose to observe failures and to choose to wait for the probes to be inserted before continuing.

This would then alleviate the need to effectively poll for when the probes were actually inserted in this test and then you could probably remove the need for the `InstallationTestStub` and `UninstallationTestStub` methods.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/FunctionProbes/FunctionProbesScenario.cs
```diff
@@ -0,0 +1,326 @@
...
+        {
+            return RunTestCaseAsync(probeManager, probeProxy, method, args, thisObj, thisObj, token);
+        }
+
+        private static Task RunTestCaseAsync(FunctionProbesManager probeManager, PerFunctionProbeProxy probeProxy, MethodInfo method, object[] args, CancellationToken token)
+        {
+            return RunTestCaseAsync(probeManager, probeProxy, method, args, null, null, token);
+        }
+
+        private static Task RunAsyncTestCaseAsync(FunctionProbesManager probeManager, PerFunctionProbeProxy probeProxy, MethodInfo method, object[] args, object thisObj, object capturedThisObj, CancellationToken token)
```
jander-msft, line 220: This method and `RunTestCaseAsync` seem to only differ in whether the `MethodInfo` has a `Task` return type or not. Could you combine these, check `method.ReturnType.IsAssignableTo(typeof(Task))`, and then conditionally return its return value or `Task.CompletedTask`?

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/FunctionProbes/FunctionProbesScenario.cs
```diff
@@ -0,0 +1,326 @@
...
+                return Task.CompletedTask;
+            },
+            new object[]
+            {
+                null
+            },
+            hasImplicitThis: false, capturedThisObj: null, token);
+        }
+
+        private static Task RunAsyncTestCaseAsync(FunctionProbesManager probeManager, PerFunctionProbeProxy probeProxy, MethodInfo method, object[] args, object thisObj, CancellationToken token)
```
jander-msft, line 200: The two overloads only differ in whether the `MethodInfo` is a static method or not. Could you rename these methods to be more clear as to which requires a "this" parameters e.g. `RunStaticMethodTestCaseAsync` and `RunInstanceMethodTestCaseAsync`? Also, please validate that `method.IsStatic` is correct for each method to help test authors know when they used the wrong test case runner with their method.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/FunctionProbes/FunctionProbesScenario.cs
```diff
@@ -0,0 +1,326 @@
...
+        {
+            return RunAsyncTestCaseAsync(probeManager, probeProxy, method, args, thisObj, thisObj, token);
+        }
+
+        private static Task RunAsyncTestCaseAsync(FunctionProbesManager probeManager, PerFunctionProbeProxy probeProxy, MethodInfo method, object[] args, CancellationToken token)
+        {
+            return RunAsyncTestCaseAsync(probeManager, probeProxy, method, args, null, null, token);
+        }
+
+        private static Task RunTestCaseAsync(FunctionProbesManager probeManager, PerFunctionProbeProxy probeProxy, MethodInfo method, object[] args, object thisObj, CancellationToken token)
```
jander-msft, line 210: Same as https://github.com/dotnet/dotnet-monitor/pull/4515/files#r1223315252

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/FunctionProbesTests.cs
```diff
@@ -0,0 +1,85 @@
...
+        public async Task RunTestScenario(Architecture targetArchitecture, string subScenario)
+        {
+            await ScenarioRunner.SingleTarget(
+                _outputHelper,
+                _httpClientFactory,
+                DiagnosticPortConnectionMode.Listen,
+                TestAppScenarios.FunctionProbes.Name,
+                appValidate: async (runner, client) =>
+                {
+                    await AppRunnerExtensions.SendCommandAsync(runner, TestAppScenarios.FunctionProbes.Commands.RunTest);
```
jander-msft, line 71: Is this command necessary? It seems nothing is being coordinated and all these tests care about is to run the app scenario to completion.

jander-msft (approved):

PR #4568, opened 2023-06-08
[Parameter Capturing] Proposal for OOM handling in profiler

schmittjoseph:
###### Summary

This PR is a proposal for how we may want to handle OOM scenarios in our profiler. There's currently a mixture of either throwing `bad_alloc` (and not catching it) or gracefully returning `E_OUTOFMEMORY`.

In this PR, any function that's marked as returning an `HRESULT` will return `E_OUTOFMEMORY` instead of throwing `bad_alloc` during OOM scenarios. I've made changes to parameter-capturing specific code to showcase this, but there's several other places in our profiler code we'd also want to standardize.

### DIFF
--- a/src/MonitorProfiler/ProbeInstrumentation/ProbeInstrumentation.cpp
+++ b/src/MonitorProfiler/ProbeInstrumentation/ProbeInstrumentation.cpp
@@ -4,10 +4,17 @@
 #include "corhlpr.h"
 #include "macros.h"
 #include "ProbeInstrumentation.h"
+#include "../Utilities/BlockingQueue.h"
 
 using namespace std;
 
 #define IfFailLogRet(EXPR) IfFailLogRet_(m_pLogger, EXPR)
+
+#ifndef DLLEXPORT
+#define DLLEXPORT
+#endif
+
+BlockingQueue<PROBE_WORKER_PAYLOAD> g_probeManagementQueue;
 
 ProbeInstrumentation::ProbeInstrumentation(const shared_ptr<ILogger>& logger, ICorProfilerInfo12* profilerInfo) :
     m_pCorProfilerInfo(profilerInfo),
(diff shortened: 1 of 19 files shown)


src/MonitorProfiler/Utilities/OomUtilities.h
```diff
@@ -0,0 +1,8 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+#pragma once
+
+#define START_NO_OOM_THROW_REGION try {
+#define END_NO_OOM_THROW_REGION } catch (const std::bad_alloc&) { return E_OUTOFMEMORY; }
+#define ReturnHResultIfOom(exp) START_NO_OOM_THROW_REGION; exp; END_NO_OOM_THROW_REGION;
```
jander-msft, line 8: The typical pattern we use for these type of macros is "If(Result)Ret" or "If(Result)Ret(Value)"; I would suggest `IfOomRetMem`.

src/MonitorProfiler/Utilities/OomUtilities.h
```diff
@@ -0,0 +1,8 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+#pragma once
```
jander-msft, line 4: Move all of these to https://github.com/dotnet/dotnet-monitor/blob/main/src/inc/macros.h; these feel common enough that they don't need their own special header file.

src/MonitorProfiler/ProbeInstrumentation/ProbeInstrumentation.cpp
```diff
@@ -189,7 +195,8 @@ HRESULT ProbeInstrumentation::InstallProbes(vector<UNPROCESSED_INSTRUMENTATION_R
     vector<ModuleID> requestedModuleIds;
```
jander-msft, line 195: I know that this is stack allocated, but the ctors may heap allocated? Do we need to guard OOM from those too?

PR #4578, opened 2023-06-08
[Parameter Capturing] Protect against faulting probes

schmittjoseph:
###### Summary

This PR protects instrumented functions against faulting probes. If an installed probe throws an exception, it is now caught and a callback into the profiler is subsequently invoked, notifying us of the fault and uninstalling the probes.

**Note:**
We are trying to minimize diffs in `ILRewriter.{cpp,h}` to make keeping it in-sync with the runtime's copy easier, so the changes in these files are specifically written to minimize diffs as opposed to being the cleanest option.

### DIFF
--- a/src/MonitorProfiler/ProbeInstrumentation/ProbeInjector.cpp
+++ b/src/MonitorProfiler/ProbeInstrumentation/ProbeInjector.cpp
@@ -86,7 +86,7 @@ HRESULT ProbeInjector::InstallProbe(
 
     // START: Try block
 
-     /* uniquifier */
+    /* uniquifier */
     pTryBegin = rewriter.NewILInstr();
     pTryBegin->m_opcode = CEE_LDC_I8;
     pTryBegin->m_Arg64 = request.uniquifier;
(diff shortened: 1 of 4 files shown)


src/MonitorProfiler/ProbeInstrumentation/ProbeInjector.cpp
```diff
@@ -138,6 +157,70 @@ HRESULT ProbeInjector::InstallProbe(
...
+    pNewInstr->m_opcode = CEE_CALLI;
+    pNewInstr->m_Arg32 = request.pAssemblyData->GetFaultingProbeCallbackSignature();
+    rewriter.InsertBefore(pInsertProbeBeforeThisInstr, pNewInstr);
+
+    pNestedTryLeave = rewriter.NewILInstr();
+    pNestedTryLeave->m_opcode = CEE_LEAVE;
+    rewriter.InsertBefore(pInsertProbeBeforeThisInstr, pNestedTryLeave);
+
+    // END: Try block (nested)
+    // START: Catch block (nested)
```
kkeirstead, line 194: nit: Just for consistency with the comments, do you want to add in `END: Catch block (nested)` and `END: Catch block`

jander-msft (dismissed):

jander-msft:
Can you write a comment somewhere that describes the faulting behavior (that all probes are removed in the event that any probe faults).

jander-msft (approved):

PR #4596, opened 2023-06-09
[Parameter Capturing] Increase profiler logging and capture it during tests

schmittjoseph:
###### Summary

Currently some of the parameter capturing tests are flaky on MacOS, add additional logs statements in the profiler to help triage. Also update the tests to capture the logs, as well as capture error-level profiler logs for **all** tests.

### DIFF
--- a/src/MonitorProfiler/ProbeInstrumentation/ProbeInstrumentation.cpp
+++ b/src/MonitorProfiler/ProbeInstrumentation/ProbeInstrumentation.cpp
@@ -247,7 +247,7 @@ HRESULT ProbeInstrumentation::InstallProbes(vector<UNPROCESSED_INSTRUMENTATION_R
     }
 
     IfFailLogRet(m_pCorProfilerInfo->RequestReJITWithInliners(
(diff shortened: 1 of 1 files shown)


src/MonitorProfiler/ProbeInstrumentation/ProbeInstrumentation.cpp
```diff
@@ -306,12 +309,15 @@ HRESULT STDMETHODCALLTYPE ProbeInstrumentation::GetReJITParameters(ModuleID modu
 {
     HRESULT hr;
 
+    m_pLogger->Log(LogLevel::Trace, _LS("ReJIT - moduleId: 0x%08x, methodDef: 0x%04x"), moduleId, methodDef);
```
jander-msft, line 312: For any log statement that is in a hot path, especially uncoditioned ones, you should check if it's enabled for the level at which you want to log e.g. `if (m_pLogger->IsEnabled(LogLevel::Trace))`

jander-msft (dismissed):

jander-msft (approved):

PR #4597, opened 2023-06-09
Add exception instance identifiers to exceptions event source

jander-msft:
###### Summary

Add payload fields for exception identifiers:
- Each exception instance _shall_ have an identifier
- Each exception instance _may_ have a single inner exception
- Each exception instance _may_ have a list of multiple inner exceptions (e.g. AggregateException.InnerExceptions)

This change only updates the event source to produce this information. Future changes will fill in the production of real exception identifier information and consume it within dotnet-monitor.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/Exceptions/Eventing/ExceptionsEventSourceTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/Exceptions/Eventing/ExceptionsEventSourceTests.cs
@@ -64,12 +64,12 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Eventing
         }
 
         [Theory]
-        [InlineData(0, null, "0,0,0")]
-        [InlineData(1, "", "1,2")]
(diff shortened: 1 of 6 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/Exceptions/Eventing/ExceptionsEventSourceTests.cs
```diff
@@ -78,16 +78,22 @@ public void ExceptionsEventSource_WriteException_Event(ulong id, string message,
 
             ulong[] frameIds = frameIdsString.Split(',', StringSplitOptions.RemoveEmptyEntries).Select(ulong.Parse).ToArray();
             DateTime timestamp = DateTime.UtcNow;
+            ulong[] innerExceptionIds = innerExceptionIdsString.Split(',', StringSplitOptions.RemoveEmptyEntries).Select(ulong.Parse).ToArray();
```
kkeirstead, line 81: Is this being used?

jander-msft: Good catch. I missed passing it in on line 83.

PR #4649, opened 2023-06-14
Capture inner exceptions in startup hook

jander-msft:
###### Summary

These changes add a new step to the exceptions processor list that demultiplexes the `Exception` typed parts of the given `Exception` instance (specifically, it will report the `InnerException` property value, the `AggregateException.InnerExceptions` property value, and the exception itself). This allows for reporting of inner exceptions through the rest of the pipeline and eventually eventing the inner exception information.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
@@ -66,7 +66,7 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Pipeline.Steps
             StackTrace exceptionStackTrace = new(exception, fNeedFileInfo: false);
 
             // Inner exceptions have complete call stacks because they were caught or do not
-            // have an call stacks because they were never thrown. Report the stack frames as-is.
+            // have any call stacks because they were never thrown. Report the stack frames as-is.
             if (isInnerException)
                 return exceptionStackTrace.GetFrames();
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
```diff
@@ -61,15 +61,21 @@ public void Invoke(Exception exception, ExceptionPipelineExceptionContext contex
             _next(exception, context);
         }
 
-        private static ReadOnlySpan<StackFrame> ComputeEffectiveCallStack(Exception exception)
+        private static ReadOnlySpan<StackFrame> ComputeEffectiveCallStack(Exception exception, bool isInnerException)
         {
+            StackTrace exceptionStackTrace = new(exception, fNeedFileInfo: false);
+
+            // Inner exceptions have complete call stacks because they were caught or do not
+            // have an call stacks because they were never thrown. Report the stack frames as-is.
```
kkeirstead, line 69:
```suggestion
            // have any call stacks because they were never thrown. Report the stack frames as-is.
```

src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/Exceptions/Pipeline/Steps/ExceptionDemultiplexerPipelineStepTests.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Collections.Generic;
+using System;
+using Xunit;
+
+namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Pipeline.Steps
+{
+    [TargetFrameworkMonikerTrait(TargetFrameworkMonikerExtensions.CurrentTargetFrameworkMoniker)]
+    public sealed class ExceptionDemultiplexerPipelineStepTests
+    {
+        private readonly Dictionary<Exception, int> _reportedExceptionCounts = new();
```
kkeirstead, line 14: Right now, it looks like we only ever have a single instance of each exception, which makes this feels overengineered - I'm assuming in the future this mechanism will be leveraged more

kkeirstead (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/Exceptions/Pipeline/Steps/ExceptionDemultiplexerPipelineStepTests.cs
jander-msft: I was potentially thinking of adding more tests that would flex this a bit more, but decided against it for now. The way it is allows us to determine the exact count of each type of exception that is reported rather than failing the step's invocation immediately.

PR #4650, opened 2023-06-14
Generate unique ID for each exception instance

jander-msft:
###### Summary

Generate unique ID for each exception instance (that doesn't have an ID already assigned to it). This allows for correlation of exceptions to inner exceptions within .NET Monitor (not yet implemented).

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
@@ -139,7 +139,7 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Pipeline.Steps
                     if (exception.Data[ExceptionIdKey] is ulong exceptionIdCandidate)
                     {
                         exceptionId = exceptionIdCandidate;
-                        return false;
+                        return true;
                     }
                 }
 
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
@@ -6,14 +6,19 @@ using Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Identification;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Threading;
 
 namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Pipeline.Steps
 {
     internal sealed class ExceptionEventsPipelineStep
     {
+        private static readonly object ExceptionIdKey = new object();
+
         private readonly ExceptionsEventSource _eventSource = new();
         private readonly ExceptionGroupIdentifierCache _identifierCache;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
```diff
@@ -105,5 +110,46 @@ private static ReadOnlySpan<StackFrame> ComputeEffectiveCallStack(Exception exce
...
+            if (exception.Data.Contains(ExceptionIdKey))
+            {
+                exceptionId = (ulong?)exception.Data[ExceptionIdKey];
+            }
+
+            if (!exceptionId.HasValue)
+            {
+                lock (exception.Data)
+                {
+                    if (!exceptionId.HasValue)
```
kkeirstead, line 129: Is this intentionally checking against the same thing as L125?

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
```diff
@@ -105,5 +110,46 @@ private static ReadOnlySpan<StackFrame> ComputeEffectiveCallStack(Exception exce
...
+                    {
+                        exceptionId = Interlocked.Increment(ref _nextExceptionId);
```
kkeirstead, line 131: Given that this starts at 1, if I'm reading this correctly the initial `exceptionId` will be 2 - is that the desired behavior, or do we not really care?

jander-msft: It'll actually create 2 as the first identifier, but it also doesn't really matter. We use this pattern everywhere in the exceptions capturing code. We can fix that separately.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
jander-msft: This is wrong, I'll go fix it. It needs to get the value from the Data property again.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
```diff
@@ -105,5 +110,56 @@ private static ReadOnlySpan<StackFrame> ComputeEffectiveCallStack(Exception exce
...
+
+            static bool TryGetExceptionId(Exception exception, out ulong exceptionId)
+            {
+                if (exception.Data.Contains(ExceptionIdKey))
+                {
+                    // The ExceptionIdKey data should only ever have a ulong
+                    if (exception.Data[ExceptionIdKey] is ulong exceptionIdCandidate)
+                    {
+                        exceptionId = exceptionIdCandidate;
+                        return false;
```
kkeirstead, line 142: I'm assuming this should probably be true?

jander-msft: (smacks head) yes

kkeirstead (approved):

PR #4657, opened 2023-06-15
Publish single file bundle to blob storage

jander-msft:
###### Summary

Theses changes allow the publishing of the single file bundles to blob storage accounts and have them marked as non-shipping. It piggybacks of the archive job to copy the built bundles from the `pub` directory to the `artifacts/bundles/Release/NonShipping/<RID>/` directory from which the Publishing.props file will detect it and write entries into BAR manifest as appropriate. Because these files are in the `pub` directory during the signing job, they are also getting signed.

Example build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2201872&view=results

An example blob storage URI for the win-x64 single file bundle (if the above build was actually uploading them) would be `https://dotnetbuilds.blob.core.windows.net/public/diagnostics/monitor/7.3.0-alpha.1.23315.4/win-x64/dotnet-monitor.exe`

### DIFF
--- a/eng/AfterSolutionBuild.targets
+++ b/eng/AfterSolutionBuild.targets
@@ -7,7 +7,7 @@
   <!-- Only publish projects after build if opt-in -->
   <Target Name="PublishProjectsAfterBuild"
           AfterTargets="Build"
-          Condition="'$(PublishProjectsAfterBuild)' == 'true' and '$(CreateArchives)' != 'true'">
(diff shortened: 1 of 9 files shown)


jander-msft:
This change makes the publish job fail likely due to it trying to upload the same named file into the same artifacts container in the AzDO pipeline. Will close this until a new solution that works around that is in place.

PR #4660, opened 2023-06-16
Publish single file bundle to blob storage

jander-msft:
###### Summary

Theses changes allow the publishing of the single file bundles to blob storage accounts and have them marked as non-shipping. It piggybacks of the archive job to copy the built bundles from the `pub` directory to the `artifacts/bundles/Release/NonShipping` directory from which the Publishing.props file will detect it and write entries into BAR manifest as appropriate. Because these files are in the `pub` directory during the signing job, they are also getting signed. The bundles in the `artifacts/bundles/Release/NonShipping` directory are named as `<BundleName>-<RID><BundleExtension>` in order to avoid collisions in the directory and when added to the `BlobArtifacts` artifact in the Azure Pipeline run.

Example build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2201872&view=results

An example blob storage URI for the win-x64 single file bundle (if the above build was actually uploading them) would be `https://dotnetbuilds.blob.core.windows.net/public/diagnostics/monitor/7.3.0-alpha.1.23315.5/dotnet-monitor-win-x64.exe`

### DIFF
--- a/src/singlefile/Directory.Build.targets
+++ b/src/singlefile/Directory.Build.targets
@@ -1,0 +1,24 @@
+<Project>
+  <Import Project="$(MSBuildThisFileDirectory)..\Directory.Build.targets" />
+
+  <PropertyGroup>
+    <BuildDependsOn>$(BuildDependsOn);PublishProjectsBeforeBundle;CopyBundle</BuildDependsOn>
+  </PropertyGroup>
+
+  <Import Project="$(RepositoryEngineeringDir)PublishProjects.targets" />
+
+  <!-- Publish projects if they were not published after build -->
+  <Target Name="PublishProjectsBeforeBundle"
+          Condition="'$(PublishProjectsAfterBuild)' != 'true'">
+    <CallTarget Targets="PublishProjects" />
+  </Target>
+
+  <Target Name="CopyBundle">
+    <MakeDir Directories="$(ArtifactsNonShippingBundlesDir)" />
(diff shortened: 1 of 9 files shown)


src/singlefile/Directory.Build.targets
```diff
@@ -0,0 +1,24 @@
...
+          Condition="'$(PublishProjectsAfterBuild)' != 'true'">
+    <CallTarget Targets="PublishProjects" />
+  </Target>
+
+  <Target Name="CopyBundle">
+    <MakeDir Directories="$(ArtifactsNonShippingBundlesDir)" />
+    <!-- e.g. dotnet-monitor.exe -> dotnet-monitor-win-x64.exe -->
+    <Copy SourceFiles="$(DotnetMonitorPublishSfbPath)$(BundleName)$(BundleSuffix)"
+          DestinationFiles="$(ArtifactsNonShippingBundlesDir)$(BundleName)-$(PackageRid)$(BundleSuffix)" />
+    <Message Text="$(BundleName)$(BundleSuffix) -> $(ArtifactsNonShippingBundlesDir)$(BundleName)-$(PackageRid)$(BundleSuffix)"
```
wiktork, line 21: Is this Message intended to stay?

jander-msft: Yes, helps with understanding where it was copied when reading logs.

wiktork (approved):

PR #4673, opened 2023-06-16
[Exceptions] Add Method Parameter Types

kkeirstead:
###### Summary

Builds on https://github.com/dotnet/dotnet-monitor/pull/4471; adds method parameters to the stack frame.


src/Microsoft.Diagnostics.Monitoring.WebApi/Stacks/NameIdentificationEvents.cs
```diff
@@ -17,6 +17,7 @@ public static class FunctionDescPayloads
             public const int ModuleId = 3;
             public const int Name = 4;
             public const int TypeArgs = 5;
+            public const int Parameters = 6;
```
jander-msft, line 20: Let's rename this to ParameterTypes to be more clear as to what they are.

src/Microsoft.Diagnostics.Monitoring.WebApi/Stacks/NameFormatter.cs
```diff
@@ -24,6 +24,8 @@ internal sealed class NameFormatter
         private const char GenericStart = '[';
         private const char GenericSeparator = ',';
         private const char GenericEnd = ']';
+        private const char OpenParentheses = '(';
```
jander-msft, line 27: Name these for what they are used, not what they literally are e.g. `MethodParametersStart` and `MethodParametersEnd`

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsPipelineTests.cs
```diff
@@ -110,7 +110,7 @@ public Task EventExceptionsPipeline_AsyncException()
                     Assert.Equal(typeof(TaskCanceledException).FullName, instance.TypeName);
                     Assert.False(string.IsNullOrEmpty(instance.Message));
 
-                    ValidateStack(instance, "ThrowForNonSuccess", "System.Private.CoreLib.dll", "System.Runtime.CompilerServices.TaskAwaiter");
+                    ValidateStack(instance, "ThrowForNonSuccess(System.Threading.Tasks.Task)", "System.Private.CoreLib.dll", "System.Runtime.CompilerServices.TaskAwaiter");
```
jander-msft, line 113: While this is technically correct, we will probably want to simplify this to just be the simple name e.g. `Task` instead of `System.Threading.Tasks.Task`; the fully qualified name is going to cause too much verbosity. I would save this for a future change.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Identification/ExceptionGroupIdentifierCache.cs
```diff
@@ -91,6 +91,14 @@ public ulong GetOrAdd(MethodBase method)
             catch (Exception) { }
 
             // RTDynamicMethod does not implement GetGenericArguments.
+
+            ParameterInfo[] parameters = Array.Empty<ParameterInfo>();
+            try
+            {
+                parameters = method.GetParameters();
```
jander-msft, line 98: Please document the specific reason why you are try...catching this. Also, make sure the comment that was above the `Type[] genericArguments = Array.Empty<Type>();` line remains just above it.

jander-msft: I can't find a reason why `MethodBase.GetParameters` would fail by reading the code, especially for `RTDynamicMethod`; it seems like it should work.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/StacksTests.cs
```diff
@@ -122,7 +122,7 @@ public Task TestJsonStacks(Architecture targetArchitecture)
                     WebApi.Models.CallStackFrame[] expectedFrames = ExpectedFrames();
                     (WebApi.Models.CallStack stack, IList<WebApi.Models.CallStackFrame> actualFrames) = GetActualFrames(result, expectedFrames.First(), expectedFrames.Length);
 
-                    Assert.NotNull(stack);
+                    Assert.NotNull(stack); // -> this is because of parentheses, need to figure out if we want this behavior change
```
kkeirstead, line 125: Wanted to call attention to this - for Exceptions' stacks, I'm currently including `()` on the end if there are no parameter types, but we don't currently include `()` for Stacks, which caused these tests to fail.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsPipelineTests.cs
```diff
@@ -53,7 +53,7 @@ public Task EventExceptionsPipeline_SingleException()
                     Assert.False(string.IsNullOrEmpty(instance.Message));
```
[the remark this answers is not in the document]
kkeirstead: I added an extra test for this - if we'd rather, I can also convert an existing test to have this parameter, since it's currently effectively a duplicate of the `CustomException` test.

jander-msft:
I think something was merged incorrectly; the build fails due to an invalid cast.

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/StackUtilities.cs
```diff
@@ -55,7 +55,12 @@ internal static Models.CallStackFrame CreateFrameModel(CallStackFrame frame, Nam
             else if (cache.FunctionData.TryGetValue(frame.FunctionId, out FunctionData functionData))
             {
                 frameModel.ModuleName = NameFormatter.GetModuleName(cache, functionData.ModuleId);
-                frameModel.MethodName = functionData.Name;
+
+                builder.Clear();
+                builder.Append(functionData.Name);
+
+                NameFormatter.BuildMethodParameterTypes(builder, cache, functionData.ParameterTypes);
```
jander-msft, line 62: @wiktork, do you think this is okay (the method name includes the name and the parameter types) or should the parameter types be another property on the CallStackFrame?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/StacksTests.cs
jander-msft: I'm not sure I understand this comment... why would `stack` be `null` with your changes and thus cause a failure here?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsPipelineTests.cs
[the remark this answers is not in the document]
jander-msft: Please simplify that test so it's sole focus is on the call stack e.g. throw a built in exception without an inner exception and rename the test to something like EsotericStackFrameTypes.

Also, consider adding `unsafe` types as one of the parameters e.g. `void*` to flex more of the special cases that are less commonly encountered.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/StacksTests.cs
kkeirstead: IIRC, `GetActualFrames` only returns a stack when it's a perfect match - since my changes add a `()` even when there are no parameters (and stacks previously didn't have this behavior), there's no match (`myMethod != myMethod()`). Thus, this was returning null for the stack, and the test was failing out here.

jander-msft: Let's wait for feedback on https://github.com/dotnet/dotnet-monitor/pull/4673#discussion_r1237598447 before attempting to fix.

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/StackUtilities.cs
jander-msft: @kkeirstead, what if you add a parameter to `CreateFrameModel` named `bool excludeMethodParameters = false`, skip the `buildMethodParameterTypes` call when this is true, and then set this parameter to true in the stacks feature usage? That way the existing behavior for stacks is preserved and can be later flipped back to false if the feature starts supporting method parameters.

wiktork: Can we keep the structured data separate and then use the param types in the method name at rendering time?

jander-msft: > Can we keep the structured data separate and then use the param types in the method name at rendering time?

That makes sense.

kkeirstead: I think my latest change addresses this point, unless I misinterpreted what you had in mind.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsPipelineTests.cs
kkeirstead: /todo [Exceptions] Simplify Representation of Method Parameter Types

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/StacksTests.cs
kkeirstead, line 125: I'll rip this out

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsPipelineTests.cs
```diff
@@ -156,6 +156,27 @@ public Task EventExceptionsPipeline_CustomException()
                 });
         }
 
+        /// <summary>
+        /// Tests that custom exceptions with ref parameters are handled correctly.
```
kkeirstead, line 160: I'll also update this comment

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/StacksTests.cs
jander-msft, line 125: Please remove comment

src/Tools/dotnet-monitor/Exceptions/ExceptionsOperation.cs
```diff
@@ -204,13 +214,19 @@ private static async Task WriteTextInnerExceptionsAndStackFrames(TextWriter writ
 
             if (null != currentInstance.CallStack)
             {
+                StringBuilder builder = new();
```
jander-msft, line 217: Use the TextWriter directly; avoid using StringBuilder here.

src/Tools/dotnet-monitor/Exceptions/ExceptionsOperation.cs
```diff
@@ -261,5 +271,26 @@ private static async Task WriteTextExceptionFormat(TextWriter writer, IException
                 await writer.WriteAsync(instance.Message);
             }
         }
+
+        public static string AssembleMethodName(string methodName, IList<string> parameterTypes)
```
jander-msft, line 275: Inline this with the caller and write directly to the TextWriter; do not use StringBuilder

kkeirstead: Now it should actually be gone 😃

jander-msft (approved):

PR #4696, opened 2023-06-19
[Parameter Capturing] Add feature flag

schmittjoseph:
###### Summary

This PR adds an experimental feature flag for parameter capturing to dotnet-monitor. Setting it causes:
- The hosting startup assembly to be loaded into the target process.
- The hosting startup assembly to register the parameter capturing service.
- The profiler to initialize the probe instrumentation code.

To support this functionality, the concept of applying in-process feature flags to the target process is added so that our in-process code can conditionally enable functionality.

Included in this PR are functional tests to validate that the hosting startup gets loaded into target processes as needed, and does not impact apps which don't support a hosting startup assembly.


src/Tools/dotnet-monitor/StartupHook/StartupHookValidator.cs
```diff
@@ -24,18 +24,26 @@ internal sealed class StartupHookValidator
...
         {
             _logger = logger;
+            _inProcessFeatures = inProcessFeatures;
             _sharedLibraryService = sharedLibraryService;
         }
 
-        public async Task<bool> CheckAsync(IEndpointInfo endpointInfo, CancellationToken token)
+        public async Task<bool> CheckAsync(IEndpointInfo endpointInfo, CancellationToken token, bool logInstructions = false)
         {
+            if (!_inProcessFeatures.IsStartupHookRequired)
```
jander-msft, line 42: I think I'd rather have the callers be responsible for this check rather than having them blindly invoke the `StartupHookValidator` and rely on its return value when startup hooks are not necessary. The return value indicates whether the startup hook was configured correctly or not. Returning false when startup hooks are not required tells the caller that startup hooks were not configured (but that wasn't really checked at all).

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/HostingStartupScenario.cs
```diff
@@ -0,0 +1,82 @@
...
+            {
+                return Task.FromResult(0);
+            }, token);
+        }
+
+        public static Task<int> VerifyAspNetAppWithoutHostingStartupAsync(ParseResult result, CancellationToken token)
+        {
+            return ScenarioHelpers.RunWebScenarioAsync<Startup>((logger, host) =>
+            {
+                Assert.Null(host.Services.GetService<InProcMonitoringSentinelService>());
```
jander-msft, line 50: Because you are referencing types directly from the hosting startup assembly, that likely means the hosting startup assembly is part of the application output and is resolved using the built in assembly resolver. I think it would be better if the test app did not reference the assembly directly and allowed the startup hook assembly to resolve it. That will be probably require removing the project reference and using reflection to access the necessary parts of the hosting startup from the test app.

This doesn't have to be done as part of this PR since this is already happening without this PR.

jander-msft: /todo [Parameter Capturing] Remove hosting startup project dependency from unit test application

src/Microsoft.Diagnostics.Monitoring.HostingStartup/HostingStartup.cs
```diff
@@ -15,8 +18,13 @@ public void Configure(IWebHostBuilder builder)
         {
             builder.ConfigureServices(services =>
             {
-                // TODO: Gate this behind an environment variable.
-                services.AddHostedService<ParameterCapturingService>();
+                // Add a no-op singleton for easy identification that this hosting startup is active.
+                services.AddSingleton<InProcMonitoringSentinelService>();
```
jander-msft, line 22: Consider using something like a `public static int InvocationCount` property on `HostingStartup` so we don't have to add an unused service to the app's composition. The tests should be able to easily see that property either directly or through reflection.

jander-msft (approved):

PR #4713, opened 2023-06-22
Move Microsoft.Extensions versions to TFM specific dependabot props

jander-msft:
###### Summary

Split the `Microsoft.Extensions.*` package versions along their corresponding .NET runtime versions. Have each project only reference the `Microsoft.Extensions.*` assembly version that corresponds with the TFM.

###### Background

The `dotnet/installer` updates started failing because a type was moved from `Microsoft.Extensions.Logging` to `Microsoft.Extensions.Logging.Abstractions`. The referenced version of `Microsoft.Extensions.Logging` was set to 7.0.0 which means that when `Microsoft.Extensions.Logging` 7.0.0 is used in concert with `Microsoft.Extensions.Logging.Abstractions` 8.0.0, ambiguous type references would occur.

To solve this, I've split all of the `Microsoft.Extensions.*` versions along their corresponding .NET runtime versions. Thus, the 6.0.*, 7.0.*, and 8.0.* lines of these assemblies are versioned independently. These versions are then conditionally included based on the target framework version of the built project. This allows the projects to include the extension assemblies that correspond with their built TFM and the entire assembly list for the extension assemblies is self consistent within the major version.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -91,15 +91,30 @@
          project references for command line builds. They should match the values in the diagnostics repo. -->
     <MicrosoftBclAsyncInterfacesVersion>6.0.0</MicrosoftBclAsyncInterfacesVersion>
     <MicrosoftDiagnosticsTracingTraceEventVersion>3.0.7</MicrosoftDiagnosticsTracingTraceEventVersion>
-    <MicrosoftExtensionsLoggingVersion>7.0.0</MicrosoftExtensionsLoggingVersion>
   </PropertyGroup>
   <PropertyGroup Label=".NET 6 Dependent" Condition=" '$(TargetFramework)' == 'net6.0' ">
     <MicrosoftAspNetCoreAuthenticationJwtBearerVersion>$(MicrosoftAspNetCoreApp60Version)</MicrosoftAspNetCoreAuthenticationJwtBearerVersion>
     <MicrosoftAspNetCoreAuthenticationNegotiateVersion>$(MicrosoftAspNetCoreApp60Version)</MicrosoftAspNetCoreAuthenticationNegotiateVersion>
+    <MicrosoftExtensionsConfigurationAbstractionsVersion>$(MicrosoftExtensionsConfigurationAbstractions60Version)</MicrosoftExtensionsConfigurationAbstractionsVersion>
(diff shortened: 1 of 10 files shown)


src/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
```diff
@@ -129,7 +129,9 @@ private static int ProcessEgressResult(EgressArtifactResult result)
             // dependency injection.
             services.TryAddSingleton<EgressProvider<TOptions>, TProvider>();
 
+#if NET7_0_OR_GREATER
```
wiktork, line 132: I take it this is a result of downgrading from 7/8 to 6 correct?

wiktork (dismissed):

src/Microsoft.Diagnostics.Monitoring.Extension.Common/EgressHelper.cs
jander-msft: Yes

eng/Versions.props
```diff
@@ -91,15 +91,42 @@
...
+  </PropertyGroup>
+  <PropertyGroup>
+    <!--
+      This property sets a baseline version for packages produced by the dotnet/runtime repository such that
+      the value is the same as that in the SDK OR OLDER. This property can be used for packages that ship as
+      part of the corresponding framework; this allows those packages to be referenced but not copied to the
+      output directory since the ref pack from the SDK should be the same version or newer (this causes a package
+      version conflict and will result in the framework reference winning; since it is part of the framework,
+      the assembly is not copied to the output directory). However, this cause the NuGet package dependency
+      for any NuGet packages that are created have this as the minimum version.
```
wiktork, line 103: One confusing element for this explanation is why the reference is required at all? If the package is already part of the runtime, is PackageReference even necessary if we do not need a higher version?

wiktork (approved):

eng/Versions.props
jander-msft: You're right, but this is the behavior that I was seeing for some reason.

I think I'm going to merge this in as-is in order to unblock the dotnet/installer updates and then do more investigation as a follow up.

PR #4733, opened 2023-06-24
Enable profiler attach and update call stack tests to cover listen-nosuspend scenario

jander-msft:
###### Summary

These changes attempt to attach the profiler and then fallback to setting the profiler as a startup profiler. The call stacks test have been updated to cover the profiler attach scenario by enabling testing of the `listen-nosuspend` combination of settings such that the target application can connect to .NET Monitor's listening diagnostic port BUT not suspend itself thus allowing managed code to more-or-less immediately execute without waiting for .NET Monitor to resume the process.

###### Release Notes Entry

Enable call stacks feature to work with processes configured with the `nosuspend` option on their diagnostic port.

### DIFF
--- a/src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
+++ b/src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
@@ -98,10 +98,31 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Profiler
                         endpointInfo.RuntimeInstanceCookie.ToString("D"),
                         cancellationToken);
 
-                    await client.SetStartupProfilerAsync(
-                        ProfilerIdentifiers.Clsid.Guid,
-                        profilerFileInfo.PhysicalPath,
-                        cancellationToken);
+                    // There's no way to ask the target process if it is suspended at startup.
+                    // Just attempt to attach the profiler; if the process is already running,
+                    // this will throw an exception. Then fallback to setting the startup profiler,
+                    // assuming that the failure occurred due to the process already running.
+                    // Can't check that assumption because ServerErrorException and related
+                    // exceptions do not report what the issue is beyond the human readable
+                    // Exception.Message property value.
+                    try
+                    {
+                        // This will wait until the profiler's ICorProfilerCallback3::InitializeForAttach
+                        // implementation completes.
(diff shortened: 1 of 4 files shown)


src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
```diff
@@ -98,10 +98,31 @@ public async Task ApplyProfiler(IEndpointInfo endpointInfo, CancellationToken ca
...
+                        await client.AttachProfilerAsync(
+                            TimeSpan.FromSeconds(10),
+                            ProfilerIdentifiers.Clsid.Guid,
+                            profilerFileInfo.PhysicalPath,
+                            Array.Empty<byte>(),
+                            cancellationToken);
+                    }
+                    catch (ServerErrorException)
+                    {
+                        await client.SetStartupProfilerAsync(
```
wiktork, line 121: I was really expecting the reverse here; we SetStartupProfiler by default, and if that fails we fall back to Attach. What error codes does SetStartupProfilerAsync return if the process is not suspended?

jander-msft: We should be able to do either way; they should be mutually exclusive.

jander-msft: I'll flip it around so that SetStartupProfiler is first; assuming processes that are connecting to the .NET Monitor server are more likely to not use nosuspend than are to use nosuspend, then this call should succeed more often than calling AttachProfiler.

jander-msft: I investigated this a bit more and we can't flip this around to what you'd expect. If the process has `nosuspend` and is already running managed code, the `SetStartupProfiler` command will succeed; this repros on .NET 6 and .NET 7 (haven't tested .NET 8, but it probably repros there too.

fyi @davmason

jander-msft: I've updated the comments to reflect what is happening and why we cannot do SetStartupProfiler first.

wiktork (approved):

PR #4746, opened 2023-06-27
Use ApplyStartupHook diagnostic command in exceptions pipeline tests

jander-msft:
###### Summary

Make use of `DiagnosticsClient.ApplyStartupHookAsync` in the exceptions pipeline tests for .NET 8+ targets instead of using the `DOTNET_STARTUP_HOOKS` environment variable.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTestCommon/EndpointInfoSourceCallback.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTestCommon/EndpointInfoSourceCallback.cs
@@ -3,12 +3,15 @@
 
 using Microsoft.Diagnostics.Monitoring.TestCommon.Runners;
 using Microsoft.Diagnostics.Monitoring.WebApi;
+using Microsoft.Diagnostics.NETCore.Client;
 using Microsoft.Diagnostics.Tools.Monitor;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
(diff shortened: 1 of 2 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTestCommon/EndpointInfoSourceCallback.cs
```diff
@@ -89,6 +100,19 @@ public Task OnRemovedEndpointInfoAsync(IEndpointInfo info, CancellationToken tok
                 token);
         }
 
+        private static Task ApplyStartupHookAsync(DiagnosticsClient client, string path, CancellationToken token)
+        {
+            // DiagnosticsClient.ApplyStartupHookAsync currently is not public
+            MethodBase applyStartupHookAsync =
+                typeof(DiagnosticsClient).GetMethod(
```
wiktork, line 107: Will this become public in the future?

wiktork (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTestCommon/EndpointInfoSourceCallback.cs
jander-msft: I hope so. Many of the APIs that we've added are marked internal pending API review. Maybe we can make an effort to make them all public in the .NET 8 timeframe.

PR #4764, opened 2023-06-27
Promote call stacks and exceptions to be supported features

jander-msft:
###### Summary

These changes convert the experimental call stacks and exceptions features into supported features in the 8.0 version:
- Removed experimental feature flags for call stacks and exceptions
- Add `CallStacks` and `Exceptions` configuration sections under the `InProcessFeatures` configuration section; each new section has an `Enabled` property.
- Tests have been updated to use the new configuration elements to verify the feature state.

Details of enabling features:
- If `InProcessFeatures:Enabled` is set to true, both of these features are enabled automatically.
- If `InProcessFeatures:Enabled` is set to false, both are disabled regardless of the settings for each feature.
- If `InProcessFeatures:<Feature>:Enabled` is set to true (but `InProcessFeatures:Enabled` is unconfigured or true), then the individual feature is enabled. **Note this is a new behavior introduced in this PR to simplify configuration of in-process features so users can opt-into individual features rather than taking the whole default set.**
- If `InProcessFeatures:<Feature>:Enabled` is set to false, then the individual feature is disabled.

Example JSON configuration that enables the call stacks feature only with minimal configuration:
```
{
  "InProcessFeatures": {
    "CallStacks": {
      "Enabled": true
    }
  }
}
```

Example JSON configuration that enables all in-process features but opts out of call stack feature:
```
{
  "InProcessFeatures": {
    "Enabled": true
    "CallStacks": {
      "Enabled": false
    }
  }
}
```

###### Release Notes Entry

Enable call stacks and exceptions and supported features.


src/Microsoft.Diagnostics.Monitoring.WebApi/IExperimentalFlags.cs
```diff
@@ -9,8 +9,5 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
 {
     internal interface IExperimentalFlags
     {
```
kkeirstead, line 11: Are we intentionally keeping this around as an empty interface?

src/Tools/dotnet-monitor/Exceptions/ExceptionsPostConfigureOptions.cs
```diff
@@ -0,0 +1,28 @@
...
+        public ExceptionsPostConfigureOptions(IConfiguration configuration)
+        {
+            _configuration = configuration;
+        }
+
+        void IPostConfigureOptions<ExceptionsOptions>.PostConfigure(string name, ExceptionsOptions options)
+        {
+            InProcessFeatureOptionsBinder.BindEnabled(
+                options,
+                ConfigurationKeys.InProcessFeatures_CallStacks,
```
kkeirstead, line 23: I believe this is a copy-paste error (as well as L25)

jander-msft: Good catch

src/Tools/dotnet-monitor/InProcessFeatures/InProcessFeatureOptionsBinder.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            bool? inProcessFeaturesEnabled = null;
+            IConfigurationSection inProcessFeaturesSection = configuration.GetSection(ConfigurationKeys.InProcessFeatures_Enabled);
+            if (!string.IsNullOrEmpty(inProcessFeaturesSection.Value))
+            {
+                inProcessFeaturesEnabled = inProcessFeaturesSection.Get<bool>();
+            }
+
+            // If in-process features are disabled, then disable the individual feature
+            if (!inProcessFeaturesEnabled.GetValueOrDefault(true))
```
kkeirstead, line 21: I'm struggling a bit with the readability flow in this method. Is it possible we could do some nesting up here? My understanding is that `inProcessFeaturesEnabled` will only have a value if the above conditional is met, so we could nest this into there. Additionally, my understanding is that `options.Enabled.HasValue` will only be satisfied if the current condition is met since that's the only place `options.Enabled` is being explicitly set thus far, so we could combine that logic as well. Feel free to correct me if I mis-read this, but I just noticed as I was reviewing this that it was requiring a lot more cognitive effort than I'd anticipated to keep everything straight.

src/Microsoft.Diagnostics.Monitoring.WebApi/IExperimentalFlags.cs
jander-msft: I'm leaving this and the experimental feature logger in for now since parameter capturing are using them and wanted to make that merge a little easier. They can be removed once that feature moves away from this interface.

src/Tools/dotnet-monitor/InProcessFeatures/InProcessFeatureOptionsBinder.cs
jander-msft: Both of the Enable properties (at the InProcessFeatures level and the individual feature level) are nullable. When one of them is null, it is deferring the final decision of whether a feature is enable or not to the other property. So here, either the user didn't specify anything (the value is null and thus deferring whether individual features are enabled to their individual Enabled properties), the user specified `true` (so we don't disable all individual features), or the user specified `false` (so we disable all individual features).

The `!inProcessFeaturesEnabled.GetValueOrDefault(true)` is effectively asking "should we disable all features?":

| `inProcessFeaturesEnabled` | `!inProcessFeaturesEnabled.GetValueOrDefault(true)` | Explanation |
|---|---|---|
| null | false | User didn't specify anything -> don't disable all features |
| true | false | User said to enable in-process features -> don't disable all features |
| false | true | User said to disable in-process features -> disable all features |

jander-msft: I should write some tests for the InProcessFeatureOptionsBuilder to validate each path within this function; that'll make sure it doesn't change accidentally and should make it easier to understand the intended outcome for each input combination.

kkeirstead:
I'm wondering whether the knobs here might be too confusing for users (for people who want more fine-grained control). I think the main crux of that is that `InProcessFeatures:Enabled` is a boolean that effectively has three possible states (on|off|unset), which have different behaviors that are influenced by `InProcessFeatures:<Feature>:Enabled`.

This is just a suggestion to prompt some discussion - I'm OK with the current approach (or something different) if others think it makes the most sense. Temporarily putting aside schema complications, on a theoretical level this is how I'd like to think about this as a user:

* `InProcessFeatures:Enabled = true`: Give me everything - I don't mind having everything turned on, as well as anything you might add in the future.
* `InProcessFeatures:Enabled = false`: I don't want `InProcessFeatures` - turn them all off.
* `InProcessFeatures:Enabled = { Feature1, Feature2, etc. }`: I want to explicitly specify what I want enabled. This protects me in the event that DM adds something else later that I don't want on by default (at which point I can explicitly add it to this list).

Everything is controlled by `InProcessFeatures:Enabled`, and it requires no understanding of how multiple variables interact with each other in terms of what overrides what. My only concern is that `InProcessFeatures:Enabled` here is basically overloaded so that it can be a bool or a list. We do have a precedent for a behavior similar to this with [templates](https://github.com/dotnet/dotnet-monitor/blob/main/documentation/collectionrules/templates.md), allowing users to specify a name (string) instead of what the schema expects, but I also recognize that we might not want that here.

jander-msft:
> * `InProcessFeatures:Enabled = { Feature1, Feature2, etc. }`: I want to explicitly specify what I want enabled. This protects me in the event that DM adds something else later that I don't want on by default (at which point I can explicitly add it to this list).

This PR already has that in the form of `InProcessFeatures:<Feature>:Enabled` e.g.
```
InProcessFeatures:CallStacks:Enabled = true
InProcessFeatures:Exceptions:Enabled = true
```

Keep in mind that these features have a strong possibility that they will have their own configuration elements in the future e.g.
```
{
  "InProcessFeatures": {
    "Enable": true
    "CallStacks": {
      "IncludeMethodParameterTypes": true,
      "StitchAsyncFrames": true
    },
    "Exceptions": {
      "JustMyCode": true
    }
  }
}
```

Using a list at the `InProcessFeatures:Enabled` property:
- Will be confusing from a schema perspective.
- It also makes it more difficult to opt into and out of features individually without modifying the entire list.
- It forces you to get the name of the features correctly in two places instead of one thus causing potential mixed state of you either specifying feature options correctly OR opting into individual features but small possibility of not doing both correctly.
- Finally, there's no facility to opt-out of an individual feature if we use the list; having this ability is helpful for devops where a feature is causing an issue for a customer and we need to recommend they disable the feature. If the customer was only using `InProcessFeatures:Enabled = true` to enable all features, then they would have to enumerate the features that they do want and exclude the problematic one e.g. `InProcessFeatures:Enabled = true -> InProcessFeatures:Enabled = [ 'CallStacks' ]`. With `InProcessFeatures:<Feature>:Enabled` properties, it is much easier to add a single property of `InProcessFeatures:<Feature>:Enabled = false` than it is to rewrite the rest of the configuration.

src/Tools/dotnet-monitor/Experimental/ExperimentalStartupLogger.cs
```diff
@@ -20,14 +20,6 @@ public ExperimentalStartupLogger(ILogger<Startup> logger, IExperimentalFlags exp
 
         public void Log()
         {
-            if (_experimentalFlags.IsCallStacksEnabled)
```
wiktork, line 23: Maybe add some comments here for the expectation of this method?

src/Tools/dotnet-monitor/InProcessFeatures/InProcessFeatureOptionsBinder.cs
```diff
@@ -0,0 +1,66 @@
...
+{
+    internal static class InProcessFeatureOptionsBinder
+    {
+        public static void BindEnabled(IInProcessFeatureOptions options, string configurationKey, IConfiguration configuration, bool enabledByDefault)
+        {
+            // Get the value of InProcessFeatures:<Feature>:Enabled. This may have already
+            // been bound by the default options binding, but this makes it consistent for
+            // anything that passes in an unbound options instance and fulfills the complete
+            // expected behavior of binding the Enabled property.
+            options.Enabled = configuration
```
wiktork, line 17: Just to be clear, if the user says "InProcessFeatures": {}, the result will be enabled = true?

src/Tools/dotnet-monitor/Exceptions/ExceptionsPostConfigureOptions.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Diagnostics.Monitoring.Options;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.Stacks
+{
+    internal sealed class ExceptionsPostConfigureOptions : IPostConfigureOptions<ExceptionsOptions>
```
wiktork, line 10: Perhaps we can have a baseclass for this since the post configure options look almost identical between features?

src/Tools/dotnet-monitor/InProcessFeatures/InProcessFeatureOptionsBinder.cs
jander-msft: No, that doesn't translate to anything in configuration. The minimal to enable in-process features is:
```
{
  "InProcessFeatures": {
    "Enabled": true
  }
}
```
This is the same as it was before.

However, I could add the ability to allow the following to enable all in-process features:
```
{
  "InProcessFeatures": true
}
```

jander-msft: Chatted offline... the "Enable" properties are good enough for now. I've removed the `InProcessFeatures:<Feature> = true/false` ability. We can add that back later if there is demand for this simplified experience.

wiktork (approved):

PR #4775, opened 2023-06-28
[Parameter Capturing] Support sending messages to managed in-proc features from dotnet-monitor

schmittjoseph:
###### Summary

This PR introduces new infrastructure to support sending messages (with arbitrary managed object payloads) from `dotnet-monitor` to our in-process features either in the `StartupHook` or the `HostingStartup` assemblies. 

Overview:
- The existing profiler communication channel has been upgraded to support arbitrarily sized payloads. Previously it only support sending a simple int32 payload.
- Managed objects are serialized/deserialized as a utf8-encoded json string. They are **NOT** deserialized by native code. They are serialized & deserialized using `System.Text.Json`.
  - As a result, messages intended purely for the profiler (e.g. callstacks) do not support these json payloads.

Example flow of sending a managed->managed message:
1. [in-proc] When the startup hook loads and initializes, it pinvokes into the profiler and register a callback to be notified of messages (iff the profiler is available).
1. [in-proc] Any in-proc features interested in receiving messages register a callback with `MonitorMessageDispatcher` for a given command and its payload object type.
1. [dotnet-monitor] A message is sent using the profiler channel, containing a specific command with a managed object as the payload.
1. [dotnet-monitor] The managed object is serialized as a utf8-encoded json string and sent to the profiler.
1. [in-proc] The profiler receives the message, and sees that no native features are interested in the command, and so it lets our managed code try to handle it by reverse pinvoking the previously registered callback (if present).
1. [in-proc] The startup hook receives the message and checks if any in-proc features are interested in the command. If so, the utf8-json string is deserialized using the previously registered type and then invokes the registered callback.

Out of scope of this PR:
- The ability to unregister a message callback with the profiler. We currently don't have a scenario that needs this functionality.
- Allowing payloads bigger than `INT32_MAX`.
- Sending back the status (profiler->dotnet-monitor) of the action taken by a message. Currently the status sent back is merely an ACK/NACK.

### DIFF
--- a/src/MonitorProfiler/MainProfiler/MainProfiler.cpp
+++ b/src/MonitorProfiler/MainProfiler/MainProfiler.cpp
@@ -384,7 +384,7 @@ STDAPI DLLEXPORT RegisterMonitorMessageCallback(
     // A lock-free approach could be used to safely update and observe the value of the callback,
     // however that would introduce the edge case where the provided callback is unregistered
     // right before it is invoked.
-    // This means that the unregistered would still be invoked, leading to potential issues
+    // This means that the unregistered callback would still be invoked, leading to potential issues
     // such as calling into an instanced method that has been disposed.
     //
     // For simplicitly just use locking for now as it prevents the above edge case.


src/Microsoft.Diagnostics.Monitoring.StartupHook/MonitorMessageDispatcher/IMonitorMessageSource.cs
```diff
@@ -0,0 +1,27 @@
...
+        }
+
+        public IpcCommand Command { get; private set; }
+        public IntPtr NativeBuffer { get; private set; }
+        public long BufferSize { get; private set; }
+    }
+
+    internal interface IMonitorMessageSource : IDisposable
+    {
+        public delegate void MonitorMessageHandler(object sender, MonitorMessageArgs args);
```
wiktork, line 24: nit: The (object sender, EventArgs) signature is not really something we have to follow here. Action<> and Func<> with strongly typed parameters should work well here too.

src/Microsoft.Diagnostics.Monitoring.StartupHook/MonitorMessageDispatcher/MonitorMessageDispatcher.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        public void UnregisterCallback(IpcCommand command)
+        {
+            _dispatchTable.TryRemove(command, out _);
+        }
+
+        private void OnMessage(object sender, MonitorMessageArgs args)
+        {
+            object? payload = null;
+            if (!_dispatchTable.TryGetValue(args.Command, out MessageDispatchEntry dispatchEntry))
```
wiktork, line 57: Any chance for some race conditions here? For example, if Unregister happens right before OnMessage, we will throw.

src/Microsoft.Diagnostics.Monitoring.StartupHook/MonitorMessageDispatcher/MonitorMessageDispatcher.cs
```diff
@@ -0,0 +1,85 @@
...
+            {
+                throw new NotSupportedException("Unsupported message type.");
+            }
+
+            unsafe
+            {
+                using UnmanagedMemoryStream memoryStream = new((byte*)args.NativeBuffer.ToPointer(), args.BufferSize);
+                payload = JsonSerializer.Deserialize(memoryStream, dispatchEntry.DeserializeType);
```
wiktork, line 65: Should we be guarding against exceptions from Deserialize?

src/Microsoft.Diagnostics.Monitoring.StartupHook/MonitorMessageDispatcher/ProfilerMessageSource.cs
```diff
@@ -0,0 +1,64 @@
...
+                {
+                    throw new ArgumentException(nameof(nativeBuffer));
+                }
+
+                ProfilerMessageSource instance = s_instance ?? throw new NotSupportedException();
+                instance.RaiseMonitorMessage(new MonitorMessageArgs(command, nativeBuffer, bufferSize));
+            }
+            catch (Exception ex)
+            {
+                return Marshal.GetHRForException(ex);
```
wiktork, line 53: You might want to peek into what GetHRForException actually works against and if it matches the exceptions you're expecting here, or are they all going to become E_FAIL

src/Microsoft.Diagnostics.Monitoring.StartupHook/MonitorMessageDispatcher/ProfilerMessageSource.cs
```diff
@@ -0,0 +1,64 @@
...
+            RegisterMonitorMessageCallback(OnProfilerMessage);
+            s_instance = this;
+        }
+
+        private void RaiseMonitorMessage(MonitorMessageArgs e)
+        {
+            MonitorMessageEvent?.Invoke(this, e);
+        }
+
+        private static int OnProfilerMessage(IpcCommand command, IntPtr nativeBuffer, long bufferSize)
```
wiktork, line 34: I was under the impression that the Marshaller should be able to handle instance methods as well. If it can't handle them implicitly, you could use Marshal.GetFunctionPointerForDelegate. Just make sure you pin your delegate first.

src/Microsoft.Diagnostics.Monitoring.StartupHook/ProfilerResolver.cs
```diff
@@ -0,0 +1,65 @@
...
+
+            Assembly assembly = type.Assembly;
+            lock (s_registeredAssembliesLocker)
+            {
+                if (s_registeredAssemblies.Contains(assembly))
+                {
+                    return;
+                }
+
+                s_registeredAssemblies.Add(assembly);
```
wiktork, line 37: nit: you can combine your Add & Contains call into one. Add returns true if it's the first add, false otherwise.

src/Microsoft.Diagnostics.Monitoring.StartupHook/StartupHook.cs
```diff
@@ -24,6 +25,14 @@ public static void Initialize()
             }
 
             s_exceptionProcessor.Start();
+
+            try
+            {
+                SharedInternals.MessageDispatcher = new MonitorMessageDispatcher(new ProfilerMessageSource());
+            }
+            catch
+            {
```
wiktork, line 34: Do we have any way of capturing errors from the Startup hook? I know that's not specific to your feature even, but I am concerned we have a lot of possible failure conditions and no way of knowing they are happening.

src/MonitorProfiler/Communication/IpcCommClient.cpp
```diff
@@ -16,13 +21,53 @@ HRESULT IpcCommClient::Receive(IpcMessage& message)
...
+    int bufferOffset = 0; 
+    message.Command = *reinterpret_cast<IpcCommand*>(&headersBuffer[bufferOffset]);
+    bufferOffset += sizeof(IpcCommand);
+
+    int payloadSize = *reinterpret_cast<int*>(&headersBuffer[bufferOffset]);
+    bufferOffset += sizeof(int);
+
+    assert(bufferOffset == sizeof(headersBuffer));
+
+    if (payloadSize != 0)
```
wiktork, line 39: Check for negative and some reasonable upper bound.

src/MonitorProfiler/Communication/IpcCommClient.cpp
```diff
@@ -59,15 +103,53 @@ HRESULT IpcCommClient::Send(const IpcMessage& message)
         return E_UNEXPECTED;
     }
 
-    char buffer[sizeof(MessageType) + sizeof(int)];
-    *reinterpret_cast<MessageType*>(buffer) = message.MessageType;
-    *reinterpret_cast<int*>(&buffer[sizeof(MessageType)]) = message.Parameters;
+    if (message.Payload.size() > UINT32_MAX)
```
wiktork, line 106: As with my previous comment, we should establish some reasonable message size and not allow anything larger.

src/MonitorProfiler/MainProfiler/MainProfiler.cpp
```diff
@@ -19,6 +19,14 @@ using namespace std;
 
 #define IfFailLogRet(EXPR) IfFailLogRet_(m_pLogger, EXPR)
 
+#ifndef DLLEXPORT
+#define DLLEXPORT
+#endif
+
+typedef INT32 (STDMETHODCALLTYPE *ManagedMessageCallback)(INT16, const BYTE*, UINT64);
+std::mutex g_messageCallbackMutex;
```
wiktork, line 27: Let's make these static but class scoped at least? Random globals are difficult to work with.

src/MonitorProfiler/MainProfiler/MainProfiler.cpp
```diff
@@ -285,15 +293,28 @@ HRESULT MainProfiler::InitializeCommandServer()
...
+        case IpcCommand::Callstack:
+            return ProcessCallstackMessage();
+        default:
+            lock_guard<mutex> lock(g_messageCallbackMutex);
+            if (g_pManagedMessageCallback == nullptr)
+            {
+                return E_FAIL;
+            }
+
+            return g_pManagedMessageCallback(
```
wiktork, line 311: is the intent to lock around the message callback itself or just access to g_pManagedMessageCallback variable? For the former, I believe we already sequentially dequeue messages. For the latter, consider std::atomic or similar lockfree calls.

src/Microsoft.Diagnostics.Monitoring.StartupHook/MonitorMessageDispatcher/IMonitorMessageSource.cs
jander-msft: Or use `public event EventHandler<MonitorMessageArgs> MonitorMessage`.

src/Microsoft.Diagnostics.Monitoring.StartupHook/MonitorMessageDispatcher/IMonitorMessageSource.cs
```diff
@@ -0,0 +1,27 @@
...
+        public event MonitorMessageHandler MonitorMessageEvent;
```
jander-msft, line 25:
```suggestion
        public event EventHandler<MonitorMessageArgs> MonitorMessage;
```

src/MonitorProfiler/Communication/CommandServer.cpp
```diff
@@ -72,16 +77,18 @@ void CommandServer::ListeningThread()
             continue;
         }
 
-        IpcMessage response;
-        response.MessageType = SUCCEEDED(hr) ? MessageType::OK : MessageType::Error;
-        response.Parameters = hr;
+        for (size_t i = 0; i < sizeof(HRESULT); i++)
```
jander-msft, line 80: This size of an HRESULT is always 4-bytes... could this be written out explicitly without looping?

```c++
response.Payload[0] = static_cast<BYTE>(hr & 0xFF);
response.Payload[1] = static_cast<BYTE>((hr >> 8) & 0xFF);
response.Payload[2] = static_cast<BYTE>((hr >> 16) & 0xFF);
response.Payload[3] = static_cast<BYTE>((hr >> 24) & 0xFF);
```

Maybe the ` & 0xFF` is necessary to avoid sign/extension problems?

Or is there a way to use memcpy? That might have endianness problems.

src/Microsoft.Diagnostics.Monitoring.WebApi/ProfilerMessage.cs
```diff
@@ -0,0 +1,51 @@
...
+
+    internal struct JsonProfilerMessage : IProfilerMessage
+    {
+        public IpcCommand Command { get; set; } = IpcCommand.Unknown;
+        public byte[] Payload { get; set; }
+
+        public JsonProfilerMessage(IpcCommand command, object payloadObject)
+        {
+            Command = command;
+            Payload = SerializePayload(payloadObject);
```
wiktork, line 31: Nit: I would prefer something like JsonProfilerMessage.FromPayload(IpcCommand, object payloadObject). Also consider immutable properties.

src/MonitorProfiler/MainProfiler/MainProfiler.cpp
```diff
@@ -348,6 +370,27 @@ HRESULT STDMETHODCALLTYPE MainProfiler::GetReJITParameters(ModuleID moduleId, md
...
     }
-    
+
+    return S_OK;
+}
+
+STDAPI DLLEXPORT RegisterMonitorMessageCallback(
+    ManagedMessageCallback pCallback)
+{
+    lock_guard<mutex> lock(g_managedMessageCallbackMutex);
```
wiktork, line 380: Add some comments here regarding why the locks are desired and what it's guarding

wiktork (dismissed):

wiktork (approved):

PR #4783, opened 2023-06-29
Integrate `ApplyStartupHook` into Dotnet-Monitor

kkeirstead:
###### Summary

Builds on @jander-msft 's changes in the [runtime](https://github.com/dotnet/runtime/pull/86813) - this allows dotnet-monitor to automatically set the startup hook (instead of requiring users to manually set it themselves).


src/Tools/dotnet-monitor/StartupHook/StartupHookValidator.cs
```diff
@@ -48,6 +48,21 @@ public async Task<bool> CheckAsync(IEndpointInfo endpointInfo, CancellationToken
...
+            {
+                try
+                {
+                    await client.ApplyStartupHookAsync(startupHookLibraryFileInfo.PhysicalPath, token);
+
+                    return true;
+                }
+                catch (Exception ex)
+                {
+                    _logger.StartupHookApplyFailed(ex);
```
kkeirstead, line 62: Currently, this logs that the automatic apply was unsuccessful before we check the environment variables - do we want this behavior, or do we only want to log if the env var also wasn't provided?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsPipelineTests.cs
```diff
@@ -273,7 +273,7 @@ public Task EventExceptionsPipeline_InnerUnthrownException()
 
                     IExceptionInstance outerInstance = instances.Skip(1).Single();
                     Assert.NotNull(outerInstance);
-                    
+
```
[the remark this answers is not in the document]
kkeirstead: I don't believe there's currently anything that expects failures for <8, but IIRC Justin added in logic to use `ApplyStartupHook` instead of setting the environment variable for .NET 8+ in [this PR](https://github.com/dotnet/dotnet-monitor/pull/4746/files), which lets us exercise the new method as well as the existing way of setting it. If people want additional testing for this PR, I can add that though.

src/Tools/dotnet-monitor/StartupHook/StartupHookValidator.cs
```diff
@@ -34,43 +33,88 @@ internal sealed class StartupHookValidator
...
 
-            IFileInfo startupHookLibraryFileInfo = managedFileProvider.GetFileInfo(StartupHookFileName);
-            if (!startupHookLibraryFileInfo.Exists)
+        public async Task<bool> CheckAsync(IEndpointInfo endpointInfo, CancellationToken token, bool logInstructions = false)
+        {
+            if (endpointInfo.RuntimeVersion.Major >= 8)
             {
-                // This would be a bug in dotnet-monitor; throw appropriate non-MonitoringException instance.
-                throw new FileNotFoundException(null, startupHookLibraryFileInfo.Name);
+                // We currently have no way of tracking whether the startup hook was applied successfully; for now,
```
kkeirstead, line 69: Joe and I chatted about this a bit yesterday - we currently have no mechanism for tracking whether `ApplyStartupHookAsync` succeeded when we call `CheckAsync`, so for now this operates on the assumption that it did succeed. I believe the consensus was this is something that probably should be addressed, but that it's outside the scope of this change.

I can see a bit of an interesting edge-case here though - if someone has the env var set on 8+, we respect that and don't attempt to call `ApplyStartupHookAsync`. However, if that fails, it won't be reflected here, since we just assume 8 succeeds. We could probably add some more nuanced logic to this check that looks for the env var - if a value is there, we check to ensure it appears valid; if not, we blindly return true.

src/Tools/dotnet-monitor/Experimental/ExperimentalFlags.cs
```diff
@@ -17,16 +17,27 @@ internal class ExperimentalFlags : IExperimentalFlags
         private const string ExperimentalPrefix = ToolIdentifiers.StandardPrefix + "Experimental_";
```
jander-msft, line 17: Please revert the changes to this file; the experimental feature flags for these two features were removed in favor of using configuration options.

src/Tools/dotnet-monitor/StartupHook/StartupHookEndpointInfoSourceCallback.cs
```diff
@@ -0,0 +1,46 @@
...
+        Task IEndpointInfoSourceCallbacks.OnAddedEndpointInfoAsync(IEndpointInfo endpointInfo, CancellationToken cancellationToken)
+        {
+            return Task.CompletedTask;
+        }
+
+        Task IEndpointInfoSourceCallbacks.OnBeforeResumeAsync(IEndpointInfo endpointInfo, CancellationToken cancellationToken)
+        {
+            if (_inProcessFeatures.IsStartupHookRequired)
+            {
+                _ = _startupHookValidator.ApplyStartupHook(endpointInfo, cancellationToken); // Not currently doing anything with the result
```
jander-msft, line 33: I think the StartupHookValidator would be simplified if you:
1) Call `CheckAsync` without logging instructions
1) If true, immediately return
1) Call `ApplyStartupHook`
1) If true, immediately return
1) Call `CheckAsync` with logging instructions

The `CheckAsync` method shouldn't check the runtime version but unconditionally check if the environment variable exists. If it make it more clear, rename this method to `CheckEnvironmentAsync`.

The `ApplyStartupHook` method shouldn't invoke the `CheckAsync` method but merely call `ApplyStartupHook` if it's 8.0 or newer; return false on exception or if lower than 8.0

kkeirstead: Sure, I can separate those concerns - if I'm interpreting this correctly, then `CheckAsync` (or `CheckEnvironmentAsync`) will fail in most cases for 8+ since the env var won't be set, even though we'll have already (hopefully) applied the startup hook - is that the behavior we want?

jander-msft: Yes, if the user configured the environment variable manually, that should take precedence over applying the startup hook through the diagnostic communication channel.

kkeirstead: I might not have explained that correctly (or maybe I'm just misunderstanding) - assuming [this call](https://github.com/dotnet/dotnet-monitor/blob/main/src/Tools/dotnet-monitor/Exceptions/ExceptionsService.cs#L70) still exists with the updated `CheckEnvironmentAsync` method, it'll fail for 8, and then `ExceptionsService` will throw a `NotSupportedException` - which means we wouldn't start the pipeline, and exceptions wouldn't work.

jander-msft: Yup, you are correct. What we need is state stored per-process to track whether a startup hook was actually validated or applied. That state can be added to the `StartupHookEndpointInfoSourceCallback` class, the class elevated to a service, and the ExceptionsService would ask that service if the startup had been applied correctly.

src/Tools/dotnet-monitor/StartupHook/StartupHookEndpointInfoSourceCallback.cs
```diff
@@ -0,0 +1,54 @@
...
+        Task IEndpointInfoSourceCallbacks.OnAddedEndpointInfoAsync(IEndpointInfo endpointInfo, CancellationToken cancellationToken)
+        {
+            return Task.CompletedTask;
+        }
+
+        Task IEndpointInfoSourceCallbacks.OnBeforeResumeAsync(IEndpointInfo endpointInfo, CancellationToken cancellationToken)
+        {
+            if (_inProcessFeatures.IsStartupHookRequired)
+            {
+                if (_startupHookValidator.CheckEnvironmentAsync(endpointInfo, cancellationToken).Result)
```
jander-msft, line 33: Please change OnBeforeResume to be `async` and await this method instead of using `.Result`.

src/Tools/dotnet-monitor/StartupHook/StartupHookEndpointInfoSourceCallback.cs
kkeirstead: The latest iteration includes something along these lines - let me know if you want me to take a different approach until we come up with a more encompassing solution.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsPipelineTests.cs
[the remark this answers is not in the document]
kkeirstead: @jander-msft I know we chatted offline about this the other day - did you have any thoughts about if we want to add functional tests for Exceptions in this PR?

src/Tools/dotnet-monitor/StartupHook/StartupHookEndpointInfoSourceCallback.cs
```diff
@@ -0,0 +1,60 @@
...
+            return Task.CompletedTask;
+        }
+
+        async Task IEndpointInfoSourceCallbacks.OnBeforeResumeAsync(IEndpointInfo endpointInfo, CancellationToken cancellationToken)
+        {
+            if (_inProcessFeatures.IsStartupHookRequired)
+            {
+                if (await _startupHookValidator.CheckEnvironmentAsync(endpointInfo, cancellationToken))
+                {
+                    return;
```
jander-msft, line 38:
```suggestion
                    ApplyStartupState[endpointInfo.RuntimeInstanceCookie] = true;
                    return;
```

src/Tools/dotnet-monitor/ToolIdentifiers.cs
```diff
@@ -16,6 +18,21 @@ internal static class ToolIdentifiers
 
         public const string DefaultSocketName = "dotnet-monitor.sock";
 
+        public const string EnvVarEnabledValue = "1";
```
jander-msft, line 21: This and the IsEnvVarEnabled are left over from an incorrect merge?

src/Tools/dotnet-monitor/Exceptions/ExceptionsService.cs
```diff
@@ -66,8 +69,11 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
                         throw new NotSupportedException();
                     }
 
+                    bool isStartupHookApplied = false;
+                    _ = _startupHookEndpointInfoSourceCallbacks.ApplyStartupState.TryGetValue(pi.EndpointInfo.RuntimeInstanceCookie, out isStartupHookApplied);
+
                     // Validate that the process is configured correctly for collecting exceptions.
-                    if (!await _startupHookValidator.CheckAsync(pi.EndpointInfo, stoppingToken))
+                    if (!isStartupHookApplied && !await _startupHookValidator.CheckEnvironmentAsync(pi.EndpointInfo, stoppingToken))
```
jander-msft, line 76: Because `_startupHookEndpointInfoSourceCallbacks.ApplyStartupState` is tracking the startup hook state, the `CheckEnvironmentAsync` call and the `_unconfiguredProcesses` list can be removed.

src/Tools/dotnet-monitor/InProcessFeatures/InProcessFeatures.cs
```diff
@@ -17,13 +16,14 @@ public InProcessFeatures(IOptions<CallStacksOptions> callStacksOptions, IOptions
             _callStacksOptions = callStacksOptions.Value;
             _exceptionsOptions = exceptionsOptions.Value;
         }
-
         private bool IsCallStacksEnabled => _callStacksOptions.GetEnabled();
 
-        private bool IsExceptionsEnabled => _exceptionsOptions.GetEnabled();
+        public bool IsExceptionsEnabled => _exceptionsOptions.GetEnabled();
```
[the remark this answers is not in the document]
kkeirstead: Not intentional - there were a few things that merged funny and I'm guessing this was one.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsPipelineTests.cs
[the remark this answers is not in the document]
jander-msft: I'm fine with a follow up PR to do this. There currently aren't any functional tests for this feature, so it would be a bit of an undertaking.

jander-msft (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsPipelineTests.cs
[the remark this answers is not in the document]
kkeirstead: /todo Add Functional Tests For Exceptions Feature

PR #4792, opened 2023-06-30
Implement `--exit-on-stdin-disconnect` flag for dotnet-monitor collect

gregg-miskelly:
###### Summary

This PR adds a new `--exit-on-stdin-disconnect` command line switch to `dotnet collect`. When specified, this will cause dotnet-monitor to watch stdin, and if it closes, will initiate graceful shutdown.

This also made some minor debuggability improvements to the unit test code to make it faster to see which processes were spawned.

###### Release Notes Entry
Adds new `--exit-on-stdin-disconnect` command line switch to `dotnet collect`

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/CollectTraceTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/CollectTraceTests.cs
@@ -16,6 +16,8 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Net.Http;
+using System.Reflection;
+using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
 using Xunit.Abstractions;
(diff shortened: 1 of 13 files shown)


gregg-miskelly:
I have to figure out why this isn't working on macOS

src/Tools/dotnet-monitor/Commands/CollectCommandHandler.cs
```diff
@@ -143,5 +149,55 @@ private static IHostBuilder Configure(this IHostBuilder builder, StartupAuthenti
...
+                using Stream inputStream = Console.OpenStandardInput();
+                byte[] buffer = new byte[100];
+
+                async Task WatchAsync()
+                {
+                    try
+                    {
+                        while (true)
+                        {
+                            if (await inputStream.ReadAsync(buffer.AsMemory(), cancellationToken) == 0)
```
wiktork, line 167: Since we are not awaiting this task, won't we dispose inputStream as soon as WatchStdinForDisconnect leaves scope? Should WatchAsync be calling Console.OpenStandardInput?

gregg-miskelly: > Since we are not awaiting this task, won't we dispose inputStream as soon as WatchStdinForDisconnect leaves scope?

No, the fact that we aren't awaiting `WatchAsync` just means that if it throws, those errors will be dropped on the floor. But otherwise, `await` has no effect. If dotnet-monitor has a prefered way of marking a task as ignored, I am happy to switch to that. I searched around, but I couldn't find anything except one or two other places where the task was assigned to the discard variable. 

> Should WatchAsync be calling Console.OpenStandardInput?

I didn't want to do that, as any exception from `WatchAsync` is dropped on the floor (since we can't await it). So I wanted to alert the caller to the fact that they specified `--exit-on-stdin-disconnect` but their stdin was corrupt, so it wouldn't work.

gregg-miskelly: Seems like I should probably at least add a comment to line 195 (`_ = WatchAsync()` line).

src/Tools/dotnet-monitor/Commands/CollectCommandHandler.cs
```diff
@@ -143,5 +149,55 @@ private static IHostBuilder Configure(this IHostBuilder builder, StartupAuthenti
...
+                    }
+
+                    lifetime.StopApplication();
+                }
+
+                _ = WatchAsync();
+            }
+            catch (Exception e)
+            {
+                throw new MonitoringException(string.Format(CultureInfo.InvariantCulture, Strings.ErrorMessage_UnableToWatchForDisconnect, e.Message), e);
```
gregg-miskelly, line 199: Oh, I misunderstood what the comment about `MonitoringException` was saying. Sounds like I should be writing the output to standard error?.

src/Tools/dotnet-monitor/Commands/CollectCommandHandler.cs
jander-msft: I think Wiktor's first point is that since you are not awaiting the `Task` returned from `WatchAsync`, then the `WatchStdinForDisconnect` method immediately exits after calling `WatchAsync`. Since `using Stream inputStream = Console.OpenStandardInput();` is in the body of `WatchStdinForDisconnect`, it will be nearly immediately disposed. This also means the try...catch isn't really going to catch much unless it happens in the opening of `WatchStdinForDisconnect` before the first await.

gregg-miskelly: Ah. Thanks. That is what I was attempting with the catch, but certainly not the stream.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/CollectTraceTests.cs
```diff
@@ -82,6 +84,49 @@ public Task StopOnEvent_DoesNotStop_WhenPayloadFieldValueMismatch()
...
+            toolRunner.DiagnosticPortPath = diagnosticPortPath;
+            toolRunner.DisableAuthentication = true;
+            toolRunner.ExitOnStdinDisconnect = true;
+            await toolRunner.StartAsync();
+            Assert.False(toolRunner.HasExited);
+
+            // Close the tool's stdin, which should cause it to exit.
+            toolRunner.StandardInput.Close();
+
+            using (CancellationTokenSource monitorExitCancellationTokenSource = new(TimeSpan.FromSeconds(10)))
```
wiktork, line 118: nit: can you use or add something to src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/TestTimeouts.cs? The test environment can be very slow here so sometimes we have to up wait times.

wiktork (dismissed):

jander-msft (approved):

gregg-miskelly:
I completed testing, so feel free to merge whenever you would like.

PR #4821, opened 2023-07-06
Improve error handling around endpoint callbacks

schmittjoseph:
###### Summary

Currently there exists multiple bad states dotnet-monitor can enter due to an exception being thrown by a registered `IEndpointInfoSourceCallbacks` callback. With this PR, the new behavior for callbacks is best-effort with logging to the user when an endpoint may be in a partially useable state.

Long term we may want to refactor how we track processes to also include metadata on what capabilities are available to better handle partially-initialized endpoints. e.g. for a given process it may have successfully initialized the profiler but not the startup hook, this is something we should likely track. However, that is out of scope of this PR.

The below issues are addressed by this PR:

- When a `OnBeforeResumeAsync` throws an exception
    1.  If the target process is suspended, dotnet-monitor will never resume the process. This is the worst case as it effectively breaks the target app.
    2. dotnet-monitor will repeatedly try initialize the endpoint for the entire lifespan of the target process.
    3. Errors are silently swallowed and never shown to the user so there's no indication of this bad state.
- When a `OnAddedEndpointInfoAsync` throws an exception
    1. dotnet-monitor will repeatedly try initialize the endpoint for the entire lifespan of the target process.
    2. Errors are silently swallowed and never shown to the user  so there's no indication of this bad state.
- When a `OnRemovedEndpointInfoAsync` throws an exception
    1. The endpoint connection is never properly terminated.
    2. Our thread processing endpoint removal is permanently torn down and future endpoints will never be removed, leaking resources associated with all future endpoints.

### DIFF
--- a/src/Tools/dotnet-monitor/LoggingExtensions.cs
+++ b/src/Tools/dotnet-monitor/LoggingExtensions.cs
@@ -511,7 +511,6 @@ namespace Microsoft.Diagnostics.Tools.Monitor
                 eventId: LoggingEventIds.EndpointRemovalFailed.EventId(),
                 logLevel: LogLevel.Debug,
                 formatString: Strings.LogFormatString_EndpointRemovalFailed);
-        
         public static void EgressProviderInvalidOptions(this ILogger logger, string providerName)
         {
             _egressProviderInvalidOptions(logger, providerName, null);


src/Tools/dotnet-monitor/LoggingEventIds.cs
```diff
@@ -106,7 +106,9 @@ internal enum LoggingEventIds
         ExtensionManifestNotParsable = 94,
         ExtensionMalformedOutput = 95,
         EgressProviderTypeNotExist = 96,
-        ProfilerRuntimeIdentifier = 97
+        ProfilerRuntimeIdentifier = 97,
+        EndpointInitializationFailed = 98,
```
wiktork, line 110: Can you start at 99? 98 is taken in https://github.com/dotnet/dotnet-monitor/pull/4792/files, but hasn't been merged to main yet.

src/Tools/dotnet-monitor/EndpointInfo/ServerEndpointInfoSource.cs
```diff
@@ -189,19 +206,29 @@ private async Task ResumeAndQueueEndpointInfo(ReversedDiagnosticsServer server,
...
                 }
                 finally
                 {
                     _activeEndpointsSemaphore.Release();
                 }
             }
-            catch (Exception)
+            catch (Exception ex)
             {
-                server?.RemoveConnection(info.RuntimeInstanceCookie);
```
wiktork, line 202: I'm a bit torn on this approach. If we have a recoverable error, a retry from the app would be useful.

jander-msft:
Would you mind providing an example of what the EndpointInitializationFailed message looks like in the console output when that occurs?

jander-msft (approved):

PR #4826, opened 2023-07-07
Enable operation tracking for trigger actions

wiktork:
###### Summary
Adds collection rule actions to the operations table.

TODO
- Collection rules do not have limits, since there is no obvious action to take when a limit is reached.
- There is no way to distinguish operation types initiated by direct api access vs. collection rules. 
- Associate operations at the /collectionrules endpoint.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationStore.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationStore.cs
@@ -32,6 +32,8 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
             public Guid OperationId { get; set; }
 
             public ISet<string> Tags { get; set; }
+
+            public TaskCompletionSource TaskCompletionSource { get; } = new();
         }
 
         private readonly Dictionary<Guid, EgressEntry> _requests = new();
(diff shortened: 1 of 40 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationStore.cs
```diff
@@ -49,6 +51,23 @@ public bool IsStoppable
...
         }
 
+        public async Task<ExecutionResult<EgressResult>> ExecuteOperation(IEgressOperation egressOperation)
+        {
+            // Collection Rules do not follow request limits.
+
+            Guid operationGuid = await AddOperation(egressOperation, RequestLimitTracker.Unlimited);
+            EgressEntry entry = null;
+
+            lock (_requests)
```
jander-msft, line 61: Could you refactor the existing `AddOperation` method to be `Task<EgressEntry> AddOperationInternal(IEgressOperation, string)` and then add back `Task<Guid> AddOperation(IEgressOperation, string)` and have the `ExecuteOperation` method call `AddOperationInternal` to avoid the second lock and dictionary lookup?

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationStore.cs
```diff
@@ -131,6 +150,8 @@ public void CancelOperation(Guid operationId)
                 entry.State = Models.OperationState.Cancelled;
                 entry.EgressRequest.CancellationTokenSource.Cancel();
                 entry.EgressRequest.Dispose();
+
+                entry.TaskCompletionSource.TrySetResult();
```
jander-msft, line 154:
```suggestion
                entry.TaskCompletionSource.TrySetCanceled();
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/HttpApi/ApiClientExtensions.cs
```diff
@@ -429,6 +429,11 @@ public static async Task<List<OperationSummary>> GetOperations(this ApiClient cl
             return await client.GetOperations(tags, timeoutSource.Token).ConfigureAwait(false);
         }
 
+        public static Task<HttpStatusCode> StopEgressOperation(this ApiClient client, Guid operationId)
+        {
+            return StopEgressOperation(client, new Uri(client.BaseAddress, FormattableString.Invariant($"operations/{operationId}")));
+        }
+
         public static async Task<HttpStatusCode> StopEgressOperation(this ApiClient client, Uri operation)
```
jander-msft, line 437: /todo Refactor ApiClient operation methods to use identifiers rather than raw URIs

jander-msft (dismissed):

jander-msft (approved):

PR #4829, opened 2023-07-07
Stop spawned test processes when testing stops

schmittjoseph:
###### Summary

When launching test processes, configure them to automatically exit whenever the test host stops. This fixes an issue where when running tests locally and you stop them early (or the test timeouts), multiple processes are left running. To achieve this, automatically inject a new startup hook into all spawned dotnet processes 

Note: This does not handle non-dotnet apps (such as `azurite`) or the scenario where we spawn a suspended process and testing stops before we have a chance to resume that process. These can be addressed in a separate PR.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/DotNetRunner.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/DotNetRunner.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Globalization;
 using System.IO;
 using System.Reflection;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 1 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/DotNetRunner.cs
```diff
@@ -93,6 +96,18 @@ public sealed class DotNetRunner : IDisposable
         /// </summary>
         public bool WaitForDiagnosticPipe { get; set; }
 
+        /// <summary>
+        /// Determines if the spawned process should be stopped when the currently executing process exits.
+        /// </summary>
+        public bool StopProcessOnExit { get; set; } = true;
```
jander-msft, line 102: `StopOnParentExit` to be more clear was to what this is doing.

jander-msft (approved):

PR #4847, opened 2023-07-11
Add Helix project and supporting targets

jander-msft:
###### Summary

These changes enable the use of Helix for executing tests on a multitude of platforms and architectures. Helix testing is enabled through the use of the `/eng/helix/Helix.proj` project and can be executed on the command line using `build.cmd/sh -projects ./eng/helix/Helix.proj -test -skipnative -skipmanaged` after a successful full build of both native and managed components.

###### Details

- The `/eng/helix/Helix.proj` project gathers the executable work item information (e.g. `dotnet test` command line) from each participating project (e.g. test projects).
- These work items (e.g. tests) are executed in a set of queues. There are several queues that are employed to get broad coverage of scenarios where .NET Monitor is used; the set of queues on which the work items are executed are determined by the build target environment e.g. a Windows x64 host can launch arm64, x64, and x86 tests on Windows queues depending on which target architecture was built.
- Test results are automatically uploaded to Azure Pipelines if this project is invoked within an Azure Pipelines build (no work was done here since this is automatically supported by the Helix SDK).
- All existing enabled tests are runnable and succeed through this project EXCEPT for the Azure blob storage tests that require Azurite to be installed; these tests are automatically skipped since Azurite is not initialized and the environment variable to indicate they should not be skipped is not applied. Enabling this is still a work in progress.
- The pipelines have not been update to use this project yet; that will come as a separate PR.

Example build where E2E testing via Helix is enabled: https://dev.azure.com/dnceng/internal/_build/results?buildId=2218286&view=results

### DIFF
--- a/eng/helix/Helix.proj
+++ b/eng/helix/Helix.proj
@@ -1,0 +1,164 @@
+<Project Sdk="Microsoft.DotNet.Helix.Sdk"
+         DefaultTargets="Test">
+
+  <PropertyGroup>
+    <HelixConfiguration>$(Configuration)</HelixConfiguration>
+    <HelixArchitecture>$(BuildArch)</HelixArchitecture>
+    <UseOpenQueues>true</UseOpenQueues>
+    <UseOpenQueues Condition="'$(HelixAccessToken)' != ''">false</UseOpenQueues>
+    <!-- Ensure dotnet correlation payloads match the tested RID -->
+    <DotNetCliRuntime>$(PackageRid)</DotNetCliRuntime>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(UseOpenQueues)' == 'true'">
+    <QueueSuffix>.open</QueueSuffix>
+    <!-- Open queues require a creator to be set. -->
+    <Creator Condition=" '$(USERNAME)' != '' ">$(USERNAME)</Creator>
+    <Creator Condition=" '$(USER)' != '' ">$(USER)</Creator>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(SYSTEM_COLLECTIONURI)' == 'https://dev.azure.com/dnceng/'">
+    <HelixSourcePrefix Condition="'$(BUILD_REASON)' == 'Manual'">pr</HelixSourcePrefix>
+    <HelixSourcePrefix Condition="'$(BUILD_REASON)' == 'PullRequest'">pr</HelixSourcePrefix>
+    <HelixSourcePrefix Condition="'$(HelixSourcePrefix)' == ''">official</HelixSourcePrefix>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <HelixSourcePrefix Condition="'$(HelixSourcePrefix)' == ''">pr</HelixSourcePrefix>
+    <HelixSource>$(HelixSourcePrefix)/dotnet/dotnet-monitor</HelixSource>
+    <HelixSource Condition="'$(BUILD_SOURCEBRANCH)' != ''">$(HelixSource)/$(BUILD_SOURCEBRANCH)</HelixSource>
+    <HelixType>test/binaries/</HelixType>
+  </PropertyGroup>
+
+  <ItemDefinitionGroup>
+    <HelixAvailableTargetQueue>
+      <AdditionalProperties>Platform=$(HelixArchitecture)</AdditionalProperties>
+    </HelixAvailableTargetQueue>
+  </ItemDefinitionGroup>
+
+  <!-- List of all available Helix machines: https://helix.dot.net/ -->
+
+  <!-- Windows queues -->
+  <ItemGroup Condition="$([MSBuild]::IsOSPlatform(Windows))">
+    <!-- win-arm64 -->
+    <HelixAvailableTargetQueue Include="windows.11.arm64$(QueueSuffix)"
+                               Condition="'$(HelixArchitecture)' == 'arm64'">
+      <TestRunName>Windows 11 $(HelixArchitecture) $(HelixConfiguration)</TestRunName>
+    </HelixAvailableTargetQueue>
+    <!-- win-x64 -->
+    <HelixAvailableTargetQueue Include="windows.10.amd64.client$(QueueSuffix)"
+                               Condition="'$(HelixArchitecture)' == 'x64'">
+      <TestRunName>Windows 10 $(HelixArchitecture) $(HelixConfiguration)</TestRunName>
+    </HelixAvailableTargetQueue>
+    <!-- win-x86 -->
+    <!-- Windows x64 clients can run dotnet x86 natively -->
+    <HelixAvailableTargetQueue Include="windows.10.amd64.client$(QueueSuffix)"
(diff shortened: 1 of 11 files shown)


src/Tests/Helix.targets
```diff
@@ -0,0 +1,17 @@
+<Project>
+  <PropertyGroup Condition="'$(TargetFramework)' != ''">
+    <HelixDotnetTestArgs>$(TargetName) $(Configuration) $(TargetFramework) $(Platform) $(TestRunnerFilterArguments)</HelixDotnetTestArgs>
+    <HelixDotnetTestCommand Condition="'$(IsHelixPosixShell)' != 'true'">call %HELIX_CORRELATION_PAYLOAD%\dotnettest.cmd $(HelixDotnetTestArgs)</HelixDotnetTestCommand>
```
jander-msft, line 4: Instead of having a bash and cmd script that do the same thing, let's try to use a single script type that is available on all targets e.g. python.

eng/helix/payload/dotnettest.cmd
```diff
@@ -0,0 +1,33 @@
...
+
+echo "Start tests..."
+
+dotnet.exe test ^
+  "%HELIX_CORRELATION_PAYLOAD%\%testAssembly%\%configuration%\%targetFramework%\%testAssembly%.dll" ^
+  --logger:"console;verbosity=normal" ^
+  --logger:"trx;LogFileName=%testAssembly%_%targetFramework%_%architecture%.trx" ^
+  --logger:"html;LogFileName=%testAssembly%_%targetFramework%_%architecture%.html" ^
+  --ResultsDirectory:%HELIX_WORKITEM_UPLOAD_ROOT% ^
+  --blame "CollectHangDump;TestTimeout=15m" ^
```
jander-msft, line 24: Can this be parameterized so that it can be shared among all types of test runs?

src/Tests/Helix.targets
```diff
@@ -0,0 +1,17 @@
...
+    <HelixDotnetTestCommand Condition="'$(IsHelixPosixShell)' == 'true'">$HELIX_CORRELATION_PAYLOAD/dotnettest.sh $(HelixDotnetTestArgs)</HelixDotnetTestCommand>
+  </PropertyGroup>
+  
+  <ItemGroup Condition="'$(TargetFramework)' != ''">
+    <!-- Helix work item that calls "dotnet test" for the test assembly. -->
+    <HelixWorkItem Include="$(MSBuildProjectName)-$(TargetFramework)-$(Platform)">
+      <PreCommands Condition="'$(HelixTargetQueue)' != ''">@(HelixWorkItemPreCommand)</PreCommands>
+      <Command Condition="'$(HelixTargetQueue)' != ''">$(HelixDotnetTestCommand)</Command>
+      <PostCommands Condition="'$(HelixTargetQueue)' != ''">@(HelixWorkItemPostCommand)</PostCommands>
+      <Timeout>00:30:00</Timeout>
```
jander-msft, line 14: Add comment why this value was chosen

src/Tests/Helix.targets
jander-msft: Will consider this separately.

eng/helix/Helix.proj
```diff
@@ -0,0 +1,164 @@
...
+    <UseOpenQueues>true</UseOpenQueues>
+    <UseOpenQueues Condition="'$(HelixAccessToken)' != ''">false</UseOpenQueues>
+    <!-- Ensure dotnet correlation payloads match the tested RID -->
+    <DotNetCliRuntime>$(PackageRid)</DotNetCliRuntime>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(UseOpenQueues)' == 'true'">
+    <QueueSuffix>.open</QueueSuffix>
+    <!-- Open queues require a creator to be set. -->
+    <Creator Condition=" '$(USERNAME)' != '' ">$(USERNAME)</Creator>
```
kkeirstead, line 16: Just a quick sanity check (I don't have enough context here to know) - is USERNAME or USER guaranteed to be non-empty (if not, based on the comment I'm assuming this wouldn't work), and USER is prioritized over USERNAME if both are present (or is it only one-or-the-other)?

eng/helix/Helix.proj
```diff
@@ -0,0 +1,164 @@
...
+  </ItemGroup>
+
+  <!-- Linux gnu libc queues -->
+  <ItemGroup Condition="$([MSBuild]::IsOSPlatform(Linux)) and !$(PackageRid.Contains(musl))">
+    <!-- linux-arm64 -->
+    <HelixAvailableTargetQueue Include="ubuntu.1804.armarch$(QueueSuffix)@mcr.microsoft.com/dotnet-buildtools/prereqs:debian-11-helix-arm64v8"
+                               Condition="'$(HelixArchitecture)' == 'arm64'">
+      <TestRunName>Debian 11 $(HelixArchitecture) $(HelixConfiguration)</TestRunName>
+    </HelixAvailableTargetQueue>
+    <HelixAvailableTargetQueue Include="ubuntu.1804.armarch$(QueueSuffix)@mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-helix-arm64v8"
```
kkeirstead, line 68: I know you mentioned this the other day that 18.04 was realistically the oldest version we should be running against - would there be any benefit to also running against a newer version (and I think the same question goes for macOS, but for an older version)?

eng/helix/Helix.proj
jander-msft: The `USERNAME` property comes from the Windows environment variable `USERNAME` and the `USER` property comes from the non-Windows environment variable `USER`. While these aren't explicitly mutually exclusive, they typically are and one of them should be set regardless of which environment this is running.

eng/helix/Helix.proj
jander-msft: I think that's a reasonable suggestion (use the oldest supported and newest support version of a distro). I'm going to hold off on this for now and get the tests running first so that way we can see how fast these things run in practice; we might even want to limit the set for PRs, do more for CI, and all variations on a scheduled basis.

jander-msft: #4911

wiktork (approved):

eng/helix/payload/dotnettest.cmd
```diff
@@ -0,0 +1,34 @@
...
+if not "%~6" == "" (
+   set filterArgs=--filter ^"%~6^"
+)
+
+set exit_code=0
+
+echo "Start tests..."
+
+dotnet.exe test ^
+  "%HELIX_CORRELATION_PAYLOAD%\%testAssembly%\%configuration%\%targetFramework%\%testAssembly%.dll" ^
```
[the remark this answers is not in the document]
jander-msft: It's not included at this time; I haven't seen perf bottlenecks while running via Helix so I'm going to not include it for now and investigate the impact it has on including it.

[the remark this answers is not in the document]
jander-msft: Following up on this... this file should already be included next to the test assemblies. This happens as part of build, which didn't change when enabling Helix testing.

PR #4866, opened 2023-07-11
Refactor in-proc eventing code for re-use

schmittjoseph:
###### Summary

Move reusable in-proc eventing code belonging to the exceptions feature into a sharable `AbstractMonitorEventSource` class. The intent of this class is to be re-used by parameter capturing.

The only functional difference is that `WriteEventCore` takes care of calling `RestartFlushingEventTimer` instead of each individual event.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/Eventing/AbstractMonitorEventSource.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/Eventing/AbstractMonitorEventSource.cs
@@ -60,14 +60,14 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.Eventing
         }
 
         [NonEvent]
-        protected unsafe void WriteEventCore(int eventId)
+        protected unsafe void WriteEventWithFlushing(int eventId)
         {
             WriteEvent(eventId);
             RestartFlushingEventTimer();
         }
 
         [NonEvent]
-        protected unsafe void WriteEventCore(int eventId, Span<EventData> data)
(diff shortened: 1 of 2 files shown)


src/Microsoft.Diagnostics.Monitoring.StartupHook/Eventing/AbstractMonitorEventSource.cs
```diff
@@ -0,0 +1,170 @@
...
+        [NonEvent]
+        private void RestartFlushingEventTimer()
+        {
+            // This will reset the timer to fire after the specified time period. If the timer is already
+            // started, it will be reset. If it already finished, it will be started again.
+            _flushEventsTimer.Change(EventSourceBufferAvoidanceTimeout, Timeout.InfiniteTimeSpan);
+        }
+
+        [NonEvent]
+        protected unsafe void WriteEventCore(int eventId)
```
jander-msft, line 63: This should not be necessary; there's already an `EventSource.WriteEvent(int)` method that can be called by deriving classes.

jander-msft (dismissed):

src/Microsoft.Diagnostics.Monitoring.StartupHook/Eventing/AbstractMonitorEventSource.cs
jander-msft: > I can rename these methods to `WriteEventWithFlushing` if that'd be more clear

That sounds good to me.

jander-msft (approved):

PR #4868, opened 2023-07-11
Use portable runtime identifier for profiler enablement when available

jander-msft:
###### Summary

If the process host reports a portable runtime identifier, use that value to determine which profiler variant to use. Additionally, log how the profiler runtime identifier was determined (`ProcessEnvironment`, `ProcessHost`, `ProcessImplicit`).

With these changes, the .NET 8 test applications have their profiler variant determined by the portable runtime identifier (since they respond to the `ProcessInfo3` command) and thus report as `ProcessHost` whereas the lower .NET versions fallback to the `ProcessImplicit` determination.

### DIFF
--- a/src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
+++ b/src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
@@ -54,50 +54,59 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Profiler
                 DiagnosticsClient client = new DiagnosticsClient(endpointInfo.Endpoint);
                 Dictionary<string, string> env = await client.GetProcessEnvironmentAsync(cancellationToken);
 
+                string runtimeIdentifierSource = "ProcessEnvironment";
                 if (!env.TryGetValue(ToolIdentifiers.EnvironmentVariables.RuntimeIdentifier, out string runtimeIdentifier))
                 {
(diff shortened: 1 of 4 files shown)


src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
```diff
@@ -54,50 +54,59 @@ public async Task ApplyProfiler(IEndpointInfo endpointInfo, CancellationToken ca
...
                 if (!env.TryGetValue(ToolIdentifiers.EnvironmentVariables.RuntimeIdentifier, out string runtimeIdentifier))
                 {
                     ProcessInfo processInfo = await client.GetProcessInfoAsync(cancellationToken);
 
-                    // This is mostly correct, except that "arm" and "armv6" are both reported as "arm32".
-                    string ridArchitecture = processInfo.ProcessArchitecture;
-                    Debug.Assert(!"arm32".Equals(ridArchitecture, StringComparison.Ordinal), "Unable to distinguish arm from armv6");
+                    // Use portable runtime identifier as reported by the runtime
+                    runtimeIdentifier = processInfo.PortableRuntimeIdentifier;
+                    runtimeIdentifierSource = "ProcessHost";
```
wiktork, line 64: nit: Should these be consts somewhere?

wiktork (approved):

PR #4876, opened 2023-07-12
[Parameter Capturing] Fix native integer parameter handling

schmittjoseph:
###### Summary

Native integers (IntPtr,UIntPtr) were incorrectly being treated as standard boxable parameters. Fix the boxing token logic to identify these and mark them as unsupported for now.

Note:
- IntPtr and UIntPtr are marked as primitive types by .net but with a TypeCode of `Object` so use this for detection.


jander-msft (approved):
I'm fine with carving this out a unsupported for now so long as we have an issue to track adding support for it later. If you want to fix it now, I'll take another look too.

PR #4883, opened 2023-07-12
Fix Collection Rule Trigger Type Console Spew

kkeirstead:
###### Summary

Currently, `dotnet monitor` emits the `CollectionRuleTriggerType` repeatedly in logs:

CollectionRuleTriggerTypeDuplication

This can lead to huge amounts of console spew as this continues to grow every time the collection rule is triggered. The issue appears to have been tied to `triggerScopeRegistration` failing to be properly disposed in a loop.

###### Release Notes Entry

Prevent overreporting of the collection rule trigger type in console output


jander-msft (approved):

jander-msft:
Does this need to be backported? We should also call this out in the release notes.

kkeirstead:
> Does this need to be backported? We should also call this out in the release notes.

I believe so - would that be release 7.1, 7.2, 7.x, and 8.x? I just added the update release notes label.

jander-msft:
> > Does this need to be backported? We should also call this out in the release notes.
> 
> I believe so - would that be release 7.1, 7.2, 7.x, and 8.x? I just added the update release notes label.

8.x will automatically get it from the merge from main; 7.1, 7.2, and 7.x would need it; check if the lack of the `using` exists and 6.x and backport it there too if needed

PR #4899, opened 2023-07-13
Compatibility with 6/7 Runtimes For Multiple `System.Diagnostics.Metrics` Sessions

kkeirstead:
###### Summary

The core change is [here](https://github.com/dotnet/diagnostics/pull/3889) - this is the only piece that requires a change to the `dotnet monitor` repo for now, but the changes in the diagnostics repo do apply for Metrics/LiveMetrics/Triggers. For more context on the problem this is trying to solve, see [here](https://github.com/dotnet/diagnostics/pull/3889#issuecomment-1629262044).

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsService.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsService.cs
@@ -58,6 +58,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
                     using IDisposable monitorListener = _optionsMonitor.OnChange((_, _) => optionsTokenSource.SafeCancel());
 
                     MetricsPipelineSettings counterSettings = MetricsSettingsFactory.CreateSettings(counterOptions, options);
+                    counterSettings.UseSharedSession = pi.EndpointInfo.RuntimeVersion.Major >= 8;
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsService.cs
```diff
@@ -58,6 +58,7 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
                     using IDisposable monitorListener = _optionsMonitor.OnChange((_, _) => optionsTokenSource.SafeCancel());
 
                     MetricsPipelineSettings counterSettings = MetricsSettingsFactory.CreateSettings(counterOptions, options);
+                    counterSettings.UseSharedSession = pi.EndpointInfo.RuntimeVersion.Major >= 8;
```
kkeirstead, line 61: This feels a little clunky by not wiring it through `CreateSettings` (same with MetricsOperationFactory), but if we make that method responsible for knowing the runtime version that'll impact a lot of callers. What I'm doing here (I think) is the most compact option that pushes the logic into `dotnet-monitor`, but then we need to manually set `UseSharedSession` on settings, which I guess I'm not sure if that's something we care about.

jander-msft (approved):

PR #4901, opened 2023-07-13
Add documentation for exceptions history feature

jander-msft:
###### Summary

Adds basic documentation for the exceptions history feature and fixes the inner exceptions placement within the exception instance JSON.

I've intentionally left out how to set `DOTNET_STARTUP_HOOKS` for now and will follow up with a subsequent PR to add that.

###### Release Notes Entry

Add first chance exceptions history feature and `/exceptions` route.


documentation/configuration/in-process-features-configuration.md
```diff
@@ -108,9 +109,81 @@ Similarly, the call stacks feature can be individually disabled by setting the s
...
+
+Similarly, the exceptions history feature can be individually disabled by setting the same property to `false`:
+
+<details>
+  <summary>JSON</summary>
+
+  ```json
+  {
+    "InProcessFeatures": {
+      "Enabled": true,
```
kkeirstead, line 162: Is it worth calling out what happens if you explicitly set `InProcessFeatures:Enabled = false`? If we add this it also might be better to live a bit higher up in the doc in the `InProcessFeatures` section

kkeirstead (approved):

wiktork (approved):

documentation/configuration/in-process-features-configuration.md
jander-msft: I'll add that in the next PR

PR #4903, opened 2023-07-13
Verify shared library files integrity

jander-msft:
###### Summary

Change the shared library initializer to validate the files in the existing shared directory or write new files instead of relying on a sentinel file and keep the file handles open with read-only shared access. This allows .NET Monitor to confidently share libraries that will be loaded into target applications and know that they have not been tampered with between runs of .NET Monitor. If any file is different than expected, shared library initialization will fail and any feature that is dependent on it will not work.

Example console output:
```
16:58:51 info: Microsoft.Diagnostics.Tools.Monitor.Startup[60]
      Tell us about your experience with dotnet monitor: https://aka.ms/dotnet-monitor-survey
16:58:51 info: Microsoft.Diagnostics.Tools.Monitor.Startup[77]
      Connection mode: Listen at '\\.\pipe\diagport'
16:58:51 warn: Microsoft.Diagnostics.Tools.Monitor.Startup[13]
      WARNING: Authentication has been disabled. This can pose a security risk and is not intended for production environments.
16:58:52 info: Microsoft.Hosting.Lifetime[14]
      Now listening on: https://localhost:49260
16:58:52 info: Microsoft.Hosting.Lifetime[14]
      Now listening on: http://localhost:49261
16:58:52 fail: Microsoft.Diagnostics.Tools.Monitor.LibrarySharing.SharedLibraryService[73]
      Failed to initialize shared library storage. Some monitoring features may be unavailable.
      System.InvalidOperationException: The shared file 'C:\Users\jander\dnmshared\libs\8.0.0-dev.23363.1\any\net6.0\Microsoft.Diagnostics.Monitoring.StartupHook.dll' is different from source file 'C:\src\dotnet\dotnet-monitor\artifacts\bin\dotnet-monitor\Debug\net6.0\shared\any\net6.0\Microsoft.Diagnostics.Monitoring.StartupHook.dll'.
         at Microsoft.Diagnostics.Tools.Monitor.LibrarySharing.DefaultSharedLibraryInitializer.Initialize() in C:\src\dotnet\dotnet-monitor\src\Tools\dotnet-monitor\LibrarySharing\DefaultSharedLibraryInitializer.cs:line 123
         at Microsoft.Diagnostics.Tools.Monitor.LibrarySharing.SharedLibraryService.ExecuteAsync(CancellationToken stoppingToken) in C:\src\dotnet\dotnet-monitor\src\Tools\dotnet-monitor\LibrarySharing\SharedLibraryService.cs:line 47
16:58:52 info: Microsoft.Hosting.Lifetime[0]
```

### DIFF
--- a/src/Tools/dotnet-monitor/LibrarySharing/DefaultSharedLibraryInitializer.cs
+++ b/src/Tools/dotnet-monitor/LibrarySharing/DefaultSharedLibraryInitializer.cs
@@ -5,20 +5,28 @@ using Microsoft.Diagnostics.Monitoring.WebApi;
 using Microsoft.Extensions.FileProviders;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
+using Microsoft.Win32.SafeHandles;
 using System;
+using System.Buffers;
+using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Reflection;
+using System.Security.Cryptography;
 
 namespace Microsoft.Diagnostics.Tools.Monitor.LibrarySharing
 {
     internal sealed class DefaultSharedLibraryInitializer :
-        ISharedLibraryInitializer
+        ISharedLibraryInitializer,
+        IDisposable
     {
         private static readonly string SharedLibrarySourcePath = Path.Combine(AppContext.BaseDirectory, "shared");
 
+        private readonly List<SafeFileHandle> _sharedFileHandles = new();
         private readonly ILogger<DefaultSharedLibraryInitializer> _logger;
         private readonly string _sharedLibraryTargetPath;
+
+        private long _disposeState;
 
         public DefaultSharedLibraryInitializer(
             IOptions<StorageOptions> _storageOptions,
@@ -26,6 +34,19 @@ namespace Microsoft.Diagnostics.Tools.Monitor.LibrarySharing
         {
             _logger = logger;
             _sharedLibraryTargetPath = _storageOptions.Value.SharedLibraryPath;
+        }
+
+        public void Dispose()
+        {
+            if (!DisposableHelper.CanDispose(ref _disposeState))
+                return;
+
+            // Release locks on shared file handles
+            foreach (SafeFileHandle handle in _sharedFileHandles)
+            {
+                handle.Dispose();
+            }
+            _sharedFileHandles.Clear();
         }
 
         public IFileProviderFactory Initialize()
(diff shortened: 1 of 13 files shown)


src/Tools/dotnet-monitor/LibrarySharing/DefaultSharedLibraryInitializer.cs
```diff
@@ -86,6 +152,22 @@ public IFileProviderFactory Initialize()
             return new Factory(sharedLibraryPath);
         }
 
+        private static int ComputeHashAndLength(SafeFileHandle handle, Span<byte> destination)
+        {
+            int length = (int)RandomAccess.GetLength(handle);
```
wiktork, line 157: Consider separating the length and hash checks. If the lengths don't match, there is no need to compute the hash.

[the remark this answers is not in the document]
jander-msft: How does that benefit this though? Also, the current usage allows me to use a stack allocated array for the hash values whereas those chunk reading APIs return a new array for each computation.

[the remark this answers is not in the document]
jander-msft: I'm think that instead of hashing both the source and target file (which requires reading of the entire content and computing a hash value for each) that I change this to doing some type of chunked simultaneous reading of the content of the source and target files and comparing the chunks; if any chunk of data is different, then immediately fail. I think this should be faster in general and fails faster in the failure case.

src/Tools/dotnet-monitor/StreamExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+
+                    // Fill buffer; read will only be less than buffer size
+                    // if encounter end of stream.
+                    int otherRead = await otherStream.ReadAtLeastAsync(
+                        otherMemory,
+                        otherMemory.Length,
+                        throwOnEndOfStream: false,
+                        cancellationToken);
+
+                    if (thisRead != otherRead || !thisMemory.Span.SequenceEqual(otherMemory.Span))
```
wiktork, line 50: can you make sure this works properly for fileSize % BufferSize == 0 scenarios?

src/Tools/dotnet-monitor/StreamExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+using System.Buffers;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    internal static class StreamExtensions
+    {
+        private const int DefaultBufferSize = 4096;
```
wiktork, line 14: Did you want to match the default FileStream buffer size here (4KiB, which coincides with the default page size)? Also consider being explicit about the FileStream buffer size.

wiktork (dismissed):

src/Tools/dotnet-monitor/StreamExtensions.cs
jander-msft: The default buffer size for FileStream is 4096: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/IO/FileStream.cs#L15

src/Tools/dotnet-monitor/StreamExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    internal static class StreamExtensions
+    {
+        public static async Task<bool> HasSameContentAsync(
+            this Stream thisStream,
+            Stream otherStream,
+            CancellationToken cancellationToken)
+        {
+            byte[] thisBuffer = ArrayPool<byte>.Shared.Rent(StreamDefaults.BufferSize);
```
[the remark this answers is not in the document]
jander-msft: In general, you cannot do that. Some streams are not seekable even though you can read through them sequentially e.g. `PipeStream` and certain `FileStream` implementations; non-seekable streams will throw if you access the Length property.

wiktork (approved):

PR #4912, opened 2023-07-14
Add Check For Existing Key For `ApplyStartupHook`

kkeirstead:
###### Summary

After talking to @schmittjoseph , `OnBeforeResumeAsync` could be called multiple times for the same endpoint. Without this check, it'll repeatedly try to apply the startup hook, even if it already applied (or failed) earlier.


jander-msft:
> After talking to @schmittjoseph , `OnBeforeResumeAsync` could be called multiple times for the same endpoint. Without this check, it'll repeatedly try to apply the startup hook, even if it already applied (or failed) earlier.

So then shouldn't this check be added to the very beginning of the method so it can short circuit the entire path rather than just on the ApplyStartupHook?

PR #4933, opened 2023-07-18
[Parameter Capturing] Add http api

schmittjoseph:
###### Summary

Add an HTTP API for capturing parameters in a monitored process. Captured parameters are logged using the process's own ILogger.

#### Overview of Changes
- Add a new HTTP api `POST /parameters` for capturing parameters in a target application. The request is tracked via our `/operations` and is also stoppable on-demand.
  - Requests can optionally include a duration for how long to capture parameters (or indefinitely). Durations are handled in the target process, which ensures reliability if we lose connection to the process and also allows us to add more complex stop-conditions in the future. 
  - Requests for a given method are described using a new `MethodDescription` model and will fuzzily match in the case of multiple matches.
  - Currently requests are all-or-nothing, if a single `MethodDescription` fails to resolve, the entire request is discarded.
  - Errors that occur with the parameter capturing operation inside the target process are recorded in operations table.
- Add a new concept of in-process egress operations.
  - Intentionally avoid the wordage of "egressed artifact" when an operation completes to make it clear that the in-process artifact did not leave the target process. 
- Add the concept of monitored processes announcing what in-process infrastructure is available for use (e.g. did the hosting startup load in this process? Is managed messaging available?). This is done via environment variables and allows the parameter capturing operations to check if a target process is capable of responding to our requests.
- `ParameterCapturingService` now receives requests via our `ProfilerMessage`+`MonitorMessageDispatcher` infrastructure and keeps dotnet-monitor notified of a request's status via a new event source (`ParameterCapturingEventSource`).

#### Example Requests and Output

<details><summary>Successful request</summary>

##### Request
```sh
curl -X 'POST' \
  'https://localhost:52323/parameters?durationSeconds=30' \
  -H 'accept: */*' \
  -H 'Content-Type: application/json' \
  -d '[
  {
    "assemblyName": "BuggyDemoWeb",
    "typeName": "BuggyDemoWeb.Controllers.HomeController",
    "methodName": "SelectDiagnsticType"
  }
]'
```
##### Logs: dotnet-monitor
```csharp
14:47:12 info: Microsoft.Diagnostics.Monitoring.WebApi.InProcessEgressOperation[18]
      => ArtifactType:parameters ArtifactSource_ProcessId:22096 ArtifactSource_RuntimeInstanceCookie:d7ded3ae8f5740a5ac4206586d36a4b2
      Finished generating in-process artifact
```

##### Logs: Target app
```csharp
info: Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing.ParameterCapturingService[0]
      Started parameter capturing for 00:00:30 on 1 method(s).
...
info: Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing.ParameterCapturingService[0]
      BuggyDemoWeb.Controllers.HomeController.SelectDiagnsticType(
        this: 'BuggyDemoWeb.Controllers.HomeController',
        route: 'exception/port-exhaustion')
...
info: Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing.ParameterCapturingService[0]
      Stopped parameter capturing.
```
</details>

<details><summary>Bad request</summary>

##### Request
```sh
curl -X 'POST' \
  'https://localhost:52323/parameters?durationSeconds=30' \
  -H 'accept: */*' \
  -H 'Content-Type: application/json' \
  -d '[
  {
    "assemblyName": "BuggyDemoWeb",
    "typeName": "BuggyDemoWeb.Controllers.HomeController",
    "methodName": "SelectDiagnsticType2"
  },
  {
    "assemblyName": "BuggyDemoWeb",
    "typeName": "BuggyDemoWeb.Controllers.HomeController",
    "methodName": "SelectDiagnsticType3"
  }
]'
```
##### Logs: dotnet-monitor
```csharp
14:41:14 fail: Microsoft.Diagnostics.Monitoring.WebApi.InProcessEgressOperation[1]
      => ArtifactType:parameters ArtifactSource_ProcessId:26020 ArtifactSource_RuntimeInstanceCookie:5d7c4270511d4cf3aea84d59274bad9e
      Request failed.
      Microsoft.Diagnostics.Monitoring.MonitoringException: Unable to resolve one or more methods:
      --> BuggyDemoWeb.dll!BuggyDemoWeb.Controllers.HomeController.SelectDiagnsticType2
      --> BuggyDemoWeb.dll!BuggyDemoWeb.Controllers.HomeController.SelectDiagnsticType3

         at Microsoft.Diagnostics.Tools.Monitor.ParameterCapturing.CaptureParametersOperation.ExecuteAsync(TaskCompletionSource`1 startCompletionSource, CancellationToken token) in C:\Users\joschmit\work\dotnet-monitor\src\Tools\dotnet-monitor\ParameterCapturing\CaptureParametersOperation.cs:line 101
         at Microsoft.Diagnostics.Tools.Monitor.ParameterCapturing.CaptureParametersOperation.ExecuteAsync(TaskCompletionSource`1 startCompletionSource, CancellationToken token) in C:\Users\joschmit\work\dotnet-monitor\src\Tools\dotnet-monitor\ParameterCapturing\CaptureParametersOperation.cs:line 104
         at Microsoft.Diagnostics.Monitoring.WebApi.InProcessEgressOperation.<>c__DisplayClass15_0.<<ExecuteAsync>b__0>d.MoveNext() in C:\Users\joschmit\work\dotnet-monitor\src\Microsoft.Diagnostics.Monitoring.WebApi\Operation\InProcessEgressOperation.cs:line 41
      --- End of stack trace from previous location ---
         at Microsoft.Diagnostics.Monitoring.WebApi.ExecutionHelper.InvokeAsync[T](Func`2 action, ILogger logger, CancellationToken token) in C:\Users\joschmit\work\dotnet-monitor\src\Microsoft.Diagnostics.Monitoring.WebApi\ActionContextExtensions.cs:line 45
```
##### Logs: Target app
```csharp
warn: Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing.ParameterCapturingService[0]
      Unable to resolve one or more methods: 
      --> BuggyDemoWeb.dll!BuggyDemoWeb.Controllers.HomeController.SelectDiagnsticType2
      --> BuggyDemoWeb.dll!BuggyDemoWeb.Controllers.HomeController.SelectDiagnsticType3
```

</details>

#### Out of scope of this PR
- Functional tests.
- A deny list for instrumenting certain methods (e.g. debugger-related).
- Limit usage to .NET 7+ target applications until https://github.com/dotnet/runtime/issues/88924 is fixed in .NET 6 (if it is).

### DIFF
--- a/src/Tools/dotnet-monitor/ParameterCapturing/CaptureParametersOperation.cs
+++ b/src/Tools/dotnet-monitor/ParameterCapturing/CaptureParametersOperation.cs
@@ -92,13 +92,14 @@ namespace Microsoft.Diagnostics.Tools.Monitor.ParameterCapturing
 
                 await _capturingStoppedCompletionSource.Task.WaitAsync(token).ConfigureAwait(false);
             }
-            catch (OperationCanceledException)
+            catch (OperationCanceledException ex)
             {
+                _ = startCompletionSource?.TrySetException(ex);
                 _ = _capturingStartedCompletionSource.TrySetCanceled(token);
                 _ = _capturingStoppedCompletionSource.TrySetCanceled(token);
 
                 using CancellationTokenSource stopCancellationToken = new(TimeSpan.FromSeconds(30));
-                await StopAsync(stopCancellationToken.Token);
+                await StopAsync(stopCancellationToken.Token).SafeAwait();
                 throw;
             }
         }


src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/MethodResolver.cs
```diff
@@ -0,0 +1,127 @@
...
+            {
+                try
+                {
+                    MethodInfo[]? allMethods = module.Assembly.GetType(methodDescription.ClassName)?.GetMethods(
+                        BindingFlags.Public |
+                        BindingFlags.NonPublic |
+                        BindingFlags.Instance |
+                        BindingFlags.Static);
+
+                    if (allMethods == null)
```
kkeirstead, line 110: I'm assuming this is just a sanity check since I believe `GetMethods` returns an empty array if there are no matches?

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/MethodResolver.cs
```diff
@@ -0,0 +1,127 @@
...
+                ClassName = methodDescription.ClassName
+            };
+
+            if (_declaringTypeToMethods.TryGetValue(declType, out List<MethodInfo>? methods))
+            {
+                return methods;
+            }
+
+            List<MethodInfo> classMethods = new();
+            if (!_dllNameToModules.TryGetValue(methodDescription.ModuleName, out List<Module>? possibleModules))
```
kkeirstead, line 94: nit: Not necessarily a recommendation, just noticed some duplication since this `if` block is the same as what we have at the bottom of this method - the `foreach` loop will only execute if `possibleModules` contains elements, which suggests we could do away with this conditional and just let everything return at the bottom.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingService.cs
```diff
@@ -1,81 +1,271 @@
...
+                UnresolvedMethodsExceptions ex = new(methodsFailedToResolve);
+                _logger!.LogWarning(ex.Message);
+                throw ex;
+            }
+
+            _probeManager!.StartCapturing(methods);
+            _logger!.LogInformation(
+                ParameterCapturingStrings.StartParameterCapturingFormatString,
+                request.Duration,
+                request.Methods.Length);
```
kkeirstead, line 171: Should this be `request.Methods.Length` or `methods.length`? In the event that there are multiple `resolvedMethods` for a single `request.Method`

src/Tools/dotnet-monitor/ParameterCapturing/CaptureParametersOperation.cs
```diff
@@ -0,0 +1,143 @@
...
+        public CaptureParametersOperation(IEndpointInfo endpointInfo, ProfilerChannel profilerChannel, ILogger logger, MethodDescription[] methods, TimeSpan duration)
+        {
+            _profilerChannel = profilerChannel;
+            _endpointInfo = endpointInfo;
+            _logger = logger;
+            _methods = methods;
+            _duration = duration;
+        }
+
+        private async Task<bool> IsEndpointIsCapableOfProcessingRequestsAsync(CancellationToken token)
```
kkeirstead, line 36: nit: Probably either `IsEndpointCapableOfProcessingRequestsAsync` or `EndpointIsCapableOfProcessingRequestsAsync` (double `is`)

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -938,4 +938,7 @@
     <value>Failed to apply the startup hook. Not all in process features will be available.</value>
     <comment>Gets a string similar to "Failed to apply the startup hook. Not all in process features will be available."</comment>
   </data>
+  <data name="ErrorMessage_ParameterCapturingNotAvailable" xml:space="preserve">
+    <value>Parameter capturing is not available in the process.</value>
```
kkeirstead, line 942: nit: Should this be more descriptive of why it's not available, or just stating that it's not?

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/MethodResolver.cs
```diff
@@ -0,0 +1,127 @@
...
+using System.Reflection;
+
+namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing
+{
+    public sealed class MethodResolver
+    {
+        private sealed class DeclaringTypeDescription : IEquatable<DeclaringTypeDescription?>
+        {
+            public string ModuleName { get; set; } = string.Empty;
+            public string ClassName { get; set; } = string.Empty;
```
jander-msft, line 17: TypeName

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/MethodResolver.cs
```diff
@@ -0,0 +1,127 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing
+{
+    public sealed class MethodResolver
+    {
+        private sealed class DeclaringTypeDescription : IEquatable<DeclaringTypeDescription?>
```
jander-msft, line 14: Using a [record class](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record) would eliminate the need to write this out.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/MethodResolver.cs
```diff
@@ -0,0 +1,127 @@
...
+            // Build a lookup table of all viable dll names to their backing reflection Module.
+            IEnumerable<Assembly> assemblies = AppDomain.CurrentDomain.GetAssemblies().Where(
+                assembly => !assembly.ReflectionOnly &&
+                !assembly.IsDynamic);
+
+            foreach (Assembly assembly in assemblies)
+            {
+                foreach (Module module in assembly.GetModules())
+                {
+                    if (!_dllNameToModules.TryGetValue(module.Name, out List<Module>? moduleList))
```
jander-msft, line 51: This dictionary doesn't represent "dll name to modules". The way it is constructed is "module name to modules". I also think you want a "assembly name to assembly" mapping here and then have the user describe the target methods in terms of (assembly name, type name, method name).

If you want to conflate the term "module" with "assembly", that's... okay, but you have to be internally consistent on the term module actually meaning assembly and thus caching a mapping of assembly names to assemblies, especially since types and methods are uniquely scoped to assemblies, not to modules e.g. you use `Assembly.GetType(name)` to get a `Type`.

FYI, practically, there's a one-to-one mapping between modules and assemblies; the only time you see an assembly to multiple module relationship is if someone gets creative with the compiler (this is not something supported by the project system): https://learn.microsoft.com/en-us/dotnet/framework/app-domains/build-multifile-assembly

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingService.cs
```diff
@@ -1,81 +1,270 @@
...
+                    FullMode = BoundedChannelFullMode.DropWrite
+                });
+                _stopRequests = Channel.CreateBounded<bool>(new BoundedChannelOptions(capacity: 1)
+                {
+                    FullMode = BoundedChannelFullMode.DropWrite
+                });
+
+                _probeManager = new FunctionProbesManager(new LogEmittingProbes(_logger));
+            }
+            catch (Exception ex)
```
jander-msft, line 84: I understand that you want to make sure that no exception bubbles out of this class so the user's app isn't affected, but the way this is done is causing all of the fields to be nullable (even though the compilation is forcing nullable attribution), which causes the rest of the code to constantly either null check or tell the compiler "no, trust me, it's not null". Let's think of a way to make everything non-nullable to avoid this.

src/Tools/dotnet-monitor/ParameterCapturing/ParameterCapturingEvents.cs
```diff
@@ -0,0 +1,48 @@
...
+            public const int Flush = 1;
+
+            public const int CapturingStart = 2;
+            public const int CapturingStop = 3;
+            public const int FailedToCapture = 4;
+            public const int ServiceNotAvailable = 5;
+        }
+
+
+        public enum ServiceNotAvailableReason : uint
```
jander-msft, line 21: Maybe `ServiceStatus` with `NotStarted = 0`, `Running`, `NotSupported`, and `GenericError` states? This would eliminate the need to make this any usage of this nullable.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingService.cs
```diff
@@ -1,81 +1,270 @@
...
+            {
+                return;
+            }
+
+            void tryToStopCapturing()
+            {
+                try
+                {
+                    StopCapturing();
+                    _eventSource.CapturingStop();
```
jander-msft, line 197: Why not make this part of the `StopCapturing` method so that it is always evented no matter what tries to stop the capture.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingService.cs
```diff
@@ -1,81 +1,270 @@
...
-        private readonly bool _isAvailable;
+
+        private ParameterCapturingEvents.ServiceNotAvailableReason? _notAvailableReason;
+        private string _notAvailableDetails = string.Empty;
 
         private readonly FunctionProbesManager? _probeManager;
+        private readonly ParameterCapturingEventSource _eventSource = new();
         private readonly ILogger? _logger;
 
+        private Channel<StartCapturingParametersPayload>? _requests;
```
jander-msft, line 34: If the design of this feature is that there is only one capturing session ever enabled at a time, then I think there should only be one channel that contains both the start and stop requests. Or maybe it just the start requests, but the request object has a completion source that can be signaled by the StopCapture method and the ExecuteAsync will just pop the request of the channel, start it, and wait for the completion source to be signaled before stopping it.

jander-msft: No, but you can put it on the payload itself so the completion is only associated with that exact request.

jander-msft: Another thing to consider is that you could add a capture ID property to both the start and stop requests. In the hosting startup, you can then have a dictionary of that ID to the request payload object; on a start request, add it to the dictionary, on a stop request, signal the completion source and remove it from the dictionary. This would absolve the need to have a notion of a singleton request except for the processing routine which would still only handle one request at a time.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/MethodResolver.cs
```diff
@@ -0,0 +1,127 @@
...
+            }
+        }
+
+        private readonly Dictionary<string, List<Module>> _dllNameToModules = new(StringComparer.InvariantCultureIgnoreCase);
+        private readonly Dictionary<DeclaringTypeDescription, List<MethodInfo>> _declaringTypeToMethods = new();
+
+        public MethodResolver()
+        {
+            // Build a lookup table of all viable dll names to their backing reflection Module.
+            IEnumerable<Assembly> assemblies = AppDomain.CurrentDomain.GetAssemblies().Where(
```
wiktork, line 43: If you haven't already, can you try this with AssemblyLoadContext?

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/MethodResolver.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            // Maintain a cache for all methods for a given assembly+type.
+            DeclaringTypeDescription declType = new(methodDescription.AssemblyName, methodDescription.TypeName);
+            if (_declaringTypeToMethods.TryGetValue(declType, out List<MethodInfo>? methods))
+            {
+                return methods;
+            }
+
+            List<MethodInfo> classMethods = new();
+            if (!_nameToAssemblies.TryGetValue(methodDescription.AssemblyName, out List<Assembly>? possibleAssemblies))
```
jander-msft, line 71: nit: I keep stumbling trying to read why it short-circuits here but that both add the same list to the same dictionary and return the same local. Consider inverting the if-condition, move the for each inside the if block, and unconditionally add the list to the dictionary and return the list.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingService.cs
```diff
@@ -1,84 +1,316 @@
...
+        {
+            if (!IsAvailable() || _initializedState == null)
+            {
+                BroadcastServiceState();
+                return;
+            }
+
+            if (payload.Methods.Length == 0)
+            {
+                _eventSource.FailedToCapture(payload.RequestId, new ArgumentException(nameof(payload.Methods)));
```
jander-msft, line 115: I think it would be best for us to be deliberate about explicitly passing the CapturingFailedReason for exceptions that we create rather than relying on the exception type. As it is, there's no distinction between an ArgumentException that we purposefully generate and one that is thrown by something that we call.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingService.cs
```diff
@@ -1,84 +1,316 @@
...
+            }
+
+            CapturingRequest request = new(payload);
+            if (!_initializedState.AllRequests.TryAdd(payload.RequestId, request))
+            {
+                _eventSource.FailedToCapture(payload.RequestId, new ArgumentException(nameof(payload.RequestId)));
+                return;
+            }
+
+            if (_initializedState.RequestQueue.Writer.TryWrite(request) != true)
```
jander-msft, line 126:
```suggestion
            if (!_initializedState.RequestQueue.Writer.TryWrite(request))
```

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingService.cs
```diff
@@ -1,84 +1,316 @@
...
+            {
+                _initializedState.AllRequests.TryRemove(payload.RequestId, out _);
```
jander-msft, line 128: I think that `request.StopRequest.TrySetCanceled` should also be called to make sure nothing is actually waiting on the abandoned request.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingService.cs
```diff
@@ -1,84 +1,316 @@
...
+        private sealed class CapturingRequest
+        {
+            public CapturingRequest(StartCapturingParametersPayload payload)
+            {
+                Payload = payload ?? throw new ArgumentNullException(nameof(payload));
+                StopRequest = new TaskCompletionSource<object?>(TaskCreationOptions.RunContinuationsAsynchronously);
+            }
+
+            public StartCapturingParametersPayload Payload { get; }
+            public TaskCompletionSource<object?> StopRequest { get; }
```
jander-msft, line 34: Make this the non-generic version since the result isn't actually used but is only meant for signalling.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingService.cs
jander-msft, line 128:
```suggestion
                _ = _initializedState.AllRequests.TryRemove(payload.RequestId, out _);
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/InProcessEgressOperation.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    internal sealed class InProcessEgressOperation : IEgressOperation
+    {
+        private readonly KeyValueLogScope _scope;
+
+        public EgressProcessInfo ProcessInfo { get; private set; }
+        public string EgressProviderName { get { return null; } }
+        public bool IsStoppable { get { return _operation?.IsStoppable ?? false; } }
```
jander-msft, line 19:
```suggestion
        public bool IsStoppable => _operation.IsStoppable;
```
I don't think you have to check for null.

src/Microsoft.Diagnostics.Monitoring.WebApi/ProfilerMessage.cs
```diff
@@ -1,17 +1,39 @@
...
+using Microsoft.Diagnostics.Monitoring.WebApi.Models;
+#endif
 using System;
 using System.Text;
 using System.Text.Json;
+using System.Threading;
 
 namespace Microsoft.Diagnostics.Monitoring
 {
+    internal sealed class EmptyPayload { }
```
jander-msft, line 16: All of these new classes shouldn't be in the WebApi assembly since nothing in it uses them.

src/Tools/dotnet-monitor/ParameterCapturing/EventParameterCapturingPipeline.cs
```diff
@@ -0,0 +1,130 @@
...
+        public string Details { get; set; }
+    }
+
+    internal sealed class EventParameterCapturingPipeline : EventSourcePipeline<EventParameterCapturingPipelineSettings>
+    {
+        public EventHandler<Guid> OnStartedCapturing;
+        public EventHandler<Guid> OnStoppedCapturing;
+        public EventHandler<Guid> OnUnknownRequestId;
+
+        public EventHandler<CapturingFailedArgs> OnCapturingFailed;
```
jander-msft, line 34: These should be properties on EventParameterCapturingPipelineSettings.

src/Tools/dotnet-monitor/ParameterCapturing/EventParameterCapturingPipeline.cs
```diff
@@ -0,0 +1,130 @@
...
+        public string Details { get; set; }
+    }
+
+    internal sealed class ServiceStateUpdateArgs
+    {
+        public ParameterCapturingEvents.ServiceState ServiceState { get; set; }
+        public string Details { get; set; }
+    }
+
+    internal sealed class EventParameterCapturingPipeline : EventSourcePipeline<EventParameterCapturingPipelineSettings>
```
jander-msft, line 28: This should be in the dotnet-monitor project

jander-msft: Weird, I could sweat GH said it was in the WebApi project. :shrug:

src/Tools/dotnet-monitor/ParameterCapturing/CaptureParametersOperation.cs
```diff
@@ -0,0 +1,190 @@
...
+
+                await _capturingStoppedCompletionSource.Task.WaitAsync(token).ConfigureAwait(false);
+            }
+            catch (OperationCanceledException)
+            {
+                _ = _capturingStartedCompletionSource.TrySetCanceled(token);
+                _ = _capturingStoppedCompletionSource.TrySetCanceled(token);
+
+                using CancellationTokenSource stopCancellationToken = new(TimeSpan.FromSeconds(30));
+                await StopAsync(stopCancellationToken.Token);
```
jander-msft, line 101:
```suggestion
                await StopAsync(stopCancellationToken.Token).SafeAwait();
```
I don't think we want the ExecuteAsync method Task to contain the failure due to stopping being cancelled but rather let the original exception be the one that ultimately fails the task.

src/Tools/dotnet-monitor/ParameterCapturing/CaptureParametersOperation.cs
```diff
@@ -0,0 +1,190 @@
...
+                token);
+
+            try
+            {
+                await _capturingStartedCompletionSource.Task.WaitAsync(token).ConfigureAwait(false);
+                startCompletionSource?.TrySetResult(null);
+
+                await _capturingStoppedCompletionSource.Task.WaitAsync(token).ConfigureAwait(false);
+            }
+            catch (OperationCanceledException)
```
jander-msft, line 95: You need call `startCompletionSource.TrySetException` so that callers who don't wait initially wait on the completion of the ExecuteAsync method (e.g. ones that are waiting for it to start) will fail instead of hanging forever.

jander-msft (approved):

jander-msft (approved):

PR #4935, opened 2023-07-18
Install NodeJs and Azurite on Helix

jander-msft:
###### Summary

These changes add nodejs and azurite to the helix machines so that the Azure blob storage extension tests can be executed.

###### Details
- Use correlation payload to acquire nodejs and unpack it onto the build machine.
- Use pre-commands to setup PATH and install azurite. Set environment variables as necessary to make the tests fail if azurite is not installed correctly.
- Skip these test on `linux-musl-*` environments since nodejs is not supported there.
- Use older version of nodejs due to glibc older version on Ubuntu 18.04, but use current version on win-arm64.


eng/Versions.props
```diff
@@ -123,4 +123,10 @@
     <MicrosoftDiagnosticsMonitoringLibraryVersion>$(MicrosoftDiagnosticsMonitoringVersion)</MicrosoftDiagnosticsMonitoringLibraryVersion>
     <MicrosoftDiagnosticsMonitoringEventPipeLibraryVersion>$(MicrosoftDiagnosticsMonitoringEventPipeVersion)</MicrosoftDiagnosticsMonitoringEventPipeLibraryVersion>
   </PropertyGroup>
+  <PropertyGroup Label="Helix">
+    <!-- Ubuntu 18.04 requires older nodejs version due to glibc version -->
```
kkeirstead, line 127: Without understanding the full scope of this, would it potentially make more sense to make a newer version the default, and then reduce the version if needed (versus in this case, where it looks like we use the oldest version anything needs, and only bump up the version if needed)?

jander-msft: Since all of Linux x64 queues (CBL Mariner 2.0, Debian 11, and Ubuntu 18.04) run from the same invocation, I can't change the version of what's installed on a per-queue basis; they all have the same correlation payload. It would require either installing multiple versions and having each queue dictate which version it will use OR refactor the project to make the groupings even more granular. I'm not sure which is better; I can log an issue for us to think about how we can move to newer Node installations even if we have to preserve v16 for Ubuntu 18.04.

jander-msft: /todo Determine how to use newer nodejs version for most Helix queues

eng/Helix.SingleTarget.targets
```diff
@@ -1,13 +1,56 @@
...
+    <!-- Set TEST_DOTNET_ROOT to same as DOTNET_ROOT (which is provided by ) -->
     <HelixWorkItemPreCommand Include="export TEST_DOTNET_ROOT=${DOTNET_ROOT}"
                              Condition="'$(IsHelixPosixShell)' == 'true'" />
     <HelixWorkItemPreCommand Include="SET TEST_DOTNET_ROOT=%DOTNET_ROOT%"
                              Condition="'$(IsHelixPosixShell)' != 'true'" />
   </ItemGroup>
 
-  <!-- TODO: Add Helix work item PreCommands to install Node and Azurite -->
+  <!-- Configure NodeJs and Npm -->
+  <ItemGroup Condition="'$(HelixRequiresNode)' == 'true'">
```
jander-msft, line 26: I might be able to put each of these blocks into cmd/sh scripts that are sourced from the dynamically generated script in order to simplify this a bit. It would also remove the semicolon workaround for prepending new paths to PATH on Windows.

eng/helix/payload/installazurite.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
jander-msft: I think the main exectuable.sh script goes through an marks everything executable, but I've updated it so that they are executable by default. For some reason, I thought I already did this, but apparently not.

jander-msft:
It appears that `npm config ...` is setting error level to non-zero even though it doesn't appear that something functionally when wrong; all other platforms seem to be working fine here. I'm going to merge this in as-is and iterate in a separate change to get Windows fixed.

PR #4944, opened 2023-07-19
[Parameter Capturing] Fix function probe tests to report failed asserts

schmittjoseph:
###### Summary

After https://github.com/dotnet/dotnet-monitor/pull/4578 , function probe tests were inadvertently broken.  Failed asserts in the test probes would trigger our fault protection code, meaning the asserts would never propagate up and cause the test to fail. Fix this regression and add a test to make sure it doesn't happen again.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/FunctionProbes/FunctionProbesScenario.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/FunctionProbes/FunctionProbesScenario.cs
@@ -259,13 +259,22 @@ namespace Microsoft.Diagnostics.Monitoring.UnitTestApp.Scenarios.FunctionProbes
 
         private static async Task Test_AssertsInProbesAreCaughtAsync(FunctionProbesManager probeManager, PerFunctionProbeProxy probeProxy, CancellationToken token)
         {
-            MethodInfo method = typeof(SampleNestedStruct).GetMethod(nameof(SampleNestedStruct.DoWork));
-            SampleNestedStruct nestedStruct = new();
+            MethodInfo method = typeof(StaticTestMethodSignatures).GetMethod(nameof(StaticTestMethodSignatures.NoArgs));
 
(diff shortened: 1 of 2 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/FunctionProbes/FunctionProbesScenario.cs
```diff
@@ -252,6 +257,17 @@ private static async Task Test_RequestInstallationOnProbeFunctionAsync(FunctionP
             await Assert.ThrowsAnyAsync<OperationCanceledException>(async () => await WaitForProbeInstallationAsync(probeManager, probeProxy, new[] { method }, timeoutSource.Token));
         }
 
+        private static async Task Test_AssertsInProbesAreCaughtAsync(FunctionProbesManager probeManager, PerFunctionProbeProxy probeProxy, CancellationToken token)
+        {
+            MethodInfo method = typeof(SampleNestedStruct).GetMethod(nameof(SampleNestedStruct.DoWork));
+            SampleNestedStruct nestedStruct = new();
+
+            await Assert.ThrowsAnyAsync<XunitException>(async () => await RunInstanceMethodTestCaseAsync(probeManager, probeProxy, method, new object[]
```
jander-msft, line 265: Can you add a comment as to why this should be expected to fail? I think it's because you are saying it has a `this` parameter, but in reality it does not, so the `Assert.Equal(thisParameterSupported ? thisObj : null, actualArgs[0]);` fails.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/FunctionProbes/PerFunctionProbeProxy.cs
```diff
@@ -59,6 +70,16 @@ public int GetProbeInvokeCount(MethodInfo method)
             return probe.GetInvokeCount();
         }
 
+        public XunitException GetProbeAssertException(MethodInfo method)
```
jander-msft, line 73: `bool TryGetProbeAssertException(MethodInfo method, out XunitException exception)`

This will be more friendly with nullability when we enable that in the future and forces the caller to think about whether the returned exception was actually provided or not.

jander-msft (approved):

PR #4947, opened 2023-07-19
[Exceptions] Include ActivityId and ActivityIdFormat

kkeirstead:
###### Summary

When an `ActivityId` is present (a non-zero guid), we now show it on an exception. Currently, the `ActivityIdFormat` is only being included on the JSON payload, but is omitted for the text payload.


src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/Exceptions/Eventing/ExceptionsEventListener.cs
```diff
@@ -115,6 +118,16 @@ private static ulong ToUInt64(object? value)
             return ToType<ulong>(value);
         }
 
+        private static T ToEnum<T>(object? value)
```
jander-msft, line 121: What's the actual type of the value when this is called? We might be able to write somethingmore succinct here.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/CurrentAppDomainExceptionSource.cs
```diff
@@ -40,7 +41,10 @@ private void CurrentDomain_FirstChanceException(object? sender, FirstChanceExcep
             {
                 _handlingException.Value = true;
 
-                RaiseExceptionThrown(e.Exception, timestamp);
+                RaiseExceptionThrown(e.Exception,
+                    timestamp,
+                    Activity.Current?.Id ?? Guid.Empty.ToString(),
```
jander-msft, line 46: These should probably be null if there is no activity information so that way we can differentiate between no activity vs custom activity identifiers.

src/Tools/dotnet-monitor/Exceptions/ExceptionsOperation.cs
```diff
@@ -103,6 +103,12 @@ await using (Utf8JsonWriter writer = new(stream, new JsonWriterOptions() { Inden
                 writer.WriteString("moduleName", instance.ModuleName);
                 writer.WriteString("message", instance.Message);
 
+                if (IncludeActivityId(instance))
+                {
+                    writer.WriteString("activityId", instance.ActivityId);
```
jander-msft, line 108: Let's write this as a sub-object so that way it can be wholesale ignored if need be or individual properties don't have too long of names:

```
"activity": {
  "id": "...",
  "idFormat": "..."
}
```

src/Tools/dotnet-monitor/Exceptions/ExceptionsOperation.cs
```diff
@@ -103,6 +103,12 @@ await using (Utf8JsonWriter writer = new(stream, new JsonWriterOptions() { Inden
...
+                    writer.WriteString("activityIdFormat", instance.ActivityIdFormat.ToString());
```
jander-msft, line 109:
```suggestion
                    writer.WriteString("activityIdFormat", instance.ActivityIdFormat.ToString("G"));
```

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/CurrentAppDomainExceptionSource.cs
kkeirstead: Do you want to preserve the `ActivityIdFormat` type as a nullable, or just convert it to its string value since that's what we ultimately present to the user?

jander-msft: I think it would be best to propagate it as `ActivityIdFormat?` and let the UX decide how to treat that (e.g. `.ToString("G")` if not null).

src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/Exceptions/Eventing/ExceptionsEventListener.cs
kkeirstead: Switched to using something specific for ActivityIdFormat - I can go back to a more generic method if we'd rather do that though.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/CurrentAppDomainExceptionSource.cs
kkeirstead: Discussed offline - can't use a nullable version of `ActivityIdFormat`, so keeping that as `Unknown` when not available.

src/Tools/dotnet-monitor/Exceptions/ExceptionsOperation.cs
```diff
@@ -282,5 +297,10 @@ private static async Task WriteTextExceptionFormat(TextWriter writer, IException
                 await writer.WriteAsync(instance.Message);
             }
         }
+
+        private static bool IncludeActivityId(IExceptionInstance instance)
+        {
+            return !string.IsNullOrEmpty(instance.ActivityId) && instance.ActivityId != Guid.Empty.ToString();
```
jander-msft, line 303:
```suggestion
            return !string.IsNullOrEmpty(instance.ActivityId);
```
I don't think we should check for Guid.Empty, especially since this is not a used sentinel value other than maybe our own tests (which should be changed to use either `null` or `string.Empty`).

src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/Exceptions/Eventing/ExceptionsEventListener.cs
jander-msft: I'd still like to know what the runtime type of the `value` parameter is; there might be more efficient code that can be used knowing what the expected type is. Please use a debugger to get the actual type.

kkeirstead: Sorry, I thought you just wanted to reduce all the verbosity with T - the type is listed as `object {int}` according to the debugger

jander-msft: With that known, it can be updated to something like:

```cs
if (value is int intValue)
{
  return (ActivityIdFormat)intValue;
}
throw new InvalidCastException();
```

No parsing or string conversion needed, just unboxing and casting.

jander-msft (approved):

PR #4964, opened 2023-07-20
Add per-process service support

jander-msft:
###### Summary

These changes allow establishing services in a per-process scope; the benefit is that features can easily establish per-process state at any point in the lifetime of the diagnostic capability of the target process. Note that this is current only support for listen mode as connect mode does NOT have stable endpoint instances and tracking over the lifetime of the target process.

Example usage of this is in https://github.com/jander-msft/dotnet-monitor/pull/104, where the exception history feature has been converted to make use of the per-process services facility; note that the exceptions service and startup hooks service are greatly simplified. Because exceptions are tracked per-process, the HTTP API can allow reporting of exception history for any target process.

### DIFF
--- a/src/Tools/dotnet-monitor/EndpointInfo/ScopedEndpointInfo.cs
+++ b/src/Tools/dotnet-monitor/EndpointInfo/ScopedEndpointInfo.cs
@@ -1,0 +1,35 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Diagnostics.Monitoring.WebApi;
+using Microsoft.Diagnostics.NETCore.Client;
+using System;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    internal sealed class ScopedEndpointInfo : IEndpointInfo
+    {
+        private IEndpointInfo _endpointInfo;
+
+        public void Set(IEndpointInfo endpointInfo)
+        {
+            _endpointInfo = endpointInfo;
+        }
+
+        public int ProcessId => _endpointInfo.ProcessId;
+
+        public Guid RuntimeInstanceCookie => _endpointInfo.RuntimeInstanceCookie;
+
+        public string CommandLine => _endpointInfo.CommandLine;
+
+        public string OperatingSystem => _endpointInfo.OperatingSystem;
+
+        public string ProcessArchitecture => _endpointInfo.ProcessArchitecture;
+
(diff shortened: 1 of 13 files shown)


src/Tools/dotnet-monitor/EndpointInfo/ScopedEndpointInfo.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Diagnostics.Monitoring.WebApi;
+using Microsoft.Diagnostics.NETCore.Client;
+using System;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
```
wiktork, line 8: So this class exists just to create a scoped service correct?

jander-msft: Correct, this allows other scoped services to to just compose the only endpoint and not have to worry about how to determine which one it should consider.

src/Tools/dotnet-monitor/IDiagnosticLifetimeService.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    /// <summary>
+    /// Callbacks for the lifetime events related to the diagnostic connection.
+    /// </summary>
+    public interface IDiagnosticLifetimeService
```
wiktork, line 12: Can this be internal?

src/Tools/dotnet-monitor/IDiagnosticLifetimeService.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        /// <summary>
+        /// Invoked just as the target process is discovered and is available for diagnostics.
+        /// </summary>
+        /// <remarks>
+        /// This is called just before the ResumeRuntime command is invoked.
+        /// </remarks>
+        ValueTask StartAsync(CancellationToken cancellationToken);
```
wiktork, line 20: is this a test only interface?

jander-msft: No, this is a callback interface for scoped services to know when we've detected a diagnosable process and when to stop it's implementation due to the process shutting down/loss of connectivity/etc. It's similar to `IEndpointInfoSourceCallbacks`, which I hope will be superseded by this interface.

wiktork (approved):

PR #4991, opened 2023-07-25
Collect exceptions from all processes and collect from startup

jander-msft:
###### Summary

This change gives the exceptions history feature the ability to collect exceptions from startup (prior behavior was that it started a few seconds after process discovery) and to collect from all processes (prior behavior only collected from the default process). There is currently one takeback with this change: when a process shuts down, the exception history information for the process is removed and no longer accessible; I would prefer to address that in a separate change.


src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/ExceptionsController.cs
```diff
@@ -21,43 +23,50 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
...
 
-            ExceptionsFormat? format = ComputeFormat(Request.GetTypedHeaders().Accept);
-            if (!format.HasValue)
+            ProcessKey? processKey = Utilities.GetProcessKey(pid, uid, name);
+
+            return InvokeForProcess(processInfo =>
             {
-                return this.NotAcceptable();
-            }
+                ExceptionsFormat format = ComputeFormat(Request.GetTypedHeaders().Accept) ?? ExceptionsFormat.PlainText;
```
kkeirstead, line 62: Is this to bring the default behavior in-line with what stacks does? No issue with the change, just wanted to know the motivation.

jander-msft: Yes, we should provide a default content type as a response if the client doesn't ask for one.

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/ExceptionsController.cs
```diff
@@ -21,43 +23,50 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
...
-            _exceptionsStore = serviceProvider.GetRequiredService<IExceptionsStore>();
-            _exceptionsOptions = serviceProvider.GetRequiredService<IOptions<ExceptionsOptions>>();
-            _operationFactory = serviceProvider.GetRequiredService<IExceptionsOperationFactory>();
+            _options = serviceProvider.GetRequiredService<IOptions<ExceptionsOptions>>();
         }
 
         /// <summary>
-        /// Gets the exceptions from the default process.
+        /// Gets the exceptions from the target process.
         /// </summary>
```
kkeirstead, line 41: nit: Do we want to add the `param name` comments here since we use that everywhere else when we have a pid/uid/name option?

src/Tools/dotnet-monitor/DiagnosticLifetimeBackgroundService.cs
```diff
@@ -0,0 +1,64 @@
...
+        public virtual ValueTask DisposeAsync()
+        {
+            _stoppingSource?.SafeCancel();
+
+            return ValueTask.CompletedTask;
+        }
+
+        public async ValueTask StartAsync(CancellationToken cancellationToken)
+        {
+            _stoppingSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
```
wiktork, line 33: Consider guarding against multiple calls

src/Tools/dotnet-monitor/StartupHook/StartupHookService.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            if (_inProcessFeatures.IsStartupHookRequired)
+            {
+                if (await _startupHookValidator.CheckEnvironmentAsync(_endpointInfo, cancellationToken))
+                {
+                    _resultSource.SetResult(true);
+                    return;
+                }
+
+                if (await _startupHookValidator.ApplyStartupHook(_endpointInfo, cancellationToken))
```
wiktork, line 39: I know this is not related to your PR, but I believe we do not properly merge startup hooks, but replace existing ones.

src/Tools/dotnet-monitor/StartupHook/StartupHookService.cs
```diff
@@ -0,0 +1,59 @@
...
+            IEndpointInfo endpointInfo,
+            IInProcessFeatures inProcessFeatures,
+            StartupHookValidator startupHookValidator)
+        {
+            _endpointInfo = endpointInfo ?? throw new ArgumentNullException(nameof(endpointInfo));
+            _inProcessFeatures = inProcessFeatures ?? throw new ArgumentNullException(nameof(inProcessFeatures));
+            _startupHookValidator = startupHookValidator ?? throw new ArgumentNullException(nameof(startupHookValidator));
+        }
+
+        public async ValueTask StartAsync(CancellationToken cancellationToken)
```
kkeirstead, line 29: Just as a sanity check since I had to modify this logic the last time - can this only be called once? Since we're not verifying that we haven't already tried to apply the startup hook.

src/Tools/dotnet-monitor/DiagnosticLifetimeBackgroundService.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// <remarks>
+    /// Similar to <see cref="Microsoft.Extensions.Hosting.BackgroundService"/>.
+    /// </remarks>
+    internal abstract class DiagnosticLifetimeBackgroundService :
+        IDiagnosticLifetimeService,
+        IAsyncDisposable
+    {
+        private Task _executingTask;
+        private CancellationTokenSource _stoppingSource;
+
```
wiktork, line 23: Would it also make sense to migrate our metrics store to operate off IDiagnosticLifetimeService?

src/Tools/dotnet-monitor/DiagnosticLifetimeBackgroundService.cs
```diff
@@ -0,0 +1,64 @@
...
+            return ValueTask.CompletedTask;
+        }
+
+        public async ValueTask StartAsync(CancellationToken cancellationToken)
+        {
+            _stoppingSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+
+            // Begin to execute but don't wait for it to complete.
+            _executingTask = ExecuteAsync(_stoppingSource.Token);
```
wiktork, line 36: There's a possible race condition here if StartAsync and StopAsync are called close to each other.

src/Tools/dotnet-monitor/StartupHook/StartupHookService.cs
jander-msft: We don't directly modify startup hooks. We use the diagnostic IPC channel to apply them. It does appendation within the runtime itself.

src/Tools/dotnet-monitor/DiagnosticLifetimeBackgroundService.cs
jander-msft: Sure, but let's do one thing at a time.

src/Tools/dotnet-monitor/StartupHook/StartupHookService.cs
jander-msft: Yes, it's only called once for the lifetime of the diagnostic session. If the target app fails diagnostic commands and continuously disconnects and reconnects for some reason, then it will show up each time.

I have an idea of storing some data about each process after each session is ended, so this could be added to that in order to prevent continuous log writing if need be.

wiktork (approved):

wiktork (approved):

PR #5004, opened 2023-07-27
Add top-level exception limit via ConfiguredExceptionsStore

jander-msft:
###### Summary

This change introduces a user-configurable limit to the exception store named the top-level limit. A top-level exception is one that is not an inner exception of another exception. By definition, any newly thrown exception is a top-level exception, but it will be demoted once it is wrapped as an inner exception. The limit only affects the number of top-level exceptions that are allowed in the store; all of the inner exceptions remain unbound as of this change.

I accomplished this by creating a new wrapper `IExceptionsStore` implementation named `ConfiguredExceptionsStore` in order to preserve the basic behavior on exception caching within the `ExceptionsStore` class. The limits are applied via a callback mechanism that modifies that drops older exceptions from the store, if necessary, before admitting new exceptions into the store, thus enforcing the limit at any observable time. A default limit of 20 was arbitrarily chosen and can adjusted based on empirical feedback.


documentation/schema.json
```diff
@@ -1063,6 +1063,15 @@
...
           "default": true
+        },
+        "TopLevelLimit": {
+          "type": [
+            "integer",
+            "null"
+          ],
+          "description": "The number of top-level exceptions to keep in the cache before automatically removing older exceptions. A top-level is exception is one that is not an inner exception of another exception.",
+          "format": "int32",
+          "default": 20
```
wiktork, line 1074: Apologies if this was already covered, but do we filter certain exception types by default? For example, I could see OperationCancelledException starving out every other exception type if we only have 20 exceptions.

jander-msft: No, we do not have any filter whatsoever at this time.

documentation/schema.json
```diff
@@ -1063,6 +1063,15 @@
           ],
           "description": "Enables or disables the exceptions feature. By default, this feature is enabled if in-process features are enabled.",
           "default": true
+        },
+        "TopLevelLimit": {
+          "type": [
+            "integer",
+            "null"
+          ],
+          "description": "The number of top-level exceptions to keep in the cache before automatically removing older exceptions. A top-level is exception is one that is not an inner exception of another exception.",
```
kkeirstead, line 1072:
```suggestion
          "description": "The number of top-level exceptions to keep in the cache before automatically removing older exceptions. A top-level exception is one that is not an inner exception of another exception.",
```

documentation/schema.json
kkeirstead, line 1074: Do we want to have max/min on this?

src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
```diff
@@ -784,4 +784,7 @@
   <data name="DisplayAttributeDescription_InProcessFeaturesOptions_Exceptions" xml:space="preserve">
     <value>Options for the in-process exceptions feature.</value>
   </data>
+  <data name="DisplayAttributeDescription_ExceptionsOptions_TopLevelLimit" xml:space="preserve">
+    <value>The number of top-level exceptions to keep in the cache before automatically removing older exceptions. A top-level is exception is one that is not an inner exception of another exception.</value>
```
kkeirstead, line 788:
```suggestion
    <value>The number of top-level exceptions to keep in the cache before automatically removing older exceptions. A top-level exception is one that is not an inner exception of another exception.</value>
```

documentation/schema.json
kkeirstead: Just noticed we do at least check for > 0 in `ConfiguredExceptionsStore.cs`, which was my primary concern.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfiguredExceptionsStoreTests.cs
```diff
@@ -0,0 +1,259 @@
...
+            Assert.Throws<ArgumentOutOfRangeException>(() => new ConfiguredExceptionsStore(0));
+        }
+
+        /// <summary>
+        /// Validates adding a single exception to the store with a limit of one will hold the exception instance.
+        /// </summary>
+        [Fact]
+        public async Task ConfiguredExceptionsStore_LimitOne_OnlyExceptionRemains()
+        {
+            int ExceptedCount = 1;
```
kkeirstead, line 43: I'm _assuming_ this is supposed to be `ExpectedCount` - if so it looks like this copy-paste propagated to a bunch of these test methods

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfiguredExceptionsStoreTests.cs
```diff
@@ -0,0 +1,259 @@
...
+            await callback.WaitForThresholdAsync(CommonTestTimeouts.GeneralTimeout);
+
+            // Assert
+            IReadOnlyList<IExceptionInstance> instances = store.GetSnapshot();
+            IExceptionInstance instance = Assert.Single(instances);
+            Assert.Equal(ExpectedId, instance.Id);
+        }
+
+        /// <summary>
+        /// Validates adding an inner exception and outer exception will hold both exceptions, even
```
kkeirstead, line 64: nit: slightly confusing comment

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfiguredExceptionsStoreTests.cs
```diff
@@ -0,0 +1,259 @@
...
+            IExceptionInstance outerInstance2 = instances[1];
+            Assert.NotNull(outerInstance2);
+            Assert.Equal(OuterException2Id, outerInstance2.Id);
+        }
+
+        /// <summary>
+        /// Validates that the last top level exception that shares an inner exception with a prior
+        /// top level exception remains in the store with a limit of one, including the inner exception.
+        /// </summary>
+        /// <returns></returns>
```
kkeirstead, line 177:
```suggestion
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfiguredExceptionsStoreTests.cs
```diff
@@ -0,0 +1,259 @@
...
+using Microsoft.Diagnostics.Tools.Monitor.Exceptions;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests
+{
+    public sealed class ConfiguredExceptionsStoreTests
```
kkeirstead, line 15: Could we add a test with a `TopLevelLimit` of 2 (or just not 1)? If we want to keep it simple we could probably just add three outer exceptions and make sure the initial one is dropped

src/Tools/dotnet-monitor/Exceptions/ConfiguredExceptionsStore.cs
```diff
@@ -0,0 +1,138 @@
...
+
+            public override void BeforeAdd(IExceptionInstance instance)
+            {
+                _outerExceptionMap.Add(instance.Id, new List<ulong>());
+                _innerExceptionMap.Add(instance.Id, instance.InnerExceptionIds.ToList());
+
+                // Make sure all of the inner exceptions are no longer top-level exceptions.
+                foreach (ulong innerExceptionId in instance.InnerExceptionIds)
+                {
+                    if (_outerExceptionMap.TryGetValue(innerExceptionId, out List<ulong> outerExceptions))
```
kkeirstead, line 85: Might be my own fault for not looking at this long enough, but maybe add a comment for this logic - I think I'm not fully grasping the use of the `_outerExceptionMap`.

Okay - I think I get it - for efficiency (traversing) reasons this gives us bidirectionality so we can reference whether an exception is an inner exception of anything else (i.e. who are all the parents of this particular exception). I'm keeping the first part of this comment in since I still think more explanation might be helpful for anyone skimming this file.

src/Tools/dotnet-monitor/Exceptions/ConfiguredExceptionsStore.cs
```diff
@@ -0,0 +1,138 @@
...
+                    {
+                        outerExceptions.Add(instance.Id);
+                    }
+
+                    _topLevelExceptions.Remove(innerExceptionId);
+                }
+
+                _topLevelExceptions.AddLast(instance.Id);
```
kkeirstead, line 93: I'll preface this by saying this is purely hypothetical (and I'm not sure if it matters) - I'm wondering if it's more performant to add to the front. My understanding is this is a DLL with head/tail pointers, _but_ the `Remove` algorithm removes the first occurrence of the value (i.e. iterating from the head pointer). This could be an incorrect assumption, but I'd guess that outer exceptions often become inner exceptions shortly after we process them (i.e., if we had a limit of 1000, we'd add exception_1 to the back...and then shortly after we'd add it's outer exception_2, meaning we now need to traverse all the way to the back to remove something that was just added). If we add to the front, searching with `Remove` in the average case (with these assumptions) would probably be faster. You can also remove at the back or front for when we hit the limit, so it shouldn't be an issue to switch to `RemoveLast`.

As previously mentioned - might be making some incorrect assumptions, and didn't run any benchmarks, but thought I'd throw it out there since that was my first thought reading through this.

src/Tools/dotnet-monitor/Exceptions/ConfiguredExceptionsStore.cs
jander-msft: Added more comments describing what an outer exception is and that it's easier/faster to track the information this way rather than having to find IExceptionInstances.

src/Tools/dotnet-monitor/Exceptions/ConfiguredExceptionsStore.cs
jander-msft: Absolutely right! This was probably a remnant of when I was using list-based tracking, but forgot to update the algorithm.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfiguredExceptionsStoreTests.cs
```diff
@@ -0,0 +1,299 @@
...
+        }
+
+        /// <summary>
+        /// Validates that the last top level exception and its inner exceptions will remain
+        /// in the store with a limit of one.
+        /// </summary>
+        [Fact]
+        public async Task ConfiguredExceptionsStore_LimitOne_LastRemainsWithInnerExceptions()
+        {
+            int ExceptedInstanceCount = 4;
```
kkeirstead, line 139: nit: `ExpectedInstanceCount`

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfiguredExceptionsStoreTests.cs
```diff
@@ -0,0 +1,299 @@
...
+        }
+
+        /// <summary>
+        /// Validates that the last top level exception that shares an inner exception with a prior
+        /// top level exception remains in the store with a limit of one, including the inner exception.
+        /// </summary>
+        [Fact]
+        public async Task ConfiguredExceptionsStore_LimitOne_LastRemainsWithSharedInnerException()
+        {
+            int ExceptedInstanceCount = 3;
```
kkeirstead, line 181: nit: `ExpectedInstanceCount`

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfiguredExceptionsStoreTests.cs
```diff
@@ -0,0 +1,299 @@
...
+        }
+
+        /// <summary>
+        /// Validates that adding a third top level exception to a store with a limit of two
+        /// will remove the first exception.
+        /// </summary>
+        [Fact]
+        public async Task ConfiguredExceptionsStore_LimitTwo_LastTwoRemain()
+        {
+            int ExceptedInstanceCount = 3;
```
kkeirstead, line 221: nit: `ExpectedInstanceCount`

src/Tools/dotnet-monitor/Exceptions/ConfiguredExceptionsStore.cs
```diff
@@ -0,0 +1,149 @@
...
+                }
+
+                _callback?.BeforeAdd(instance);
+            }
+
+            private void RemoveIfNoOuterExceptions(ulong exceptionId)
+            {
+                // Remove the exception if it no longer belongs to any other exception in the store
+                // as an inner exception.
+                if (_outerExceptionMap.TryGetValue(exceptionId, out List<ulong> outerExceptionIds))
```
kkeirstead, line 123: I know we had previously discussed this briefly but I don't think I saw it in here (sorry if I missed it) - do we want any mechanism for limiting inner-exception depth (i.e. we only have 20 top-level exceptions, but they go 1000 inner exceptions deep). Not saying that limiting needs to be added, just wanted to make sure it was brought up.

jander-msft: That's something I can add as a separate change (don't want to complicate this one further).

I think the easiest limit to explain to customers (but not necessarily the easiest to implement) is to say that every exception can only have a certain inner exception depth and everything outside of that depth would be eligible for removal.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ConfiguredExceptionsStoreTests.cs
jander-msft: I thought I fixed these, but apparently not.

kkeirstead (approved):
I don't think I had anything else - feel free to hold off on merging if you want to get more feedback from others.

PR #5017, opened 2023-07-28
[Exceptions] Simplify Types in Plain Text Representation

### LINKED ISSUE #4714: [Exceptions] Simplify Representation of Method Parameter Types
https://github.com/dotnet/dotnet-monitor/pull/4673#discussion_r1239077757

kkeirstead:
###### Summary

In Plain Text, generic type args and method parameters are now trimmed to improve readability; the JSON format stays the same for machine readability. There are currently a few rough edges (one of which being the logic for getting the namespace in the profiler - if anyone has guidance there, let me know), so I'll likely be pushing some small changes before taking this out of draft.

Text (Before)

```text
First chance exception at 2023-08-01T17:45:22.5728285Z
System.NullReferenceException: Oh no something was null!
   at StartupHookTestAppConsole.Program.RandomGeneric[StartupHookTestAppConsole.Program+T,StartupHookTestAppConsole.Program+U,StartupHookTestAppConsole.Program+V](System.String)
   at StartupHookTestAppConsole.Program.Main(System.String[])
```

Text (After)
```text
First chance exception at 2023-08-01T17:49:57.6324409Z
System.NullReferenceException: Oh no something was null!
   at StartupHookTestAppConsole.Program.RandomGeneric[Program+T,Program+U,Program+V](String)
   at StartupHookTestAppConsole.Program.Main(String[])
```

Closes #4714


src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
```diff
@@ -184,15 +184,18 @@ protected override void Dispose(bool disposing)
             ulong ModuleId,
             uint Token,
             uint OuterToken,
-            string Name)
+            string Name,
+            string TokenNamespace)
```
jander-msft, line 188: Rename all `TokenNamespace` to `Namespace`. I think adding `Token` to the front is going to cause folks to ask "what is a token namespace?" when such a thing does not exist.

kkeirstead: It initially was namespace, but then iirc I had issues with the reserved word `namespace` (hence the rename). I don't like the current name either though so I'd be in favor of changing it

jander-msft: `Namespace` with a capital `N` shouldn't cause any compilation issues.

kkeirstead: While that is true here, there are other places where the name `tokenNamespace` is used that would become `namespace` (which is how I hit the problem in the first place). I can technically change those to use `N` as well - it'll just break the naming convention in those locations (such as in `AddToken` in `EventExceptionsPipelineNameCache.cs`)

jander-msft: You can also use `@namespace` in those few places where you'd use the lowercase formatting. The `@` escapes keywords to be used as identifiers.

src/MonitorProfiler/Utilities/TypeNameUtilities.cpp
```diff
@@ -196,7 +199,20 @@ HRESULT TypeNameUtilities::GetTypeDefName(NameCache& nameCache, ModuleID moduleI
...
             IfFailRet(pMDImport->GetNestedClassProps(tokenToProcess, &outerTokenType));
         }
-        nameCache.AddTokenData(moduleId, tokenToProcess, outerTokenType, tstring(wName));
+        else
+        {
+            std::string::size_type found = tstring(wName).find_last_of(L'.');
+
+            if (found != std::string::npos)
+            {
+                wcsncpy_s(wNamespace, wName, found);
```
kkeirstead, line 208: This area seems to be causing build failures on non-windows platforms - I'm not sure if this is the approach we'll continue using for getting the namespace (is there something cleaner than just parsing for the `.`?)

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/StackUtilities.cs
```diff
@@ -20,15 +20,24 @@ internal enum StackFormat
 
     internal static class StackUtilities
     {
-        public static Models.CallStack TranslateCallStackToModel(CallStack stack, NameCache cache)
+        public static Models.CallStack TranslateCallStackToModel(CallStack stack, NameCache cache, bool methodNameIncludesGenericParameters = true)
```
jander-msft, line 23: Some suggestions to simplify this a bit:
- Add a `TypeFormat` parameter to this method, make the caller decide up front which format it wants the type information for the model to be in.
- Remove `CallStackFrame.FullTypeArgs` and `CallStackFrame.FullParameterTypes`. Populate the existing `CallStackFrame.TypeArgs` and `CallStackFrame.ParameterTypes` based on the `TypeFormat` parameter.

src/MonitorProfiler/Utilities/TypeNameUtilities.cpp
jander-msft: Because you are ultimately passing `tstring` values to `AddTokenData`, just use the methods on `tstring` such as [substr](https://en.cppreference.com/w/cpp/string/basic_string/substr).

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/StackUtilities.cs
kkeirstead: My understanding is that for Exceptions we do this translation when we receive an exception (not when the user hits `/exceptions`) - as a result we don't know whether they'll want JSON/text. We don't have that problem for `/stacks`, but without changing when we do this translation from the stack to the model so that it happens on-demand, I'm not sure how we would know the `TypeFormat`?

src/MonitorProfiler/Utilities/TypeNameUtilities.cpp
```diff
@@ -183,6 +183,9 @@ HRESULT TypeNameUtilities::GetTypeDefName(NameCache& nameCache, ModuleID moduleI
         }
 
         WCHAR wName[256];
+        WCHAR wNamespace[256];
```
jander-msft, line 186: This isn't necessary now and should be removed. `tstring` should have a parameterless ctor to produce an initial empty string.

src/MonitorProfiler/Utilities/TypeNameUtilities.cpp
```diff
@@ -192,11 +195,25 @@ HRESULT TypeNameUtilities::GetTypeDefName(NameCache& nameCache, ModuleID moduleI
...
+        tstring wNamespaceString = tstring(wNamespace);
+
         if (IsTdNested(dwTypeDefFlags))
         {
             IfFailRet(pMDImport->GetNestedClassProps(tokenToProcess, &outerTokenType));
         }
-        nameCache.AddTokenData(moduleId, tokenToProcess, outerTokenType, tstring(wName));
+        else
+        {
+            std::string::size_type found = tstring(wName).find_last_of(L'.');
```
jander-msft, line 208:
```suggestion
            std::string::size_type found = tstring(wName).find_last_of(_T('.'));
```
Might need it to be this.

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/StackUtilities.cs
jander-msft: Good point, I'll think about this some more and see if a unified model can still be done.

src/Microsoft.Diagnostics.Monitoring.WebApi/Stacks/NameFormatter.cs
```diff
@@ -20,24 +20,31 @@ internal sealed class NameFormatter
         private const string ArrayType = "_ArrayType_";
         private const string CompositeType = "_CompositeType_";
 
+        private const char DotSeparator = '.';
         private const char NestedSeparator = '+';
         private const char GenericStart = '[';
         private const char GenericSeparator = ',';
         private const char GenericEnd = ']';
 
+        internal enum TypeFormat
```
jander-msft, line 29: I'm not certain that we'll have some other format, so maybe change this to be a `bool` in the parameters e.g. `bool useSimpleName`.

src/Microsoft.Diagnostics.Monitoring.WebApi/Stacks/NameFormatter.cs
```diff
@@ -98,37 +112,45 @@ private static void BuildClassName(StringBuilder builder, NameCache cache, ulong
...
 
-        public static void BuildGenericParameters(StringBuilder builder, NameCache cache, ulong[] parameters)
+        public static void BuildGenericTypeNames(StringBuilder builder, NameCache cache, ulong[] parameters, TypeFormat typeFormat = TypeFormat.FullName)
+        {
+            IList<string> typeNames = GetTypeNames(cache, parameters, typeFormat);
+
+            WriteTypeNamesList(builder, typeNames);
+        }
+
+        public static void WriteTypeNamesList(StringBuilder builder, IList<string> typeNames, char startChar = GenericStart, char endChar = GenericEnd, char separationChar = GenericSeparator)
```
jander-msft, line 122: Rather than just default values, could you add two more methods that are explicit about what they are formatting: `BuildGenericArgTypes` and `BuildMethodParameterTypes` that will specify these three pieces?

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/CallStackResults.cs
```diff
@@ -17,9 +17,17 @@ public class CallStackFrame
         [JsonPropertyName("moduleName")]
         public string ModuleName { get; set; }
 
+        [JsonIgnore]
+        internal IList<string> TypeArgs { get; set; } = new List<string>();
```
jander-msft, line 21: Let's make these:
- `SimpleGenericArgTypes`
- `FullGenericArgTypes`
- `SimpleParameterTypes`
- `FullParameterTypes`
to make clear the `Simple` vs `Full` symmetry and the fact that these are all types.

src/Microsoft.Diagnostics.Monitoring.WebApi/Stacks/NameFormatter.cs
kkeirstead: Unless you feel strongly about it, I think I'd rather use the enum - I initially prototyped this using a bool, but I feel like this gives us a lot more future flexibility (without refactoring everything) without introducing much more verbosity. In the most recent change I did update this to use `Full` and `Simple` though.

jander-msft (approved):

PR #5049, opened 2023-08-01
Shutdown process scopes services during tool shutdown.

jander-msft:
###### Summary

In #4964, per-process-service scope was added to allow services to be executed in a service scope for each target process. The service scope is disposed after detecting that the process is no longer running/responsive/etc. However, what was missed is that these service scopes were not disposed when .NET Monitor shuts down, thus leaving any kind of operations that were active still in flight (e.g. if a trace is started as part of the service scope, the trace would not be shut down and keep running in the target app).

These changes collect the service scopes as part of `ServerEndpointInfoSource` disposal where each of the scopes is disposed. Additional locking was added to make sure that the service scopes dictionary is not accessed simultaneously.

### DIFF
--- a/src/Tools/dotnet-monitor/EndpointInfo/ServerEndpointInfoSource.cs
+++ b/src/Tools/dotnet-monitor/EndpointInfo/ServerEndpointInfoSource.cs
@@ -109,7 +109,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor
             }
 
             // Only dispose services and not notify of connection disconnects. Don't want services to spend
-            // spend time saving off state, completing sessions, etc but they should aggressively (yet correctly)
+            // time saving off state, completing sessions, etc but they should aggressively (yet correctly)
             // stop their operations at this point.
             foreach (AsyncServiceScope serviceScope in serviceScopes)
             {
--- a/src/Tools/dotnet-monitor/EndpointInfo/ServerEndpointInfoSource.cs
+++ b/src/Tools/dotnet-monitor/EndpointInfo/ServerEndpointInfoSource.cs
@@ -109,7 +109,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor
             }
 
             // Only dispose services and not notify of connection disconnects. Don't want services to spend
-            // spend time saving off state, completing sessions, etc but they should aggresively (yet correctly)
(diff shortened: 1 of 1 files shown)


src/Tools/dotnet-monitor/EndpointInfo/ServerEndpointInfoSource.cs
```diff
@@ -77,11 +82,45 @@ internal sealed class ServerEndpointInfoSource : BackgroundService, IEndpointInf
...
+            if (!DisposableHelper.CanDispose(ref _disposalState))
+                return;
+
+            // Makes sure the background task is canceled.
+            Dispose();
 
             _pendingRemovalWriter.TryComplete();
+
+            // Wait for background services to complete.
+            await ExecuteTask.SafeAwait().WaitAsync(CancellationToken.None);
```
kkeirstead, line 96: Just double checking that we want `CancellationToken.None` for these - I know we typically avoid doing that.

src/Tools/dotnet-monitor/EndpointInfo/ServerEndpointInfoSource.cs
```diff
@@ -77,11 +82,45 @@ internal sealed class ServerEndpointInfoSource : BackgroundService, IEndpointInf
...
+                serviceScopes = new List<AsyncServiceScope>(_activeEndpointServiceScopes.Values);
+                _activeEndpointServiceScopes.Clear();
+            }
+            finally
+            {
+                _activeEndpointsSemaphore.Release();
+            }
+
+            // Only dispose services and not notify of connection disconnects. Don't want services to spend
+            // spend time saving off state, completing sessions, etc but they should aggressively (yet correctly)
```
kkeirstead, line 112:
```suggestion
            // time saving off state, completing sessions, etc but they should aggressively (yet correctly)
```

src/Tools/dotnet-monitor/EndpointInfo/ServerEndpointInfoSource.cs
jander-msft: We don't have access to a specified token and creating our own `CancellationTokenSource` will likely not fire in time when shutting down the tool. I think the `ExecuteTask` is going to complete fairly quickly and "always" complete. But what you've pointed out is a good thing to remember in general.

PR #5070, opened 2023-08-03
[Parameter Capturing] Refactor ParameterCapturingService for easier unit testing

schmittjoseph:
###### Summary

This PR is in preparation for adding profiler callbacks on probe installation/uninstallation completion. Refactor the `ParameterCapturingService` for easier unit testing. Of note, split it into 2:
- `ParameterCapturingPipeline` is responsible for acting on requests to start and stop parameter capturing. It has no concept of service state and doesn't need to worry about throwing exceptions in its ctor. 
- `ParameterCapturingService` wraps `ParameterCapturingPipeline` in a safe manner and takes care of tracking the service's state and broadcasting that state back to dotnet-monitor when needed.


src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingPipeline.cs
```diff
@@ -0,0 +1,195 @@
...
+                    await request.StopRequest.Task.WaitAsync(cts.Token).ConfigureAwait(false);
+                }
+                catch (OperationCanceledException)
+                {
+
+                }
+
+                _logger.LogInformation(ParameterCapturingStrings.StopParameterCapturing);
+                _probeManager.StopCapturing();
+                ParameterCapturingEventSource.Log.CapturingStop(request.Payload.RequestId);
```
jander-msft, line 74: Rather than accessing a static instance, create a callback interface that the ParameterCapturingService (or an inner class) can implement and effectively forwards to the event source. This would also better separate this implementation so that it can be tested without automatically involving the event source and could allow tests to verify state of the pipeline by implementing its own callback. I believe this would also alleviate the need for the event source to be a static instance, and thus can be a member instance that is actually disposed properly.

src/Tests/Microsoft.Diagnostics.Monitoring.HostingStartup.UnitTests/ParameterCapturing/Pipeline/ParameterCapturingPipelineTests.cs
```diff
@@ -0,0 +1,292 @@
...
+            TestParameterCapturingCallbacks callbacks = new(
+                onCapturingStart: (payload, _) =>
+                {
+                    onStartCallbackSource.TrySetResult(payload.RequestId);
+                });
+
+            ParameterCapturingPipeline pipeline = new(probeManager, callbacks);
+            StartCapturingParametersPayload payload = CreateStartCapturingPayload(Timeout.InfiniteTimeSpan);
+
+            using CancellationTokenSource cts = new();
```
jander-msft, line 119: For each of these tests, consider moving this to the beginning, setting a generous timeout on it, and cancelling each of the `TaskCompletionSource` instances when this expires to prevent tests from hanging forever.

src/Tests/Microsoft.Diagnostics.Monitoring.HostingStartup.UnitTests/ParameterCapturing/Pipeline/ParameterCapturingPipelineTests.cs
```diff
@@ -0,0 +1,292 @@
...
+
+            // Act
+            Task pipelineTask = pipeline.RunAsync(cts.Token);
+            pipeline.SubmitRequest(payload);
+
+            // Assert
+            try
+            {
+                await pipelineTask;
```
jander-msft, line 259:
```suggestion
               await Assert.ThrowsAsync<Exception>(pipelineTask);
```
or something to that effect?

jander-msft: I think that also returns the Exception instance, so you can do:

```
Exception ex = await Assert.ThrowsAsync(...);
Assert.Equals(throwException, ex);
```

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Pipeline/ParameterCapturingPipeline.cs
```diff
@@ -0,0 +1,169 @@
...
+
+        public void SubmitRequest(StartCapturingParametersPayload payload)
+        {
+            if (payload.Methods.Length == 0)
+            {
+                throw new ArgumentException(nameof(payload.Methods));
+            }
+
+            CapturingRequest request = new(payload);
+            if (!_allRequests.TryAdd(payload.RequestId, request))
```
wiktork, line 140: It seems that in some cases the pipeline returns true/false and talks to the callback (which then talks to the eventsource), while in other cases the pipeline throws and the caller catches and then talks to the eventsource.

jander-msft (approved):

PR #5071, opened 2023-08-03
Report unhandled exceptions through ExceptionsEventSource

jander-msft:
###### Summary

Refactor exception source, pipeline, and event source to allow for processing unhandled exceptions through a separate pipeline instance but reporting them through the same event source instance used for first chance exceptions. This was done by:
- Refactoring `ExceptionSourceBase` to be agnostic of the source of the exception. Add `GuardedExceptionSourceBase` to shared between the first chance and unhandled sources.
- Move ownership of the `ExceptionsEventSource` and exception ID generation to the processor so that it can be shared between the first chance pipeline and the unhandled pipeline.
- Add new `ExceptionInstanceUnhandled` event to `ExceptionsEventSource`.

I've left the handling of the `ExceptionInstanceUnhandled` event in the `EventExceptionsPipeline` alone for now in order to keep this PR focused on providing the infrastructure to react to unhandled exceptions.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/ExceptionSourceBase.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/ExceptionSourceBase.cs
@@ -7,18 +7,18 @@ using System.Diagnostics;
 namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions
 {
     /// <summary>
-    /// Represents a source of thrown exceptions.
+    /// Represents a source of exceptions.
     /// </summary>
     internal abstract class ExceptionSourceBase
     {
-        protected void RaiseExceptionThrown(Exception ex, DateTime timestamp, string? activityId, ActivityIdFormat format)
+        protected void RaiseException(Exception ex, DateTime timestamp, string? activityId, ActivityIdFormat format)
         {
-            ExceptionThrown?.Invoke(this, new ExceptionEventArgs(ex, timestamp, activityId, format));
+            Exception?.Invoke(this, new ExceptionEventArgs(ex, timestamp, activityId, format));
         }
 
         /// <summary>
-        /// Event that is raised each time an exception is thrown.
+        /// Event that is raised each time an exception is made available.
         /// </summary>
-        public event EventHandler<ExceptionEventArgs>? ExceptionThrown;
+        public event EventHandler<ExceptionEventArgs>? Exception;
     }
 }
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/ExceptionSourceBase.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/ExceptionSourceBase.cs
@@ -13,12 +13,12 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions
     {
         protected void RaiseException(Exception ex, DateTime timestamp, string? activityId, ActivityIdFormat format)
         {
-            Exception?.Invoke(this, new ExceptionEventArgs(ex, timestamp, activityId, format));
+            ExceptionAvailable?.Invoke(this, new ExceptionAvailableEventArgs(ex, timestamp, activityId, format));
         }
 
         /// <summary>
         /// Event that is raised each time an exception is made available.
         /// </summary>
-        public event EventHandler<ExceptionEventArgs>? Exception;
+        public event EventHandler<ExceptionAvailableEventArgs>? ExceptionAvailable;
     }
 }
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/GuardedExceptionSourceBase.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/GuardedExceptionSourceBase.cs
@@ -1,8 +1,8 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System.Diagnostics;
 using System;
+using System.Diagnostics;
 using System.Threading;
 
 namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions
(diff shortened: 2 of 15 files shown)


src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionIdSource.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            if (null == exception)
+                return 0;
+
+            if (TryGetExceptionId(exception, out ulong exceptionId))
+                return exceptionId;
+
+            lock (exception.Data)
+            {
+                if (TryGetExceptionId(exception, out exceptionId))
```
kkeirstead, line 25: I'm assuming I'm missing something here (and I see we had this logic before) - why do we call this method twice in a row, once without the lock and once with the lock?

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/UnhandledExceptionEventsPipelineStep.cs
```diff
@@ -0,0 +1,39 @@
...
+            _eventSource = eventSource;
+            _idSource = idSource;
+            _next = next;
+        }
+
+        public void Invoke(Exception exception, ExceptionPipelineExceptionContext context)
+        {
+            ArgumentNullException.ThrowIfNull(exception);
+
+            // Do not send via the EventSource unless an listener is active.
```
kkeirstead, line 30:
```suggestion
            // Do not send via the EventSource unless a listener is active.
```

src/Tools/dotnet-monitor/Exceptions/EventExceptionsPipeline.cs
```diff
@@ -87,6 +87,10 @@ private void Callback(TraceEvent traceEvent)
                         traceEvent.GetPayload<string>(ExceptionEvents.ExceptionInstancePayloads.ActivityId),
                         traceEvent.GetPayload<ActivityIdFormat>(ExceptionEvents.ExceptionInstancePayloads.ActivityIdFormat));
                     break;
+                case "ExceptionInstanceUnhandled":
+                    // TODO: Advertise unhandled exception as necessary
+                    ulong exceptionId = traceEvent.GetPayload<ulong>(ExceptionEvents.ExceptionInstancePayloads.ExceptionId);
```
kkeirstead, line 92:
```suggestion
                    ulong exceptionId = traceEvent.GetPayload<ulong>(ExceptionEvents.ExceptionInstanceUnhandledPayloads.ExceptionId);
```

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionIdSource.cs
jander-msft: In case some other thread assigned the exception an identifier while it held the lock and this thread was waiting on the lock; not likely to happen, but can happen if an exception is created on one thread, and thrown from multiple threads at the same time.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/ExceptionSourceBase.cs
```diff
@@ -7,18 +7,18 @@
...
-            ExceptionThrown?.Invoke(this, new ExceptionEventArgs(ex, timestamp, activityId, format));
+            Exception?.Invoke(this, new ExceptionEventArgs(ex, timestamp, activityId, format));
         }
 
         /// <summary>
-        /// Event that is raised each time an exception is thrown.
+        /// Event that is raised each time an exception is made available.
         /// </summary>
-        public event EventHandler<ExceptionEventArgs>? ExceptionThrown;
+        public event EventHandler<ExceptionEventArgs>? Exception;
```
kkeirstead, line 22: Any concern about this name-change with readability?

kkeirstead (dismissed):
I didn't have anything else, feel free to wait on other reviewers.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/GuardedExceptionSourceBase.cs
```diff
@@ -3,28 +3,16 @@
...
-
-        public CurrentAppDomainExceptionSource()
-        {
-            AppDomain.CurrentDomain.FirstChanceException += CurrentDomain_FirstChanceException;
-        }
-
-        private void CurrentDomain_FirstChanceException(object? sender, FirstChanceExceptionEventArgs e)
+        protected void RaiseExceptionGuarded(Exception exception)
         {
             DateTime timestamp = DateTime.UtcNow;
```
wiktork, line 17: nit: It is more common to implement IDisposable in the base class and Dispose(bool disposing) in derived classes.

wiktork (approved):

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/GuardedExceptionSourceBase.cs
jander-msft: I think you wanted to leave this comment somewhere. In any event, I largely disagree with that approach in general because you shouldn't force concrete implementations to have the disposal pattern if there isn't anything to dispose. I also largely leave that out of any interface derivations because that consumer of the interface in most cases should not have knowledge of the disposability of an implementation nor be responsible for it, unless you are making it part of the contract that something that acquires that instance is now responsible for the lifetime of it.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/ExceptionSourceBase.cs
jander-msft: Changed to ExceptionAvailable

kkeirstead (approved):

PR #5072, opened 2023-08-03
[Exceptions] Add Functional Tests

kkeirstead:
###### Summary

This PR adds basic functional tests for Exceptions - these tests validate that the JSON/text results that come back for a single exception are as-expected.

There is also a test that does not set the startup hook env var that is exclusively for .NET8 to confirm that the `ApplyStartupHook` logic is working.

I wanted to put this up for draft before investing in additional scenarios - let me know if you'd like to see these tests exercise other things.


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/MonitorRunner.cs
```diff
@@ -165,8 +165,11 @@ public virtual async Task StartAsync(string command, string[] args, Cancellation
             // Override the user config directory
             _adapter.Environment.Add("DotnetMonitorTestSettings__UserConfigDirectoryOverride", UserConfigDirectoryPath);
 
-            // Ensures that the TestStartupHook is loaded early so it helps resolve other test assemblies
-            _adapter.Environment.Add(ToolIdentifiers.EnvironmentVariables.StartupHooks, TestStartupHookPath);
+            if (!_adapter.Environment.ContainsKey(ToolIdentifiers.EnvironmentVariables.StartupHooks))
```
jander-msft, line 168: Why is this needed? VS might put in startup hooks of its own, so you might be failing to run .NET Monitor for local testing.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/ExceptionsScenario.cs
```diff
@@ -57,6 +57,9 @@ public static CliCommand Command()
             reflectionTypeLoadExceptionCommand.SetAction(ReflectionTypeLoadExceptionAsync);
 
             CliCommand scenarioCommand = new(TestAppScenarios.Exceptions.Name);
+
+            scenarioCommand.SetAction(ExecuteAsync);
```
jander-msft, line 61: Why is there a handler on the Exceptions command? I would think that the tests should be using one of the sub commands.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/ScenarioRunner.cs
```diff
@@ -26,7 +26,8 @@ internal static class ScenarioRunner
             Action<AppRunner> configureApp = null,
             Action<MonitorCollectRunner> configureTool = null,
             bool disableHttpEgress = false,
-            string profilerLogLevel = null)
+            string profilerLogLevel = null,
+            string customStartupHookPath = null)
```
jander-msft, line 30: Remove this and set it in the `configureTool` and `configureApp` callbacks.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Runners/ScenarioRunner.cs
```diff
@@ -38,6 +39,7 @@ internal static class ScenarioRunner
             toolRunner.DiagnosticPortPath = diagnosticPortPath;
             toolRunner.DisableAuthentication = true;
             toolRunner.DisableHttpEgress = disableHttpEgress;
+            toolRunner.CustomStartupHookPath = customStartupHookPath;
```
jander-msft, line 42: Why does the `ToolRunner` have a setting for specifying the startup hook? We shouldn't be applying startup hooks to .NET Monitor.

kkeirstead: Yep - I was having a lot of issues initially getting these tests to run end-to-end and was experimenting to figure out the issue, but never went back and pulled out some of the unnecessary stuff like this - it should be removed now.

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/AppRunner.cs
```diff
@@ -167,6 +169,16 @@ public async Task StartAsync(CancellationToken token, bool waitForReady = true)
                     ProfilerIdentifiers.EnvironmentVariables.StdErrLogger_Level, ProfilerLogLevel);
             }
 
+            if (EnableMonitorStartupHook)
+            {
+                string startupHookPath = AssemblyHelper.GetAssemblyArtifactBinPath(
```
jander-msft, line 174: Move this to the ctor so that you can calculate the path relative to the test assembly rather than relative to this utility assembly; it's likely trying to take something like `artifacts/bin/<TestAssembly>/Debug/net8.0/<TestCommonAssembly>.dll` and turning it into `artifacts/bin/<TestAssembly>/Debug/net8.0/<StartupHookAssembly>.dll`, which is not correct.

kkeirstead: So, I can change this if it's broken, but this was actually a direct copy from the `feature/parameter-capturing` branch https://github.com/dotnet/dotnet-monitor/pull/5072#discussion_r1283743205

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/ExceptionsTests.cs
```diff
@@ -0,0 +1,178 @@
...
+            await ScenarioRunner.SingleTarget(
+                _outputHelper,
+                _httpClientFactory,
+                DiagnosticPortConnectionMode.Listen,
+                TestAppScenarios.Exceptions.Name,
+                appValidate: async (appRunner, apiClient) =>
+                {
+                    await GetExceptions(apiClient, appRunner, ExceptionFormat.PlainText);
+
+                    var exceptionsLines = exceptionsResult.Split(new string[] { "\r\n", "\n" }, StringSplitOptions.None);
```
jander-msft, line 63:
```suggestion
                    var exceptionsLines = exceptionsResult.Split(Environment.NewLine, StringSplitOptions.None);
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/ExceptionsTests.cs
```diff
@@ -0,0 +1,178 @@
...
+            await appRunner.SendCommandAsync(TestAppScenarios.Exceptions.Commands.Begin);
+
+            int processId = await appRunner.ProcessIdTask;
+
+            await RetryUtilities.RetryAsync(
+                () => CaptureExtensions(apiClient, processId, format),
+                shouldRetry: (Exception ex) => ex is ArgumentException,
+                maxRetryCount: 5,
+                outputHelper: _outputHelper);
+        }
```
jander-msft, line 159: I don't see where `TestAppScenarios.Exceptions.Commands.End` is sent. I would assume it would be sent here.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests.csproj
```diff
@@ -86,6 +86,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\Microsoft.Diagnostics.Monitoring.Options\Microsoft.Diagnostics.Monitoring.Options.csproj" />
+    <ProjectReference Include="..\..\Microsoft.Diagnostics.Monitoring.StartupHook\Microsoft.Diagnostics.Monitoring.StartupHook.csproj" />
```
jander-msft, line 89: Why is there a ProjectReference to the startup hook assembly?

eng/Helix.SingleTarget.targets
```diff
@@ -15,8 +15,12 @@
     <!-- Set TEST_DOTNET_ROOT to same as DOTNET_ROOT (which is provided by the dotnet cli correlation payload) -->
     <HelixWorkItemPreCommand Include="export TEST_DOTNET_ROOT=${DOTNET_ROOT}"
                              Condition="'$(IsHelixPosixShell)' == 'true'" />
+    <HelixWorkItemPreCommand Include="export LANG=en_US.UTF-8"
+                             Condition="'$(IsHelixPosixShell)' == 'true'" />
     <HelixWorkItemPreCommand Include="SET TEST_DOTNET_ROOT=%DOTNET_ROOT%"
                              Condition="'$(IsHelixPosixShell)' != 'true'" />
+    <HelixWorkItemPreCommand Include="SET LANG=en_US.UTF-8"
```
jander-msft, line 22: This is not needed for Windows. It should only remain as a workaround for the incorrectly specified value for the Linux containers.

eng/Helix.SingleTarget.targets
```diff
@@ -15,8 +15,12 @@
     <!-- Set TEST_DOTNET_ROOT to same as DOTNET_ROOT (which is provided by the dotnet cli correlation payload) -->
     <HelixWorkItemPreCommand Include="export TEST_DOTNET_ROOT=${DOTNET_ROOT}"
                              Condition="'$(IsHelixPosixShell)' == 'true'" />
+    <HelixWorkItemPreCommand Include="export LANG=en_US.UTF-8"
```
jander-msft, line 18: Add a comment referring to the issue that we logged about needing to fix this value and make this only apply when it's targeting the Linux machines e.g. adding `$([MSBuild]::IsOSPlatform(Linux))` to the condition is sufficient.

PR #5073, opened 2023-08-03
[Parameter Capturing] Cancel initializing profilers on unsupported runtimes

schmittjoseph:
###### Summary

Ensure that our profilers only initialize on supported runtimes (.net core, and not desktop). This is to help avoid problems where our profilers are loaded into a process with both the desktop and core CLRs and they attempt to act on global objects (e.g. the parameter-capturing request queue).

This is a follow-up on earlier parameter capturing feedback:  https://github.com/dotnet/dotnet-monitor/pull/4428#discussion_r1197002931

### DIFF
--- a/src/Profilers/CommonMonitorProfiler/ProfilerBase.cpp
+++ b/src/Profilers/CommonMonitorProfiler/ProfilerBase.cpp
@@ -10,16 +10,14 @@ ProfilerBase::ProfilerBase() :
 {
 }
 
-HRESULT ProfilerBase::IsRuntimeSupported(ICorProfilerInfo12* pCorProfilerInfo, bool& supported)
+HRESULT ProfilerBase::IsRuntimeSupported(bool& supported)
 {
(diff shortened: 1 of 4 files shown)


src/Profilers/CommonMonitorProfiler/CommonUtilities/RuntimeUtilities.h
```diff
@@ -0,0 +1,31 @@
...
+
+#pragma once
+
+#include "macros.h"
+#include "corprof.h"
+
+class RuntimeUtilities
+{
+    public:
+        static HRESULT IsRuntimeSupported(ICorProfilerInfo12* pCorProfilerInfo, BOOL& supported)
```
jander-msft, line 12: Could you put this in `ProfilerBase`? Also, use `bool` instead of `BOOL`

jander-msft (approved):

jander-msft (approved):

PR #4474, opened 2023-05-23
[Parameter Capturing] Add in-proc service to capture parameters on demand

schmittjoseph:
###### Summary

This PR adds the in-proc managed layer for parameter capturing and logging. A new hosted service gets injected into a target application which tries to retrieve an `ILogger` instance. If successful, it then sets up the infrastructure for function probes leveraging our new profiler API surface. The layer ontop of this service to drive an end-to-end scenario is not included.

Out of scope of this PR:
- Functional tests, as they also require https://github.com/dotnet/dotnet-monitor/pull/4428 to be merged.
- Changes in the dotnet-monitor tool to drive an end-to-end scenario using this new service.
- More complex parameter type handling (e.g. value-type type refs) and argument serialization (e.g. enumerables). These will be added in smaller separate PRs.

PR #5074, opened 2023-08-03
[Parameter Capturing] Improve unknown declaring type representation

schmittjoseph:
###### Summary

When constructing the template string for logging in function probes, represent unknown declaring types as `<unknown_type>` instead of an empty string.

This is a follow up on earlier parameter capturing feedback: https://github.com/dotnet/dotnet-monitor/pull/4474#discussion_r1213500057

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingStrings.resx
+++ b/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingStrings.resx
@@ -147,6 +147,9 @@
   <data name="UnknownArgumentValue" xml:space="preserve">
     <value>unknown</value>
   </data>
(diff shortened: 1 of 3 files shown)


src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingStrings.resx
```diff
@@ -147,6 +147,9 @@
   <data name="UnknownArgumentValue" xml:space="preserve">
     <value>unknown</value>
   </data>
+  <data name="UnknownDeclaringType" xml:space="preserve">
+    <value>&lt;unknown_type&gt;</value>
```
jander-msft, line 151: I know that this code is somewhat inconsistent on this, but I don't think this should be a localizable resource. This type of information is likely to be machine parsable, so it should use well-known constants.

PR #5119, opened 2023-08-09
Fixup unstable test

wiktork:

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingLogger.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingLogger.cs
@@ -77,11 +77,13 @@ namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing
         {
             try
             {
-                while (!DisposableHelper.IsDisposed(ref _disposedState))
+                while (_messages.TryTake(out (string format, string[] args) entry, Timeout.InfiniteTimeSpan))
                 {
-                    (string format, string[] args) = _messages.Take();
-                    Log(_systemLogger, format, args);
+                    Log(_systemLogger, entry.format, entry.args);
                 }
+            }
+            catch (ObjectDisposedException)
+            {
             }
             catch
             {
@@ -90,8 +92,7 @@ namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing
 
         public void Complete()
         {
-            //TODO: Should this also be called during shutdown to ensure all messages are logged?
-
+            // NOTE We currently do not wait for the background thread in production code
             _messages.CompleteAdding();
             _thread.Join();
         }
@@ -104,7 +105,7 @@ namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing
             {
                 return;
             }
-
+            _messages.CompleteAdding();
             _messages.Dispose();
         }
     }
--- a/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingLogger.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingLogger.cs
@@ -58,6 +58,8 @@ namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing
 
         public void Log(ParameterCaptureMode mode, string format, string[] args)
         {
+            DisposableHelper.ThrowIfDisposed<ParameterCapturingLogger>(ref _disposedState);
+
             if (mode == ParameterCaptureMode.Inline)
             {
                 Log(_userLogger, format, args);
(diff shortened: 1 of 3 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.HostingStartup.UnitTests/ParameterCapturing/SystemLoggingTests.cs
```diff
@@ -29,6 +29,8 @@ public void TestLoggingCategories()
             {
                 logger.Log(ParameterCaptureMode.Inline, inlineMessage, Array.Empty<string>());
                 logger.Log(ParameterCaptureMode.Background, backgroundMessage, Array.Empty<string>());
+
+                logger.Complete();
```
jander-msft, line 33: I don't understand how this currently helps with tests; the shutdown ordering and logic remains the same.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingLogger.cs
```diff
@@ -86,6 +88,14 @@ private void ThreadProc()
             }
         }
 
+        public void Complete()
+        {
+            //TODO: Should this also be called during shutdown to ensure all messages are logged?
+
+            _messages.CompleteAdding();
```
jander-msft, line 95: This is no longer called by product code, so this change as-is seems like it would a regression.

src/Tests/Microsoft.Diagnostics.Monitoring.HostingStartup.UnitTests/ParameterCapturing/SystemLoggingTests.cs
wiktork: I will add more comments in the code, but effectively `Complete()` will cause all the messages to be processed prior to the Dispose operation. Prior to this change, the dispose flag would get set immediately and potentially cancel the background thread before it finished processing.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingLogger.cs
wiktork: The previous Disposal code wasn't getting called at all due to a merge conflict. But putting that aside, the TODO here is meant to address how we want to cancel background logging at the end of the ParameterCapturingService. Do we force logging of all remaining background types (potentially delaying process shutdown to do so), or do we make Disposal immediately stop processing all messages? I'm not sure which direction I wanted to go with it, but want to address the test flakiness first.

src/Tests/Microsoft.Diagnostics.Monitoring.HostingStartup.UnitTests/ParameterCapturing/SystemLoggingTests.cs
jander-msft: Okay, I see that now in `ThreadProc` it is checking the disposal state and keying off that to determine whether it should stop processing entries, which is... interesting. Not sure how I feel about that since it's somewhat conflating completion with disposal. I think it would be better if it checked for `_messages.IsAddingCompleted` or (maybe better, if we want it to finish processing all of the messages) `_messages.IsCompleted`.

jander-msft: Or even use `while (_messages.TryTake(out (string format, string[] args), Timeout.InfiniteTimeSpan))` to have it wait for an item and return false when no more items are available rather than likely throwing an exception due to waiting in `_messages.Take` when it gets completed or disposed.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingLogger.cs
jander-msft: > The previous Disposal code wasn't getting called at all due to a merge conflict.

Okay, I see that. But it feels like you are just "kicking the can down the road" here.

> But putting that aside, the TODO here is meant to address how we want to cancel background logging at the end of the ParameterCapturingService. Do we force logging of all remaining background types (potentially delaying process shutdown to do so), or do we make Disposal immediately stop processing all messages? I'm not sure which direction I wanted to go with it, but want to address the test flakiness first.

I have no problem with deciding that later.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingLogger.cs
```diff
@@ -95,8 +106,6 @@ public void Dispose()
                 return;
             }
```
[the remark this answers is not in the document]
wiktork: > Shouldn't we still call `.CompleteAdding()` here so that `while (_messages.TryTake(out (string format, string[] args) entry, Timeout.InfiniteTimeSpan))` returns?

Good catch, `Dispose` does not interrupt TryTake but does prevent another `TryTake` from occuring.

PR #5125, opened 2023-08-10
[Parameter Capturing] Make probe management APIs asynchronous

schmittjoseph:
###### Summary

This PR converts the probe management APIs to be asynchronous. The prior behavior was to submit requests to our profiler without any feedback on if the requests succeeded or failed. Now the results of requests are reported back to our managed code so we can act on it accordingly.

High-level changes:
- Move the probe faulting behavior from the mutating profiler into `ParameterCapturingService`.
- Support registering probe management callbacks with the mutating profiler. These callbacks will be invoked by a new dedicated thread (See `ProbeInstrumentation::InitBackgroundService` for an explanation of why this is necessary).
- `FunctionProbesManager` now maintains an internal state to properly handle callbacks and transition its new task completion sources as necessary.

Out of scope of this PR:
- The dotnet-monitor side changes for logging probe faults

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/FunctionProbesManager.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/FunctionProbesManager.cs
@@ -200,7 +200,7 @@ namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing.Fun
             {
                 using IDisposable _ = cts.Token.Register(() =>
                 {
-                    _uninstallationTaskSource?.TrySetCanceled(cts.Token);
+                    _uninstallationTaskSource.TrySetCanceled(cts.Token);
                 });
                 await _uninstallationTaskSource.Task.ConfigureAwait(false);
             }
@@ -305,7 +305,7 @@ namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing.Fun
                         }
                     }
 
-                    _installationTaskSource?.TrySetCanceled(cts.Token);
+                    _installationTaskSource.TrySetCanceled(cts.Token);
                 });
 
                 await _installationTaskSource.Task.ConfigureAwait(false);


src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/FunctionProbesManager.cs
```diff
@@ -25,50 +28,172 @@ internal sealed class FunctionProbesManager : IFunctionProbesManager
...
+            _ = _installationTaskSource?.TrySetException(ex);
+            _ = _uninstallationTaskSource?.TrySetException(ex);
+            _ = _probeRegistrationTaskSource?.TrySetException(ex);
+        }
+
+        public async Task StopCapturingAsync(CancellationToken token)
+        {
+            if (ProbeStateInstalled != Interlocked.CompareExchange(ref _probeState, ProbeStateUninstalling, ProbeStateInstalled))
+            {
+                throw new InvalidOperationException();
```
wiktork, line 150: Perhaps add a reason related to the unexpected state?

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/FunctionProbesManager.cs
```diff
@@ -25,50 +28,172 @@ internal sealed class FunctionProbesManager : IFunctionProbesManager
...
+
+            FunctionProbesStub.InstrumentedMethodCache = null;
+            RequestFunctionProbeUninstallation();
+        }
+
+
+        public async Task StartCapturingAsync(IList<MethodInfo> methods, CancellationToken token)
         {
             if (methods.Count == 0)
             {
```
wiktork, line 183: Need to check disposed state as well.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Pipeline/ParameterCapturingPipeline.cs
```diff
@@ -50,7 +50,23 @@ public async Task RunAsync(CancellationToken stoppingToken)
...
             {
                 CapturingRequest request = await _requestQueue.Reader.ReadAsync(stoppingToken);
-                if (!TryStartCapturing(request.Payload))
+
+                void onFault(object? sender, InstrumentedMethod faultingMethod)
+                {
+                    _callbacks.ProbeFault(request.Payload.RequestId, faultingMethod);
+                }
+
+                using ScopeGuard sg = new(
```
wiktork, line 59: Consider just using a try/finally block here rather than creating ScopeGuard

src/Profilers/MutatingMonitorProfiler/ProbeInstrumentation/ProbeInstrumentation.cpp
```diff
@@ -113,15 +214,13 @@ void ProbeInstrumentation::DisableIncomingRequests()
 void ProbeInstrumentation::ShutdownBackgroundService()
 {
     DisableIncomingRequests();
+    m_managedCallbackQueue.Complete();
+    m_probeManagementThread.join();
```
wiktork, line 218: did you mean managedCallbackThread here?

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/FunctionProbesManager.cs
```diff
@@ -25,50 +28,176 @@ internal sealed class FunctionProbesManager : IFunctionProbesManager
...
         private long _disposedState;
-        private bool _isCapturing;
+
+        public event EventHandler<InstrumentedMethod>? OnProbeFault;
 
         public FunctionProbesManager(IFunctionProbes probes)
         {
             ProfilerResolver.InitializeResolver<FunctionProbesManager>();
 
+            RegisterFunctionProbeCallbacks(OnRegistration, OnInstallation, OnUninstallation, OnFault);
```
wiktork, line 67: Persisted pointers to functions across managed to native boundaries usually require pinning. OnRegistration will be pinned by the marshaller for the duration of the RegisterFunctionProbeCallbacks call, but after that function ends, the address may get moved around in memory during garbage collection.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/FunctionProbesManager.cs
```diff
@@ -94,14 +223,23 @@ public void StartCapturing(IList<MethodInfo> methods)
                 }
 
                 FunctionProbesStub.InstrumentedMethodCache = new ReadOnlyDictionary<ulong, InstrumentedMethod>(newMethodCache);
+
+                _installationTaskSource = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
```
wiktork, line 227: This looks like it could race with a Dispose call. For example, the first set of state checks could succeed, dispose could be called, and then a new _installationTaskSource could be created.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/FunctionProbesManager.cs
```diff
@@ -26,49 +29,197 @@ internal sealed class FunctionProbesManager : IFunctionProbesManager
...
+            }
+            catch
+            {
+                _uninstallationTaskSource = null;
+                _probeState = ProbeStateInstalled;
+                throw;
+            }
+
+            using CancellationTokenSource cts = CancellationTokenSource.CreateLinkedTokenSource(_disposalTokenSource.Token, token);
+            using IDisposable _ = cts.Token.Register(() =>
```
wiktork, line 184: What's the behavior here if we are already disposed? Will the token registration action run immediately?

wiktork (approved):

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/FunctionProbesManager.cs
```diff
@@ -26,49 +29,201 @@ internal sealed class FunctionProbesManager : IFunctionProbesManager
...
+                throw;
+            }
+
+            using CancellationTokenSource cts = CancellationTokenSource.CreateLinkedTokenSource(_disposalToken, token);
+            using IDisposable _ = cts.Token.Register(() =>
+            {
+                _uninstallationTaskSource.TrySetCanceled(cts.Token);
+            });
+            await _uninstallationTaskSource.Task.ConfigureAwait(false);
+        }
```
jander-msft, line 193: Should `_uninstallationTaskSource` be nulled out here?

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/FunctionProbesManager.cs
```diff
@@ -94,29 +249,71 @@ public void StartCapturing(IList<MethodInfo> methods)
...
+                    catch
+                    {
+                    }
+                }
+
+                _installationTaskSource.TrySetCanceled(cts.Token);
+            });
+
+            await _installationTaskSource.Task.ConfigureAwait(false);
         }
```
jander-msft, line 292: Should `_installationTaskSource` be nulled out?

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Pipeline/ParameterCapturingPipeline.cs
```diff
@@ -50,33 +50,48 @@ public async Task RunAsync(CancellationToken stoppingToken)
...
+                    try
+                    {
+                        await request.StopRequest.Task.WaitAsync(cts.Token).ConfigureAwait(false);
+                    }
+                    catch (OperationCanceledException)
+                    {
+
+                    }
+
+                    await _probeManager.StopCapturingAsync(stoppingToken).ConfigureAwait(false);
```
jander-msft, line 80: From an implementation-is-opaque perspective, this method will immediately cancel without doing any work if `stoppingToken` is already cancelled. Is that the behavior you want here? You might want an arbitrary but low timeout in the case that shutdown is requested so that probes can be uninstalled; there may be scenarios where an application is performing reload of the entire service collection and will restart the host without ending the process.

src/Profilers/MutatingMonitorProfiler/ProbeInstrumentation/ProbeInstrumentation.cpp
```diff
@@ -47,9 +63,84 @@ HRESULT ProbeInstrumentation::RegisterFunctionProbe(FunctionID enterProbeId)
 HRESULT ProbeInstrumentation::InitBackgroundService()
 {
     m_probeManagementThread = thread(&ProbeInstrumentation::WorkerThread, this);
+    //
+    // Create a dedicated thread for managed callbacks.
+    // Performing the callbacks will taint the calling thread preventing it
+    // from using certain ICorProfiler APIs marked as unsafe.
+    // Those functions will fail with CORPROF_E_UNSUPPORTED_CALL_SEQUENCE.
+    //
+    m_managedCallbackThread = thread(&ProbeInstrumentation::ManagedCallbackThread, this);
```
jander-msft, line 72: nit: Is there a way we can name both of these threads so we can identify them under the debugger?

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingService.cs
```diff
@@ -82,7 +48,7 @@ public ParameterCapturingService(IServiceProvider services)
                     IpcCommand.StopCapturingParameters,
                     OnStopMessage);
 
-                ILogger serviceLogger = services.GetService<ILogger<DotnetMonitor.ParameterCapture.Service>>()
+                _logger = services.GetService<ILogger<DotnetMonitor.ParameterCapture.Service>>()
```
jander-msft, line 51: Save this for a separate PR... I don't see error handling for when we try to load into a application that doesn't have logging configured; I know this is very unlikely, but it looks like we might crash the startup of the application.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/FunctionProbesManager.cs
jander-msft: I'm thinking in terms of letting GC collect it since we don't need it anymore.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingService.cs
jander-msft: Okay, I was missing the last catch and only saw the NotSupportException catch. Sorry!

jander-msft (approved):

PR #5131, opened 2023-08-11
[Exceptions] Request-Level Filtering

kkeirstead:
###### Summary

Adds basic filtering to `/exceptions` via a POST. This does not restrict which exceptions are stored (that will come in a later PR), just which ones are shown for a given request.

There are a few areas in this PR where I expect some discussion (and potentially changes) - I'll try to call a few of them out explicitly down below.

Also, this PR looks a lot bigger than it is (a lot of it is docs/tests) - most of the real changes are confined to a handful of files.

###### Release Notes Entry

Add the ability to filter which exceptions are displayed on the `/exceptions` route


src/Microsoft.Diagnostics.Monitoring.WebApi/Exceptions/ExceptionsConfiguration.cs
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        private static bool CompareValues(string configurationValue, string actualValue)
+        {
+            return actualValue.Contains(configurationValue, StringComparison.CurrentCultureIgnoreCase);
+        }
+    }
+
+    public class ExceptionConfiguration
+    {
```
kkeirstead, line 109: I believe there was discussion earlier about aligning this with `MethodDescription` IIRC - the way I have it here is designed to directly correlate to the `/exceptions` JSON payload for consistency and ease-of-use, which references methodName/className/moduleName. I could also see this growing in the future (for example, to include a list of parameter types), but I know we're starting simple for now

src/Microsoft.Diagnostics.Monitoring.WebApi/Exceptions/ExceptionsConfiguration.cs
```diff
@@ -0,0 +1,134 @@
...
+        {
+            if (exclude != false && !string.IsNullOrEmpty(configurationValue))
+            {
+                exclude = CompareValues(configurationValue, actualValue);
+            }
+        }
+
+        private static bool CompareValues(string configurationValue, string actualValue)
+        {
+            return actualValue.Contains(configurationValue, StringComparison.CurrentCultureIgnoreCase);
```
kkeirstead, line 104: In the original discussion for this, there was some debate over what level of fuzzy-matching we wanted - whether we wanted to check for an exact match unless a wildcard was provided. I did prototype it this way (and it'd be easy to swap it back), but this to me feels more user-friendly - when in doubt, I'd rather over-show than under-show and have people try to make sure they've got their names exactly right...e.g. I want to be able to say "InvalidOperationException" instead of needing to do "System.InvalidOperationException"

documentation/api/exceptions.md
```diff
@@ -3,7 +3,7 @@
 
 # Exceptions History - Get
 
-Captures a history of first chance exceptions that were thrown in the [default process](defaultprocess.md).
+Captures a history of first chance exceptions that were thrown in the specified process.
```
kkeirstead, line 6: I think this just fell through the cracks from the recent behavior change - it doesn't specifically apply to this change, but I figured I'd update it here.

src/Microsoft.Diagnostics.Monitoring.WebApi/Exceptions/ExceptionsConfiguration.cs
kkeirstead: I think standardization between the features would be good - what I was getting at is that, at the very least, I want `/exceptions` to be consistent with itself (hence why it's currently using module/class/method)

src/Microsoft.Diagnostics.Monitoring.WebApi/Exceptions/ExceptionsConfiguration.cs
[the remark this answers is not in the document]
kkeirstead: I covered a few of these things in https://github.com/dotnet/dotnet-monitor/pull/5131#discussion_r1291765141 - but in summary I'd rather bias this toward more results than less. In those examples, a user starts with something simple - if they get more results than they wanted, they can further refine their query. One case where this does fall apart (which may/may not be a big issue, and I can change this if we feel that it is) is users who have methods called "MyMethod", "MyMethod2", "MyMethod3", etc. - any search for "MyMethod" inherently includes "MyMethod2/3/etc.", unless you explicitly exclude "MyMethod2/3/etc." (which is a bit tedious).

In this case I'm not sure if the standardization makes sense because it's just for easier consumption, not because we actually need to know exactly what method the user means. For something like this, having to remember to type "MyNameSpace.MyClass+MyNestedType" instead of "MyClass" feels excessive - it's less precise, but that feels OK to me for this use-case...I think this will still do a good job of using broad strokes to eliminate most of what you don't care about

src/Microsoft.Diagnostics.Monitoring.WebApi/Exceptions/ExceptionsConfiguration.cs
```diff
@@ -0,0 +1,183 @@
...
+                {
+                    return true;
+                }
+            }
+
+            return fullValue.Equals(configurationValue, StringComparison.Ordinal);
+        }
+    }
+
+    public class ExceptionConfiguration
```
jander-msft, line 157: ExceptionFilter

src/Microsoft.Diagnostics.Monitoring.WebApi/Exceptions/ExceptionsConfiguration.cs
```diff
@@ -0,0 +1,183 @@
...
+        /// <summary>
+        /// By default, filters will check for a namespace.name match or a name match for ease-of-use.
+        /// This toggle allows users to limit the filter to only check for an exact namespace.name match.
+        /// </summary>
+        [JsonPropertyName("allowSimplifiedNames")]
+        public bool AllowSimplifiedNames { get; set; } = true;
+
+        private static bool CheckConfiguration(IExceptionInstance exception, List<ExceptionConfiguration> filterList, Func<ExceptionConfiguration, CallStackFrame, bool> evaluateFilterList)
+        {
+            var topFrame = exception.CallStack.Frames.Any() ? exception.CallStack.Frames.First() : null;
```
jander-msft, line 37:
```suggestion
            var topFrame = exception.CallStack.Frames.FirstOrDefault();
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Exceptions/ExceptionsConfiguration.cs
```diff
@@ -0,0 +1,183 @@
...
+                    CompareExcludeValues(configuration.ExceptionType, GetSimplifiedName(exception.TypeName), exception.TypeName, ref exclude);
+                }
+
+                return exclude ?? false;
+            };
+
+            return CheckConfiguration(exception, Exclude, evaluateFilterList);
+        }
+
+        private static string GetSimplifiedName(string name)
```
jander-msft, line 93: Take a look at using and return `ReadOnlySpan<char>`; as it is now, it allocates 3 new strings for each call to this method and the only thing it is doing is slicing the original string.

jander-msft, line 93: Can you write some examples in comments of the input and expected output?

src/Microsoft.Diagnostics.Monitoring.WebApi/Exceptions/ExceptionsConfiguration.cs
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        internal bool ShouldInclude(IExceptionInstance exception)
+        {
+            Func<ExceptionConfiguration, CallStackFrame, bool> evaluateFilterList = (configuration, topFrame) =>
+            {
+                bool include = true;
+                if (topFrame != null)
+                {
+                    CompareIncludeValues(configuration.MethodName, GetSimplifiedName(topFrame.MethodName), topFrame.MethodName, ref include);
```
jander-msft, line 57: Rather than always computing the simplified name, can you do something like `AllowSimplifiedNames ? GetSimplifiedName(topFrame.MethodName) ? topFrame.MethodName` and not have the called methods care about switching between the two values? This should speed up execution time and minimize allocations when they aren't used.

jander-msft: I see that there is some kind of fallback where if `AllowSimplifiedNames` is true and the value doesn't match the simple name, then it falls back to checking the full name.

kkeirstead:
Summarizing what was discussed earlier regarding fuzzy-matching - the current iteration of this change no longer includes simple names or fuzzy matching (everything must be fully written out). To improve our standardization, omitting this for now gives us more flexibility to add it (potentially using wildcards) in the future, without locking us into anything for this PR or establishing a one-off mechanism.

documentation/api/definitions.md
```diff
@@ -169,6 +169,26 @@ Object describing the list of event providers, keywords, event levels, and addit
 }
 ```
 
+## ExceptionConfiguration
```
jander-msft, line 172: nit: Update to `ExceptionFilter`

src/Microsoft.Diagnostics.Monitoring.WebApi/Exceptions/ExceptionsConfiguration.cs
```diff
@@ -0,0 +1,134 @@
...
+        public List<ExceptionFilter> Include { get; set; } = new();
+
+        /// <summary>
+        /// The list of exception configurations that determine which exceptions should be shown.
+        /// Each configuration is a logical OR, so if any of the configurations match, the exception isn't shown.
+        /// </summary>
+        [JsonPropertyName("exclude")]
+        public List<ExceptionFilter> Exclude { get; set; } = new();
+
+        private static bool CheckConfiguration(IExceptionInstance exception, List<ExceptionFilter> filterList, Func<ExceptionFilter, CallStackFrame, bool> evaluateFilterList)
```
jander-msft, line 28: At best, these should be extension methods in a separate class. Let's not mix serialization classes with business logic.

src/Microsoft.Diagnostics.Monitoring.WebApi/Exceptions/ExceptionsConfiguration.cs
```diff
@@ -0,0 +1,134 @@
...
+
+using Microsoft.Diagnostics.Monitoring.WebApi.Models;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi.Exceptions
+{
+    public class ExceptionsConfiguration
```
jander-msft, line 12: Does this need to be public?

src/Microsoft.Diagnostics.Monitoring.WebApi/Exceptions/ExceptionsConfiguration.cs
```diff
@@ -0,0 +1,134 @@
...
+            }
+        }
+
+        private static bool CompareValues(string configurationValue, string actualValue)
+        {
+            return actualValue.Equals(configurationValue, StringComparison.Ordinal);
+        }
+    }
+
+    public class ExceptionFilter
```
jander-msft, line 108:
```suggestion
    public sealed class ExceptionFilter
```

src/Microsoft.Diagnostics.Monitoring.WebApi/IExceptionsOperationFactory.cs
```diff
@@ -13,6 +14,6 @@ internal interface IExceptionsOperationFactory
         /// <summary>
         /// Creates an operation that produces an exceptions artifact.
         /// </summary>
-        IArtifactOperation Create(ExceptionFormat format);
+        IArtifactOperation Create(ExceptionFormat format, ExceptionsConfiguration configuration);
```
jander-msft, line 17: This couples the artifact implementation to the HTTP API implementation. Rather than using the same type, please either create another "settings" type, pass each piece of data as separate parameters, etc; adding a parameter such as `IEnumerable<ExceptionFilter> include` would also "violate" this idea.

src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/TestAppScenarios.cs
```diff
@@ -49,6 +49,7 @@ public static class Exceptions
             public static class SubScenarios
             {
                 public const string SingleException = nameof(SingleException);
+                public const string FilteringExceptions = nameof(FilteringExceptions);
```
jander-msft, line 52: nit: There isn't anything in this scenario that relates to filtering. It should be renamed to something like `MultipleExceptions`

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/HttpApi/ApiClient.cs
```diff
@@ -534,12 +535,27 @@ private async Task<ResponseStreamHolder> CaptureMetricsAsync(int processId, int
             throw await CreateUnexpectedStatusCodeExceptionAsync(responseBox.Value).ConfigureAwait(false);
         }
 
-        public async Task<ResponseStreamHolder> CaptureExceptionsAsync(int processId, ExceptionFormat format, CancellationToken token)
+        public async Task<ResponseStreamHolder> CaptureExceptionsAsync(HttpMethod method, ExceptionsConfiguration configuration, int processId, ExceptionFormat format, TimeSpan timeout)
```
jander-msft, line 538:
```suggestion
        public async Task<ResponseStreamHolder> CaptureExceptionsAsync(ExceptionsConfiguration configuration, int processId, ExceptionFormat format, TimeSpan timeout)
```

This is always a POST, no need to add a parameter.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/HttpApi/ApiClient.cs
```diff
@@ -534,12 +535,27 @@ private async Task<ResponseStreamHolder> CaptureMetricsAsync(int processId, int
...
+
+            return await CaptureExceptionsAsync(
+                method,
+                new StringContent(json, Encoding.UTF8, ContentTypes.ApplicationJson),
+                processId,
+                format,
+                timeoutSource.Token);
+        }
+
+        public async Task<ResponseStreamHolder> CaptureExceptionsAsync(HttpMethod method, HttpContent content, int processId, ExceptionFormat format, CancellationToken token)
```
jander-msft, line 552: This should be a private method and the original `CaptureExceptionsAsync(int processId, ExceptionFormat format, CancellationToken token)` should be added back that calls this method.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/HttpApi/ApiClientExtensions.cs
```diff
@@ -336,20 +337,12 @@ public static async Task<ResponseStreamHolder> CaptureMetricsAsync(this ApiClien
         }
 
         /// <summary>
-        /// GET /exceptions
+        /// GET or POST /exceptions
         /// </summary>
-        public static Task<ResponseStreamHolder> CaptureExceptionsAsync(this ApiClient client, int processId, ExceptionFormat format)
+        public static Task<ResponseStreamHolder> CaptureExceptionsAsync(this ApiClient client, ExceptionsConfiguration configuration, int processId, ExceptionFormat format)
```
jander-msft, line 342: Remove this in favor of the comments in ApiClient.cs

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsPipelineTests.cs
```diff
@@ -34,6 +34,43 @@ public sealed class ExceptionsPipelineTests
             "Microsoft.Diagnostics.Monitoring.StartupHook",
             TargetFrameworkMoniker.Net60);
 
+        private static ExceptionFilter simpleInvalidOperationException = new()
```
jander-msft, line 37: nit: all of these should be `readonly` and should be capitalized.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsPipelineTests.cs
```diff
@@ -421,6 +458,126 @@ public Task EventExceptionsPipeline_ReflectionTypeLoadException()
...
+                expectedInstanceCount: 1,
+                validate: instances =>
+                {
+                    IExceptionInstance instance = Assert.Single(instances);
+
+                    ExceptionsConfiguration full = new ExceptionsConfiguration()
+                    {
+                        Include = new() { fullInvalidOperationException }
+                    };
+                    Assert.True(full.ShouldInclude(instance));
```
jander-msft, line 478: For the Assert.True and Assert.False calls, add a text description as a second parameter describing the failed condition. This way, we can more easily identify from the tests reports what the issue is rather than just seeing "Expeceted: true, Actual: false"

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsPipelineTests.cs
jander-msft, line 37: Is simple vs full name still applicable?

src/Tools/dotnet-monitor/Exceptions/ExceptionsOperation.cs
```diff
@@ -159,10 +200,13 @@ await using (Utf8JsonWriter writer = new(stream, new JsonWriterOptions() { Inden
             await stream.WriteAsync(JsonRecordDelimiter, token);
         }
 
-        private static async Task WriteText(Stream stream, IReadOnlyList<IExceptionInstance> instances, CancellationToken token)
+        private async Task WriteText(Stream stream, IReadOnlyList<IExceptionInstance> instances, CancellationToken token)
         {
-            Dictionary<ulong, IExceptionInstance> priorInstances = new(instances.Count);
-            foreach (IExceptionInstance currentInstance in instances)
+            var filteredInstances = FilterExceptions(_configuration, instances);
```
jander-msft, line 205: This will cause inner exceptions to be filtered out and not reported in any manner.

src/Tools/dotnet-monitor/Exceptions/ExceptionsOperation.cs
```diff
@@ -85,12 +87,51 @@ public Task StopAsync(CancellationToken token)
 
         private async Task WriteJson(Stream stream, IReadOnlyList<IExceptionInstance> instances, CancellationToken token)
         {
-            foreach (IExceptionInstance instance in instances)
+            foreach (IExceptionInstance instance in FilterExceptions(_configuration, instances))
```
jander-msft, line 90: Same issue in https://github.com/dotnet/dotnet-monitor/pull/5131/files#r1300421481, but this is more problematic because we'd be reporting an ID for an exception that doesn't exist in the output.

kkeirstead: This is interesting...I actually viewed that as a feature and not a bug, but I guess it depends on how people would expect to use this. My thinking was that if we included the associated inner exceptions, that means users are seeing things that didn't match their filter's spec (and if they wanted to see a broader swath that might include the inner exceptions, you'd just expand your filter). I guess you could either be annoyed that there's a lot of inner exceptions that you didn't ask for in your filter, or you could be annoyed that you only have the top-level exceptions and you need to adjust your query (or GET /exceptions) to see the inner ones...

This starts to feel like you _could_ offer a knob for whether you want to include inner exceptions, but I'm not sure that we want to do that (at least not initially). If you weren't sure, it's probably safer from a breaking perspective to only include the top-level and then add in inner exceptions later if people ask - the other direction involves yanking out things that previously showed up (or adding a knob). But, if we know we definitely want the inner exceptions, then we can include them now.

src/Microsoft.Diagnostics.Monitoring.WebApi/Exceptions/ExceptionsConfiguration.cs
kkeirstead: I believe it does - since it's part of the API surface, when I tried switching it to internal it complained. I also confirmed that the configuration we used for logs is public.

src/Microsoft.Diagnostics.Monitoring.WebApi/IExceptionsOperationFactory.cs
kkeirstead: Let me know if you would prefer different naming/folder locations for this (iirc it's currently ExceptionsConfigurationSettings)

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsPipelineTests.cs
kkeirstead: I wasn't sure what you wanted for the message, but I figure the only variable thing is what exception we actually capture so that's what I'm showing - you can compare that against the configuration for a particular test to see any potential mismatch. Let me know if you want something different though.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsPipelineTests.cs
kkeirstead: Strictly speaking, no...I kept variants of these tests to make sure that all the logic for not allowing partial names had been removed (basically just confirming that the functionality was correctly pulled out). They're still in there, but I can pull these out if you'd like since it's probably not something we have to repeatedly test for regressions.

src/Tools/dotnet-monitor/Exceptions/ExceptionsOperation.cs
kkeirstead: See https://github.com/dotnet/dotnet-monitor/pull/5131#discussion_r1300695223

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/ExceptionsConfiguration.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// The list of exception configurations that determine which exceptions should be shown.
+        /// Each configuration is a logical OR, so if any of the configurations match, the exception isn't shown.
+        /// </summary>
+        [JsonPropertyName("exclude")]
+        public List<ExceptionFilter> Exclude { get; set; } = new();
+    }
+
+    public sealed class ExceptionFilter
+    {
+        /// <summary>
```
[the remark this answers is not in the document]
kkeirstead: Integrated with `MethodDescription` - there was an issue with `MethodDescription` requiring all fields to be provided (filtering doesn't have this requirement), so I introduced an `IMethodDescription` interface. If that's not an approach we want, I can change it to something else

src/Tools/dotnet-monitor/Exceptions/ExceptionsOperation.cs
kkeirstead: Latest changes add a sentinel value indicating that the inner exception was filtered out

documentation/api/exceptions-custom.md
```diff
@@ -0,0 +1,234 @@
...
+# Exceptions History - Custom
+
+Captures a history of first chance exceptions that were thrown in the specified process, with the ability to filter which exceptions are included in the response.
+
+>**Note**: This feature is not enabled by default and requires configuration to be enabled. The [in-process features](./../configuration/in-process-features-configuration.md) must be enabled since the exceptions history feature uses shared libraries loaded into the target application for collecting the exception information.
+
+## HTTP Route
+
+```http
+POST /exceptions HTTP/1.1
```
jander-msft, line 12:
```suggestion
POST /exceptions?pid={pid}&uid={uid}&name={name}&egressProvider={egressProvider}&tags={tags} HTTP/1.1
```

src/Tools/dotnet-monitor/Exceptions/ExceptionsOperation.cs
```diff
@@ -85,12 +87,51 @@ public Task StopAsync(CancellationToken token)
...
+            {
+                shouldInclude = false;
+                // filter out exceptions that don't match the filter
+                if (configuration.ShouldInclude(instance))
+                {
+                    shouldInclude = true;
+                }
+            }
+
+            if (configuration.Exclude.Count > 0)
```
jander-msft, line 123: If you process the exclude list first and the include list second, then you don't need a local variable to maintain whether the exception should ultimately be included.

jander-msft (dismissed):

jander-msft (approved):

PR #5142, opened 2023-08-14
Add `Starting` state to operations

schmittjoseph:
###### Summary

Add a new `Starting` state to operations. This is their new default state and will transition to `Running` once the operation has started. These changes are needed to provide a reliable way of adding functional tests for parameter capturing.

Collection rule actions already had a concept of a `startCompletionSource`. Generalize this to an `IStartable.Started` task that both collection rules and operations implement.  Doing so enables easy passthrough of the starting task with our wrapper operations (`HttpResponseEgressOperation` and `EgressOperation`), and easy handling of collection rule actions that are tracked as operations.

###### Release Notes Entry
Before an operation has started running, it now has a state of `Starting` in the `/operations` API.

### DIFF
--- a/src/Tools/dotnet-monitor/CollectionRules/Actions/CollectionRuleEgressActionBase.cs
+++ b/src/Tools/dotnet-monitor/CollectionRules/Actions/CollectionRuleEgressActionBase.cs
@@ -17,7 +17,6 @@ namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Actions
 
         protected EgressOperationStore EgressOperationStore { get; }
 
-
         protected CollectionRuleEgressActionBase(IServiceProvider serviceProvider, IProcessInfo processInfo, TOptions options)
             : base(processInfo, options)
         {


src/Microsoft.Diagnostics.Monitoring.WebApi/Operation/EgressOperationService.cs
```diff
@@ -46,7 +46,12 @@ private async Task ExecuteEgressOperation(EgressRequest egressRequest, Cancellat
 
                 try
                 {
-                    ExecutionResult<EgressResult> result = await egressRequest.EgressOperation.ExecuteAsync(_serviceProvider, token);
+                    Task<ExecutionResult<EgressResult>> executeTask = egressRequest.EgressOperation.ExecuteAsync(_serviceProvider, token);
+
+                    await egressRequest.EgressOperation.Started.WithCancellation(token);
```
jander-msft, line 51: nit: Let's use `WaitAsync` instead of `WithCancellation`; the latter is an extension method that we added to allow this cancellation notion in pre-.NET 6 versions and shouldn't be necessary anymore.

jander-msft, line 51: Why can't this and the marking the operation as running be part of `EgressOperation.ExecuteAsync`? This currently feels like it's leaking the details of the operation execution outside of where it needs to be understood.

jander-msft: Nevermind, sorry, I misunderstood that the execution state is actually stored within the operation store.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/HttpApi/ApiClientExtensions.cs
```diff
@@ -496,5 +493,28 @@ public static async Task<OperationStatusResponse> PollOperationToCompletion(this
 
             return operationResult;
         }
+
+        public static Task<OperationStatusResponse> PollOperationToRunning(this ApiClient apiClient, Uri operationUrl)
```
jander-msft, line 497: `PollOperationToAfterStarted`? I don't like this name, but it better describes the what the method is trying to achieve. When I first saw this method called before looking at the implementation, I immediately thought "but will this complete if it missed the `Running` state?"; I'm glad it doesn't care about the state being `Running`.

jander-msft (approved):

PR #5157, opened 2023-08-15
[Parameter Capturing] Fix handling of an assembly loaded multiple times via `AssemblyLoadContext` 

schmittjoseph:
###### Summary

Fix a bug in resolving methods that exist in multiple loaded instances of a given assembly via a custom `AssemblyLoadContext`.


src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/MethodResolver.cs
```diff
@@ -31,7 +31,11 @@ public MethodResolver()
                     continue;
                 }
 
-                if (!_nameToAssemblies.TryGetValue(assemblySimpleName, out List<Assembly>? assemblyList))
+                if (_nameToAssemblies.TryGetValue(assemblySimpleName, out List<Assembly>? assemblyList))
+                {
+                    assemblyList.Add(assembly);
```
wiktork, line 36: So do we instrument every instance of the assembly?

wiktork (approved):

PR #5169, opened 2023-08-17
[Parameter Capturing] Require .net 7+

schmittjoseph:
###### Summary

Due to https://github.com/dotnet/runtime/issues/88924 , require net7+ on a target process to use parameter capturing. 

In additional to blocking requests for unsupported runtime versions, this PR also blocks enabling the feature inside the target process (preventing the mutating profiler from being loaded and our in-proc `ParameterCapturingService` from being added).

### DIFF
--- a/src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
+++ b/src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
@@ -4,6 +4,7 @@
 using Microsoft.Diagnostics.Monitoring.WebApi;
 using Microsoft.Diagnostics.NETCore.Client;
 using Microsoft.Diagnostics.Tools.Monitor.LibrarySharing;
+using Microsoft.Diagnostics.Tools.Monitor.ParameterCapturing;
 using Microsoft.Extensions.FileProviders;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
@@ -98,7 +99,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Profiler
                     await ApplyNotifyOnlyProfilerAsync(client, nativeFileProvider, cancellationToken);
                 }
 
-                if (_inProcessFeatures.IsMutatingProfilerRequired)
+                if (IsMutatingProfilerNeeded(endpointInfo))
                 {
                     await ApplyMutatingProfilerAsync(client, nativeFileProvider, cancellationToken);
                 }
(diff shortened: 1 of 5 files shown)


src/Tools/dotnet-monitor/InProcessFeatures/InProcessFeaturesService.cs
```diff
@@ -30,7 +31,7 @@ public async Task ApplyInProcessFeatures(IEndpointInfo endpointInfo, Cancellatio
             try
             {
                 DiagnosticsClient client = new DiagnosticsClient(endpointInfo.Endpoint);
-                if (_parameterCapturingOptions.GetEnabled())
+                if (_parameterCapturingOptions.GetEnabled() && CaptureParametersOperation.IsEndpointRuntimeSupported(endpointInfo))
```
wiktork, line 34: Which version of the profiler is going to get loaded for these circumstances? We should not be loading the mutating profiler if we cannot use parameter capturing.

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -944,4 +944,7 @@
   <data name="ErrorMessage_ParameterCapturingNotAvailable" xml:space="preserve">
     <value>Parameter capturing is not available in the process.</value>
   </data>
+  <data name="ErrorMessage_ParameterCapturingRequiresAtLeastNet7" xml:space="preserve">
+    <value>Parameter capturing is not available in the process, the process needs to be using .net 7 or greater.</value>
```
jander-msft, line 948: nit: .NET 7 (capitalized)

src/Tools/dotnet-monitor/InProcessFeatures/InProcessFeaturesService.cs
jander-msft: I think we should add what @wiktork was getting at: don't bother to load the mutating profiler if parameter capturing is not applicable; this will help avoid customer reports that may blame us when we really aren't involved in the actual problem. Ideally, we would have a capability and requirement description system to declare this, but we don't at this time, so just add it as an explicit check in the profiler service.

src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
```diff
@@ -181,6 +182,20 @@ private async Task<string> CalculateRuntimeIdentifierAsync(DiagnosticsClient cli
             return runtimeIdentifier;
         }
 
+        private bool IsMutatingProfilerNeeded(IEndpointInfo endpointInfo)
```
jander-msft, line 185: Consider inlining this with the caller and then log something when a feature callers for the mutating profiler but it is not supported thus will not be loaded.

jander-msft (approved):

PR #5178, opened 2023-08-21
[Parameter Capturing] Add functional API tests

schmittjoseph:
###### Summary

Add functional tests for the `/parameters` API. Testing is scoped to basic end-to-end API usage since there's already extensive test coverage on the function probes, capturing of parameters, log statement formats, etc.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/ParameterCapturingTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/ParameterCapturingTests.cs
@@ -44,17 +44,20 @@ namespace Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests
             {
                 int processId = await appRunner.ProcessIdTask;
 
-                MethodDescription[] methods = new MethodDescription[]
+                CaptureParametersConfiguration config = new()
                 {
-                    new MethodDescription()
+                    Methods = new MethodDescription[]
                     {
-                        AssemblyName = Guid.NewGuid().ToString("D"),
-                        TypeName = Guid.NewGuid().ToString("D"),
-                        MethodName = Guid.NewGuid().ToString("D")
+                        new MethodDescription()
+                        {
+                            ModuleName = Guid.NewGuid().ToString("D"),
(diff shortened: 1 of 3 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/ParameterCapturingTests.cs
```diff
@@ -0,0 +1,191 @@
...
+            _httpClientFactory = serviceProviderFixture.ServiceProvider.GetService<IHttpClientFactory>();
+            _outputHelper = outputHelper;
+        }
+
+#if NET7_0_OR_GREATER
+        [Theory]
+        [MemberData(nameof(ProfilerHelper.GetArchitecture), MemberType = typeof(ProfilerHelper))]
+        public async Task UnresolvableMethodsFailsOperation(Architecture targetArchitecture)
+        {
+            await RunTestCaseCore(TestAppScenarios.ParameterCapturing.SubScenarios.AspNetApp, targetArchitecture, async (appRunner, apiClient) =>
```
jander-msft, line 43: nit: For every of these async methods, return the `Task` from the called method rather than awaiting; there's no need to use `async`/`await` and compile a state machine.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/ScenarioHelpers.cs
```diff
@@ -69,8 +69,8 @@ public static async Task<int> RunScenarioAsync(Func<ILogger, Task<int>> func, Ca
             return result;
         }
 
-        public static async Task<int> RunWebScenarioAsync<TStartup>(Func<ILogger, Task<int>> func, CancellationToken token)
-            where TStartup : class
+        public static async Task<int> RunWebScenarioAsync<TStartup>(Action<IServiceCollection> configureServices, Func<ILogger, Task<int>> func, CancellationToken token)
```
jander-msft, line 72: nit: Let's move the new parameter to the end and make it null by default so that the overload is not necessary.

jander-msft (dismissed):

jander-msft (approved):

PR #5181, opened 2023-08-22
[Parameter Capturing] Standardize `MethodDescription` fields

schmittjoseph:
###### Summary

In our `MethodDescription` model, switch from `assemblyName`->`moduleName` to standardize with our other features. See https://github.com/dotnet/dotnet-monitor/pull/5131#discussion_r1301871995


jander-msft:
I'd rather see `className` renamed to `typeName`; not all types are classes.

jander-msft:
> > I'd rather see `className` renamed to `typeName`; not all types are classes.
> 
> While technically true, "className" would seem to be the common way of referring to these. e.g. In VS there's the "Class View" which shows structs.

That might be due to historical reasons e.g. not all debugging capabilities were initially available for value types as compared to references types at the beginning of .NET Framework. Also, to be pedantic, it would be more accurate to have named it "Reference Type View" in order to be language agnostic, but most people think of the term as "class".

Using "type" is encompassing of what is supported and is unambiguous (e.g. no one will think "I wonder if structs are supported?").

jander-msft:
> > > > I'd rather see `className` renamed to `typeName`; not all types are classes.
> > > 
> > > 
> > > While technically true, "className" would seem to be the common way of referring to these. e.g. In VS there's the "Class View" which shows structs.
> > 
> > 
> > That might be due to historical reasons e.g. not all debugging capabilities were initially available for value types as compared to references types at the beginning of .NET Framework. Also, to be pedantic, it would be more accurate to have named it "Reference Type View" in order to be language agnostic, but most people think of the term as "class".
> > Using "type" is encompassing of what is supported and is unambiguous (e.g. no one will think "I wonder if structs are supported?").
> 
> We can certainly keep `typeName` in this PR, however if we do that I'd argue we need to standardize all of our other features to use `typeName` as well (e.g. `/stacks` and `/exceptions`) before we GA 8.0 as it'll be breaking changes.

100% agree

wiktork (dismissed):

jander-msft (approved):

PR #5186, opened 2023-08-23
[Parameter Capturing] Add deny list

schmittjoseph:
###### Summary

Add a simple deny list based on parent type. See `src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingService.cs` for the list, and `src/Tests/Microsoft.Diagnostics.Monitoring.HostingStartup.UnitTests/ParameterCapturing/TypeUtilsTests.cs` for examples of how it'd be enforced.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingStrings.resx
+++ b/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingStrings.resx
@@ -142,6 +142,6 @@
     <value>Too many requests</value>
   </data>
   <data name="UnresolvedMethodsFormatString" xml:space="preserve">
-    <value>Unable to resolve one or more methods: {0}</value>
+    <value>Unable to resolve one or more method descriptions: {0}</value>
   </data>
 </root>
--- a/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingStrings.Designer.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingStrings.Designer.cs
@@ -134,7 +134,7 @@ namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Unable to resolve one or more methods: {0}.
+        ///   Looks up a localized string similar to Unable to resolve one or more method descriptions: {0}.
         /// </summary>
         internal static string UnresolvedMethodsFormatString {
             get {


src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Pipeline/ParameterCapturingPipeline.cs
```diff
@@ -154,13 +156,36 @@ public bool TryComplete()
             return _requestQueue.Writer.TryComplete();
         }
 
+        private void ValidateMethods(MethodDescription[] methods)
```
jander-msft, line 159: While this works for this moment, if we ever add fuzzy/partial matching, this gets complicated real quickly. It might be better to move this to after where the actual methods are computed and run this validation on the methods.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Pipeline/ParameterCapturingPipeline.cs
```diff
@@ -154,13 +156,36 @@ public bool TryComplete()
...
+        }
+
         public void SubmitRequest(StartCapturingParametersPayload payload)
         {
             if (payload.Methods.Length == 0)
             {
                 throw new ArgumentException(nameof(payload.Methods));
             }
 
+            ValidateMethods(payload.Methods);
```
jander-msft, line 187: Could this be made into a singleton service? This would alleviate the need for the pipeline to understand this policy and focus on technical execution and would allow easier direct testing of the actual behavior (e.g. the service implementation is tested rather than just the helper method).

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingStrings.resx
```diff
@@ -117,6 +117,9 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="DeniedMethodsFormatString" xml:space="preserve">
+    <value>The following methods are not allowed to have their parameters captured: {0}</value>
```
jander-msft, line 121: This comment refers to the methods that are not allowed, but the usage of this string is in terms of the method descriptions. Consider better describing that the method descriptions are not allowed.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/DeniedMethodsExceptions.cs
```diff
@@ -0,0 +1,34 @@
...
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text;
+
+namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing
+{
+    internal sealed class DeniedMethodsExceptions : ArgumentException
+    {
+
```
wiktork, line 14: nit extra empty line

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingService.cs
```diff
@@ -32,6 +32,13 @@ internal sealed class ParameterCapturingService : BackgroundService, IParameterC
 
         private readonly ILogger? _logger;
 
+        private static readonly string[] s_typeDenyList = {
```
wiktork, line 35: Maybe add some comments here to denote this is a parent namespace or parent type and partials are ok along the ./+ boundaries only.

wiktork (approved):

jander-msft (approved):

PR #5200, opened 2023-08-25
[Parameter Capturing] Add more helpful messages when feature unavailable

### LINKED ISSUE #4996: Add more descriptive log messages for Parameter Capturing when unavailable
https://github.com/dotnet/dotnet-monitor/pull/4933#discussion_r1274145781

schmittjoseph:
###### Summary

When parameter capturing is not available in the target process, include the reason why in the exception.

Closes #4996


jander-msft:
Could you also log a warning message [here](https://github.com/dotnet/dotnet-monitor/blob/2bc993040b7a400d083ccfb5d7e910ad02e0b26d/src/Tools/dotnet-monitor/Profiler/ProfilerService.cs#L196) when the runtime is not supported? We can tell them up front that it won't work in addition to the changes you have here.

jander-msft (approved):

PR #5213, opened 2023-08-28
Change ExceptionsStore to allow multiple callbacks

jander-msft:
###### Summary

Refactor the `ExceptionsStore` callback notion in order to allow multiple callbacks and separate the callback behavior from the store implementation. These changes allow future callbacks to be added without having to chain exception store implementations together and to separate the callback concerns from the store concerns.

Overview of Changes
- Introduce `IExceptionsStoreCallback` and `IExceptionsStoreCallbackFactory` interfaces; the latter allows breaking circular dependencies between exception stores and callback implementations that would prevent DI composition.
- Refactor `ConfiguredExceptionsStore` into `ExceptionsStoreLimitsCallback` and `ExceptionsStoreLimitsCallbackFactory`.
- Refactor `ConfiguredExceptionsStoreTests` into `ExceptionsStoreTests` and `ExceptionsStoreLimitsCallbackTests` to separate testing concerns between testing the now-only exceptions store implementation and the callback implementation.


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsStoreTests.cs
```diff
@@ -0,0 +1,385 @@
...
+            using (CallbackRegistration.Register(aggregateCallback, thresholdCallback2))
+            {
+                RemoveExceptionInstance(store, ExpectedId2);
+
+                await thresholdCallback2.WaitForThresholdsAsync(CommonTestTimeouts.GeneralTimeout);
+            }
+
+            // Assert
+            IReadOnlyList<IExceptionInstance> instances = store.GetSnapshot();
+            Assert.Equal(ExpectedCount, instances.Count);
```
kkeirstead, line 143: nit: Could probably Assert Single here since `ExpectedCount` is 1

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsStoreTests.cs
```diff
@@ -0,0 +1,385 @@
...
+
+            IExceptionInstance instance1 = instances[0];
+            Assert.NotNull(instance1);
+            Assert.Equal(ExpectedId1, instance1.Id);
+        }
+
+        /// <summary>
+        /// Validates adding exceptions and removing all exceptions
+        /// will invoke the callback and the store be empty.
```
kkeirstead, line 152:
```suggestion
        /// will invoke the callback and the store will be empty.
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsStoreTests.cs
```diff
@@ -0,0 +1,385 @@
...
+            ulong ExpectedId3 = 3;
+
+            // Arrange
+            AggregateCallback aggregateCallback = new();
+            await using ExceptionsStore store = new(new[] { new PassThroughCallbackFactory(aggregateCallback) });
+
+            IExceptionsNameCache cache = CreateCache();
+
+            // Act
+            ThresholdCallback thresholdCallback1 = new(3, 0);
```
kkeirstead, line 168: nit: For readability, consider naming these more descriptively (one adds, the other removes)

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsStoreTests.cs
```diff
@@ -0,0 +1,385 @@
...
+                await thresholdCallback2.WaitForThresholdsAsync(CommonTestTimeouts.GeneralTimeout);
+            }
+
+            // Assert
+            Assert.Empty(store.GetSnapshot());
+        }
+
+        /// <summary>
+        /// Validates adding exceptions and removing all exceptions in an unordered manner
+        /// will invoke the callback and the store be empty.
```
kkeirstead, line 194:
```suggestion
        /// will invoke the callback and the store will be empty.
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/ExceptionsStoreTests.cs
```diff
@@ -0,0 +1,385 @@
...
+
+            private int _addCount;
+            private int _removeCount;
+
+            public ThresholdCallback(int addThreshold, int removeThreshold = 0)
+            {
+                _addThreshold = addThreshold;
+                _removeThreshold = removeThreshold;
+
+                SetIfAtTarget(_addCompletionSource, addThreshold, 0);
```
kkeirstead, line 308: I don't think this matters from a functionality standpoint, but this felt flipped while I was reading this (i.e. it should be `SetIfAtTarget(_addCompletionSource, 0, addThreshold`). It only applies if they're both 0 so it doesn't matter, but this states that the threshold is the value instead of the target (which is how this is represented in `AfterAdd` and `AfterRemove`). Maybe I'm misinterpreting it, but it seemed off when I was reading this.

src/Tools/dotnet-monitor/Exceptions/ExceptionsStore.cs
```diff
@@ -20,19 +21,27 @@ internal sealed class ExceptionsStore : IExceptionsStore, IAsyncDisposable
...
         private readonly KeyedCollection<ulong, ExceptionInstance> _instances = new ExceptionInstanceCollection();
         private readonly Task _processingTask;
 
         private long _disposalState;
 
-        public ExceptionsStore(ExceptionsStoreCallback callback)
+        public ExceptionsStore(IEnumerable<IExceptionsStoreCallbackFactory> factories)
         {
-            _callback = callback;
+            ArgumentNullException.ThrowIfNull(factories);
```
kkeirstead, line 34: Confirming whether we only care if it's null, but not if it's empty.

src/Tools/dotnet-monitor/Exceptions/ExceptionsStoreLimitsCallback.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        public override void BeforeAdd(IExceptionInstance instance)
+        {
+            _outerExceptionMap.Add(instance.Id, new List<ulong>());
+            _innerExceptionMap.Add(instance.Id, instance.InnerExceptionIds.ToList());
+
+            // Make sure all of the inner exceptions are no longer top-level exceptions.
+            foreach (ulong innerExceptionId in instance.InnerExceptionIds)
+            {
+                if (_outerExceptionMap.TryGetValue(innerExceptionId, out List<ulong> outerExceptions))
```
kkeirstead, line 54: I know this code was just moved from an existing file, but I wanted to double check that we didn't overlook a scenario in the current version. If `_outerExceptionMap` doesn't contain the key of `innerExceptionId`, don't we still want to set `_outerExceptionMap[innerExceptionId] = new() { instance.Id }`?

jander-msft: That would mean that an inner exception was not reported previously because each exception gets added to the `_outerExceptionMap` at the begin of `BeforeAdd`, but this should not be possible unless events are dropped or exceptions are filtered at the collection site. I can take a look at mocking up this scenario in a separate change and addressing it there.

jander-msft: /todo Exception store callbacks should gracefully handle inner exceptions that were not reported

kkeirstead (approved):

PR #5217, opened 2023-08-28
[Exceptions] Configuration-Level Filtering

kkeirstead:
###### Summary

This PR complements #5131, allowing users to add configuration to control which exceptions are kept (and which are immediately discarded without entering the store). 

###### Release Notes Entry

Add the ability to filter which exceptions are collected by `dotnet-monitor`.


src/Microsoft.Diagnostics.Monitoring.WebApi/Exceptions/ExceptionsSettingsFactory.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace Microsoft.Diagnostics.Monitoring.WebApi.Exceptions
+{
+    /// <summary>
+    /// Utility class to create ExceptionsConfigurationSettings.
+    /// </summary>
+    internal static class ExceptionsSettingsFactory
+    {
+        public static ExceptionsConfigurationSettings ConvertExceptionsConfiguration(ExceptionsConfiguration configuration)
+        {
+            ExceptionsConfigurationSettings configurationSettings = new();
```
kkeirstead, line 15: Consider adding a null check here

documentation/configuration/in-process-features-configuration.md
```diff
@@ -187,3 +187,66 @@ Similarly, the exceptions history feature can be individually disabled by settin
     value: "false"
   ```
 </details>
+
+### Filtering
+
+Which exceptions are collected and stored can be filtered via configuration using [ExceptionsConfiguration](../api/definitions.md#ExceptionsConfiguration). This can be useful for noisy exceptions that are not useful to capture - an example of this may be disregarding any `TaskCanceledException` that the target application produces.
+
+Note that this is different than [real-time filtering](../api/exceptions-custom.md), which does **not** restrict the collection of exceptions and is solely responsible for determining which exceptions are displayed when invoking the `/exceptions` endpoint.
```
jander-msft, line 195:
```suggestion
Note that this is different than [real-time filtering](../api/exceptions-custom.md), which does **not** restrict the collection of exceptions and is solely responsible for determining which exceptions are displayed when using the `/exceptions` route.
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Exceptions/ExceptionsOptions.cs
```diff
@@ -22,5 +23,7 @@ internal sealed class ExceptionsOptions :
         [Range(1, int.MaxValue)]
         [DefaultValue(ExceptionsOptionsDefaults.TopLevelLimit)]
         public int? TopLevelLimit { get; set; }
+
+        public ExceptionsConfiguration Filters { get; set; }
```
jander-msft, line 27: `CollectionFilters` to make it clear that these are filters are applied at collection time.

src/Microsoft.Diagnostics.Monitoring.WebApi/Exceptions/ExceptionsOptions.cs
```diff
@@ -1,11 +1,12 @@
...
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using Microsoft.Diagnostics.Monitoring.WebApi;
+using Microsoft.Diagnostics.Monitoring.Options;
+using Microsoft.Diagnostics.Monitoring.WebApi.Models;
 using System.ComponentModel;
 using System.ComponentModel.DataAnnotations;
 
-namespace Microsoft.Diagnostics.Monitoring.Options
+namespace Microsoft.Diagnostics.Monitoring.WebApi.Exceptions
```
jander-msft, line 9: Why are all of these classes being moved to the WebApi assembly? They are first and foremost options. Leaving them in the Options assembly should make them available to all other assemblies.

jander-msft (commented):
There are merge conflicts that need to be fixed as well.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/OptionsExtensions.cs
```diff
@@ -89,6 +89,19 @@ public static RootOptions EnableInProcessFeatures(this RootOptions options)
             return options;
         }
 
+        public static RootOptions SetExceptionFiltering(this RootOptions options, ExceptionsConfiguration configuration)
```
jander-msft, line 92: nit: This method should only do enough to set the configuration elements to provide the filter configuration; it shouldn't enable in-process features and it shouldn't enable exceptions itself. Let the caller do those (and technically, only exceptions needs to be enabled, not all of in-process features).

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/ExceptionsTests.cs
```diff
@@ -457,6 +457,288 @@ public async Task Exceptions_FilterIncludeDetailed(Architecture targetArchitectu
...
+                configureApp: runner =>
+                {
+                    runner.Architecture = targetArchitecture;
+                    runner.EnableMonitorStartupHook = true;
+                },
+                configureTool: runner =>
+                {
+                    ExceptionsConfiguration configuration = new();
+                    configuration.Include.Add(
+                        new()
```
jander-msft, line 489: It's non-obvious what the type is here, so use the type name rather than inferring it. This applies to all of the tests (new and existing).

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/ExceptionsTests.cs
```diff
@@ -457,6 +457,288 @@ public async Task Exceptions_FilterIncludeDetailed(Architecture targetArchitectu
...
+                appValidate: async (appRunner, apiClient) =>
+                {
+                    await GetExceptions(apiClient, appRunner, ExceptionFormat.PlainText);
+
+                    ValidateSingleExceptionText(
+                        SystemInvalidOperationException,
+                        ExceptionMessage,
+                        FrameClassName,
+                        FrameMethodName,
+                        new() { FrameParameterType, FrameParameterType });
```
jander-msft, line 478: Please use type name

src/Tools/dotnet-monitor/Exceptions/ExceptionsStore.cs
```diff
@@ -26,10 +28,17 @@ internal sealed class ExceptionsStore : IExceptionsStore, IAsyncDisposable
         private readonly CancellationTokenSource _disposalSource = new();
         private readonly KeyedCollection<ulong, ExceptionInstance> _instances = new ExceptionInstanceCollection();
         private readonly Task _processingTask;
+        private ExceptionsConfigurationSettings _configuration;
 
         private long _disposalState;
 
-        public ExceptionsStore(IEnumerable<IExceptionsStoreCallbackFactory> factories)
+        public ExceptionsStore(IEnumerable<IExceptionsStoreCallbackFactory> factories, IOptions<ExceptionsOptions> options)
+            : this(factories, options.Value.CollectionFilters ?? new())
```
jander-msft, line 36: Let's push the unrolling of the options into factory, replace this ctor with one that just takes the factories (and update the tests that are passing in an empty configuration to call this ctor), and have the factory call the second ctor.

src/Tools/dotnet-monitor/Exceptions/ExceptionsStore.cs
```diff
@@ -26,10 +28,17 @@ internal sealed class ExceptionsStore : IExceptionsStore, IAsyncDisposable
...
+        {
+
+        }
+
+        public ExceptionsStore(IEnumerable<IExceptionsStoreCallbackFactory> factories, ExceptionsConfiguration configuration)
```
jander-msft, line 41:
```suggestion
        public ExceptionsStore(IEnumerable<IExceptionsStoreCallbackFactory> factories, ExceptionsConfigurationSettings settings)
```

src/Tools/dotnet-monitor/Exceptions/ExceptionsStore.cs
kkeirstead: I'm not sure I'm understanding what you have in mind here - by pushing the responsibility of unrolling the options to the `IExceptionsStoreCallbackFactory`, won't we repeatedly do this for each factory (instead of once for the `IEnumerable` of factories)? Additionally, I'm unclear on "have the factory call the second ctor" - if we're expecting an `IEnumerable` of factories in the second ctor, I wouldn't expect us to be calling that from the factory itself. Let me know if I'm misinterpreting your comment and I'll go ahead and refactor it

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/ExceptionsTests.cs
jander-msft: This is a similar ask as https://github.com/dotnet/dotnet-monitor/pull/5217#discussion_r1319172815. It's not obvious what `new ()` is creating here.

src/Tools/dotnet-monitor/Exceptions/ExceptionsStore.cs
jander-msft: Sorry, I forgot that `ExceptionsStore` is in the service container thus it would need the `IOptions<>` instance on it directly. Please disregard.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/ExceptionsTests.cs
kkeirstead: Oh, I totally interpreted that the wrong way *face palms* - I'll add that in the next revision.

jander-msft (approved):

PR #5230, opened 2023-08-31
[Parameter Capturing] Introduce cache for parameter formatting

schmittjoseph:
###### Summary

This PR is in preparation for a new way of formatting parameter values that is more expensive and so requires a caching layer. 

There's no user-facing changes in this PR.

Overview of changes:
- Break out the `PrettyPrinter.FormatObject` helper method into a series of formatters (see `ObjectFormatting/Formatters/RuntimeFormatters.cs`).
- Introduce initial format specifier support (a subset of https://learn.microsoft.com/visualstudio/debugger/format-specifiers-in-csharp#format-specifiers) to supplement current behavior while future proofing this refactor for new formatters.
- Refactor the `FunctionProbesStub.InstrumentMethodCache` into a general-purpose cache for parameter capturing (see `FunctionProbes/FunctionProbesCache.cs`. This cache's lifetime is still tied to the parameter capturing session.
- Rename `PrettyPrinter.cs` -> `MethodTemplateStringGenerator.cs`

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/FunctionProbesManager.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/FunctionProbesManager.cs
@@ -240,11 +240,10 @@ namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing.Fun
                 throw new InvalidOperationException(string.Format(CultureInfo.InvariantCulture, ParameterCapturingStrings.ErrorMessage_ProbeStateMismatchFormatString, ProbeStateUninstalled, _probeState));
             }
 
+            ObjectFormatterCache newObjectFormatterCache = new();
+            Dictionary<ulong, InstrumentedMethod> newMethodCache = new(methods.Count);
             try
             {
-                Dictionary<ulong, InstrumentedMethod> newMethodCache = new(methods.Count);
-                ObjectFormatterCache newObjectFormatterCache = new();
-
                 List<ulong> functionIds = new(methods.Count);
                 List<uint> argumentCounts = new(methods.Count);
                 List<uint> boxingTokens = new();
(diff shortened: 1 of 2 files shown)


src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/FunctionProbesManager.cs
```diff
@@ -277,7 +282,7 @@ public async Task StartCapturingAsync(IList<MethodInfo> methods, CancellationTok
             }
             catch
             {
-                FunctionProbesStub.InstrumentedMethodCache = null;
+                FunctionProbesStub.Cache = null;
```
wiktork, line 285: Did we want something more deterministic here? For example, call Clear() first and then remove the reference?

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ObjectFormatting/Formatters/RuntimeFormatters.cs
```diff
@@ -0,0 +1,40 @@
...
+    internal static class RuntimeFormatters
+    {
+        public static string IConvertibleFormatter(object obj, FormatSpecifier formatSpecifier)
+        {
+            if (obj is not string)
+            {
+                formatSpecifier |= FormatSpecifier.NoQuotes;
+            }
+
+            string formatted = ((IConvertible)obj).ToString(CultureInfo.InvariantCulture);
```
jander-msft, line 18: For this and IFormattableFormatter, could you check the cast and throw a more informative exception? e.g.

```
if (obj is not IConvertable convertable)
{
    throw new ArgumentException();
}
```

Better yet, change the parameter to be IConvertable and have the caller test using this casting methodology rather than checking with IsAssignableTo/From.

jander-msft: Ignore the second part, I see the factory is provided the type and not the actual object.

jander-msft (approved):

PR #5260, opened 2023-09-05
Allow Helix tests to use internal assets

jander-msft:
###### Summary

Allow Helix test setup to get the blob storage internal container token to read packages and files from the container.

### DIFF
--- a/eng/helix/Helix.proj
+++ b/eng/helix/Helix.proj
@@ -113,6 +113,14 @@
     <HelixTargetQueue Include="@(HelixAvailableTargetQueue)" />
   </ItemGroup>
 
+  <!-- Additional package feeds -->
+  <ItemGroup>
+    <AdditionalDotNetPackageFeed Include="https://dotnetbuilds.blob.core.windows.net/internal"
+                                 Condition="'$(SYSTEM_TEAMPROJECT)' == 'internal'">
+      <SasToken>$([System.Environment]::GetEnvironmentVariable('DotNetBuildsInternalContainerReadToken'))</SasToken>
+    </AdditionalDotNetPackageFeed>
+  </ItemGroup>
+
   <!-- Correlation Payload: SDK (to use "dotnet test") -->
   <PropertyGroup>
     <IncludeDotNetCli>true</IncludeDotNetCli>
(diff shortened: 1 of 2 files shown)


eng/pipelines/jobs/test-binaries.yml
```diff
@@ -101,6 +101,7 @@ jobs:
         -projects $(Build.SourcesDirectory)/eng/helix/Helix.proj
         -skipmanaged
         -skipnative
+        /p:DotNetBuildsInternalContainerReadToken=$(dotnetbuilds-internal-container-read-token-base64)
```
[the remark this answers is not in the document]
jander-msft: Good point. I'll conditionally include this.

eng/helix/Helix.proj
```diff
@@ -113,6 +113,14 @@
     <HelixTargetQueue Include="@(HelixAvailableTargetQueue)" />
   </ItemGroup>
 
+  <!-- Additional package feeds -->
+  <ItemGroup>
+    <AdditionalDotNetPackageFeed Include="https://dotnetbuilds.blob.core.windows.net/internal"
```
jander-msft, line 118: Conditionally include if DotNetBuildsInternalContainerReadToken is not empty.

jander-msft: Actually, this is not needed since internal builds will always have this token and it's already conditioned to only be included for internal builds.

eng/helix/Helix.proj
```diff
@@ -113,6 +113,14 @@
...
+                                 Condition="'$(SYSTEM_TEAMPROJECT)' == 'internal'">
+      <SasToken>$([System.Environment]::GetEnvironmentVariable('DotNetBuildsInternalContainerReadToken'))</SasToken>
```
wiktork, line 120: I was under the impression that enironment variables are automatically mapped to properties. Is this not the case for the token?

wiktork (approved):

eng/helix/Helix.proj
jander-msft: This is how ASP.NET does it (I basically copied their solution). I can make a follow up change that switches to using the property if that works.

jander-msft: You were right, it needs to be a property. ASP.NET must be setting it via the environment block as well somewhere. Fixed in https://github.com/dotnet/dotnet-monitor/pull/5262

PR #5289, opened 2023-09-07
[Parameter Capturing] Add a DebuggerDisplay object formatter

schmittjoseph:
###### Summary

This PR adds the ability to format objects using their [`DebuggerDisplayAttribute`](https://learn.microsoft.com/dotnet/api/system.diagnostics.debuggerdisplayattribute).

Due the amount of complexity involved with the expression parsing and binding, I've opted to have this PR just contain the technical ability and not exposing the feature to the user. Adding the ability to use this feature (and documentation) will be done in a separate PR.

Initial limitations:
- Complex expressions are not supported (they are strongly discouraged from being used in a `DebuggerDisplayAttribute` anyways). Only properties, fields, and parameter-less methods are currently supported. Chained expressions are supported so long as each segment of the chain meets the above criteria.
- The `Name` field on the `DebuggerDisplayAttribute` won't be used.
-  Only the `nq` (NoQuotes) and `nse` (NoSideEffects) format specifiers are supported for now. Support for additional ones can be added in the future.
-  There's limited visibility into the expression parsing and binding when dealing with malformed expressions. If a failure occurs, we silently handled it and fall back to a runtime based object formatter.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ObjectFormatting/Formatters/DebuggerDisplay/DebuggerDisplayFormatter.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ObjectFormatting/Formatters/DebuggerDisplay/DebuggerDisplayFormatter.cs
@@ -21,37 +21,50 @@ namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing.Obj
                 return null;
             }
 
-            DebuggerDisplayAttributeValue? attribute = GetDebuggerDisplayAttribute(objType);
-            if (attribute == null)
-            {
-                return null;
-            }
-
             //
-            // We found an attribute.
-            // The last encompassing type will be the source of the attribute.
-            // Check if we've already processed this base type, and if so return the precomputed result.
+            // Guard against exceptions from any of these methods as we're dealing with user-provided input.
+            // None of these methods should throw but since we're potentially running inside of user code
+            // we want to be safe here.
             //
-            if (formatterCache != null &&
-                formatterCache.TryGetFormatter(attribute.Value.EncompassingTypes[^1], out ObjectFormatterFunc? precachedFormatter) &&
-                precachedFormatter != null)
+            try
             {
-                return new FormatterFactoryResult(precachedFormatter, attribute.Value.EncompassingTypes);
-            }
+                DebuggerDisplayAttributeValue? attribute = GetDebuggerDisplayAttribute(objType);
+                if (attribute == null || attribute.Value.EncompassingTypes.Count == 0)
+                {
+                    return null;
+                }
 
-            ParsedDebuggerDisplay? parsedDebuggerDiplay = DebuggerDisplayParser.ParseDebuggerDisplay(attribute.Value.Text);
-            if (parsedDebuggerDiplay == null)
+                //
+                // We found an attribute.
+                // The last encompassing type will be the source of the attribute.
+                // Check if we've already processed this base type, and if so return the precomputed result.
+                //
+                if (formatterCache != null &&
+                    formatterCache.TryGetFormatter(attribute.Value.EncompassingTypes[^1], out ObjectFormatterFunc? precachedFormatter) &&
+                    precachedFormatter != null)
+                {
+                    return new FormatterFactoryResult(precachedFormatter, attribute.Value.EncompassingTypes);
+                }
+
+                ParsedDebuggerDisplay? parsedDebuggerDiplay = DebuggerDisplayParser.ParseDebuggerDisplay(attribute.Value.Text);
+                if (parsedDebuggerDiplay == null)
+                {
+                    return null;
+                }
+
+                ObjectFormatterFunc? formatter = ExpressionBinder.BindParsedDebuggerDisplay(objType, parsedDebuggerDiplay.Value);
+                if (formatter == null)
+                {
+                    return null;
+                }
+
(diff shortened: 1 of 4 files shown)


src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ObjectFormatting/Formatters/DebuggerDisplay/ExpressionBinder.cs
```diff
@@ -0,0 +1,234 @@
...
+
+            if (field == null)
+            {
+                return null;
+            }
+
+            return new ExpressionEvaluator(
+                field.IsStatic
+                    ? (_) => field.GetValue(obj: null)
+                    : field.GetValue,
```
jander-msft, line 193: Doesn't this need the `obj` to be passed in order to get the value of an instance field?

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ObjectFormatting/ObjectFormatter.cs
```diff
@@ -40,6 +43,11 @@ private static class Internal
 
         public static string FormatObject(ObjectFormatterFunc formatterFunc, object obj, FormatSpecifier formatSpecifier = FormatSpecifier.None)
         {
+            if ((formatSpecifier & FormatSpecifier.NoSideEffects) != 0)
```
wiktork, line 46: So to be clear, the idea is that if someone specifies nfe we will error out since we cannot evaluate that way?

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ObjectFormatting/Formatters/DebuggerDisplay/DebuggerDisplayParser.cs
```diff
@@ -0,0 +1,215 @@
...
+            List<Expression> expressions = new();
+
+            for (int i = 0; i < debuggerDisplay.Length; i++)
+            {
+                char c = debuggerDisplay[i];
+                switch (c)
+                {
+                    case '\\':
+                        // Escape sequence is not currently supported.
+                        return null;
```
wiktork, line 36: Rather than returning null in cases that are malformed or not supported, could we actually return a short error message /code as the format string itself?  This would make it a lot easier to understand when making syntax errors.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ObjectFormatting/Formatters/DebuggerDisplay/DebuggerDisplayParser.cs
```diff
@@ -0,0 +1,215 @@
...
+            //
+            charsRead = 0;
+            if (expression.IsEmpty)
+            {
+                return null;
+            }
+
+            // Ensure the first char is the start of an expression.
+            ReadOnlySpan<char> spanExpression = expression.Span;
+            if (spanExpression[0] != '{')
```
wiktork, line 91: nit: consts for token literals

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ObjectFormatting/Formatters/DebuggerDisplay/DebuggerDisplayFormatter.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Reflection;
+using static Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing.ObjectFormatting.Formatters.DebuggerDisplay.DebuggerDisplayParser;
+
+namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing.ObjectFormatting.Formatters.DebuggerDisplay
+{
+    internal static class DebuggerDisplayFormatter
+    {
+        internal record struct DebuggerDisplayAttributeValue(string Text, IList<Type> EncompassingTypes);
+
+        public static FormatterFactoryResult? GetDebuggerDisplayFormatter(Type objType, ObjectFormatterCache? formatterCache = null)
```
wiktork, line 17: Once we integrate this into callable code, please make sure to guard against exceptions from parsing.

src/Tests/Microsoft.Diagnostics.Monitoring.HostingStartup.UnitTests/ParameterCapturing/ObjectFormatting/Formatters/DebuggerDisplay/ExpressionBinderTests.cs
```diff
@@ -0,0 +1,75 @@
...
+        [InlineData("GetCountAsString()", true, "10")]
+        [InlineData("DoesntExist()", false, null)]
+        [InlineData("WithArgs(Count)", false, null)]
+        [InlineData("Count", true, 10)]
+        [InlineData("Field", true, 10)]
+        [InlineData("NoReturnType()", true, null)]
+        // Bad chain
+        [InlineData(".", false, null)]
+        // Chained expression with implicit this type change
+        [InlineData("Recursion().RecursionProp.MyUri.Host", true, "www.bing.com")]
```
wiktork, line 52: IIRC example.com example.net example.org are all reserved names for purposes like this

wiktork (dismissed):

jander-msft (approved):

PR #5385, opened 2023-09-19
Preserve scope information in structured logs

schmittjoseph:
###### Summary

Our current `KeyValueLogScope` implementation is incompatible with structured logging. Fullfill the expected `IReadOnlyList<KeyValuePair<string, object>>` contract to fix this.

**NOTE**: An extension method was added to polyfill the existing usage that relied on having a backing dictionary for the scope.


jander-msft:
> Our current KeyValueLogScope implementation is incompatible with structured logging.

Can you elaborate on this? I don't know specifically what you are trying to fix. And update the comments as to why IReadOnlyList is needed instead of IReadOnlyCollection?

jander-msft (approved):

PR #5391, opened 2023-09-21
[Exceptions] `CollectExceptions` Action Filtering

kkeirstead:
###### Summary

Complements #5131 and #5217 - adds the ability to include/exclude exceptions from the result of a `CollectExceptions` action. Note that this does not limit which exceptions are collected, just which ones are presented. Also added some testing for the `CollectExceptions` action.

###### Release Notes Entry

Add the ability to filter which exceptions are included with the `CollectExceptions` action.


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/ExceptionsTests.cs
```diff
@@ -472,7 +472,7 @@ public async Task Exceptions_FilterIncludeBasic_Configuration(Architecture targe
                     ValidateSingleExceptionText(
                         SystemInvalidOperationException,
                         ExceptionMessage,
-                        FrameClassName,
+                        FrameTypeName,
```
jander-msft, line 475: This should already be fixed; try to rebase/merge.

jander-msft (approved):

PR #5392, opened 2023-09-21
[Parameter Capturing] Generate log scopes with additional data

schmittjoseph:
###### Summary

When logging parameters, capture additional data that can help with parameter correlation and publish it via log scopes. This applies to both user code and system code. Additional data captured:
- Activity Id + Format
- Timestamp (important in the case of system code which will be logged later on)
- Thread Id
- Information about the method that parameters came from (module, type, method).

To help with testing, update our test logger with the ability to track active scope data and associate it with log records.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/TestLogger.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/TestLogger.cs
@@ -61,7 +61,6 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon
                 if (scope.State is IReadOnlyList<KeyValuePair<string, object>> keyValueScope)
                 {
                     keyValueScopes.Add(keyValueScope);
-                    continue;
                 }
             }
             _events.Add(new LogRecordEntry(id, category, message, keyValueScopes));


src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/TestLogger.cs
```diff
@@ -3,17 +3,69 @@
...
-        public void Add(EventId id, string category, string message) => _events.Add(new LogRecordEntry(id, category, message));
+        public void Add(EventId id, string category, string message, ScopeTracker scopeTracker)
+        {
+            List<IReadOnlyList<KeyValuePair<string, object>>> keyValueScopes = new();
+            foreach (Scope scope in scopeTracker.Scopes.Values)
+            {
+                if (scope.State is IReadOnlyList<KeyValuePair<string, object>> keyValueScope)
+                {
+                    keyValueScopes.Add(keyValueScope);
+                    continue;
```
jander-msft, line 64: I don't think this `continue` is necessary.

src/Tests/Microsoft.Diagnostics.Monitoring.HostingStartup.UnitTests/ParameterCapturing/ParameterCapturingLoggerTests.cs
```diff
@@ -0,0 +1,102 @@
...
+            // Act
+            IList<LogRecordEntry> entries = TestCore(mode);
+
+            // Assert
+            LogRecordEntry entry = Assert.Single(entries);
+            IReadOnlyList<KeyValuePair<string, object>> rawScope = Assert.Single(entry.Scopes);
+
+            Dictionary<string, object> scopeData = new(rawScope);
+
+            // Validate variant data first.
```
jander-msft, line 66: The "better" way to do this is to replace the usage of `DateTime.Now` and the other "Now" variants with `TimeProvider`:
- .NET 8+: https://learn.microsoft.com/en-us/dotnet/api/system.timeprovider?view=net-8.0
- Custom shim for lower TFMs: https://github.com/dotnet/dotnet-monitor/blob/main/src/Tools/dotnet-monitor/TimeProvider.cs

You'd add a TimeProvider parameter to the `ParameterCapturingLogger` ctor. In product code, you'd pass `TimeProvider.System`. In test code, you'd pass your test-derived version that you can manually control the time.

I won't push you to do it here, since the time component is fairly simple, but this is mostly just a heads up. If the code evolves to be more complex than just snapshotting the timestamp, then I'd push for using `TimeProvider`.

jander-msft (dismissed):

wiktork (approved):

jander-msft (approved):

PR #5397, opened 2023-09-22
Update GET livemetrics route to use metrics configuration

### LINKED ISSUE #5393: Proposal: Change GET /livemetrics to use metrics configuration by default
## Problem

Today, the `GET /livemetrics` route will serve up metrics from the default providers at the [global counter interval](https://github.com/dotnet/dotnet-monitor/blob/main/documentation/configuration/metrics-configuration.md#global-counter-interval) configuration value. None of the [metrics configuration](https://github.com/dotnet/dotnet-monitor/blob/main/documentation/configuration/metrics-configuration.md#custom-metrics) is respected by this route. There is an "escape" hatch by using the `POST /livemetrics` route with the metrics configuration in the request body.

This is in contrast to the `GET /metrics` route, which serves up metrics from the default providers AND those providers can be overridden by [metrics configuration](https://github.com/dotnet/dotnet-monitor/blob/main/documentation/configuration/metrics-configuration.md#custom-metrics).

It is understood that the `/metrics` route and `/livemetrics` route serve slightly different purposes: the former is primary a Prometheus scenario enablement, whereas the latter is for collecting metrics in real-time. However, there are two issues at hand:
1. A customer is likely to initially look at `/metrics`, configure it as needed and be happy they got it to whatever desired state they want. If they ever have the thought of collecting metrics in real-time, they then look at `/livemetrics`, but it doesn't collect what they've configured! There is a stark behavior difference when first trying out `/livemetrics` and finding that the configured metrics are not appearing in its output. It's not clear to the customer why this is happening; I'm not certain every customer is going to realize that it's only collecting the default providers, then think they need to use `POST /livemetrics`, and respecify the metrics configuration that they already configured.
1. There's currently no way to express "I want `GET /livemetrics` to collect the providers that I've configured". If they want that, they need to replicate same configuration as the request body for the `POST /livemetrics` route. This will be an issue if the operator entity (the one who invokes the `/livemetrics` route) is different that the configurating entity (the one who specified that metrics configuration). Imagine an external tool wanting to collect real-time metrics, a tool that did not set up the initial metrics configuration; it would have no idea what the metrics configuration is, if it is different than the default providers, etc.

## Proposal

- Change `GET /livemetrics` to use the configured metrics providers instead of only the default metrics providers. This makes the usage of the route fall into the expected behavior of switching between `/metrics` and `/livemetrics` and allows customers to "fall into the pit of success". **This is a breaking change.**

Customers who want the prior behavior can use `POST /livemetrics` with `{"includeDefaultProviders": "true"}` as the request body.

### Proposal Variant

The above proposal is implemented. An additional configuration element is added to the `Metrics` configuration block that opts `/livemetrics` back into the existing behavior.

## Alternative 1

Update the `POST /livemetrics` route to allow an indication that the metrics configuration should be respected without having to fully respecify it.

Pro: This avoids breaking change.
Con: All existing cons remain expect for point 2 in the problem specification.

### Alternative 1 Variant 1

- Add an additional element to the request body for `POST /livemetrics` that indicates it should use the metrics providers from configuration. This element is mutually exclusive to all of the other elements, namely `includeDefaultProviders`, `providers`, and `meters`.

### Alternative 1 Variant 2

- Allow the body of `POST /livemetrics` to be optional. If not specified, this indicates that the metrics providers from configuration should be used. The current behavior today of invoking the `/POST /livemetrics` route without a request body is to respond with an HTTP 400 indicating that the configuration is required.

In either case, if no custom metrics configuration is available, it would still fall back to using the default providers.

jander-msft:
###### Summary

This changes the pipeline settings of the `GET /livemetrics` route to use the Metrics configuration instead of using the default event counters providers. Reasons for change:
- Both `GET /metrics` and `GET /livemetrics` now collect the same providers/counters/meters/instruments/etc. Customers "fall into the pit of success" more easily and don't need to wonder why `/livemetrics` collected a different data specification than `/metrics`.
- Customers are not required to respecify their metrics configuration via `POST /livemetrics` for something they already configured and naturally applies to all metrics scenarios.
- External tooling can now collect live metrics with the same specification that was already configured without having to dig through configuration sources, if has access to those sources. Customers do not need to tell those external tools how to replicate that configuration.

**This is a breaking change** If a customer wants to revert to the prior behavior, they can invoke `POST /livemetrics` with a request body of `{"includeDefaultProviders": "true"}`

###### Release Notes Entry

Update `GET /livemetrics` route to use configuration from Metrics section instead of only using the default event counters providers. Update `CollectLiveMetrics` action to use configuration from Metrics section by default.

TODO

- [x] Update `CollectLiveMetrics` action to have same behavior
- [x] Update documentation with behavioral differences

closes #5393

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.Metrics.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.Metrics.cs
@@ -45,8 +45,8 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
 
             MetricsPipelineSettings settings = MetricsSettingsFactory.CreateSettings(
                 _counterOptions.CurrentValue,
-                includeDefaults: true,
-                durationSeconds: durationSeconds);
+                durationSeconds,
+                _metricsOptions.CurrentValue);
 
             return InvokeForProcess(
                 processInfo => Result(
--- a/documentation/api/definitions.md
+++ b/documentation/api/definitions.md
@@ -126,7 +126,7 @@ Describes custom metrics.
 |---|---|---|
 | `includeDefaultProviders` | bool | Determines if the default counter providers should be used (such as System.Runtime). |
 | `providers` | [EventMetricsProvider](#eventmetricsprovider)[] | Array of counter providers for metrics to collect. |
-| `meters` | [EventMetricsMeter](#eventmetricsmeter)[] | Array of meters for metrics to collect. |
+| `meters` | [EventMetricsMeter](#eventmetricsmeter)[] | (7.1+) Array of meters for metrics to collect. |
 
 ## EventMetricsMeter
 
--- a/documentation/api/livemetrics-get.md
+++ b/documentation/api/livemetrics-get.md
@@ -3,7 +3,9 @@
 
 # Livemetrics - Get
 
-Captures metrics for a chosen process.
+Captures metrics for a chosen process for a duration of time.
+
+> **Note**: Starting in 8.0, the [metrics configuration](../configuration/metrics-configuration.md#metrics-configuration) is used as the collection specification. Prior to 8.0, the collection specification only included the [default providers](../configuration/metrics-configuration.md#default-providers) and could not be changed.
 
 > **Note**: For Prometheus style metrics, use the [metrics](./metrics.md) endpoint.
 
--- a/documentation/configuration/collection-rule-configuration.md
+++ b/documentation/configuration/collection-rule-configuration.md
@@ -762,12 +762,15 @@ Usage that collects a CPU trace for 30 seconds and egresses it to a provider nam
 
 An action that collects live metrics for the process that the collection rule is targeting.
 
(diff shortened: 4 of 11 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.Metrics.cs
```diff
@@ -45,8 +45,8 @@ partial class DiagController
 
             MetricsPipelineSettings settings = MetricsSettingsFactory.CreateSettings(
                 _counterOptions.CurrentValue,
```
wiktork, line 47: So previously we took the counter options and the defaults correct, but not the sdm defualts correct?

wiktork (approved):

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.Metrics.cs
jander-msft: Yes, expect there are no defaults for SDM; another way to put it, our SDM defaults is to not use SDM.

PR #5398, opened 2023-09-22
[Do Not Merge] Dotnet Counters + Dotnet Monitor Unification

kkeirstead:
###### Summary

This PR was co-developed with @wiktork.

Dotnet-Counters and the metrics functionality within Dotnet-Monitor currently share a lot of common logic, and keeping the code separate has resulted in a lot of duplicated work as new features are added to both tools.

This PR encompasses the unification work that takes place in dotnet-monitor. If you notice any behavior that seems to be inconsistent with what you'd expect, just let us know and we'll look into it - the intent is for this to _not_ be a breaking change for either tool. I plan to do a final round of manual testing prior to merging in these changes, once all other feedback has been resolved.

The change in the diagnostics repo can be found [here](https://github.com/dotnet/diagnostics/pull/4254).


wiktork:
@kkeirstead I've folded this PR into https://github.com/dotnet/dotnet-monitor/pull/5533

PR #5429, opened 2023-09-27
[Exceptions] Fix Bug For Missing Callstacks With Eclipsing Exceptions

kkeirstead:
###### Summary

For context: https://github.com/dotnet/runtime/issues/91125

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
@@ -103,31 +103,20 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Pipeline.Steps
             while (index < threadStackFrames.Length)
             {
                 StackFrame threadStackFrame = threadStackFrames[index];
-                if (throwingFrame.GetMethod() == threadStackFrame.GetMethod() &&
-                    throwingFrame.GetILOffset() == threadStackFrame.GetILOffset())
+
+                // Workaround for https://github.com/dotnet/runtime/issues/91125
+                // The threadStackFrame's ILOffset for eclipsing exceptions
+                // thrown from within the catch block does not match the
+                // ILOffset in the throwingFrame. Unconditionally swap out the
+                // threadStackFrame for the matching throwingFrame to ensure
+                // the correct ILOffset is used.
+                if (throwingFrame.GetMethod() == threadStackFrame.GetMethod())
                 {
+                    threadStackFrames[index] = throwingFrame;
                     break;
                 }
 
                 index++;
-            }
-
-            if (index >= threadStackTrace.FrameCount)
-            {
-                // Workaround for https://github.com/dotnet/runtime/issues/91125
-                // Fall back to only checking GetMethod() and ignoring the ILOffset()
-                index = 0;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
```diff
@@ -112,6 +112,24 @@ private static ReadOnlySpan<StackFrame> ComputeEffectiveCallStack(Exception exce
...
+            {
+                // Workaround for https://github.com/dotnet/runtime/issues/91125
+                // Fall back to only checking GetMethod() and ignoring the ILOffset()
+                index = 0;
+                while (index < threadStackFrames.Length)
+                {
+                    StackFrame threadStackFrame = threadStackFrames[index];
+                    if (throwingFrame.GetMethod() == threadStackFrame.GetMethod())
+                    {
+                        threadStackFrames[index] = throwingFrame;
```
kkeirstead, line 125: @jander-msft I wasn't sure if this approach would be problematic with swapping out the incorrect frame - it appeared to work, but I wanted to call your attention to it since you have more context here.

Relatedly, I'm currently choosing to use this as a fallback - I know we had discussed axing the ILOffset check entirely, so if you'd rather go that route I can update this.

jander-msft: I think the case of an eclipsing exception, the throwing frame is wrong, but the corresponding frame from the thread's call stack is correct. I'll take a look to verify that, and if that's correct, then all we have to do is just not check the IL offset anymore and assume the first encountered frame with the matching method is correct; this should be correct since this is observing first chance exceptions, so the frame with the same method must be the correct frame.

kkeirstead: Let me know if you end up taking a look at this - my interpretation of this was the opposite, based on what I saw debugging through it:

**For the exception thrown in the try:**
Offsets are 106/406 for both the throwing frame and the last thread stack frame

**For the exception thrown in the catch:**
Offsets are 149/749 for the throwing frame, and 106/406 for the last thread stack frame.

I would assume that the last thread stack frame's values shouldn't be the same in the try and the catch, in which case the 149/749 from the throwing frame would replace the 106/406 in the thread stack frame.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/ExceptionEventsPipelineStep.cs
```diff
@@ -112,6 +112,24 @@ private static ReadOnlySpan<StackFrame> ComputeEffectiveCallStack(Exception exce
                 index++;
             }
 
+            if (index >= threadStackTrace.FrameCount)
```
jander-msft, line 115: Instead of iterating the stack a second item, update the original code to remove the IL offset check so it finds the first matching frame solely based on the method, and then unconditionally swap out the frame.

jander-msft:
Please add a second eclipsing test where the catch block calls another method that throws an exception. This appears to cause the thread stack to be updated as originally expected, but let's add this test to make sure we cover the scenario.

jander-msft (approved):

PR #5471, opened 2023-10-03
Use central package management

jander-msft:
###### Summary

Use central package management. This allows for:
- uniformity of the restored package versions
- package consumers to not have to specify which version of the package to be restored
- pinning of transitive packages without forcing package references

Other noteworthy changes:
- Some packages were implicitly referenced, so they've been removed as explicit package references.
- Projects that reference packages that change their versions based on the current TFM must use `TargetFrameworks` property, even if they do not have multiple TFM for which they are building. An example is Microsoft.Diagnostics.Monitoring.ConfigurationSchema.UnitTests.csproj.
- The Microsoft.Diagnostics.Monitoring.TestCommon/Microsoft.Diagnostics.Monitoring.TestCommon.csproj requires an override of the xunit.assert package since itself is not a unit test project. I'll try to investigate further if this can be better augmented after this PR is merged.
- The release tool project has been excluded from CPM to avoid making this PR larger than it already is.
- The dependabot versions have not been moved to CPM since they are means for acquiring version updates rather than enforcing uniform package version usage.

### DIFF
--- a/eng/dependabot/net7.0/Packages.props
+++ b/eng/dependabot/net7.0/Packages.props
@@ -5,6 +5,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="$(MicrosoftExtensionsConfigurationAbstractions70Version)" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsLogging70Version)" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsLoggingAbstractions70Version)" />
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="$(MicrosoftExtensionsLoggingConsole70Version)" />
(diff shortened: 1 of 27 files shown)


eng/dependabot/net7.0/Packages.props
```diff
@@ -5,6 +5,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="$(MicrosoftExtensionsConfigurationAbstractions70Version)" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsLogging70Version)" />
```
wiktork, line 7: So version here will override the directory.packages.props correct?

wiktork: Actually shouldn't this be VersionOverride?

wiktork: Actually nvm, I see dependabot is still on regular versioning.

wiktork (approved):

PR #5486, opened 2023-10-05
Enable AzureAD tests in helix

schmittjoseph:
###### Summary

Enable AzureAD integration tests in internal helix builds. The `DOTNET_MONITOR_AZURE_AD_TESTS_*` environment variables are only populated in internal builds, so public builds simply blank out these environment variables, skipping the tests.

- [Public build](https://dev.azure.com/dnceng-public/public/_build/results?buildId=429115&view=results)
- [Internal build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2284592&view=results)


jander-msft (approved):
I assume these are blank in public builds by the fact that these environment variables are missing from the build agent due to the lack of variable group, correct?

PR #5487, opened 2023-10-05
Cache Node.js installations for helix tests

schmittjoseph:
###### Summary

Node.js needs to be installed on the helix test containers due to our test dependency on Azurite. However, the current mechanism of downloading from the official Node.js mirror is resulting in ~60% pipeline success rate due to transient errors.

This PR fixes this by:
- Downloading the copy of Node.js needed on the pipeline agent machine and including it as part of the payload instead of having helix download it on our behalf.
- Introduce a blob storage caching layer to avoid pulling from the Node.js mirror when possible. This caching layer will automatically roll forward on new Node.js versions (inside the same pinned major version), so we no longer need to keep our pinned version up-to-date.

The caching layer works by hashing together various attributes needed to uniquely identify a copy of Node.js (the OS, architecture, fully qualified version, etc) and then checking blob storage to see if a corresponding copy of Node.js has been cached.
- On cache miss: The `InstallNodejs.ps1` script is ran, downloading the new version of Node.js. At the end of the pipeline run, the cache is automatically updated with the results from this operation so future pipeline runs will use the cache.
- On cache hit: the cached Node.js installation downloaded. `InstallNodejs.ps1` still runs, but no-ops since there's nothing to do.
- Cache entries unused for more than 7 days will be automatically removed.

### DIFF
--- a/eng/helix/payload/installazurite.sh
+++ b/eng/helix/payload/installazurite.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 
-npm install -g azurite || exit 1
+helix_npm install -g azurite || exit 1
 export TEST_AZURITE_MUST_INITIALIZE=1


eng/helix/GetNodejsVersion.ps1
```diff
@@ -0,0 +1,31 @@
+[CmdletBinding(PositionalBinding=$false)]
+Param(
+    [int] $MajorVersion,
```
jander-msft, line 3:
```suggestion
        [Parameter(Mandatory=$true)][int] $MajorVersion,
```
Let's validate that the required parameters are actually specified in each of these scripts.

eng/helix/InstallNodejs.ps1
```diff
@@ -0,0 +1,52 @@
...
+    [ValidateSet("x86","x64","arm64")][string]$Architecture,
+    [string] $Version,
+    [string] $Mirror = "https://nodejs.org/dist",
+    [string] $DestinationFolder
+)
+
+Set-StrictMode -Version Latest
+$ErrorActionPreference = "Stop"
+
+if ($IsMacOS) {
```
jander-msft, line 12: I didn't know about these variables. That seems useful without having to invoke RuntimeInformation.IsOSPlatform

eng/pipelines/jobs/test-binaries.yml
```diff
@@ -66,7 +71,32 @@ jobs:
         artifactName: Build_Binaries_${{ parameters.configuration }}_${{ parameters.osGroup }}_${{ parameters.architecture }}
         targetPath: '$(Build.SourcesDirectory)/artifacts'
 
-    - ${{ if ne(parameters.useHelix, 'true')}}:
+    - ${{ if eq(parameters.useHelix, 'true')}}:
```
jander-msft, line 74: Should all of this be skipped for musl-libc since official nodejs only runs on glibc?

PR #5532, opened 2023-10-13
Change storage account staging root to build version

jander-msft:
###### Summary

This changes the staging directory on the staging account to use the unstable build version instead of the build number. The motivation for this change is to make it easier to look up the assets purely based on the build version instead of also requiring the build number e.g. dotnet-docker updates can rely on build version only if the updater needs to put checksums from the staging account.

Example difference:

```diff
-https://<account>.blob.core.windows.net/dotnet-monitor/20231012.2/BlobAssets/dotnet-monitor-base-8.0.0-rtm.23512.2-linux-x64.tar.gz
+https://<account>.blob.core.windows.net/dotnet-monitor/8.0.0-rtm.23512.2/BlobAssets/dotnet-monitor-base-8.0.0-rtm.23512.2-linux-x64.tar.gz
```

In the above example, the number after the `dotnet-monitor` container changed from the build number (20231012.2) to the build version (8.0.0-rtm.23512.2).

Validation build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2290054&view=results
Validation release: https://dev.azure.com/dnceng/internal/_build/results?buildId=2290231&view=results


eng/release/Scripts/PublishToBlobAccounts.ps1
```diff
@@ -26,7 +26,7 @@ function Generate-Source-Uri{
         [Parameter(Mandatory=$true)][string]$AssetType
     )
 
-    return "https://$sourceAccountName.blob.core.windows.net/$sourceContainerName/$BuildNumber/${AssetType}Assets/*"
```
wiktork, line 29: Will the docker side also need to be updated for this?

wiktork (approved):

eng/release/Scripts/PublishToBlobAccounts.ps1
jander-msft: Yes, dotnet-docker will need to be updated. It currently doesn't know how to get .NET Monitor from the staging account, so these changes don't really impact that ability, but I will be fixing the updater to understand the layout.

PR #5540, opened 2023-10-16
Prevent managed messaging initialization if profiler is not loaded

jander-msft:
###### Summary

When parameter capturing is not enabled and no other feature is enabled that requires the profiler, the startup hook will throw a DllNotFoundException exception, which appears in the `/exceptions` history backlog. This happens because the startup hook is attempting to initialize the managed messaging dispatcher, which requires the profiler to have been loaded.

Overview of changes:
- Add an additional check of whether the profiler is loaded before attempting to initialize the dispatcher.
- Add tests to verify the availability of the dispatcher based on the existence of the profiler.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/StartupHookTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/StartupHookTests.cs
@@ -1,0 +1,84 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Diagnostics.Monitoring.TestCommon;
+using Microsoft.Diagnostics.Monitoring.TestCommon.Runners;
+using Microsoft.Diagnostics.NETCore.Client;
+using Microsoft.Diagnostics.Tools.Monitor.HostingStartup;
+using System;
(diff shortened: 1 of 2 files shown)


wiktork (dismissed):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/StartupHookTests.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        private ITestOutputHelper _outputHelper;
+
+        public StartupHookTests(ITestOutputHelper outputHelper)
+        {
+            _outputHelper = outputHelper;
+        }
+
+        // It appears that the profiler isn't loading on musl libc distros for this tests.
+        [ConditionalTheory(typeof(TestConditions), nameof(TestConditions.IsNotAlpine))]
```
wiktork, line 30: consider a github issue to track bringing this back

wiktork (approved):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/StartupHookTests.cs
jander-msft: /todo Enable StartupHook_WithProfiler_HasManagedMessaging for Alpine

PR #5550, opened 2023-10-17
Ensure diagnostics are only bootstrapped once in target application

jander-msft:
###### Summary

This change is a small refactor to make sure that diagnostics are only initialized once for the lifetime of the target application. This is to hedge against the startup hook being accidentally configured to run multiple times (e.g. process connects to .NET Monitor diagnostic server, connection in pruned, and process reconnects again).

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/StartupHook.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/StartupHook.cs
@@ -2,54 +2,35 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using Microsoft.Diagnostics.Monitoring.StartupHook;
-using Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions;
-using Microsoft.Diagnostics.Monitoring.StartupHook.MonitorMessageDispatcher;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Diagnostics.Monitoring.StartupHook/StartupHook.cs
```diff
@@ -2,68 +2,47 @@
...
 internal sealed class StartupHook
 {
-    private static CurrentAppDomainExceptionProcessor? s_exceptionProcessor;
-    private static AspNetHostingStartupLoader? s_hostingStartupLoader;
+    private static DiagnosticsBootstrapper? s_bootstrapper;
+    private static object s_lock = new object();
 
     public static void Initialize()
     {
-        try
```
wiktork, line 21: I take it there are no guards for this in the runtime itself?

wiktork (approved):

src/Microsoft.Diagnostics.Monitoring.StartupHook/StartupHook.cs
jander-msft: Correct. The runtime will only load the assembly once but it will call the `Initialize` method each time.

PR #5589, opened 2023-10-24
Enforce editorconfig at build time

jander-msft:
###### Summary

- Enforce editorconfig code style at build time
- Fixup existing violations of the editorconfig
- Only have C# linting run on 7.x and lower

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -8,6 +8,8 @@
     <LangVersion>Latest</LangVersion>
     <WarningLevel>4</WarningLevel>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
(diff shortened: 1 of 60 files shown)


Directory.Build.props
```diff
@@ -8,6 +8,8 @@
     <LangVersion>Latest</LangVersion>
     <WarningLevel>4</WarningLevel>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
wiktork, line 12: Do we need the documentation file?

wiktork (approved):

Directory.Build.props
jander-msft: _Need_ is a strong word here. The editor config enforcement will error if documentation generation is disabled; that could be mitigated by disabling all of the editor config features that check the doc comments. I opt-ed to go the other direction to enable it across the board and fix all of the documentation problems.

PR #5594, opened 2023-10-26
Use final stable version

jander-msft:
###### Summary

Change `release/8.x` to produce stable versions.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -9,8 +9,7 @@
     <RepositoryUrl>https://github.com/dotnet/dotnet-monitor</RepositoryUrl>
     <VersionPrefix>8.0.0</VersionPrefix>
     <PreReleaseVersionLabel>rtm</PreReleaseVersionLabel>
-    <!-- Disable final version kind until merged into release branch. -->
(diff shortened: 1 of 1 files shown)


eng/Versions.props
```diff
@@ -9,8 +9,7 @@
     <RepositoryUrl>https://github.com/dotnet/dotnet-monitor</RepositoryUrl>
     <VersionPrefix>8.0.0</VersionPrefix>
     <PreReleaseVersionLabel>rtm</PreReleaseVersionLabel>
-    <!-- Disable final version kind until merged into release branch. -->
-    <!--<DotnetFinalVersionKind>release</DotnetFinalVersionKind>-->
+    <DotnetFinalVersionKind>release</DotnetFinalVersionKind>
```
wiktork, line 12: Maybe let's keep the comment here? We'll want to revert this for 9 previews

wiktork (approved):

eng/Versions.props
jander-msft: This is targeting `release/8.x`; the comment remains in the `main` branch.

Issue #5140, opened 2023-08-14
Add support for Metric/Instrument tags and scopes


wiktork:
For dotnet 8, meters now support additional information as part of the meter itself (rather than each individual value). See https://github.com/dotnet/runtime/pull/90417/files#diff-8797eae6d3e644ed047f952e10d4240fe18175f754586666823b555f46968672L1435-R1486. 

This includes:
- Description
- Scope hash
- Tags

The data can be gathered as part of the BeginInstrumentReporting TraceEvent.  See https://github.com/dotnet/diagnostics/blob/9eee79a44914fdb703664ee6461298161339cb4f/src/Microsoft.Diagnostics.Monitoring.EventPipe/Counters/TraceEventExtensions.cs#L82.

PR #5624, opened 2023-11-01
[Do Not Merge] Add Meter Tags, Instrument Tags, and Scopes to System Diagnostics Metrics

kkeirstead:
###### Summary

This PR goes along with https://github.com/dotnet/diagnostics/pull/4324 in the diagnostics repo.

Closes https://github.com/dotnet/dotnet-monitor/issues/5140

###### Release Notes Entry

Add Support for Meter Tags and Instrument Tags for System Diagnostics Metrics.


src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/LiveMetricsTestConstants.cs
```diff
@@ -11,7 +11,12 @@ public static class LiveMetricsTestConstants
         public const string HistogramName2 = "test-histogram-2";
         public const string ProviderName1 = "P1";
         public const string ProviderName2 = "P2";
+        public const string ProviderName3 = "P3";
         public const string MetadataKey = "key1";
         public const string MetadataValue = "value1";
+        public const string MeterMetadataKey = "mkey1";
```
schmittjoseph, line 17: Can we switch these to names that won't trigger the spell checker? e.g. `mKey1`

src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsStore.cs
```diff
@@ -165,10 +165,10 @@ public async Task SnapshotMetrics(Stream outputStream, CancellationToken token)
 
         private static string GetMetricLabels(ICounterPayload metric, double? quantile)
         {
-            string metadata = metric.Metadata;
+            string allMetadata = metric.CombineTags();
```
wiktork, line 168: So for prometheus export, we will combine meter/instrument/instance tags all into one?

kkeirstead: That's the current behavior, but if we aren't sure we want to put all the tags together we could display them differently (not sure what that would look like for PEF), or omit them temporarily if we're concerned about future breaking changes.

wiktork (dismissed):

PR #5646, opened 2023-11-07
Exclude old 7.* version from updates and scanning

jander-msft:
###### Summary

The `7.2` minor version line is out-of-support on November 8th, so remove it from dependency updates. Also exclude older 7 versions from component scans.


eng/pipelines/dotnet-monitor-cg.yml
```diff
@@ -4,6 +4,9 @@ schedules:
   branches:
     include:
     - shipped/*
+    exlucde:
```
schmittjoseph, line 7:
```suggestion
    exclude:
```

schmittjoseph (approved):

PR #5667, opened 2023-11-13
Increase stopping event test timeouts

schmittjoseph:
###### Summary

The currently stopping event timeout (10 seconds) isn't always enough in our CI pipelines. Switch to using the general timeout (3 minutes) for all stopping event tests that expect to be stopped early.

For tests that expect to not be stopped by an event (and thus take the full timeout), use a timeout of 15 seconds to avoid slowing down successful test runs.


jander-msft (approved):

jander-msft:
I believe stopping event support was added in 7.1, so please backport to `release/7.x`

PR #5680, opened 2023-11-14
Pin Az.Storage module version

jander-msft:
###### Summary

The `Az.Storage` PowerShell module released a new version today that has breaking changes regarding the generation of SAS tokens. This change initially blocked the ability to copy between blob storage accounts that are secured with SAS tokens. To mitigate this, pin the module version until changes can be made to adapt to the breaking change.

### DIFF
--- a/eng/pipelines/dotnet-monitor-release.yml
+++ b/eng/pipelines/dotnet-monitor-release.yml
@@ -163,8 +163,8 @@ stages:
           - powershell: Install-PackageProvider -Name NuGet -Force -Scope CurrentUser
             displayName: Install NuGet PowerShell Package Provider
 
-          - powershell: Install-Module Az.Storage -Force -Scope CurrentUser -AllowClobber -Verbose
-            displayName: Install Az.Storage Module
+          - powershell: Install-Module Az.Storage -Force -Scope CurrentUser -AllowClobber -Verbose -RequiredVersion 5.10.1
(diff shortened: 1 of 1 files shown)


eng/pipelines/dotnet-monitor-release.yml
```diff
@@ -163,8 +163,8 @@ stages:
           - powershell: Install-PackageProvider -Name NuGet -Force -Scope CurrentUser
             displayName: Install NuGet PowerShell Package Provider
 
-          - powershell: Install-Module Az.Storage -Force -Scope CurrentUser -AllowClobber -Verbose
-            displayName: Install Az.Storage Module
+          - powershell: Install-Module Az.Storage -Force -Scope CurrentUser -AllowClobber -Verbose -RequiredVersion 5.10.1
```
schmittjoseph, line 166: Is the breaking change a on major / minor version bump? If so does `Install-Module` support wildcards here, e.g. `-RequiredVeresion 5.10.*`?

Given this is an interim solution I'm fine with it as-if if not.

jander-msft: It doesn't seem to accept wildcards (`Cannot convert value '5.10.*' to type 'System.Version'`) and trying `5.10` will install `5.10.0`, so I think we're stuck with the exact version.

schmittjoseph (approved):

PR #5707, opened 2023-11-17
Add compliance pipeline and scripts

jander-msft:
###### Summary

Add a pipeline for running ApiScan on the shipping assets. Shipping assets are downloaded, unpacked, and filtered to what is applicable to ApiScan. All results are uploaded as an artifact on the pipeline. This pipeline requires manual invocation at this time.

Example run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2318464&view=results

### DIFF
--- a/eng/pipelines/scripts/Copy-ApiScanEligible.ps1
+++ b/eng/pipelines/scripts/Copy-ApiScanEligible.ps1
@@ -1,0 +1,97 @@
+param(
+  [Parameter(Mandatory=$true)][string] $SourcePath,
+  [Parameter(Mandatory=$true)][string] $TargetPath,
+  [Parameter(ValueFromRemainingArguments=$true)][String[]]$properties
+)
+
+$ErrorActionPreference = 'Stop'
+Set-StrictMode -Version 2.0
+
+if ($null -ne $properties) {
+  Write-Host "Unexpected extra parameters: $properties."
+  exit 1
+}
+
+$copyCount = 0
+$skipCount = 0
+
+# Copy files from source to target directory while preserving the relative directory structure
+function Copy-File() {
+  param(
+    [System.IO.FileInfo] $FileInfo
+  )
+  Write-Host "Copying $($FileInfo.FullName)"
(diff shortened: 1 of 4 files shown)


eng/pipelines/scripts/Extract-Assets.ps1
```diff
@@ -0,0 +1,66 @@
...
+  [System.IO.Compression.ZipFile]::ExtractToDirectory($FileInfo.FullName, $extractionPath)
+}
+
+function Extract-TarGz() {
+  param(
+    [System.IO.FileInfo] $FileInfo
+  )
+  $extractionPath = Create-PackageDir -FileName $FileInfo.Name
+  Write-Host "Extracting $($FileInfo.Name)"
+  tar -xf $FileInfo.FullName -C $extractionPath
```
schmittjoseph, line 41:
```suggestion
  tar -xzf $FileInfo.FullName -C $extractionPath
```

We should be specifying the compression type of the archive

jander-msft: I've pushed this change and am running another validation build for it.

eng/pipelines/scripts/Copy-ApiScanEligible.ps1
```diff
@@ -0,0 +1,97 @@
...
+  Copy-Item -Path $FileInfo.FullName -Destination $fileTargetPath
+  $script:copyCount++
+}
+
+# Check if the file is a ARM64 PE file
+function Test-PeArm64() {
+  param(
+    [System.IO.FileInfo] $FileInfo
+  )
+  $stream = [System.IO.File]::OpenRead($FileInfo.FullName)
```
schmittjoseph, line 37: Didn't think I'd ever see reading a file's COFF header in powershell .... I don't have a better suggestion since Windows doesn't have *nix's `file` utility and a dependency on `dumpbin` seems excessive since it might not always be present. LGMT.

jander-msft: Yeah, I tried looking for anything better than this, but there isn't any easy utility for it without pulling in some custom PowerShell modules.

schmittjoseph (approved):

PR #5732, opened 2023-11-27
Filter Monitor exceptions by default

schmittjoseph:
###### Summary

By default, filter out first-chance exceptions caused by in-proc dotnet-monitor features from being reported via the `/exceptions` route. This is done as part of the exception processing pipeline and can be disabled in a new top-level configuration node (`DotnetMonitorDebug`) which holds configuration settings to help debug dotnet-monitor features. These flags are meant for use by folks working on dotnet-monitor or helping diagnose issues. Options under this node aren't officially supported and are subject to change.

Example configuration to include dotnet-monitor exceptions in reporting:
```json
{
    "DotnetMonitorDebug": {
        "Exceptions": {
            "IncludeMonitorExceptions": true
        }
    }
}
```


documentation/schema.json
```diff
@@ -1649,6 +1660,36 @@
...
+              "type": "null"
+            },
+            {
+              "$ref": "#/definitions/ExceptionsDebugOptions"
+            }
+          ]
+        }
+      }
+    },
+    "ExceptionsDebugOptions": {
```
schmittjoseph, line 1680: Note that any properties under `DotnetMonitorDebug` are **not** documented as they are not officially supported and are subject to change.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/FilterMonitorExceptionPipelineStep.cs
```diff
@@ -0,0 +1,35 @@
...
+        public FilterMonitorExceptionPipelineStep(ExceptionPipelineDelegate next)
+        {
+            ArgumentNullException.ThrowIfNull(next);
+
+            _next = next;
+        }
+
+        public void Invoke(Exception exception, ExceptionPipelineExceptionContext context)
+        {
+            ArgumentNullException.ThrowIfNull(exception);
```
schmittjoseph, line 25: NOTE: Exceptions thrown during the exception pipeline are currently **not** filtered out. If we want to filter these that should be done as a separate PR.

documentation/schema.json
```diff
@@ -149,6 +149,17 @@
           "$ref": "#/definitions/TemplateOptions"
         }
       ]
+    },
```
schmittjoseph, line 152: Given how verbose adding a new configuration option is to dotnet-monitor, I can split this PR into 2 if folks would prefer that:
- One to add the new exception pipeline step (that doesn't get added)
- One to add this new configuration option

src/Microsoft.Diagnostics.Monitoring.StartupHook/DiagnosticsBootstrapper.cs
```diff
@@ -15,25 +15,28 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook
...
-        private readonly CurrentAppDomainExceptionProcessor _exceptionProcessor = new();
+        private readonly CurrentAppDomainExceptionProcessor _exceptionProcessor;
         private readonly AspNetHostingStartupLoader? _hostingStartupLoader;
 
         private long _disposedState;
 
         public DiagnosticsBootstrapper()
         {
+            _exceptionProcessor = new(ToolIdentifiers.IsEnvVarEnabled(InProcessFeaturesIdentifiers.EnvironmentVariables.Exceptions.IncludeMonitorExceptions));
+            _exceptionProcessor.Start();
```
wiktork, line 26: We won't be able to dynamically change the value of IncludeMonitorExceptions with this approach. There might be some ways to do that, but does not have to be a part of this PR.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/MonitorExecutionContextTracker.cs
```diff
@@ -0,0 +1,80 @@
...
+                _monitorScopeCount.Value++;
+            }
+            else
+            {
+                //
+                // Underflows should never happen since this method is private and only called to decrement
+                // if there's a corresponding MonitorScopeTracker being disposed which would have incremented.
+                // Gracefully handle regardless.
+                //
+                if (_monitorScopeCount.Value == uint.MinValue)
```
wiktork, line 68: would <= be a reasonable check here?

schmittjoseph: We're using a `uint` here so it'll either be 0 or underflow back up to the max value. I can change it to `== 0` if you think that'd be more clear.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests.csproj
```diff
@@ -19,6 +19,9 @@
     <None Update="ExpectedConfigurations\CollectionRuleDefaults.json">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
+    <None Update="ExpectedConfigurations\DotnetMonitorDebug.json">
```
wiktork, line 22: I'm not sure about this myself, but have you considered updating the test runners to enable the internal exceptions by default?

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/Steps/FilterMonitorExceptionPipelineStep.cs
```diff
@@ -0,0 +1,35 @@
...
+
+            if (MonitorExecutionContextTracker.IsInMonitorContext())
```
wiktork, line 27: Does the invocation of AppDomain.Exception events happen completely synchronously? Is there any possibility that an exception thrown within a bounded region of our code will get a callback _after_ the Dispose call?

schmittjoseph: Our exception handling is done synchronously so that scenario should never happen.

ref: 
> After the event has been raised and event handlers have been invoked, the common language runtime (CLR) begins to search for a handler for the exception.

https://learn.microsoft.com/en-us/dotnet/api/system.appdomain.firstchanceexception?view=net-8.0#remarks

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/MonitorExecutionContextTracker.cs
```diff
@@ -0,0 +1,80 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Diagnostics;
+using System.Threading;
+
+namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions
+{
+    internal static class MonitorExecutionContextTracker
```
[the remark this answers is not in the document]
schmittjoseph: Agreed that we'll have to keep this in mind whenever writing in-proc features. I don't have a better suggestion unfortunately as any other approach I've thought would either incur a performance penalty or miss a lot of cases.

PR #5768, opened 2023-12-18
Lower CPU threshold in collection rule integration tests

jander-msft:
###### Summary

The collection rule tests that use a CPU usage trigger are have sporadically not completion, which causes the test to wait for a long time and eventually timeout. Attempt to mitigate by lowering the CPU threshold for these tests.


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/CollectionRuleDescriptionTests.cs
```diff
@@ -195,10 +195,10 @@ public async Task CollectionRuleDescription_ActionLimitTest(DiagnosticPortConnec
                     runner.ConfigurationFromEnvironment.CreateCollectionRule(NonStartupRuleName)
                         .SetEventCounterTrigger(options =>
                         {
-                            // cpu usage greater that 5% for 2 seconds
+                            // cpu usage greater than 1% for 2 seconds
                             options.ProviderName = "System.Runtime";
                             options.CounterName = "cpu-usage";
-                            options.GreaterThan = 5;
+                            options.GreaterThan = 1;
```
wiktork, line 201: Maybe we could also lower the sliding window duration?

wiktork (dismissed):

wiktork (approved):

PR #5795, opened 2024-01-03
[Parameter Capturing] Refactor probe management API to accept more detailed boxing instructions

schmittjoseph:
###### Summary

Currently the probe management API's managed layer precomputes all boxing tokens and passes only those to the profiler. This has the restriction of not allowing us to send additional information to the profiler for special boxing instruction. e.g.:
- What parameters need an indirect load.
- What parameters are typespecs and need a new metadata token emitted by the profiler.

This PR refactors those APIs, extending the ABI to accept a c-style array of structs (type: `PARAMETER_BOXING_INSTRUCTIONS`). This struct contains a discriminator telling the profiler the type of boxing instruction followed by a fixed-size set of data (to be expanded on in the future).

As an added benefit we can replace the current mechanism for specifying primitive tokens (where we create a fake metadata table with an id of `0x7f000000`  and fill it with our own enum) with an easier-to-understand system that doesn't overload the metadata token format. 

> [!IMPORTANT]
> There are **no** functional differences in this PR, follow up PRs will build upon this to add support for more parameter types.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/BoxingInstructions.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/BoxingInstructions.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing
             //
             // A signature decoder will used to determine boxing tokens for parameter types that cannot be determined from standard
             // reflection alone. The boxing tokens generated from this decoder should only be used to fill in these gaps
-            // as it is not a comprehensive decoder and will produce UnsupportedParameterToken for any types not explicitly mentioned
+            // as it is not a comprehensive decoder and will produce an unsupport boxing instruction for any types not explicitly mentioned
             // in BoxingTokensSignatureProvider's summary.
             // 
             Lazy<ParameterBoxingInstructions[]?> ancillaryInstructions = new(() => GetAncillaryBoxingInstructionsFromMethodSignature(method));
@@ -144,23 +144,8 @@ namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing
                 MethodDefinitionHandle methodDefHandle = (MethodDefinitionHandle)MetadataTokens.Handle(method.MetadataToken);
                 MethodDefinition methodDef = mdReader.GetMethodDefinition(methodDefHandle);
 
-                MethodSignature<uint?> methodSignature = methodDef.DecodeSignature(new BoxingTokensSignatureProvider(), genericContext: null);
-
-                ParameterBoxingInstructions[] instructions = new ParameterBoxingInstructions[methodSignature.ParameterTypes.Length];
-                for (int i = 0; i < methodSignature.ParameterTypes.Length; i++)
-                {
-                    uint? mdToken = methodSignature.ParameterTypes[i];
-                    if (mdToken.HasValue)
-                    {
-                        instructions[i] = mdToken.Value;
-                    }
-                    else
-                    {
-                        instructions[i] = SpecialCaseBoxingTypes.Unknown;
-                    }
-                }
-
-                return instructions;
+                MethodSignature<ParameterBoxingInstructions> methodSignature = methodDef.DecodeSignature(new BoxingTokensSignatureProvider(), genericContext: null);
+                return [.. methodSignature.ParameterTypes];
             }
             catch (Exception)
             {
--- a/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/BoxingInstructions.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/BoxingInstructions.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing
             //
             // A signature decoder will used to determine boxing tokens for parameter types that cannot be determined from standard
             // reflection alone. The boxing tokens generated from this decoder should only be used to fill in these gaps
-            // as it is not a comprehensive decoder and will produce an unsupport boxing instruction for any types not explicitly mentioned
+            // as it is not a comprehensive decoder and will produce an unsupported boxing instruction for any types not explicitly mentioned
             // in BoxingTokensSignatureProvider's summary.
             // 
             Lazy<ParameterBoxingInstructions[]?> ancillaryInstructions = new(() => GetAncillaryBoxingInstructionsFromMethodSignature(method));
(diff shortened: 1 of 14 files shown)


src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/ProfilerAbi.cs
```diff
@@ -0,0 +1,57 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing.FunctionProbes
+{
+    public enum SpecialCaseBoxingTypes : uint
```
schmittjoseph, line 6: Should be internal

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/BoxingInstructions.cs
```diff
@@ -172,77 +144,90 @@ public static uint[] GetBoxingTokens(MethodInfo method)
                 MethodDefinitionHandle methodDefHandle = (MethodDefinitionHandle)MetadataTokens.Handle(method.MetadataToken);
                 MethodDefinition methodDef = mdReader.GetMethodDefinition(methodDefHandle);
 
-                MethodSignature<uint> methodSignature = methodDef.DecodeSignature(new BoxingTokensSignatureProvider(), genericContext: null);
+                MethodSignature<uint?> methodSignature = methodDef.DecodeSignature(new BoxingTokensSignatureProvider(), genericContext: null);
```
jander-msft, line 147: If you change `BoxingTokenSignatureProvider` to return `SpecialCaseBoxingTypes.Unknown` instead of returning `null` for unsupported types, then this code path can be reverted and no nullability checks are required.

jander-msft: Changing the provider to return `ParameterBoxingInstructions` rather than `uint` would further simplify this.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/ProfilerAbi.cs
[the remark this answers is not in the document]
jander-msft: Maybe a general comment on this file since there are multiple types that should be kept in sync.

jander-msft (dismissed):

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/BoxingInstructions.cs
```diff
@@ -58,20 +30,20 @@ public static bool[] AreParametersSupported(uint[] tokens)
...
+        public static ParameterBoxingInstructions[] GetBoxingInstructions(MethodInfo method)
         {
             List<Type> methodParameterTypes = method.GetParameters().Select(p => p.ParameterType).ToList();
-            List<uint> boxingTokens = new List<uint>(methodParameterTypes.Count + (method.HasImplicitThis() ? 1 : 0));
-
+            List<ParameterBoxingInstructions> instructions = new(methodParameterTypes.Count + (method.HasImplicitThis() ? 1 : 0));
             //
             // A signature decoder will used to determine boxing tokens for parameter types that cannot be determined from standard
             // reflection alone. The boxing tokens generated from this decoder should only be used to fill in these gaps
             // as it is not a comprehensive decoder and will produce UnsupportedParameterToken for any types not explicitly mentioned
```
kkeirstead, line 40: nit: It looks like `UnsupportedParameterToken` is no longer being defined/referenced in other places, so this should probably be updated

jander-msft (dismissed):

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/ProfilerAbi.cs
```diff
@@ -0,0 +1,63 @@
...
+    };
+
+    internal enum InstructionType : ushort
+    {
+        Unknown = 0,
+        SpecialCaseToken,
+        MetadataToken
+    }
+
+    internal struct ParameterBoxingInstructions
```
wiktork, line 39: At the very least, you should have LayoutKind.Sequential here. It might be useful to add a comment regarding the fact that Token is a union on the native side.

wiktork: You may need to look into Pack as well to make sure packsize is correct in both managed and native.

schmittjoseph: structs are LayoutKind.Sequential by default

jander-msft: It probably should be `LayoutKind.Explicit`: https://learn.microsoft.com/en-us/dotnet/standard/native-interop/customize-struct-marshalling#customize-structure-layout

schmittjoseph: As for controlling the alignment, structs use the platform's default alignment by default. This is the same as our profiler code so unless we're somehow loading a 32 bit profiler into a 64 bit .net app we shouldn't need to worry about hard coding alignment.

schmittjoseph: > It probably should be `LayoutKind.Explicit`: https://learn.microsoft.com/en-us/dotnet/standard/native-interop/customize-struct-marshalling#customize-structure-layout

I don't know if I agree with that assertion. sequential makes the most sense here since we have no need to override the field offsets since no unions are used.

jander-msft: From the docs: "DO only use LayoutKind.Explicit in marshalling when your native struct also has an explicit layout, such as a union."

The native struct `PARAMETER_BOXING_INSTRUCTIONS` has a union, thus it has an explicit layout, correct? Thus this rule says you should use `LayoutKind.Explicit`.

schmittjoseph: All of the fields in the union are the same size, so explicit layout isn't needed to account for that. I'll add a static assert to ensure we don't accidentally change the union's size in the future

schmittjoseph: Discussed offline, I'll add a comment to the token field and explicitly call out the struct layout

jander-msft (approved):

wiktork (approved):

PR #5800, opened 2024-01-08
[Parameter Capturing] Refactor reflection-only boxing logic into new method

schmittjoseph:
###### Summary

With the upcoming support for typespec related parameters, the current `GetBoxingInstructions` method becomes responsible for too much and will be difficult to read.

To address this, move reflection-only boxing logic into a new method (`GetBoxingInstructionsFromReflection`). Also reorganizes the boxing-related files into a subfolder to help group things together.

There are **no** functional changes in this PR.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Boxing/BoxingInstructions.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Boxing/BoxingInstructions.cs
@@ -89,7 +89,7 @@ namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing.Box
             {
                 ParameterInfo paramInfo = formalParameters[i];
 
-                ParameterBoxingInstructions paramBoxingInstructions = GetBoxingInstructionFromReflection(method, paramInfo.ParameterType, out bool canUseSignatureDecoder);
+                ParameterBoxingInstructions paramBoxingInstructions = GetBoxingInstructionsFromReflection(method, paramInfo.ParameterType, out bool canUseSignatureDecoder);
                 if (canUseSignatureDecoder &&
                     !IsParameterSupported(paramBoxingInstructions) &&
                     signatureDecoderInstructions.Value != null)
@@ -103,7 +103,7 @@ namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing.Box
             return instructions;
         }
 
-        private static ParameterBoxingInstructions GetBoxingInstructionFromReflection(MethodInfo method, Type paramType, out bool canUseSignatureDecoder)
+        private static ParameterBoxingInstructions GetBoxingInstructionsFromReflection(MethodInfo method, Type paramType, out bool canUseSignatureDecoder)
         {
             canUseSignatureDecoder = false;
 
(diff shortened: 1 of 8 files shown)


src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Boxing/BoxingInstructions.cs
```diff
@@ -60,74 +57,102 @@ public static ParameterBoxingInstructions[] GetBoxingInstructions(MethodInfo met
...
-                {
-                    instructions.Add(SpecialCaseBoxingTypes.Unknown);
-                }
-                else if (paramType.IsGenericParameter)
+                ParameterBoxingInstructions[]? instructions = GetAncillaryBoxingInstructionsFromMethodSignature(method);
+                if (instructions != null &&
+                    instructions.Length != formalParameters.Length)
                 {
-                    instructions.Add(SpecialCaseBoxingTypes.Unknown);
+                    Debug.Fail("Signature decoder results have an unexpected number of formal parameters");
```
kkeirstead, line 82: If `instructions` is null, do we also want this debug line? As it currently stands, it'll return null either way, but the message would be omitted

schmittjoseph: `GetAncillaryBoxingInstructionsFromMethodSignature` returning null is an expected code path so it doesn't indicate a problem (whereas returning a mismatch on the number of parameters should *never* happen but we still need to sanity check it).

There is one code path inside `GetAncillaryBoxingInstructionsFromMethodSignature` where returning null is unexpected, so I'll go ahead an add a debug message over there

kkeirstead (approved):
Feel free to wait for another review from someone who knows this space better if you want, but I didn't have anything else I wanted to call out.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Boxing/BoxingInstructions.cs
```diff
@@ -60,74 +57,102 @@ public static ParameterBoxingInstructions[] GetBoxingInstructions(MethodInfo met
...
                 }
+
+                instructions[index++] = thisBoxingInstructions;
             }
 
-            foreach (Type paramType in methodParameterTypes)
+            //
+            // A signature decoder will be used to determine boxing tokens for formal parameter types that cannot be determined from standard
+            // reflection alone. The boxing tokens generated from this decoder should only be used to fill in these gaps
+            // as it is not a comprehensive decoder and will produce an unsupported boxing instruction for any types not explicitly mentioned
```
jander-msft, line 73: Can we make the decoder comprehensive? As it is right now, there is a shared understanding between `GetBoxingInstructions` method and the decode of what is and is not supported; it seems like a careful balancing act that might bite later if one is updated and not the other.

schmittjoseph: This is something I've looked into previously. I was able to prototype a fairly comprehensive decoder that in theory should remove the need for reflection-based boxing logic for **formal** parameters. 

There's a couple important notes though with that approach:
- A reflection-based approach would still be needed for the implicit this parameter as it's not described in a method's signature.
- I haven't done enough vetting of this approach to be comfortable switching us over to it wholesale yet. I'd want to dedicate a little more time to investigating it more before doing so.

I agree that the shared understanding is brittle and can be an easy source of future errors. Short of switching wholesale over to the signature decoder approach, I can introduce some unit tests for the decoder and its consumer to better solidify the contract here.

jander-msft (approved):

PR #5812, opened 2024-01-10
[Parameter Capturing] Add support for generics, tuples, and nullable value types

schmittjoseph:
###### Summary

This PR was co-developed with @kkeirstead .

This PR adds support for generics, tuples, and nullable value types. All of these parameters are considered TypeSpecs. To support boxing them we need to identify their signature blobs and then get a metadata token that represents it. 

We can leverage our managed signature decoder to find the signature blobs for these parameters and then let the mutating profiler convert these information into a metadata token (sometimes the profiler will need to produce a new token if there already isn't one for the given TypeSpec).

###### Release Notes Entry
Support capturing the following parameter types: generics, tuples, and nullable value types


src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/ProfilerAbi.cs
```diff
@@ -50,6 +51,9 @@ internal struct ParameterBoxingInstructions
         //
         public uint Token;
 
+        public byte* SignatureBufferPointer;
+        public uint SignatureBufferLength;
```
schmittjoseph, line 55: I *could* remove this field and make it re-use our existing uint field above, but I'd prefer to keep these fields separate for now for readability. If folks have strong preferences for keeping this struct's memory size to an absolute minimum let me know and we can look into removing this field.

wiktork: I'd rather not store buffer sizes in unions, let's keep them separate.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/FunctionProbes/FunctionProbesScenario.cs
```diff
@@ -148,6 +148,39 @@ private static async Task Test_CaptureNativeIntegersAsync(FunctionProbesManager
...
+                testTuple
+            }, token);
+        }
+
+        private static async Task Test_CaptureGenericsAsync(FunctionProbesManager probeManager, PerFunctionProbeProxy probeProxy, CancellationToken token)
+        {
+            MethodInfo method = Type.GetType($"{nameof(SampleMethods)}.GenericTestMethodSignatures`2").GetMethod("GenericParameters");
+            Assert.NotNull(method);
+
+            List<List<Uri>> uris = [[new Uri("https://example.com")]];
```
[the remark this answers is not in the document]
schmittjoseph: Updated this test case

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/FunctionProbes/FunctionProbesScenario.cs
```diff
@@ -148,6 +148,40 @@ private static async Task Test_CaptureNativeIntegersAsync(FunctionProbesManager
...
+                testTuple
+            }, token);
+        }
+
+        private static async Task Test_CaptureGenericsAsync(FunctionProbesManager probeManager, PerFunctionProbeProxy probeProxy, CancellationToken token)
+        {
+            MethodInfo method = Type.GetType($"{nameof(SampleMethods)}.GenericTestMethodSignatures`2").GetMethod("GenericParameters");
+            Assert.NotNull(method);
+
+            List<List<int?>> ints = [[10]];
```
wiktork, line 168: Doesn't have to be part of this checkin, but it may be worthwhile to add a broader set of generics in the unit test.

wiktork (approved):

PR #5845, opened 2024-01-17
Change to version 9.0

wiktork:

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -77,12 +77,12 @@
   <PropertyGroup Label="Runtime Versions">
     <MicrosoftAspNetCoreApp31Version>$(MicrosoftNETCoreApp31Version)</MicrosoftAspNetCoreApp31Version>
     <MicrosoftNETCoreApp50Version>5.0.17</MicrosoftNETCoreApp50Version>
-    <MicrosoftNETCoreApp90Version>9.0.0-alpha.1.24061.6</MicrosoftNETCoreApp90Version>
+    <MicrosoftNETCoreApp90Version>$(MicrosoftNETCoreAppRuntimewinx64Version)</MicrosoftNETCoreApp90Version>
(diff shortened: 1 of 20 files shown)


eng/Versions.props
```diff
@@ -110,6 +110,14 @@
     <MicrosoftExtensionsLoggingVersion>$(MicrosoftExtensionsLogging80Version)</MicrosoftExtensionsLoggingVersion>
     <MicrosoftExtensionsLoggingAbstractionsVersion>$(MicrosoftExtensionsLoggingAbstractions80Version)</MicrosoftExtensionsLoggingAbstractionsVersion>
     <MicrosoftExtensionsLoggingConsoleVersion>$(MicrosoftExtensionsLoggingConsole80Version)</MicrosoftExtensionsLoggingConsoleVersion>
+  </PropertyGroup>  
+  <PropertyGroup Label=".NET 9 Dependent" Condition=" '$(TargetFramework)' == 'net9.0' ">
+    <MicrosoftAspNetCoreAuthenticationJwtBearerVersion>$(MicrosoftAspNetCoreApp80Version)</MicrosoftAspNetCoreAuthenticationJwtBearerVersion>
```
jander-msft, line 115: All of these versions should be referencing a `MicrosoftNETCoreApp90Version` or `MicrosoftAspNetCoreApp90Version` property (note these have `90` instead of `80` in their names).

Those two properties should be referencing `MicrosoftNETCoreAppRuntimewinx64Version` and `MicrosoftAspNetCoreAppRuntimewinx64Version` respectively.

And those two properties need to have their corresponding entries (e.g. https://github.com/dotnet/dotnet-monitor/blob/feature/9.x/eng/Version.Details.xml#L53) to be fixed up to have 9.0 versions that are coherent with the `Microsoft.Dotnet.Sdk.Internal` entry (which also needs to be fixed up to have a 9.0 version).

jander-msft (approved):

PR #5897, opened 2024-01-26
Update sdk version

wiktork:


jander-msft (approved):

jander-msft:
I think the mismatched runtime problem in the failing tests can be fixed by adding the .NET 9 entry for the Helix payload here: https://github.com/dotnet/dotnet-monitor/blob/feature/9.x/eng/helix/Helix.proj#L126

PR #5904, opened 2024-01-29
Learning Paths Automation

kkeirstead:
###### Summary

This PR adds new workflows to aid in keeping learning paths up-to-date. This is a simplified version of what I previously demoed (I removed the Issue step in the middle now that we're auto-applying the suggestions), with the following flow:

* Once a week (or on a different interval), "Check Learning Path Links" kicks off. This workflow compares the current state of the codebase with the previous one, as determined by a placeholder file "learning-path-sha.txt".
* Upon completion, "Submit Learning Path PR" kicks off. This creates a PR that includes details about modified files, things to manually review, automatic suggestions, and files that are 'out-of-sync' (i.e. not tied to the correct SHA). The commit hashes within the learning path files are automatically updated, and the suggestions are automatically applied (via line-number changes to links).
* This PR needs to be manually reviewed by whoever is handling the release and dependabot updates - outside of large refactors, this should be as simple as clicking on the PR's hyperlinks to ensure everything looks correct (and manually changing any of the links that need manual review due to more drastic changes).

There are a number of ways this could be improved (such as stable tags instead of commit hashes or increasing what this can handle automatically), but I'm trying not to over-engineer V1 of this so we can start using it here and in other repos. If we notice pain points, updates can be made to address them, but hopefully this should reduce the burden of keeping the learning paths up to date. By not running on every PR, this should hopefully alleviate some of the concerns from the first prototype about this getting in the way for every commit.

[This is an example run on my fork](https://github.com/kkeirstead/dotnet-monitor/pull/333) - this simulates miscellaneous changes being made over the course of a week, and what PR would be auto-generated. The auto-applied suggestions handle in-file adjustments, and files that have ambiguous links or were renamed/deleted are listed under manually review. The "Out Of Sync" section shows that `configuration.md` has a link referencing main instead of the correct hash.


.github/actions/apply-learning-path-updates/index.js
```diff
@@ -0,0 +1,73 @@
...
+
+function AppendModifiedFiles(path)
+{
+  modifiedFiles.push(path)
+  core.setOutput('modifiedFiles', modifiedFiles.join(' '))
+}
+
+function ReplaceOldWithNewText(content, oldText, newText)
+{
+  return content.replace(new RegExp(oldText, 'g'), newText);
```
schmittjoseph, line 15: We should avoid creating a regex expression from an arbitrary variable here. `context.replaceAll(oldText, newText)` should work here instead

.github/workflows/check-learning-path-links.yml
```diff
@@ -0,0 +1,78 @@
...
+          changedFilePaths: ${{ env.updated_files }}
+          sourceDirectoryName: 'src'
+          oldHash: ${{ env.prev_sha }}
+          newHash: ${{ env.new_sha }}
+          
+      - name: Generate artifacts (Comment)
+        working-directory: ./head
+        run: |
+          mkdir -p ./learning-path-review
+          echo -n "${{ steps.check-links.outputs.modifiedFiles }}" > ./learning-path-review/modifiedFiles
```
schmittjoseph, line 69: Using substitution expressions here are a script injection risk. All of these should be passed in via an `env` block instead to avoid the risk.

Example:
https://github.com/dotnet/dotnet-monitor/blob/341ff64a6097fe0bc66950e254e6160abcf77b84/.github/workflows/update-release-version.yml#L89-L96

.github/workflows/check-learning-path-links.yml
```diff
@@ -0,0 +1,78 @@
+name: 'Check Learning Path Links'
+on:
+  schedule: # Run once a week
+    - cron: '0 0 * * 1'
```
schmittjoseph, line 4: I'd be in favor of this running every other week (or once a month). As it stands every week is a bit of noise for a manual review PR since there will always be changes everytime the HEAD commit is updated.

.github/workflows/check-learning-path-links.yml
```diff
@@ -0,0 +1,78 @@
...
+    name: 'Check Learning Path Links'
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+
+    steps:
+      - name: Checkout head
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: true
```
schmittjoseph, line 20: What's the reason for persisted credentials? This is a security risk, especially since the token has elevated permissions and this step is ran first. Depending on the reason we can refactor this a couple different ways to minimize the risk.

.github/workflows/submit-learning-path-pr.yml
```diff
@@ -0,0 +1,117 @@
...
+          oldHash: ${{ env.prev_sha }}
+          newHash: ${{ env.new_sha }}
+
+      - name: Create PR Body
+        run: |
+          manuallyReview=$(cat ./learning-path-review/manuallyReview)
+          manuallyReview=${manuallyReview//,/<br />}
+          manuallyReviewSection="<h2>Manually Review:</h2>${manuallyReview}<br />"
+          if test -z "$manuallyReview"; then
+          manuallyReviewSection=""
```
schmittjoseph, line 83: nit: formatting, let's fix the indentation on the conditionals to make it easier to read.

.github/workflows/submit-learning-path-pr.yml
```diff
@@ -0,0 +1,117 @@
...
+          suggestions: ${{ env.suggestions }}
+          oldHash: ${{ env.prev_sha }}
+          newHash: ${{ env.new_sha }}
+
+      - name: Create PR Body
+        run: |
+          manuallyReview=$(cat ./learning-path-review/manuallyReview)
+          manuallyReview=${manuallyReview//,/<br />}
+          manuallyReviewSection="<h2>Manually Review:</h2>${manuallyReview}<br />"
+          if test -z "$manuallyReview"; then
```
schmittjoseph, line 82: Let's switch all of these if statements to use the bracket commands instead of explicit `test` as that's more conventional in bash scripts.

e.g. `if [ -z "$manuallyReview" ]`

schmittjoseph (changes requested):
Blocking on security related comments

.github/workflows/submit-learning-path-pr.yml
```diff
@@ -0,0 +1,117 @@
+name: 'Submit Learning Path PR'
```
schmittjoseph, line 1: I'm not sure I see the value of having split workflows here. It's adding a lot of complexity (and potential security risks) without adding value now that the proposed issue step has been removed.

Let's consolidate the split workflows and actions into a single one instead. If we consolidate them, then the action can also be responsible for generating a report (pr body) instead of having a bash script do it which is more prone to script injections.

kkeirstead:
Closing this to prevent churn while testing/debugging new changes...I made the mistake of merging directly from my fork's main branch where I had been running the workflow. Will make sure comments are not lost.

PR #5940, opened 2024-01-31
Learning Paths Automation

kkeirstead:
###### Summary

Closed out https://github.com/dotnet/dotnet-monitor/pull/5904 to make some changes to my fork - the primary change with the new version is that the workflows have been unified (there is no longer a "submit PR" workflow, but the logic for it has been moved into the "Check For Staleness" workflow). Additional small fixes were made - I will individually reference them in comments so no feedback is lost.


kkeirstead:
> We should avoid creating a regex expression from an arbitrary variable here. `content.replaceAll(oldText, newText)` should work here instead


https://github.com/dotnet/dotnet-monitor/pull/5904#discussion_r1470090712 -> resolved

kkeirstead:
> I'd be in favor of this running every other week or once a month. As it stands every week is a bit of noise for a manual review PR since there will always be changes everytime the HEAD commit is updated.

https://github.com/dotnet/dotnet-monitor/pull/5904#discussion_r1470098266 -> should now be set to once a month (we can always tune this later as well)

kkeirstead:
> nit: formatting, let's fix the indentation on the conditionals to make it easier to read.
---and---
> Let's switch all of these if statements to use the bracket commands instead of explicit `test` as that's more conventional in bash scripts.
> 
> e.g. `if [ -z "$manuallyReview" ]`

https://github.com/dotnet/dotnet-monitor/pull/5904#discussion_r1470102736 and https://github.com/dotnet/dotnet-monitor/pull/5904#discussion_r1470110002 should be resolved

kkeirstead:
> I'm not sure I see the value of having split workflows here. It's adding a lot of complexity (and potential security risks) without adding value now that the proposed issue step has been removed.
> 
> Let's consolidate the split workflows and actions into a single one instead. If we consolidate them, then the action can simply generate a report (pr body) instead of outputting several variables and having a bash script do it which is more prone to script injections.

https://github.com/dotnet/dotnet-monitor/pull/5904#discussion_r1470123474 Just realized I think you wanted the actions combined as well (I only combined the workflows) - I can work on putting those together.

kkeirstead:
> > I'm not sure I see the value of having split workflows here. It's adding a lot of complexity (and potential security risks) without adding value now that the proposed issue step has been removed.
> > Let's consolidate the split workflows and actions into a single one instead. If we consolidate them, then the action can simply generate a report (pr body) instead of outputting several variables and having a bash script do it which is more prone to script injections.
> 
> [#5904 (comment)](https://github.com/dotnet/dotnet-monitor/pull/5904#discussion_r1470123474) Just realized I think you wanted the actions combined as well (I only combined the workflows) - I can work on putting those together.

I've condensed everything down into a single action (and a single workflow) -> considering this resolved

.github/workflows/check-learning-path-links.yml
```diff
@@ -0,0 +1,115 @@
...
+        with:
+          repoURLToSearch: 'https://github.com/dotnet/dotnet-monitor'
+          learningPathsDirectory: 'documentation/learningPath'
+          changedFilePaths: ${{ env.updated_files }}
+          sourceDirectoryName: 'src'
+          oldHash: ${{ env.prev_sha }}
+          newHash: ${{ env.new_sha }}
+          learningPathHashFile: '.github/learning-path-sha.txt'
+
+      - name: Create PR Body
```
schmittjoseph, line 71: This seems like something that the GitHub action should generate. It's output isn't really be used for anything except generating this PR body so I'd recommend we just generate it in the action and avoid introducing a bash script here

.github/workflows/check-learning-path-links.yml
```diff
@@ -0,0 +1,115 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          MODIFIED_FILES: ${{ steps.check-links.outputs.modifiedFiles }}
+          MANUALLY_REVIEW: ${{ steps.check-links.outputs.manuallyReview }}
+          SUGGESTIONS: ${{ steps.check-links.outputs.suggestions }}
+          OUT_OF_SYNC: ${{ steps.check-links.outputs.outOfSync }}
+
+      - name: Open PR
+        uses: ./.github/actions/open-pr
+        with:
+          files_to_commit: ${{ steps.check-links.outputs.modifiedFilesToCommit }}
```
schmittjoseph, line 109: We shouldn't actually need this output from the GitHub action, we can just tell this `open-pr` action to commit any files that have changed after `check-link` was ran.

e.g.


```yaml
      - name: Open PR
        uses: ./.github/actions/open-pr
        with:
          files_to_commit: "*"
```

.github/workflows/check-learning-path-links.yml
kkeirstead: Agreed - this was originally meant to give flexibility to anyone consuming the action to do what they wanted with the output, but realistically this makes more sense and is cleaner for our use case.

.github/workflows/check-learning-path-links.yml
kkeirstead: I did try this for a run, but it appears that because I have two versions of the repo checked out (the main and the "prev"), it wasn't happy with the `*` (here's the run for reference https://github.com/kkeirstead/dotnet-monitor/actions/runs/7743591239/job/21115353520). If we want to avoid the current solution I can look into this more, but figured I'd wait until you weighed in.

schmittjoseph: From what I can tell this is because the "prev" repo is being check out inside of "main" as opposed to next to it

How it looks now:
```
-> /home/runner/work/dotnet-monitor/
--> /dotnet-monitor (this is "main")
-----> /prev
```

What it should be:
```
-> /home/runner/work/dotnet-monitor/
--> /dotnet-monitor
--> /prev
```

If you would like, I can take a look at the options for actions/checkout to see how we should configure it

.github/workflows/check-learning-path-links.yml
```diff
@@ -0,0 +1,80 @@
...
+  workflow_dispatch:
+
+permissions: {}
+  
+jobs:
+  check-learning-path-links:
+    name: 'Check Learning Path Links'
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
```
schmittjoseph, line 14: Leftover permission?

.github/workflows/check-learning-path-links.yml
kkeirstead: So, there is a way to check out side-by-side (that was what I originally was doing before I switched to this approach), but then I was having issues with the open-pr action not working. I spent most of this morning trying to get this to work, but I'm still hitting snags where I get close but it's just not quite right. I don't doubt there's a way to do this - I just don't think it's worth me spending more time on it right now for the same end result.

IMO, unless there's still more security concerns, I'm really trying to wrap up this PR. There's been so much churn from the original prototype that this has taken way more time than I wanted to allocate to it. We can iterate on it as-needed for any edge-cases I missed or other optimizations to make it cleaner or work better.

schmittjoseph: > We can iterate on it as-needed for any edge-cases I missed or other optimizations to make it cleaner or work better.

That sounds good to me. As a short term solution you should just be able to use `--all -- :!prev` as the `files_to_commit`, which will add everything except files from the `prev` folder

.github/actions/learning-path-staleness-check/index.js
```diff
@@ -0,0 +1,288 @@
...
+function AppendLineNumber(text, lineNumber)
+{
+  if (!lineNumber) { return text }
+
+  return text + " " + linePrefix + lineNumber
+}
+
+function CheckForEndOfLink(str, startIndex)
+{
+  const illegalCharIndex = str.substr(startIndex).search("[(), '\`\"\}\{]|\. "); // This regex isn't perfect, but should cover most cases.
```
schmittjoseph, line 95: Do you mind providing some examples of what links you expect to handle here? This regex doesn't look correct.

E.g. Are you only handles markdown links for now? `[friendly text](URL)`, or are you trying to handle any link that appears in the file, regardless of it its annotated?

.github/actions/learning-path-staleness-check/index.js
```diff
@@ -0,0 +1,288 @@
...
+  var modifiedFilesSection = "<h2>Modified Files:</h2>" + Array.from(modifiedFiles).join("<br />") + "<br />";
+  if (modifiedFiles.size === 0) { modifiedFilesSection = ""; }
+
+  var body = modifiedFilesSection + manuallyReviewSection + outOfSyncSection + suggestionsSection;
+  console.log("body=" + body);
+
+  core.setOutput('outputText', body);
+}
+
+function ValidateLinks(learningPathContents, repoURLToSearch, modifiedPRFiles, learningPathFile, oldHash, newHash, sourceDirectoryName, excludeLinksArray)
```
schmittjoseph, line 133: This can be addressed in a separate PR, but this method is fairly hard to follow due to how much its doing and how many cases it's trying to handle. I'd recommend splitting this up into several functions that have distinct responsibilities (and clear input/output)

.github/actions/learning-path-staleness-check/index.js
schmittjoseph: To expand on this, ignoring the correctness of the regex, depending on the examples of urls you are trying to handle this regex likely is not necessary in the first place

.github/actions/action-utils.js
```diff
@@ -51,4 +51,7 @@ module.exports.friendlyDateFromISODate = function(isoDate) {
 
 module.exports.splitVersionTag = splitVersionTag;
 module.exports.readFile = (fileName) => util.promisify(fs.readFile)(fileName, 'utf8');
-module.exports.writeFile = (fileName, contents) => util.promisify(fs.writeFile)(fileName, contents);
\ No newline at end of file
+module.exports.readFileSync = (fileName) => fs.readFileSync(fileName, 'utf8');
```
schmittjoseph, line 54: Sync variants of existing methods shouldn't be added here, the only reason we're exporting `readFile` here is because we need to promisify the existing `fs.readFile` and this is a common thing for all actions. `{readFile,writeFile}Sync` aren't adding anything when the stock `fs` version works

.github/actions/learning-path-staleness-check/action.yml
```diff
@@ -0,0 +1,30 @@
...
+    description: 'The hash to be inserted into the updated learning path'
+    required: true
+  learningPathHashFile:
+    description: 'The file where the commit hash being used for the learning paths is kept'
+    required: true
+  excludeLinks:
+    description: 'Comma separated strings to exclude from processing'
+    required: false
+runs:
+  using: 'node16'
```
schmittjoseph, line 29: Outside of this PR we will want to upgrade all of our actions to node20 as node16 is being deprecated

.github/actions/learning-path-staleness-check/index.js
```diff
@@ -0,0 +1,288 @@
...
+    return actionUtils.readFileSync(path)
+  }
+  catch (error) {}
+
+  return undefined;
+}
+
+function ConstructOutputText(core)
+{
+  var manuallyReviewSection = "<h2>Manually Review:</h2>" + Array.from(manuallyReview).join("<br />") + "<br />";
```
schmittjoseph, line 115: nit: How it's written is a far bit harder to follow (and more prone to easy-to-miss bugs in variable usage)  than building the PR body as we go,


e.g.
```javascript
var body =""
if (manuallyReview.size) {
  body += "<h2>Manually Review:</h2>" + Array.from(manuallyReview).join("<br />") + "<br />"
}

...


```

.github/actions/learning-path-staleness-check/index.js
```diff
@@ -0,0 +1,288 @@
...
+        {
+          let updatedLink = StripLineNumber(link.replace(oldHash, newHash), linePrefixIndex) + linePrefix + newLineNumberFirst;
+          UpdateSuggestions(fileName, link, updatedLink, learningPathFile, learningPathLineNumber, oldLineNumber, newLineNumberFirst);
+        }
+      }
+    }
+  }
+}
+
+const main = async () => {
```
schmittjoseph, line 210: Similar feedback to `ValidateLinks`, and can also be addressed in a seperate PR.

This method is doing too much so it's hard to follow. Ideally a method is doing one thing with clear input and output. In the case of `main` or other methods that orchestrate a series of actions, only having them contain flow-control logic and not doing implementation themselves keeps things easier to follow and often less prone to bugs that'd get missed in a review.

.github/workflows/check-learning-path-links.yml
```diff
@@ -0,0 +1,79 @@
...
+        with:
+          persist-credentials: false
+          fetch-depth: 0 # Fetch the entire repo for the below git commit graph operations
+          ref: main
+
+      - name: Get current SHA
+        id: get_new_sha
+        run: |
+          new_sha=$(git rev-parse HEAD)
+          echo "new_sha=$new_sha" >> $GITHUB_ENV
```
schmittjoseph, line 29: This step isn't necessary. GitHub (and other pipeline platforms) provide predefined variables to the agent describing the run. 

For GitHub you can just use `GITHUB_SHA` (https://docs.github.com/en/actions/learn-github-actions/variables contains a full list)

.github/workflows/check-learning-path-links.yml
```diff
@@ -0,0 +1,79 @@
...
+
+      - name: Get previous update SHA
+        id: get_sha
+        run: |
+          cd .github
+          prev_sha=$(cat learning-path-sha.txt)
+          echo "prev_sha=$prev_sha" >> $GITHUB_ENV
+          cd ../
```
schmittjoseph, line 37: Not necessary to change the working directory here, we can just use:
`prev_sha=$(cat .github/learning-path-sha.txt)`

.github/workflows/check-learning-path-links.yml
```diff
@@ -0,0 +1,79 @@
...
+    permissions:
+      contents: write
+      pull-requests: write
+
+    steps:
+      - name: Checkout head
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+          fetch-depth: 0 # Fetch the entire repo for the below git commit graph operations
```
schmittjoseph, line 22: This shouldn't be needed, there's no git graph operations that are being performed

.github/workflows/check-learning-path-links.yml
```diff
@@ -0,0 +1,79 @@
...
+          cd .github
+          prev_sha=$(cat learning-path-sha.txt)
+          echo "prev_sha=$prev_sha" >> $GITHUB_ENV
+          cd ../
+
+      - name: Checkout head
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: true
+          ref: main
```
schmittjoseph, line 43: We already have `main` checked out. I'm assuming this was in response to my comment here: https://github.com/dotnet/dotnet-monitor/pull/5904#discussion_r1470101252

Having `persist-credentials: true` on the first checkout is okay since we aren't calling code we don't own/trust, but we should at least have a comment explaining that this is intention. Here's an example from another workflow:
https://github.com/dotnet/dotnet-monitor/blob/31385e3dceb2084584324fd011c9678e49439e2d/.github/workflows/cleanup-releases.yml#L23

.github/workflows/check-learning-path-links.yml
```diff
@@ -0,0 +1,79 @@
...
+
+      - name: Open PR
+        uses: ./.github/actions/open-pr
+        with:
+          files_to_commit: ${{ steps.check-links.outputs.modifiedFilesToCommit }}
+          title: "[REQUIRES MANUAL REVIEW] Update Learning Paths"
+          commit_message: Update Learning Paths
+          body: This PR was auto generated and will not be automatically merged in - adjustments should be made manually as-needed.<br />${{ steps.check-links.outputs.outputText }}
+          branch_name: learningPathUpdates/${{ env.new_sha }}
+          fail_if_files_unchanged: true
```
schmittjoseph, line 78: This doesn't need to be addressed in this PR, but it should be addressed soon.

This will eventually lead to workflow failures if there were no new commits to a repro between workflow runs (so `HEAD` and `learning-path-sha` are the same). I'd suggest breaking up this pipeline into distinct jobs (one to calculate the set of files changed between commits, and one to update the links / open a pr).

.github/actions/learning-path-staleness-check/index.js
```diff
@@ -0,0 +1,288 @@
...
+    });
+
+    actionUtils.writeFileSync(learningPathHashFile, newHash);
+    AppendModifiedFilesToCommit(learningPathHashFile, core)
+
+    // Scan each file in the learningPaths directory
+    actionUtils.readdir(learningPathDirectory, (_, files) => {
+      files.forEach(learningPathFile => {
+        try {
+          const fullPath = learningPathDirectory + "/" + learningPathFile
```
schmittjoseph, line 251: nit: we should be using `path.join` here instead of concatenation

Example:
https://github.com/dotnet/dotnet-monitor/blob/31385e3dceb2084584324fd011c9678e49439e2d/.github/actions/generate-release-notes/index.js#L42

.github/actions/learning-path-staleness-check/index.js
```diff
@@ -0,0 +1,288 @@
+const actionUtils = require('../action-utils.js');
+const prevPathPrefix = "prev/";
+const linePrefix = "#L";
+const separator = " | ";
+
+modifiedFilesDict = {};
```
schmittjoseph, line 6: I'm not really sure what this lookup is storing, the path seems to be the modified file path, but I'm not sure about the value. Let's rename it to something more descriptive (like you did for `modifiedFilesUrlToFileName`)

schmittjoseph (changes requested):
I just realized my PR comments were submitted as individual comments instead of a single review. I've updated my feedback to mention which ones we can delay for a later PR

.github/workflows/check-learning-path-links.yml
kkeirstead: The comment was inaccurate, but I actually added this intentionally in my testing because there was an instance of my prior commit not being found if the fetch depth wasn't set.

.github/actions/learning-path-staleness-check/index.js
kkeirstead: Though in the case of our learning docs I believe basically everything does follow the `[]()` format, this should also handle normal links in text with a trailing space (I also made an adjustment to look for newlines after periods in the latest version)

.github/workflows/check-learning-path-links.yml
kkeirstead: I was aware of this, but wasn't too worried given 1) that would suggest no new commits over the course of a month (which is incredibly unlikely at least for our repo), and 2) the workflow failing wouldn't have any consequences. But yes, this could be addressed later.

.github/actions/learning-path-staleness-check/index.js
kkeirstead: Though the whole thing could use some refactoring (as mentioned in another comment), this in particular was a side effect of compressing the two actions into one, and I just put it here for simplicity when I did that.

.github/actions/learning-path-staleness-check/index.js
kkeirstead: I would be open to improving this separately, though I remember in my previous attempts back in October this was pretty challenging without passing tons of parameters around.

.github/actions/learning-path-staleness-check/index.js
```diff
@@ -0,0 +1,274 @@
...
+  var linkIndices = [];
+  for(var pos = learningPathContents.indexOf(repoURLToSearch); pos !== -1; pos = learningPathContents.indexOf(repoURLToSearch, pos + 1)) {
+      linkIndices.push(pos);
+  }
+
+  for(let startOfLink of linkIndices)
+  {
+    // Clean up the link, determine if it has a line number suffix
+    let endOfLink = startOfLink + CheckForEndOfLink(learningPathContents, startOfLink)
+    if (endOfLink === -1) { endOfLink = learningPathContents.length; } // If no illegal characters are found, the link is at the end of the file
```
schmittjoseph, line 135: We should be checking the return value of `CheckForEndOfLink(...)` for `-1` and not `endOfLink`, since otherwise this check will only work if `startOfLink = 0` and `CheckForEndOfLink` returns `-1`.

schmittjoseph: Other than this 1 thing changes LGTM

kkeirstead: For the sake of not having to name another temp variable, I just checked endOfLink against startOfLink, which should have the same effect.

schmittjoseph (approved):

PR #6039, opened 2024-02-16
[DO NOT REVIEW] Fix dependabot

schmittjoseph:
###### Summary

TODO


eng/dependabot/net8.0/Directory.Build.props
```diff
@@ -1,4 +1,8 @@
 <Project>
   <!-- Dependendabot is configured to read package versions from Directory.Build.props, so just use this file as a shim. -->
   <Import Project="Versions.props" />
+  <Import Project="Packages.props" />
+  <PropertyGroup>
+    <_SuppressSdkImports>true</_SuppressSdkImports>
```
jander-msft, line 6: The other versioned Directory.Build.props files do not have this setting. Is the needed for this version for some particular reason?

jander-msft (dismissed):

eng/dependabot/net8.0/Directory.Build.props
schmittjoseph: This was a bad copy-paste, good catch. Fixed net8.0

jander-msft (approved):

PR #6060, opened 2024-02-20
[Parameter Capturing] Add capture limit option

schmittjoseph:
###### Summary

Add a new option to parameter capturing that will stop the capturing session after a certain number of times parameters are captured. Note that this refers to the number of "sets" of parameters collected, not the total number of individual parameters. For example if `MyMethod(int a, int b, int c)` is called twice, then that would be 2 sets of parameters.

###### Release Notes Entry
Update `POST /parameters` to support stopping after a certain number of times parameters are captured.

### DIFF
--- a/documentation/api/definitions.md
+++ b/documentation/api/definitions.md
@@ -97,6 +97,7 @@ Object describing the list of methods to capture parameters for.
 |---|---|---|
 | `methods` | [MethodDescription](#methoddescription)[] | Array of methods to capture parameters for. |
 | `useDebuggerDisplayAttribute` | bool | Determines if parameters should be formatted using their [`DebuggerDisplayAttribute`](https://learn.microsoft.com/dotnet/api/system.diagnostics.debuggerdisplayattribute) if available and supported. Expressions in attributes may consist of properties, fields, methods without parameters, or any combination of these. |
+| `captureLimit` | int | The number of times to capture parameters before stopping. Note that parameters may continue to be captured for a short amount of time after this limit is reached. |
 
 ## DotnetMonitorInfo
 
--- a/documentation/api/definitions.md
+++ b/documentation/api/definitions.md
@@ -97,7 +97,7 @@ Object describing the list of methods to capture parameters for.
 |---|---|---|
 | `methods` | [MethodDescription](#methoddescription)[] | Array of methods to capture parameters for. |
 | `useDebuggerDisplayAttribute` | bool | Determines if parameters should be formatted using their [`DebuggerDisplayAttribute`](https://learn.microsoft.com/dotnet/api/system.diagnostics.debuggerdisplayattribute) if available and supported. Expressions in attributes may consist of properties, fields, methods without parameters, or any combination of these. |
-| `captureLimit` | int | The number of times to capture parameters before stopping. Note that parameters may continue to be captured for a short amount of time after this limit is reached. |
+| `captureLimit` | int | The number of times to capture parameters before stopping. If the specified duration elapses the operation will stop even if the capture limit is not yet reached. Note that parameters may continue to be captured for a short amount of time after this limit is reached. |
 
 ## DotnetMonitorInfo
 
--- a/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Pipeline/CaptureLimitPolicyProbes.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Pipeline/CaptureLimitPolicyProbes.cs
@@ -10,20 +10,19 @@ using System.Threading.Tasks;
 
 namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing.Pipeline
 {
-    internal sealed class PolicyEnforcerProbe : IFunctionProbes
+    internal sealed class CaptureLimitPolicyProbes : IFunctionProbes
     {
         private readonly IFunctionProbes _probes;
         private readonly int _captureLimit;
-
-        private int _captureCount;
         private readonly TaskCompletionSource _stopRequest;
 
+        private int _captureCount;
         private bool _stopped;
 
-        public PolicyEnforcerProbe(IFunctionProbes probes, int captureLimit, TaskCompletionSource stopRequest)
+        public CaptureLimitPolicyProbes(IFunctionProbes probes, int captureLimit, TaskCompletionSource stopRequest)
         {
-            _captureLimit = captureLimit;
             _probes = probes;
+            _captureLimit = captureLimit;
             _stopRequest = stopRequest;
         }
 
--- a/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Pipeline/CaptureLimitPolicyProbes.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Pipeline/CaptureLimitPolicyProbes.cs
@@ -7,7 +7,6 @@ using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 
-
 namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing.Pipeline
 {
     internal sealed class CaptureLimitPolicyProbes : IFunctionProbes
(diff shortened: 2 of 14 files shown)


src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Pipeline/CaptureLimitPolicyProbes.cs
```diff
@@ -0,0 +1,49 @@
...
+            if (_stopped)
+            {
+                return false;
+            }
+
+            bool didCapture = _probes.EnterProbe(uniquifier, args);
+            if (didCapture && Interlocked.Increment(ref _captureCount) == _captureLimit)
+            {
+                _stopped = true;
+                _ = _stopRequest.TrySetResult();
```
schmittjoseph, line 43: Note: By setting this task completion source we will automatically update dotnet-monitor's operation status for the current parameter capturing session. (i.e. it will go from `Running` -> `Succeeded`, even if the duration hasn't expired yet).

documentation/api/definitions.md
```diff
@@ -97,6 +97,7 @@ Object describing the list of methods to capture parameters for.
 |---|---|---|
 | `methods` | [MethodDescription](#methoddescription)[] | Array of methods to capture parameters for. |
 | `useDebuggerDisplayAttribute` | bool | Determines if parameters should be formatted using their [`DebuggerDisplayAttribute`](https://learn.microsoft.com/dotnet/api/system.diagnostics.debuggerdisplayattribute) if available and supported. Expressions in attributes may consist of properties, fields, methods without parameters, or any combination of these. |
+| `captureLimit` | int | The number of times to capture parameters before stopping. Note that parameters may continue to be captured for a short amount of time after this limit is reached. |
```
kkeirstead, line 100: Could we also update the examples in https://github.com/dotnet/dotnet-monitor/blob/main/documentation/api/parameters.md so that we have one with the `captureLimit`?

kkeirstead: Something I just thought of that I don't think there's a solution for, but maybe we can try to get ahead of it in the docs - people might not realize that the duration seconds still defaults to 30s when they use the capture limit. I can imagine setting a limit of 10 and being really confused when it stops before that, not realizing the duration has to be explicitly set to something higher (or indefinite) since it's whichever finishes first.

schmittjoseph: Good point. Perhaps for the description we say something like:
` The number of times to capture parameters before stopping. If the specified duration elapses the operation will stop even if the capture limit is not yet reached`

src/Tests/Microsoft.Diagnostics.Monitoring.HostingStartup.UnitTests/ParameterCapturing/Pipeline/ParameterCapturingPipelineTests.cs
```diff
@@ -341,6 +330,51 @@ public async Task Request_StopsAfterDuration()
             Assert.Equal(payload.RequestId, stoppedRequest);
         }
 
+        [Fact]
+        public async Task Request_StopsAfterCaptureLimit()
```
jander-msft, line 334: Consider parameterizing this test with a capture limit parameter. I'd like for it to test "the corners" (e.g. capture limit 0, 1, 2) to make sure no assumptions are made about the limit.

src/Tests/Microsoft.Diagnostics.Monitoring.HostingStartup.UnitTests/ParameterCapturing/Pipeline/CaptureLimitPolicyProbesTests.cs
```diff
@@ -0,0 +1,84 @@
...
+    {
+        [Fact]
+        public void EnterProbe_RequestStop_OnLimitReached()
+        {
+            // Arrange
+            TaskCompletionSource requestStop = new();
+            CaptureLimitPolicyProbes probes = new(new TestFunctionProbes(), captureLimit: 1, requestStop);
+
+            // Act
+            probes.EnterProbe(1, []);
```
jander-msft, line 23: Probably should... (ahem) _capture_... the return value and assert the expected value.

schmittjoseph: In general unit tests should only be testing one thing at a time so they should only fail for one reason. I can add another test to validate these probes pass through the uniqufier / arguments, but I'd rather not update the other tests to also validate that.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Pipeline/CaptureLimitPolicyProbes.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            // In addition to the stop request, use a flag to more quickly react to the limit being reached,
+            // limiting the amount of extra data being captured which is important in the case of hot paths.
+            if (_stopped)
+            {
+                return false;
+            }
+
+            bool didCapture = _probes.EnterProbe(uniquifier, args);
+            if (didCapture && Interlocked.Increment(ref _captureCount) == _captureLimit)
```
wiktork, line 40: I'm assuming EnterProbe is not locked correct? If two threads do this at the same time, one of them will receive a value greater than capture limit. Change this test to >=

schmittjoseph: We shouldn't need to do that -- if two threads enter at the same time one of the threads will still pass this check. Only 1 thread needs to signal the capture limit has been reached

src/Tests/Microsoft.Diagnostics.Monitoring.HostingStartup.UnitTests/ParameterCapturing/Pipeline/CaptureLimitPolicyProbesTests.cs
```diff
@@ -0,0 +1,135 @@
...
+            probes.EnterProbe(2, []);
+
+            // Assert
+            Assert.Equal(1, invokeCount);
+        }
+
+        [Fact]
+        public void EnterProbe_PassesThrough_CacheMethods()
```
schmittjoseph, line 88: Wrong name..

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Pipeline/CaptureLimitPolicyProbes.cs
```diff
@@ -0,0 +1,49 @@
...
+            _stopRequest = stopRequest;
+        }
+
+        public void CacheMethods(IList<MethodInfo> methods) => _probes.CacheMethods(methods);
+
+        public bool EnterProbe(ulong uniquifier, object[] args)
+        {
+            // In addition to the stop request, use a flag to more quickly react to the limit being reached,
+            // limiting the amount of extra data being captured which is important in the case of hot paths.
+            if (_stopped)
```
jander-msft, line 34: I'm trying to think if there's a way we can get the probe invocation and capture count exactly right (and not invoke/count more or less), but I'm fine with this as-is; it's better to over-collect than to under-collect.

jander-msft (approved):

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Pipeline/CaptureLimitPolicyProbes.cs
schmittjoseph: There's a couple ways of ensuring over-collecting doesn't happen, but it's important to note that above all else we want to prioritize minimizing perf impact here, so some minor over-collecting is an acceptable tradeoff for now IMO.

There's likely some larger refactoring we could do to avoid over-collecting while still preserving our perf (e.g. performing the capture limit enforcement in `ParameterCapturingLogger`). I'd prefer to keep things simple for now though until we need to better solve this issue. In the interim if occasional over-collection is a problem, the consumer of this data could always drop the extra messages.

PR #6130, opened 2024-02-27
Have automated merge commits respect branch-specific files

schmittjoseph:
###### Summary

This PR updates the (code) sync-branch workflow to understand and preserve branch-specific files during the automated merge. This means a seconds commit to re-apply branch-specific files is **usually** no longer needed. This works by overriding the merge driver for branch-specific files to always prefer the current branch's version of files.

Example PR: https://github.com/schmittjoseph/dotnet-monitor/pull/567

However, merge drivers are only involved if merge conflicts are present for a given file. That means it's possible for purely additive changes in `main` to still be merged into `feature/9.x` via this approach. To handle this we also perform the old approach of re-applying branch-specific files after a merge. If there were any files where this edge case occurs, they will be handled by this step.

Note that this requires us to maintain the list of branch-specific files in 2 places. We could likely consolidate this via some additional scripting in the workflow but I'd prefer that to be a separate piece of work if we end up wanting it.

### DIFF
--- a/.github/workflows/sync-branches.yml
+++ b/.github/workflows/sync-branches.yml
@@ -78,6 +78,14 @@ jobs:
           git config merge.ours.driver true
           git merge "$base_branch" --strategy=ort --strategy-option=theirs
 
+          git checkout "origin/${{ matrix.branch }}" -- \
+            "global.json" \
+            "eng/branch-vscode-config" \
+            "eng/common" \
+            "eng/Common.props" \
+            "eng/Versions.props" \
+            "eng/Version.Details.xml"
+
       - name: Open PR
         uses: ./.github/actions/open-pr
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/sync-branches.yml
```diff
@@ -74,9 +74,11 @@ jobs:
         run: |
           git config user.name "github-actions[bot]"
           git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
-
+          # Activate the ours merge driver to respect any branch specific files
+          git config merge.ours.driver true
           git merge "$base_branch" --strategy=ort --strategy-option=theirs
 
+          # Must be kept in sync with .gitattributes
```
schmittjoseph, line 81: See the PR description for details about consolidating this duplicate list -- tl;dr is that we could likely do it but it should be a separate effort.

jander-msft (approved):

PR #6146, opened 2024-02-29
Replace SwaggerUI and SwaggerScope option with OpenAPI document

jander-msft:
###### Summary

Remove the interactive SwaggerUI and SwaggerScope configuration option in favor of returning the OpenAPI document directly from the root route.


documentation/schema.json
```diff
@@ -1768,6 +1761,7 @@
           ]
         },
         "Filters": {
+          "description": "The filters that determine which exceptions should be included/excluded when collecting exceptions.",
```
schmittjoseph, line 1764: Any idea why this has changed?

src/Tools/dotnet-monitor/Startup.cs
```diff
@@ -109,12 +104,11 @@ public static void Configure(IApplicationBuilder app, IWebHostEnvironment env, I
             {
                 builder.MapControllers();
 
-                // Use a redirect to the Swagger UI if the request hits the default endpoint.
-                // This means that the swagger endpoint can have a permanent address, even if we decide to host
-                // something else at the root later.
-                builder.MapGet("/", (HttpResponse response) =>
+                builder.MapGet("/", (HttpResponse response, ISwaggerProvider provider) =>
```
schmittjoseph, line 107: Double checking -- this route doesn't require authentication to access, correct?

jander-msft: Correct

documentation/schema.json
jander-msft: If you look at `OptionsDisplayStrings.resx`, you'll see that `DisplayAttributeDescription_CollectExceptionsOptions_Filters` was missing from the resx file but it exists in the backing cs file. This caused the string property to return `null` so the openapi generation would see a description value of `null`. This is probably the result of a bad merge.

When I removed `DisplayAttributeDescription_AzureAdOptions_SwaggerScope` from the resx file, it regenerated the backing cs file and removed the `DisplayAttributeDescription_CollectExceptionsOptions_Filters` property. So I added it back to the resx file with the expected value.

I can try to back out the changes around `DisplayAttributeDescription_CollectExceptionsOptions_Filters` and attempt to fix them separately.

schmittjoseph: I'm fine with this as-is, it looks like our branches have this issue as well so we'd want both fixes backported anyways.

schmittjoseph (approved):

PR #6272, opened 2024-03-18
Replace LogEmittingProbes with EventSourceEmittingProbes and add support for artifact egress on the 'POST /parameters' route.

clguiman:
###### Summary
* Updated the 'POST /parameters' route to produce an artifact similar to all other routes.
  * Updated `CaptureParametersOperation` to inherit from `IArtifactOperation`. I removed `IInProcessOperation` since no one is using it anymore.
* Replaced `LogEmittingProbes` with `EventSourceEmittingProbes`. The new probe sends the captured parmaters back to dotnet-monitor through EventSource on a dedicated thread (see `AsyncParameterCapturingEventSource`).
* Updated `EventParameterCapturingPipeline` to read the captured parameters from event source and store them in the `ParameterCapturingStore`
* Updated the documentation.

###### Release Notes Entry
Parameters can now be sent to an egress provider and are no longer logged inside the target application.


src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/FunctionProbesStub.cs
```diff
@@ -10,8 +11,7 @@ public static class FunctionProbesStub
         private delegate void EnterProbeDelegate(ulong uniquifier, object[] args);
         private static readonly EnterProbeDelegate s_fixedEnterProbeDelegate = EnterProbeStub;
 
-        [ThreadStatic]
-        private static bool s_inProbe;
+        private static readonly AsyncLocal<int> s_inProbeCount = new();
```
schmittjoseph, line 14: This need to remain a `ThreadStatic` otherwise we risk recursing probes. `*Local` types in .NET (e.g. `ThreadLocal` and `AsyncLocal`) operate via managed code outside of our control. That managed code may call methods which we have probes installed into, so the recursing check would cause recursion.

schmittjoseph: > Updated FunctionProbes to support pausing capturing across async operations and added the NoProbeContext class to easily define a scope in which capturing is disabled.

Why is this change needed? Is it because event source writes out events using tasks?

clguiman: I noticed we were tracking String.Concat and other in `FunctionProbesManager.StartCapturingAsync`

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -592,28 +592,75 @@ public ActionResult<DotnetMonitorInfo> GetInfo()
...
+                    processInfo,
+                    requestId => _captureParametersFactory.Create(requestId, processInfo.EndpointInfo, configuration, duration),
+                    tags);
+            }, processKey, Utilities.ArtifactType_RequestParameters);
         }
 
-        [HttpGet("stacks", Name = nameof(CaptureStacks))]
-        [ProducesWithProblemDetails(ContentTypes.ApplicationJson, ContentTypes.TextPlain, ContentTypes.ApplicationSpeedscopeJson)]
+        [HttpGet("parameters", Name = nameof(GetCapturedParameters))]
+        [ProducesWithProblemDetails(ContentTypes.ApplicationJson, ContentTypes.TextPlain)]
```
clguiman, line 604: I should remove GetCapturedParameters and rework CaptureParameters to write the captured parameters to the specified egress provider

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/FunctionProbesStub.cs
clguiman: I reverted back to `ThreadStatic`, but I kept the ref count. We discussed offline for having a collection of excluded threads, but I think the ref count approach is simpler and doesn't involve any locking.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/EventSourceEmittingProbes.cs
```diff
@@ -29,18 +33,20 @@ public void CacheMethods(IList<MethodInfo> methods)
...
-            // All probes from the known console logger processor thread do not log.
-
-            // Possible additional guards in the future:
-            // If we instrument Modules instead of specific methods, we may need to exclude certain types to prevent noise. (such as System.String)
-            // If other custom loggers create background threads, we may need a way to specify exclusions for those threads.
+            try
+            {
+                return EnterProbeInternal(uniquifier, args);
+            }
+            catch (Exception ex)
```
schmittjoseph, line 40: This changes the probe faulting behavior -- the current behavior is that if a probe faults for whatever reason, all probes will automatically get uninstalled. Let's restore that original behavior (by not swallowing the exception). If we want to change the faulting behavior I'd prefer that to be a separate PR with more discussion around it.

As for keeping this try/catch and reporting back the exception, what does this PR do with the `FailedToCapture` event?

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Eventing/AsyncParameterCapturingEventSource.cs
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        private void ThreadLoop()
+        {
+            using IDisposable _ = MonitorExecutionContextTracker.MonitorScope();
+            using IDisposable __ = new NoProbeScope();
+
+            try
+            {
+                while (_pendingEvents.TryTake(out Action<CancellationToken>? eventAction, Timeout.Infinite, _cts.Token))
```
wiktork, line 115: Extract the token first from _cts. Otherwise, disposing asynchronously will cause .Token to throw ObjectDisposedException instead of cancelling as expected.

wiktork: Actually nvm, you join this thread before disposing.

src/Microsoft.Diagnostics.Monitoring.WebApi/ParameterCapturing/ICapturedParameters.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi.ParameterCapturing
+{
+    internal interface ICapturedParameters
+    {
+        Guid RequestId { get; }
+
+        public string ActivityId { get; }
```
wiktork, line 13: remove protection levels

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/FunctionProbesStub.cs
```diff
@@ -11,7 +11,7 @@ public static class FunctionProbesStub
         private static readonly EnterProbeDelegate s_fixedEnterProbeDelegate = EnterProbeStub;
 
         [ThreadStatic]
-        private static bool s_inProbe;
+        private static int s_inProbeCount;
```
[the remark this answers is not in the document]
clguiman: I've created #6330

documentation/api/parameters.md
```diff
@@ -1,12 +1,12 @@
...
 
-# Parameters - Post (experimental feature)
+# Parameters (experimental feature)
 
-Captures parameters for one or more methods each time they are called. Parameters are logged inside the target application using its [`ILogger`](https://learn.microsoft.com/dotnet/api/microsoft.extensions.logging.ilogger).
+Captures parameters for one or more methods each time they are called.
 
 > [!NOTE]
-> Unlike other artifacts, parameters do **not** support being sent to an egress provider.
+> Before version 8.0 Preview 3, parameters were logged inside the target application using its [`ILogger`](https://learn.microsoft.com/dotnet/api/microsoft.extensions.logging.ilogger).
```
kkeirstead, line 9:
```suggestion
> Before version 9.0 Preview 3, parameters were logged inside the target application using its [`ILogger`](https://learn.microsoft.com/dotnet/api/microsoft.extensions.logging.ilogger).
```

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Eventing/AsyncParameterCapturingEventSource.cs
```diff
@@ -0,0 +1,122 @@
...
+                IsBackground = true,
+                Name = BackgroundLoggingThreadName
+            };
+            _thread.Start();
+        }
+
+        public void Dispose()
+        {
+            _pendingEvents.CompleteAdding();
+            _cts.Cancel();
```
kkeirstead, line 40: Should this be wrapped in a try/catch, or is this safe here?

src/Microsoft.Diagnostics.Monitoring.WebApi/ParameterCapturing/CapturedParametersFormatter.cs
```diff
@@ -0,0 +1,115 @@
...
+            await writer.WriteAsync(builder, token);
+
+#if NET8_0_OR_GREATER
+            await writer.FlushAsync(token);
+#else
+            await writer.FlushAsync();
+#endif
+        }
+
+        private static string GetValueOrUnknown(string value) => string.IsNullOrEmpty(value) ? " <unknown>" : value;
```
kkeirstead, line 87: Is there a redundant space before `<unknown>`?

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/HttpApi/ApiClientExtensions.cs
```diff
@@ -516,16 +516,24 @@ public static async Task<OperationStatusResponse> PollOperationToCompletion(this
...
-        public static async Task<OperationResponse> CaptureParametersAsync(this ApiClient client, int pid, TimeSpan duration, CaptureParametersConfiguration config, TimeSpan timeout)
+        public static Task<OperationResponse> CaptureParametersAsync(this ApiClient client, int pid, TimeSpan duration, CaptureParametersConfiguration config, string egressProvider)
+        {
+            return client.CaptureParametersAsync(pid, duration, egressProvider, config, TestTimeouts.HttpApi);
+        }
+
+        /// <summary>
+        /// POST /parameters
+        /// </summary>
+        public static async Task<OperationResponse> CaptureParametersAsync(this ApiClient client, int pid, TimeSpan duration, string egressProvider, CaptureParametersConfiguration config, TimeSpan timeout)
```
kkeirstead, line 533: nit: For readability can we keep the ordering of parameters consistent here - it looks like this flips `egressProvider` and `config` compared to the method right above it

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/ParameterCapturingTests.cs
```diff
@@ -131,23 +130,52 @@ public async Task StopsProducingLogStatementsAfterOperationCompletes(Architectur
...
-                Assert.Equal(OperationState.Failed, operationResult.OperationStatus.Status);
+                Assert.Equal(HttpStatusCode.Created, operationResult.StatusCode);
+                Assert.Equal(OperationState.Succeeded, operationResult.OperationStatus.Status);
+
+                Assert.True(File.Exists(operationResult.OperationStatus.ResourceLocation));
+                using FileStream outputStream = new(operationResult.OperationStatus.ResourceLocation, FileMode.Open);
+                CapturedParametersResult captureResult = await JsonSerializer.DeserializeAsync<CapturedParametersResult>(outputStream);
+
+                Assert.NotNull(captureResult);
+                Assert.Equal(1, captureResult.CapturedMethods.Count);
```
kkeirstead, line 161: nit: `Assert.Single`

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/EventSourceEmittingProbes.cs
```diff
@@ -82,11 +85,22 @@ public bool EnterProbe(ulong uniquifier, object[] args)
                     {
                         value = ObjectFormatter.FormatObject(_objectFormatterCache.GetFormatter(args[i].GetType()), args[i]);
```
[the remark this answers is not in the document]
clguiman: I'll address that in a separate PR

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/EventSourceEmittingProbes.cs
clguiman: It was an artifact of development, it shouldn't have made it in the PR. I reverted the change

documentation/api/parameters.md
```diff
@@ -90,8 +93,6 @@ The [examples](#examples) include a mixture of `UserCode` and `SystemCode` to he
 The following logger categories are used inside the target application when capturing parameters:
 | Category Name | Description |
 | -- | -- |
-| `DotnetMonitor.ParameterCapture.UserCode` | Parameters captured in methods considered `UserCode`. |
```
schmittjoseph, line 93: Also remove the `## UserCode vs SystemCode` section from the documentation since it no longer applies

documentation/api/definitions.md
```diff
@@ -87,6 +87,48 @@ Enumeration that describes the current state of the collection rule.
...
 
+## CapturedMethod
+
+First Available: 9.0 Preview 3
+
+Object describing a captured method and its parameters.
+
+| Name | Type | Description |
+|---|---|---|
+| `requestId` | guid | Unique identifier for the capture request operation. |
```
schmittjoseph, line 98: We shouldn't expose this, it's an implementation detail and doesn't mean anything to the consumer of this API. Captured parameters will inherently be grouped by "capture request operation" by the fact that they are being egressed together to the same destination.

documentation/api/definitions.md
```diff
@@ -87,6 +87,48 @@ Enumeration that describes the current state of the collection rule.
...
+| `activityId` | string | An identifier for the current activity at the time of the capture. For more information see [Activity.Id](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity.id).|
+| `capturedDateTime` | DateTime | Time when the method call was captured. |
```
schmittjoseph, line 100: nit: just `dateTime` the captured part is implied because of the context

documentation/api/definitions.md
```diff
@@ -87,6 +87,48 @@ Enumeration that describes the current state of the collection rule.
...
+| `module` | string | The method module name. |
+| `type` | string | The method type name. |
```
schmittjoseph, line 102: We should also be recording `threadId` and `activityIdFormat` to preserve parity with the current set of ambient data that parameters collect.

The current set of ambient data is defined here https://github.com/dotnet/dotnet-monitor/blob/40ff7284fae2b72728d7a9025be021c682420457/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingLogger.cs#L20

And collected here:  
https://github.com/dotnet/dotnet-monitor/blob/40ff7284fae2b72728d7a9025be021c682420457/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingLogger.cs#L104

documentation/api/parameters.md
```diff
@@ -123,23 +113,62 @@ Authorization: Bearer fffffffffffffffffffffffffffffffffffffffffff=
...
-        this: 'SampleWebApp.Controllers.HomeController',
-        number: 10)
-info: DotnetMonitor.ParameterCapture.SystemCode[0]
-      System.String.Concat(
-        str0: 'firstString',
-        str1: '.secondString')
+{
+    "captures": [
+        {
+            "requestId": "ca17b977-e5f4-46c5-98ca-17046ece998a",
```
schmittjoseph, line 121: Update these examples with the new model

documentation/api/README.md
```diff
@@ -24,6 +24,6 @@ The following are the root routes on the HTTP API surface.
 | [`/info`](info.md) | Gets info about `dotnet monitor`. | 6.0 |
 | [`/operations`](operations.md) | Gets egress operation status or cancels operations. | 6.0 |
 | [`/collectionrules`](collectionrules.md) | Gets the current state of collection rules. | 6.3 |
-|  [`/parameters`](parameters.md) | Captures parameters for one or more methods each time they are called. | 8.0 RC 1 |
+| [`/parameters`](parameters.md) | Captures parameters for one or more methods each time they are called. | 8.0 RC 1 (breaking change in 9.0 Preview 3) |
```
schmittjoseph, line 27: Open question for more of the dotnet-monitor folks, do we want to call out breaking changes for experimental features? I'm personally in favor of not doing so since experimental features provide no support guarantees are subject to change until they are no longer experimental.

documentation/api/definitions.md
```diff
@@ -87,6 +87,49 @@ Enumeration that describes the current state of the collection rule.
 | `Throttled` | Indicates that the collection rule is temporarily throttled because the ActionCountLimit has been reached within the ActionCountSlidingWindowDuration. |
 | `Finished` | Indicates that the collection rule has completed and will no longer trigger. |
 
+## CapturedMethod
+
+First Available: 9.0 Preview 3
```
schmittjoseph, line 92: reminder to update these `First Available: 9.0 Preview 3` with relevant version once the code has been approved

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Eventing/AsyncParameterCapturingEventSource.cs
```diff
@@ -0,0 +1,122 @@
...
+{
+    internal sealed class AsyncParameterCapturingEventSource : IDisposable
+    {
+        private readonly Thread _thread;
+        private readonly CancellationTokenSource _cts = new();
+        private readonly BlockingCollection<Action<CancellationToken>> _pendingEvents;
+        private readonly ParameterCapturingEventSource _eventSource;
+
+        private const int MaxPendingEventsCollectionCapacity = 1024;
+        private const string BackgroundLoggingThreadName = "[dotnet-monitor] Probe EventSource Thread";
```
schmittjoseph, line 20:
```suggestion
        private const string BackgroundLoggingThreadName = "[dotnet-monitor] Parameter Capturing EventSource";
```

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Eventing/AsyncParameterCapturingEventSource.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        public AsyncParameterCapturingEventSource(ParameterCapturingEventSource eventSource)
+        {
+            _pendingEvents = new BlockingCollection<Action<CancellationToken>>(MaxPendingEventsCollectionCapacity);
```
schmittjoseph, line 24: nit: Let's initialize this inline with its declaration just like the `CancellationTokenSource`

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Eventing/AsyncParameterCapturingEventSource.cs
```diff
@@ -0,0 +1,122 @@
...
+            _thread = new(ThreadLoop)
+            {
+                Priority = ThreadPriority.BelowNormal,
+                IsBackground = true,
+                Name = BackgroundLoggingThreadName
+            };
+            _thread.Start();
+        }
+
+        public void Dispose()
```
schmittjoseph, line 36: This will throw if called multiple times, lets use our existing `DisposableHelper` class here to handle this.

Example:
https://github.com/dotnet/dotnet-monitor/blob/81667980f8fc0212d1cb0c1727e5b4922accb185/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ParameterCapturingLogger.cs#L158-L166

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/MethodSignature.cs
```diff
@@ -0,0 +1,126 @@
...
+                    public const string Unknown = Internal.Prefix + "unknown" + Internal.Postfix;
+                }
+            }
+        }
+
+        public string ModuleName { get; } = method.Module.Name;
+
+        public string? TypeName { get; } = EmitTypeName(method.DeclaringType);
+
+        public string MethodName { get; } = GetMethodName(method);
```
schmittjoseph, line 47: Will this cache the result of `GetMethodName(...)` or will it invoke `GetMethodName` each time someone accesses the property?

src/Tools/dotnet-monitor/ParameterCapturing/CaptureParametersOperation.cs
```diff
@@ -33,13 +40,21 @@ internal sealed class CaptureParametersOperation : IInProcessOperation
...
+            CaptureParametersConfiguration configuration,
+            TimeSpan duration,
+            CapturedParameterFormat format)
         {
             _profilerChannel = profilerChannel;
             _endpointInfo = endpointInfo;
             _logger = logger;
             _configuration = configuration;
             _duration = duration;
+            _parameterCapturingStore = endpointInfo.ServiceProvider.GetRequiredService<IParameterCapturingStore>();
```
schmittjoseph, line 56: nit: Let's retrieve the store via the `CaptureParametersOperationFactory` constructor DI instead of reaching into the service provider here to be consistent with how we do DI elsewhere

src/Microsoft.Diagnostics.Monitoring.WebApi/ParameterCapturing/CapturedParametersFormatter.cs
```diff
@@ -0,0 +1,116 @@
...
+        private static Task WriteParametersAsJson(IEnumerable<ICapturedParameters> parameters, Stream outputStream, CancellationToken token)
+        {
+            CapturedParametersResult result = BuildResultModel(parameters);
+
+            return JsonSerializer.SerializeAsync(outputStream, result, cancellationToken: token);
+        }
+
+        private const string Indent = "  ";
+
+        private static async Task WriteParametersAsPlainText(IEnumerable<ICapturedParameters> parameters, Stream outputStream, CancellationToken token)
```
schmittjoseph, line 42:
```suggestion
        private static async Task WriteParametersAsPlainText(IEnumerable<ICapturedParameters> parameters, Stream outputStream, CancellationToken token)
```
```suggestion
        private static async Task WriteParametersAsPlainTextAsync(IEnumerable<ICapturedParameters> parameters, Stream outputStream, CancellationToken token)
```

src/Microsoft.Diagnostics.Monitoring.WebApi/ParameterCapturing/CapturedParametersFormatter.cs
```diff
@@ -0,0 +1,116 @@
...
+            CapturedParametersResult result = BuildResultModel(parameters);
+
+            await using StreamWriter writer = new StreamWriter(outputStream, Encoding.UTF8, leaveOpen: true);
+            await writer.WriteLineAsync("Captured Parameters:");
+
+            StringBuilder builder = new();
+
+            foreach (CapturedMethod capture in result.CapturedMethods)
+            {
+                builder.AppendLine($"[{capture.CapturedDateTime}][thread {capture.ThreadId}] {GetValueOrUnknown(capture.ActivityId)}[format: {capture.ActivityIdFormat}]");
```
schmittjoseph, line 53: All of these interpolated strings should be using `FormattableString.Invariant` so we don't have unexpected formatting from the current culture provider

src/Tests/Microsoft.Diagnostics.Monitoring.HostingStartup.UnitTests/ParameterCapturing/MethodSignatureTests.cs
```diff
@@ -0,0 +1,280 @@
...
+using Xunit;
+
+namespace Microsoft.Diagnostics.Monitoring.HostingStartup.UnitTests.ParameterCapturing
+{
+    [TargetFrameworkMonikerTrait(TargetFrameworkMonikerExtensions.CurrentTargetFrameworkMoniker)]
+    public class MethodSignatureTests
+    {
+        [Theory]
+        [MemberData(nameof(GetMethodSignatureTestData))]
+        public void MethodSignature(
```
schmittjoseph, line 19: IMO the `[MemberData(nameof(GetMethodSignatureTestData))]` is harder to identify the cases being tested, even looking at the method it takes some mental overhead to figure out what each data set is trying to test.

Let's refactor this method into one called `TestCore` that does the AAA, and each set of data into a `[Fact]` test method with a descriptive name of what is being tested

src/Tools/dotnet-monitor/ParameterCapturing/EventParameterCapturingPipeline.cs
```diff
@@ -27,6 +30,8 @@ internal sealed class ServiceStateUpdateArgs
 
     internal sealed class EventParameterCapturingPipeline : EventSourcePipeline<EventParameterCapturingPipelineSettings>
     {
+        private readonly EventParameterCapturingPipelineCache _cache = new();
```
schmittjoseph, line 33: I don't think a cache is the right concept here -- it has a few undesired behaviors such as:
- Unbounded memory growth when dealing with long running parameter capturing sessions
- Possible data inconsistency, e.g. let's say the sender is buggy and somehow we receive an `CapturedParameter` event with a requestId that corresponds to a previously seen `CapturedParameter/Stop` event. We'd be updated an already stopped request.


My suggestion:
The cache should become more of a builder (`CapturedParametersBuilder` or something similar) that lets you:
- Start building a new set of captured parameters for a specific `requestId`.
- Add parameters to an existing set of captured parameters via requestId`.
- Finalize a set of captured parameters, returning the fully built set and removing the builders knowledge of it to help bound memory usage.


It's pretty similar to what you have now, just with the notion of finalizing sets of parameters.

src/Tools/dotnet-monitor/ParameterCapturing/EventParameterCapturingPipelineCache.cs
```diff
@@ -0,0 +1,57 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Diagnostics.Monitoring.WebApi.ParameterCapturing;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Microsoft.Diagnostics.Tools.Monitor.ParameterCapturing
+{
+    internal sealed class EventParameterCapturingPipelineCache
```
schmittjoseph, line 11: See my comment in the `EventParameterCapturingPipeline`, this should become more of a builder-style class.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/ParameterCapturingTests.cs
```diff
@@ -131,23 +130,52 @@ public async Task StopsProducingLogStatementsAfterOperationCompletes(Architectur
...
         }
-#else // NET7_0_OR_GREATER
+
         [Theory]
         [MemberData(nameof(ProfilerHelper.GetArchitecture), MemberType = typeof(ProfilerHelper))]
-        public async Task Net6AppFailsOperation(Architecture targetArchitecture)
+        public async Task CapturesParameters(Architecture targetArchitecture)
         {
-            await RunTestCaseCore(TestAppScenarios.ParameterCapturing.SubScenarios.AspNetApp, targetArchitecture, async (appRunner, apiClient) =>
+            await RunTestCaseCore(TestAppScenarios.ParameterCapturing.SubScenarios.DoNotExpectLogStatement, targetArchitecture, async (appRunner, apiClient) =>
```
schmittjoseph, line 138: See my above comment. Given the size of this PR I'm fine if you want to revert changes to these files, skip the tests, and rework these tests in a seperate PR.

schmittjoseph (changes requested):
See earlier comments

src/Tools/dotnet-monitor/ParameterCapturing/EventParameterCapturingPipeline.cs
schmittjoseph: Thinking about this further, the way you currently have `AsyncParameterCapturingEventSource` written there will only ever be 1 set of captured parameters being written to the event source at a time (e.g. no interweaving of different sets of parameters). This simplifies the requirements here quite a bit as we don't need to store multiple requests at a time. 

We really only need to store a single `CapturedParameters` and Guid for the active `requestId` at any given time.  If we ever receive `CapturedParameter` for a non-active `requestId` we can just throw away that event data, and if we ever receive a `CapturedParameter/Start` for a different `requestId` while another one is still active we can throw out the current request being built.

With all that said, I'm fine either way here and my earlier suggestions for `CapturedParametersBuilder` would be more versatile with smaller changes to your PR so I'll defer to you on which approach you want to take.

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/MethodSignature.cs
clguiman: I can't seem to find explicit documentation, but auto-implementation initialization should be called only once. You can see the backing field initialization with ILSpy if you use C# 5.0 decompilation or look directly at the IL code.

src/Tools/dotnet-monitor/ParameterCapturing/EventParameterCapturingPipeline.cs
clguiman: I tried to follow the same pattern as in the exceptions operations: the factory is transient and the store is scoped to the request (see `ServiceCollectionExtensions.ConfigureParameterCapturing`), so because the pipeline's lifetime is tied to the `CaptureParametersOperation`, the cache will never grow unbounded and it will get destroyed once the request completes.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/ParameterCapturingTests.cs
clguiman: This test verifies the capture output.
I kept or modified the loging tests because we technically still have access to an ILogger object and we might accidently log. I was planning on removing them once we remove the HostingStartup logic completly.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/ParameterCapturingTests.cs
```diff
@@ -80,28 +83,24 @@ public async Task NonAspNetAppFailsOperation(Architecture targetArchitecture)
...
+                Assert.Equal(HttpStatusCode.BadRequest, validationException.StatusCode);
 
                 await appRunner.SendCommandAsync(TestAppScenarios.ParameterCapturing.Commands.Continue);
             });
         }
 
         [Theory]
         [MemberData(nameof(ProfilerHelper.GetArchitecture), MemberType = typeof(ProfilerHelper))]
-        public async Task DoesProduceLogStatements(Architecture targetArchitecture)
+        public async Task DoesNotProduceLogStatements(Architecture targetArchitecture)
```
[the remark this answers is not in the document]
clguiman: See my comment below; I think there's still some value in having tests that confirm we don't accidently log. We'll eventually remove them with the HostingStartup cleanup.

src/Tools/dotnet-monitor/ParameterCapturing/EventParameterCapturingPipeline.cs
schmittjoseph: > so because the pipeline's lifetime is tied to the CaptureParametersOperation, the cache will never grow unbounded and it will get destroyed once the request completes.

It will grow unbounded though given a parameter capturing session with no capture limit, as each time a method with probes in encountered we grow our cache here with no way of freeing that memory until the session is done. The difference between this and the exceptions operation is the exception operation needs to maintain a cache for type names (which is inherently bounded by the number of types in the target application). We don't need to maintain data about prior parameter sets so we shouldn't continue to store that data after it's useful.

clguiman: Sure, I can implement your proposal, but we'll still grow the memory in an unbounded manner on a higher layer, in `ParameterCapturingStore`

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/ParameterCapturingTests.cs
schmittjoseph: > I kept or modified the loging tests because we technically still have access to an ILogger object and we might accidently log.

Sure but we technically have access to everything inside the process so I'm still not sure what we're gaining here -- we don't do anything with the ILogger in the probes so this test isn't doing anything of value.  Given the fact that we're planning on removing the ILogger altogether and this PR is already a bit too unweidly in size to properly review let's just disable these tests and rework them in a separate PR

src/Tools/dotnet-monitor/ParameterCapturing/EventParameterCapturingPipeline.cs
schmittjoseph: >but we'll still grow the memory in an unbounded manner on a higher layer, in ParameterCapturingStore

And that's also a problem :) `ParameterCapturingStore` should be bounded (I'd argue we shouldn't have it in the first place but I'll leave a separate comment for that).

src/Tools/dotnet-monitor/ParameterCapturing/CaptureParametersOperation.cs
```diff
@@ -114,6 +141,7 @@ public async Task ExecuteAsync(CancellationToken token)
                 settings.OnCapturingFailed += OnCapturingFailed;
                 settings.OnServiceStateUpdate += OnServiceStateUpdate;
                 settings.OnUnknownRequestId += OnUnknownRequestId;
+                settings.OnParametersCaptured += OnParametersCaptured;
```
schmittjoseph, line 144: We shouldn't use a store here to buffer all the parameters as it will:
- Lead to unbounded memory growth for infinite duraiton sessions
- Not write out data as its available like all of the other artifacts that are on-demand

Ideally when we receive a new set of parameters we immediately format them and write them to the output stream, letting the caller deal with memory restrictions, eviction policies, etc. 

I'm assuming this is modeled after how exceptions does this. Exceptions (and metrics) is a bit different from the other artifacts as it is continuously collecting data unprompted and also needs to do correlations between the pieces of data as new information is learned (e.g. inner exceptions).

Parameters are on-demand (and now) just like logs and traces, so the user has to request them and has to actively listen for the results, or results are getting sent to an egress provider. Because of this there's no need to cache data on dotnet-monitors side.

documentation/api/README.md
schmittjoseph: @clguiman We discussed this offline and we decided on not mentioning breaking changes in our documentation for experimental features (i.e. omit `breaking change in 9.0 Preview 3` here, and don't make mention of how things used to work before this change)

src/Tools/dotnet-monitor/ParameterCapturing/CaptureParametersOperation.cs
clguiman: I refactored the code to use a `CapturedParametersWriter` instead of the store. The writer queues the captured methods on a separate channel when the `OnParametersCaptured` event gets triggered and on a separate loop the data is written to the stream through the abstract `CapturedParametersFormatter` (has two implementations: `CapturedParametersTextFormatter` and `CapturedParametersJsonFormatter`)
I also introduced the `NewlineDelimitedJson` result format since it might be easier to parse if we expect data to come out in chunks.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/ParameterCapturingTests.cs
clguiman: I reverted the tests and will refactor them in a separate PR

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/ParameterCapturingTests.cs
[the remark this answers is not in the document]
clguiman: Reverted the tests

src/Tools/dotnet-monitor/ParameterCapturing/EventParameterCapturingPipelineCache.cs
clguiman: I replaced the cache with `CapturedParametersBuilder`

src/Tools/dotnet-monitor/ParameterCapturing/CaptureParametersOperation.cs
schmittjoseph: > I also introduced the NewlineDelimitedJson result format since it might be easier to parse if we expect data to come out in chunks.

We should mirror how `/logs` work here for consistency instead of rolling something different. `/logs` produces json sequence (or ndjson), see this file for how it does the formatting:

https://github.com/dotnet/dotnet-monitor/blob/8b4c2545bdd5abdfdc16a42abbfdb0c32554ed4e/src/Microsoft.Diagnostics.Monitoring.WebApi/StreamingLogger.cs#L98

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -564,8 +564,9 @@ public ActionResult<DotnetMonitorInfo> GetInfo()
         }
 
         [HttpPost("parameters", Name = nameof(CaptureParameters))]
-        [ProducesWithProblemDetails(ContentTypes.ApplicationJson)]
+        [ProducesWithProblemDetails(ContentTypes.ApplicationNdJson, ContentTypes.ApplicationJson, ContentTypes.TextPlain)]
```
schmittjoseph, line 567: Similar to my earlier comment, we shouldn't offer standard json here, only `ContentTypes.ApplicationNdJson`, `ContentTypes.ApplicationJsonSequence`, and `ContentTypes.TextPlain` to be consistent with our other APIs.

documentation/api/parameters.md
```diff
@@ -1,12 +1,12 @@
...
 
-# Parameters - Post (experimental feature)
+# Parameters (experimental feature)
 
-Captures parameters for one or more methods each time they are called. Parameters are logged inside the target application using its [`ILogger`](https://learn.microsoft.com/dotnet/api/microsoft.extensions.logging.ilogger).
+Captures parameters for one or more methods each time they are called.
 
 > [!NOTE]
-> Unlike other artifacts, parameters do **not** support being sent to an egress provider.
+> Before version 9.0 Preview 3, parameters were logged inside the target application using its [`ILogger`](https://learn.microsoft.com/dotnet/api/microsoft.extensions.logging.ilogger).
```
kkeirstead, line 9:
```suggestion
> Before version 9.0 Preview 4, parameters were logged inside the target application using its [`ILogger`](https://learn.microsoft.com/dotnet/api/microsoft.extensions.logging.ilogger).
```

src/Tools/dotnet-monitor/ParameterCapturing/CapturedParametersWriter.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        private async Task WriterLoop()
+        {
+            await _formatter.StartAsync(_cancellationToken);
+
+            while (await _parameters.Reader.WaitToReadAsync(_cancellationToken))
+            {
+                if (_parameters.Reader.TryRead(out ICapturedParameters parameter))
+                {
+                    await _formatter.WriteParameters(parameter, _cancellationToken);
```
schmittjoseph, line 54: It seems that if an exception were to be thrown here we'd never propagate up that exception nor recover while continuing to buffer up parameters to be egressed.

I'd recommend we expose the writer task outside of this class and make the `CaptureParametersOption` wait for either the `_capturingStoppedCompletionSource` OR this writer task to complete and propagate up any errors. That way if something goes wrong we report the error and stop the operation.

src/Tests/Microsoft.Diagnostics.Monitoring.HostingStartup.UnitTests/ParameterCapturing/MethodSignatureTests.cs
```diff
@@ -0,0 +1,262 @@
...
+                    {
+                        Name = "i",
+                        Type = "System.Int32",
+                        Attributes = ParameterAttributes.None,
+                        IsByRef = false
+                    },
+                ]);
+
+        private static void TestCore(
+                                                                    MethodInfo methodInfo,
```
kkeirstead, line 221: At least in the GitHub view, the indentation seems off here - might just be me

clguiman: I think it's Github, it looks ok on my computer. Having said that, I did update the formatting a bit by moving the method call on a new line.

documentation/api/parameters.md
clguiman: Oops, I forgot about that. I've completly removed the note since the team decided not to mention the old behavior in documentation.

src/Tools/dotnet-monitor/ParameterCapturing/CapturedParametersWriter.cs
clguiman: That's a good catch! Please let me know if my implementation is what you had in mind.

src/Tools/dotnet-monitor/ParameterCapturing/CaptureParametersOperation.cs
```diff
@@ -160,7 +160,14 @@ public async Task ExecuteAsync(Stream outputStream, CancellationToken token)
                     token);
 
                 await _capturingStartedCompletionSource.Task.WaitAsync(token).ConfigureAwait(false);
-                await _capturingStoppedCompletionSource.Task.WaitAsync(token).ConfigureAwait(false);
+
+                /* parametersWriter.WaitAsync() successfully completes only after parametersWriter.DisposeAsync() is called.
+                   It will only complete earlier if there's a fault and in that case we want to catch the exception and stop capturing.
+                */
+                await Task.WhenAny([
```
schmittjoseph, line 167: Do we need the `.WaitAsync()` on the tasks here? My understanding is that that we can omit those?

We should also `.Unwrap()` the return value here allow any exceptions from the returning task to properly propagate up (that or capture the task returning from this line and explicitly await it)

e.g. `await Task.WhenAny([.., ...]).Unwrap().ConfigureAwait(false)`

schmittjoseph (dismissed):
Signing off on the capture parameter operation <-> hosting startup code. @jander-msft  or @wiktork can you review the HTTP api json output formatting portion of this?

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/CapturedParametersResults.cs
```diff
@@ -0,0 +1,65 @@
...
+        public string ActivityId { get; set; }
+
+        [JsonConverter(typeof(JsonStringEnumConverter))]
+        [JsonPropertyName("activityIdFormat")]
+        public ActivityIdFormat ActivityIdFormat { get; set; }
+
+        [JsonPropertyName("threadId")]
+        public int ThreadId { get; set; }
+
+        [JsonPropertyName("dateTime")]
```
jander-msft, line 50:
```suggestion
        [JsonPropertyName("timestamp")]
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/CapturedParametersResults.cs
```diff
@@ -0,0 +1,65 @@
...
+        public DateTime CapturedDateTime { get; set; }
+
+        [JsonPropertyName("module")]
```
jander-msft, line 53:
```suggestion
        [JsonPropertyName("moduleName")]
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/CapturedParametersResults.cs
```diff
@@ -0,0 +1,65 @@
...
+        public string ModuleName { get; set; }
+
+        [JsonPropertyName("type")]
```
jander-msft, line 56:
```suggestion
        [JsonPropertyName("typeName")]
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/CapturedParametersResults.cs
```diff
@@ -0,0 +1,65 @@
...
+        public string TypeName { get; set; }
+
+        [JsonPropertyName("method")]
```
jander-msft, line 59:
```suggestion
        [JsonPropertyName("methodName")]
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/CapturedParametersResults.cs
```diff
@@ -0,0 +1,65 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi.Models
+{
+    public class CapturedParameter
+    {
+        [JsonPropertyName("name")]
```
jander-msft, line 13:
```suggestion
        [JsonPropertyName("parameterName")]
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/CapturedParametersResults.cs
```diff
@@ -0,0 +1,65 @@
...
+        public string Name { get; set; }
+
+        [JsonPropertyName("type")]
```
jander-msft, line 16:
```suggestion
        [JsonPropertyName("typeName")]
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/CapturedParametersResults.cs
```diff
@@ -0,0 +1,65 @@
...
+        public string Type { get; set; }
+
+        [JsonPropertyName("module")]
```
jander-msft, line 19:
```suggestion
        [JsonPropertyName("moduleName")]
```

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/CapturedParametersResults.cs
```diff
@@ -0,0 +1,65 @@
...
+        public string TypeModuleName { get; set; }
+
+        [JsonPropertyName("value")]
+        public string Value { get; set; }
+
+        [JsonPropertyName("isIn")]
+        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
+        public bool IsInParameter { get; set; }
```
jander-msft, line 27: What's the usage of these three `In*` properties? While these cover most of the common modifiers, there are others such as `params`, `ref readonly`, and `scoped` that are not captured. There's also nullability and default values too.

I'm wondering if these should be left off (for JSON serialization, I'm fine with the text serialization) until they can be fully captured and uniformly described.

documentation/api/parameters.md
```diff
@@ -123,23 +111,62 @@ Authorization: Bearer fffffffffffffffffffffffffffffffffffffffffff=
 ### Sample Response
 
 ```http
-HTTP/1.1 202 Accepted
+HTTP/1.1 200 OK
 Content-Type: application/json
```
jander-msft, line 115: This sample is no longer valid, correct? It doesn't return `application/json` anymore and there is not `captures` property on the root object that I can find.

clguiman: Oops, I forgot to update.

schmittjoseph (approved):
Re-signing off on the capture parameter operation <-> hosting startup code. @jander-msft can you re-review the HTTP api json output formatting portion of this?

jander-msft (approved):
API surface LGTM

PR #6316, opened 2024-03-26
Add command sets to profiler communication channel

schmittjoseph:
###### Summary

This PR upgrades the profiler channel's messages to now contain a `command set` field to help distinguish between different receivers of the IPC message. There are currently 3 distinct command sets: `ServerResponse`, `Profiler`, and `StartupHook` (currently used by the HostingStartup, but this will become StartupHook-only in the future with the parameter capturing upgrades).

This change also allows multiple instances of `MonitorMessageDispatcher` to exist, where each is subscribed to a different command set, though this isn't done in this PR.

New IPC message format:
| Field | Size (bytes) |
| ----- | ----- |
| Command set| 2 |
| Command | 2 |
| Payload size | 4 |
| Payload | `%Payload size%` |

Other changes in this PR:
- Update `Command` to be `unsigned`.
- Update the native code to use fixed width integer types when handling the IPC messages instead of the portable types. Portable types don't guarantee the actual size of the integer, just the minimum size so it'd be technically possible for the headers buffer to be incorrectly sized on certain platforms.


src/Profilers/MonitorProfiler/MainProfiler/MainProfiler.cpp
```diff
@@ -247,28 +246,22 @@ HRESULT MainProfiler::InitializeCommandServer()
 
 HRESULT MainProfiler::MessageCallback(const IpcMessage& message)
 {
-    m_pLogger->Log(LogLevel::Debug, _LS("Message received from client %d"), message.Command);
+    m_pLogger->Log(LogLevel::Debug, _LS("Message received from client %hu:%hu"), message.CommandSet, message.Command);
 
-    switch (message.Command)
+    if (message.CommandSet == static_cast<unsigned short>(CommandSet::Profiler))
```
jander-msft, line 251: Instead of special casing the profiler command set, could it be registered as a callback in `g_managedMessageCallbacks`? This would eliminate the specialization and possibly allow for consistent testing if unit testing were applied.

I would then suggest renaming `ManagedMessageCallbackManager` to `MessageCallbackManager` and `ManagedMessageCallback` to `MessageCallback`.

src/Profilers/MonitorProfiler/Communication/ManagedMessageCallbackManager.h
```diff
@@ -0,0 +1,24 @@
...
+
+#pragma once
+
+#include <unordered_map>
+#include <mutex>
+#include "cor.h"
+#include "corprof.h"
+#include "Messages.h"
+
+typedef INT32 (STDMETHODCALLTYPE *ManagedMessageCallback)(UINT16, const BYTE*, UINT64);
```
jander-msft, line 12:
```suggestion
typedef HRESULT (STDMETHODCALLTYPE *ManagedMessageCallback)(UINT16, const BYTE*, UINT64);
```
nit: I know this returned a `INT32` prior to this change, but it probably should be `HRESULT`. I know when it all gets resolved it ends up basically being the same, but all of the implementers are return `HRESULT` and the caller is returning it as an `HRESULT`.

src/Microsoft.Diagnostics.Monitoring.WebApi/ProfilerMessage.cs
```diff
@@ -7,28 +7,48 @@
 
 namespace Microsoft.Diagnostics.Monitoring
 {
-    internal enum IpcCommand : short
+    internal enum CommandSet : ushort
+    {
+        DotnetMonitor,
```
jander-msft, line 12: nit: `ServerResponse` to be clear what the intended use of the commands within the command set.

src/Microsoft.Diagnostics.Monitoring.WebApi/ProfilerMessage.cs
```diff
@@ -7,28 +7,48 @@
...
+        Profiler,
+        ManagedInProc
```
jander-msft, line 14: nit: `StartupHook` to be more clear about what is providing the command set.

src/Profilers/MonitorProfiler/Communication/Messages.h
```diff
@@ -5,15 +5,26 @@
...
+enum class DotnetMonitorCommand : unsigned short
+{
+    Status
+};
+
+enum class CommandSet : unsigned short
+{
+    DotnetMonitor,
+    Profiler,
+    ManagedInProc
```
jander-msft, line 22: I believe that this isn't reference anywhere in the native code but I understand that this is probably a placeholder so that the value isn't accidentally taken by a new command set. I would suggest adding a comment to that effect.

src/Profilers/MonitorProfiler/MainProfiler/MainProfiler.cpp
schmittjoseph: I initially considered that, however doing so introducing a new set of considerations:

The `ManagedMessageCallbackManager` instance needs to be global so that PINVOKEs can access it. If we then place a callback in it for the profiler we'd effectively create a global reference to the `MainProfiler` instance. 

We could try to clear out the profiler's callback from `MessageCallbackManager` when we want to detach the profiler (or under some other stop conditions) but I'm personally not aware of all of the conditions when we'd want to do so. Having this special casing for the profiler command set was the most straight forward. Thoughts?

jander-msft: > We could try to clear out the profiler's callback from MessageCallbackManager when we want to detach the profiler (or under some other stop conditions) but I'm personally not aware of all of the conditions when we'd want to do so.

We currently don't detach the profiler (although, we probably should, but that would a complete separate set of work that I would want us to wholistically think about), so I'm not too concerned about that.

There is a shutdown notion already where that can be plugged into: https://github.com/dotnet/dotnet-monitor/blob/main/src/Profilers/MonitorProfiler/MainProfiler/MainProfiler.cpp#L48

I'm okay with it as-is given that's how it was originally and there's nothing technically wrong with it. I'm usually an advocate for generalizing code when there's opportunity.
 
_Aside_
Ideally, the command server and handling would be completely separated from the profiler into its own dll/so such that any other dll/so can use it as a singleton, but that's well beyond the scope of this change. However, making this change (de-specialization of the command set) would move it slightly in that direction.

wiktork (approved):

jander-msft (approved):

PR #6330, opened 2024-03-27
ParameterCapturing: Use the MonitorExecutionContextTracker to skip capturing dotnet-monitor calls.

clguiman:
This is work that will be used further in #6272 for not probing internal parameter capturing infrastructure.

* Moved `MonitorExecutionContextTracker` in the `Microsoft.Diagnostics.Monitoring.StartupHook` namespace.
* `FunctionProbesStub.EnterProbeStub()` now checks if it's called from inside the monitor execution context and skips forwarding the call.
* Add a monitor scope in `AbstractMonitorEventSource.FlushTimerTick()`.


src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Eventing/ParameterCapturingEventSource.cs
```diff
@@ -82,6 +83,9 @@ public void FailedToCapture(Guid RequestId, Exception ex)
         [Event(ParameterCapturingEvents.EventIds.Flush)]
         protected override void Flush()
         {
+            // This method is called on a timer and we shouldn't track any calls inside so we don't flood the user with unimportant captures.
+            using IDisposable _ = MonitorExecutionContextTracker.MonitorScope();
```
schmittjoseph, line 87: Let's add a test to enforce validate this new behavior.

This unit test should give a pretty good example of where / how to test it.:

https://github.com/dotnet/dotnet-monitor/blob/d7bb04ff9bba038c37d67bde9dfb5d95331e7f99/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/FunctionProbes/FunctionProbesScenario.cs#L320-L333


You might need to update the StartupHook project's `InternalsVisibleTo` to include the `UnitTestApp` project and you'll want to add a new entry around here to get the test to run: https://github.com/dotnet/dotnet-monitor/blob/d7bb04ff9bba038c37d67bde9dfb5d95331e7f99/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/FunctionProbes/FunctionProbesScenario.cs#L57

schmittjoseph: Looks like I commented on the wrong file, I meant to comment on FunctionprobesStub.cs

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/FunctionProbes/FunctionProbesStub.cs
```diff
@@ -28,6 +29,12 @@ public static void EnterProbeStub(ulong uniquifier, object[] args)
                 return;
             }
 
+            // Don't probe our own code.
+            if (MonitorExecutionContextTracker.IsInMonitorContext())
```
schmittjoseph, line 33: We'll want to move this check after `s_inProbe = true`  since the call to `MonitorExecutionContextTracker.IsInMonitorContext()` could indirectly call into another function with probes in it.

schmittjoseph (dismissed):

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/Eventing/ParameterCapturingEventSource.cs
```diff
@@ -82,6 +83,9 @@ public void FailedToCapture(Guid RequestId, Exception ex)
         [Event(ParameterCapturingEvents.EventIds.Flush)]
         protected override void Flush()
         {
+            // This method is called on a timer and we shouldn't track any calls inside so we don't flood the user with unimportant captures.
```
schmittjoseph, line 86: Let's actually move this to `AbstractMonitorEventSource.FlushTimerTick` so that all of our in-proc event sources pick this up.

Other LGTM

schmittjoseph (approved):

Issue #4606, opened 2023-06-09
Unify trace profiles between dotnet-trace and dotnet-monitor


jander-msft:
With a unified trace profile infrastructure that both dotnet-trace and dotnet-monitor can leverage, both tools will contain the same subset of default trace profile names and capabilities. This will allow users of one tool to easily capture the same type of trace from the other tool when switch tools or configurations.

PR #6348, opened 2024-03-28
Support collecting GC collect only trace using a simplified profile

### LINKED ISSUE #4605: Add gc-collect trace profile
Same as dotnet-collect's [gc-collect](https://github.com/dotnet/diagnostics/blob/main/src/Tools/dotnet-trace/CommandLine/Commands/ListProfilesCommandHandler.cs#L65) profile, offer the same capability in dotnet-monitor via the `GET /trace` route and `CollectTrace` actions.

cshung:
###### Summary
Fixes #4605

To make collecting gc-collect only traces easier with dotnet-monitor. I have made this change so that you can have a gc profile.

~Ideally, the class `GcCollectConfiguration` should be part of `Microsoft.Diagnostics.Monitoring.EventPipe.dll`. it will be after https://github.com/dotnet/diagnostics/pull/4568~

> Edit: Merged and consumed.

###### Release Notes Entry
Add `GcCollect` trace profile for collecting lightweight GC collection events, same as the `gc-collect` profile in the `dotnet-trace` tool.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
     [ProducesResponseType(typeof(void), StatusCodes.Status401Unauthorized)]
     public partial class DiagController : DiagnosticsControllerBase
     {
-        private const TraceProfile DefaultTraceProfiles = TraceProfile.Cpu | TraceProfile.Http | TraceProfile.Metrics;
+        private const TraceProfile DefaultTraceProfiles = TraceProfile.Cpu | TraceProfile.Http | TraceProfile.Metrics | TraceProfile.GcCollect;
 
         private readonly IOptions<DiagnosticPortOptions> _diagnosticPortOptions;
         private readonly IOptions<CallStacksOptions> _callStacksOptions;
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Models/TraceProfile.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Models/TraceProfile.cs
@@ -11,6 +11,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Models
         Cpu = 0x1,
         Http = 0x2,
         Logs = 0x4,
-        Metrics = 0x8
+        Metrics = 0x8,
+        GcCollect = 0x10,
     }
 }
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/TraceUtilities.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/TraceUtilities.cs
@@ -11,6 +11,30 @@ using System.Linq;
 
 namespace Microsoft.Diagnostics.Monitoring.WebApi
 {
+    internal sealed class GcCollectConfiguration : MonitoringSourceConfiguration
+    {
+        public GcCollectConfiguration()
+        {
+            RequestRundown = false;
+        }
+
+        public override IList<EventPipeProvider> GetProviders() =>            
+            new EventPipeProvider[] {
+                    new EventPipeProvider(
+                        name: "Microsoft-Windows-DotNETRuntime",
+                        eventLevel: System.Diagnostics.Tracing.EventLevel.Informational,
+                        // keywords: (long)ClrTraceEventParser.Keywords.GC
+                        keywords: 1
+                    ),
+                    new EventPipeProvider(
+                        name: "Microsoft-Windows-DotNETRuntimePrivate",
+                        eventLevel: System.Diagnostics.Tracing.EventLevel.Informational,
+                        // keywords: (long)ClrTraceEventParser.Keywords.GC
+                        keywords: 1
+                    )
+            };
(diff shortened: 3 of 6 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/TraceUtilities.cs
```diff
@@ -11,6 +11,30 @@
 
 namespace Microsoft.Diagnostics.Monitoring.WebApi
 {
+    internal sealed class GcCollectConfiguration : MonitoringSourceConfiguration
```
wiktork, line 14: This will be removed after https://github.com/dotnet/diagnostics/pull/4568 correct?

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/TraceUtilities.cs
```diff
@@ -44,6 +68,10 @@ public static MonitoringSourceConfiguration GetTraceConfiguration(Models.TracePr
 
                 configurations.Add(new MetricSourceConfiguration(options.GetIntervalSeconds(), defaultProviders));
             }
+            if (profile.HasFlag(Models.TraceProfile.Gc))
+            {
+                configurations.Add(new GcCollectConfiguration());
```
wiktork, line 73: IIRC our current merging logic relies on having mutually exclusive configurations across different profiles. Please double check that combined profiles work as expected when they share a provider (cpu + gc comes to mind since I believe they both use Microsoft-Windows-DotNETRuntime)

src/Microsoft.Diagnostics.Monitoring.WebApi/Controllers/DiagController.cs
```diff
@@ -33,7 +33,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Controllers
     [ProducesResponseType(typeof(void), StatusCodes.Status401Unauthorized)]
     public partial class DiagController : DiagnosticsControllerBase
     {
-        private const TraceProfile DefaultTraceProfiles = TraceProfile.Cpu | TraceProfile.Http | TraceProfile.Metrics;
+        private const TraceProfile DefaultTraceProfiles = TraceProfile.Cpu | TraceProfile.Http | TraceProfile.Metrics | TraceProfile.Gc;
```
wiktork, line 36: I have concerns about making this the default; how expensive/verbose is the data? Does it make sense to pair gc analysis together with cpu analysis?

cshung: GC Collect Only traces are typically very lightweight. It only emits a handful of events per GC. Especially so when compared with CPU traces, where the runtime typically emit an event with stack per millisecond.

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/TraceUtilities.cs
cshung: While the [diagnostics code](https://github.com/dotnet/diagnostics/blob/fe7763cdbd0ccbec8304887130bf3514a7f4f087/src/Microsoft.Diagnostics.Monitoring.EventPipe/Configuration/AggregateSourceConfiguration.cs#L22) says about de-duplicating the providers, it isn't strictly necessary.

The [runtime code](https://github.com/dotnet/runtime/blob/34d13b27a4f0d62367145bcc1f67f7044ad2328a/src/native/eventpipe/ep-config.c#L65) will perform the proper merging anyway, it will take the bitwise or for the keyword and take the max for the level. It needs to do that when there are multiple tools connected to it anyway.

And the combination of profiles works (in the sense that we get both types of events)

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/TraceProfile.cs
```diff
@@ -11,6 +11,7 @@ public enum TraceProfile
         Cpu = 0x1,
         Http = 0x2,
         Logs = 0x4,
-        Metrics = 0x8
+        Metrics = 0x8,
+        Gc = 0x10,
```
jander-msft, line 15:
```suggestion
        GcCollect = 0x10,
```
Let's match the name with the same profile name in the dotnet-trace tool: https://github.com/dotnet/diagnostics/blob/fe7763cdbd0ccbec8304887130bf3514a7f4f087/src/Tools/dotnet-trace/CommandLine/Commands/ListProfilesCommandHandler.cs#L65

There is a longer-term, there is a desire to make these exactly the same (https://github.com/dotnet/dotnet-monitor/issues/4606), but that should be a separate effort.

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/TraceUtilities.cs
cshung: The PR https://github.com/dotnet/diagnostics/pull/4568 is merged, but I still can't remove this class just yet. It appears that we will need to publish a new nuget package there that contains the new class before this can be removed.

jander-msft:
Could you add the new TraceProfile enum value to https://github.com/dotnet/dotnet-monitor/blob/main/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTestCommon/ActionTestsHelper.cs#L60 so that the `CollectTraceAction_ProfileSuccess` test gets some coverage for that profile?

PR #6456, opened 2024-04-15
Add issuer and expiration validation for API keys

jander-msft:
###### Summary

Enable validation for the `iss` and `exp` fields in the API key token. These fields are now required to be in the token when using API key authentication.

The `generatekey` command will produce a token with the `iss` field set to `https://github.com/dotnet/dotnet-monitor/generatekey+MonitorApiKey` (this is an existing behavior) and the `exp` field set to expire 7 days from when the token was generated (this is a new behavior). The expiration can be overridden with the `--expiration` option.

In the case where a customer is generating their own JWT keys to be used with .NET Monitor, they can override the expected issuer in configuration by setting `Authentication__MonitorApiKey__Issuer` to the value of the issuer that is expected. Setting this is optional; its default value is `https://github.com/dotnet/dotnet-monitor/generatekey+MonitorApiKey`.

This is a breaking change because both the `iss` and `exp` fields are now required and validated. This will not cause disruption to most usage of the API key authentication as it is expected that customers will use the `generatekey` command to create the appropriate settings and the key will have a lifetime of 7 days. Those customers who are creating their own API key must now set both `iss` and `exp` in the key and configure `Authentication__MonitorApiKey__Issuer` within .NET Monitor for the tool to accept the API key.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/OptionsExtensions.ApiKey.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/OptionsExtensions.ApiKey.cs
@@ -6,6 +6,7 @@ using Microsoft.Diagnostics.Monitoring.WebApi;
 using Microsoft.Diagnostics.Tools.Monitor;
 using Microsoft.IdentityModel.Tokens;
 using System;
+using System.Globalization;
 using System.IdentityModel.Tokens.Jwt;
 using System.Security.Claims;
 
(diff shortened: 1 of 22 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/Options/OptionsExtensions.ApiKey.cs
```diff
@@ -0,0 +1,68 @@
...
+    {
+        /// <summary>
+        /// Sets API Key authentication. Use this overload for most operations, unless specifically testing Authentication
+        /// </summary>
+        public static RootOptions UseApiKey(this RootOptions options, string algorithmName, Guid subject, out string token)
+        {
+            string subjectStr = subject.ToString("D");
+            Claim audClaim = new Claim(AuthConstants.ClaimAudienceStr, AuthConstants.ApiKeyJwtAudience);
+            long expirationSecondsSinceEpoch = EpochTime.GetIntDate(DateTime.UtcNow + AuthConstants.ApiKeyJwtDefaultExpiration);
+            Claim expClaim = new Claim(AuthConstants.ClaimExpirationStr, expirationSecondsSinceEpoch.ToString());
```
wiktork, line 24: nit: did you need an explicit formatter on the number to string conversion?

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -942,4 +942,11 @@
   <data name="ParameterCapturingNotAvailable_Reason_HotReload" xml:space="preserve">
     <value>The process has Hot Reload enabled.</value>
   </data>
+  <data name="ErrorMessage_ExpirationMustBePositive" xml:space="preserve">
+    <value>The expiration offset must be a positive time span value.</value>
+  </data>
+  <data name="HelpDescription_Expiration" xml:space="preserve">
+    <value>The expiration time on or after the generated key will no longer be accepted. This is a time span offset (e.g. "1:00:00" for one-hour) that will be added to the current date time to create the expiration date time.</value>
```
wiktork, line 949: nit: can you produce an example that uses days? I expect that to be the preferred amount.

wiktork (dismissed):

wiktork (approved):

PR #6458, opened 2024-04-15
Type visibility changes

schmittjoseph:
###### Summary

This PR adjusts some of the type visibilities. I've taken the approach of changing things at the file-level instead of trying to make micro adjustments on specific methods / properties for a given type.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/ProfilerChannel.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/ProfilerChannel.cs
@@ -4,6 +4,7 @@
 using Microsoft.Extensions.Options;
 using System;
 using System.Diagnostics;
+using System.Globalization;
 using System.IO;
 using System.Net.Sockets;
 using System.Runtime.InteropServices;
@@ -30,7 +31,13 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
         {
             if (message.Payload.Length > MaxPayloadSize)
             {
-                throw new ArgumentException(null, nameof(message));
+                throw new ArgumentException(
+                    string.Format(
+                        CultureInfo.InvariantCulture,
+                        Strings.ErrorMessage_ProfilerPayloadTooLarge,
+                        message.Payload.Length,
+                        MaxPayloadSize),
+                    nameof(message));
             }
 
             string channelPath = ComputeChannelPath(endpointInfo);
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/ProfilerChannel.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/ProfilerChannel.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
     /// <summary>
     /// Communicates with the profiler, using a Unix Domain Socket.
     /// </summary>
-    internal sealed class ProfilerChannel
+    public sealed class ProfilerChannel
     {
         private const int MaxPayloadSize = 4 * 1024 * 1024; // 4 MiB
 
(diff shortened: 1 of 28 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/ProfilerChannel.cs
```diff
@@ -30,7 +31,13 @@ public async Task SendMessage(IEndpointInfo endpointInfo, IProfilerMessage messa
         {
             if (message.Payload.Length > MaxPayloadSize)
             {
-                throw new ArgumentException(nameof(message));
+                throw new ArgumentException(
```
schmittjoseph, line 34: NOTE: When making this `class` public it appears to enable additional analyzers. Specifically an analyzer that detects if `ArgumentException` ctors are being used incorrectly (the first argument should be the message, NOT the argument name). I've found other instances of this issue throughout our code but that should be a separate cleanup effortt.

src/Microsoft.Diagnostics.Monitoring.WebApi/ProfilerMessage.cs
```diff
@@ -5,42 +5,47 @@
 using System.Text;
 using System.Text.Json;
 
+
+#if STARTUPHOOK || HOSTINGSTARTUP
```
schmittjoseph, line 9: NOTE: This namespace adjustment is a result of `Microsoft.Diagnostics.Monitoring.TestCommon` having a dependency on `Microsoft.Diagnostics.Monitoring.WebApi` meaning tests that have nothing to do with the web api / models are still having it pulled in.

I briefly looked into removing that dependency but it'd cause several additional changes and so that should be a separate effort.

wiktork: Have you considered the approach of refactoring public types to a specific assembly?

src/Tools/dotnet-monitor/IDiagnosticLifetimeService.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor
     /// <summary>
     /// Callbacks for the lifetime events related to the diagnostic connection.
     /// </summary>
-    internal interface IDiagnosticLifetimeService
```
wiktork, line 12: Is this the correct assembly for this type? Having references to the tool directly now may make this more complicated to refactor to in the future.

wiktork (commented):
Looks good, added a few comments/questions.

src/Microsoft.Diagnostics.Monitoring.WebApi/ProfilerMessage.cs
schmittjoseph: I have, unfortunately we aren't very consistent with where types live so things are pretty spread out between assemblies currently. I'd prefer that any refactoring around where types live be a separate, more conscious, effort to ensure it makes sense. We aren't locked into any of the current assemblies where public types are in -- we can refactor whenever we like without issue

src/Tools/dotnet-monitor/IDiagnosticLifetimeService.cs
schmittjoseph: We have a couple types here that we need to expose from this assembly (also a new one that will get added in a follow-up PR). 

Similar to my other comment: We aren't very consistent with where types live so things are pretty spread out between assemblies currently. I'd prefer that any refactoring around where types live be a separate, more conscious, effort to ensure it makes sense. We aren't locked into any of the current assemblies where public types are in -- we can refactor whenever we like without issue

wiktork (approved):

PR #6502, opened 2024-04-23
Add wrapper for `EventSourcePipeline` with increased visibility

schmittjoseph:
###### Summary

Add a new public `EventSourcePipelineProxy` class. It provides a simple wrapper around the dotnet/diagnostic's internal-only `EventSourcePipeline<T>` class. Changing the visibility of `EventSourcePipeline<T>` class should be a separate and more intentional piece of work. This wrapper will allow us to rapidly make progress and any needed changes in the interim.

### DIFF
--- a/src/Tools/dotnet-monitor/AbstractEventSourcePipeline.cs
+++ b/src/Tools/dotnet-monitor/AbstractEventSourcePipeline.cs
@@ -1,0 +1,57 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Diagnostics.Monitoring.WebApi;
+using Microsoft.Diagnostics.Monitoring.EventPipe;
+using System;
+using Microsoft.Diagnostics.NETCore.Client;
+using Microsoft.Diagnostics.Tracing;
+using System.Threading.Tasks;
+using System.Threading;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    public abstract class AbstractEventSourcePipeline : IAsyncDisposable
+    {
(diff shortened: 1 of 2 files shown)


src/Tools/dotnet-monitor/AbstractEventSourcePipeline.cs
```diff
@@ -0,0 +1,57 @@
...
+using Microsoft.Diagnostics.Monitoring.EventPipe;
+using System;
+using Microsoft.Diagnostics.NETCore.Client;
+using Microsoft.Diagnostics.Tracing;
+using System.Threading.Tasks;
+using System.Threading;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    public abstract class AbstractEventSourcePipeline : IAsyncDisposable
```
schmittjoseph, line 14: Chatted with @wiktork, will rename to `EventSourcePipelineBase` to follow the naming pattern we use elsewhere

jander-msft (approved):

src/Tools/dotnet-monitor/EventSourcePipelineProxy.cs
```diff
@@ -0,0 +1,57 @@
...
+using Microsoft.Diagnostics.Monitoring.EventPipe;
+using System;
+using Microsoft.Diagnostics.NETCore.Client;
+using Microsoft.Diagnostics.Tracing;
+using System.Threading.Tasks;
+using System.Threading;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    public abstract class EventSourcePipelineProxy : IAsyncDisposable
```
wiktork, line 14: Can you add more comments as to the motivation of this class here in the code?

PR #6548, opened 2024-04-29
ParameterCapturing: Remove quotes from string evaluations, include failure reason information make the value nullable

clguiman:
###### Summary

* Updated the captured parameter result so the value parameter and activity id are nullable strings.
* Introduced `ObjectFormatterResult` which includes the evaluated and formatted parameter value and evaluation metadata flags (`IsNull`, `FailedEval`, `UnsupportedEval`, `EvalHasSideEffects`). `ObjectFormatter now returns objects of the new type.
* `EventSourceEmittingProbes.EnterProbe` doesn't request adding quotes to the formatted value anymore. The quotes were added to easily parse the ILogger output, but we don't have that problem anymore.
* Updated `ParameterCapturingEventSource` to send the new evaluation flags as part of the captured paramater.
* Updated `WebApi.Models.CapturedParameter` to include a new optional property: `EvalFailReason`.
* Updated the formatter tests to support the new behavior.

### DIFF
--- a/documentation/api/definitions.md
+++ b/documentation/api/definitions.md
@@ -116,6 +116,8 @@ Object describing a captured parameter.
 | `value` | string | The parameter value. |
 | `typeName` | string | The parameter type name. |
 | `moduleName` | string | The parameter type module name. |
+| `evalFailReason` | string | The reason why evaluation failed. If missing the evaluation was successful. |
+| `isNull` | bool | Whether the evaluated parameter is null (value is omitted if false). |
 
 ## CaptureParametersConfiguration
 
--- a/documentation/api/definitions.md
+++ b/documentation/api/definitions.md
@@ -95,7 +95,7 @@ Object describing a captured method and its parameters.
 
 | Name | Type | Description |
 |---|---|---|
-| `activityId` | string | An identifier for the current activity at the time of the capture. For more information see [Activity.Id](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity.id).|
+| `activityId` | string? | An identifier for the current activity at the time of the capture. For more information see [Activity.Id](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity.id).|
 | `activityIdFormat` | string | The activity Id format. For more information see [Activity.IdFormat](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity.idformat).|
 | `threadId` | int | The managed thread id where the method was called.|
 | `timestamp` | DateTime | Time when the method call was captured. |
@@ -113,11 +113,10 @@ Object describing a captured parameter.
 | Name | Type | Description |
 |---|---|---|
 | `parameterName` | string | The parameter name. |
-| `value` | string | The parameter value. |
+| `value` | string? | The parameter value. |
 | `typeName` | string | The parameter type name. |
 | `moduleName` | string | The parameter type module name. |
 | `evalFailReason` | string | The reason why evaluation failed. If missing the evaluation was successful. |
-| `isNull` | bool | Whether the evaluated parameter is null (value is omitted if false). |
 
 ## CaptureParametersConfiguration
 
--- a/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ObjectFormatting/Formatters/DebuggerDisplay/ExpressionBinder.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ObjectFormatting/Formatters/DebuggerDisplay/ExpressionBinder.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Reflection;
 using static Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing.ObjectFormatting.Formatters.DebuggerDisplay.DebuggerDisplayParser;
+using static Microsoft.Diagnostics.Tools.Monitor.ParameterCapturing.ParameterCapturingEvents;
 
 namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing.ObjectFormatting.Formatters.DebuggerDisplay
 {
@@ -26,7 +27,7 @@ namespace Microsoft.Diagnostics.Monitoring.HostingStartup.ParameterCapturing.Obj
 
             if (debuggerDisplay.Expressions.Count == 0)
             {
-                return (_, _) => debuggerDisplay.FormatString;
+                return (_, _) => new() { FormattedValue = debuggerDisplay.FormatString, Flags = ParameterEvaluationFlags.None };
             }
 
             ExpressionEvaluator[] evaluators = new ExpressionEvaluator[debuggerDisplay.Expressions.Count];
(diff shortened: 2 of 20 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.HostingStartup.UnitTests/ParameterCapturing/ObjectFormatting/Formatters/DebuggerDisplay/DebuggerDisplayFormatterTests.cs
```diff
@@ -75,10 +84,31 @@ public void GetDebuggerDisplayFormatter_ReturnsWorkingFormatter()
...
-            string formattedResult = factoryResult.Formatter(testObj);
+            ObjectFormatterResult formattedResult = factoryResult.Formatter(testObj);
+
+            // Assert
+            Assert.Equal(FormattableString.Invariant($"'Count = {testObj.Count}'"), formattedResult.FormattedValue);
+            Assert.Equal(ParameterEvaluationFlags.None, formattedResult.Flags);
+        }
+
+        [Fact]
+        public void GetDebuggerDisplayFormatter_FormatsNull()
```
schmittjoseph, line 95: This test is really testing behavior of `ExpressionBinder` and not `GetDebuggerDisplayFormatter`. Let's update the `ExpressionBinderTests` instead with this new null case.

I'd recommend just adding a new `InlineData` for this test case (and updating `DebuggerDisplayClass` with a new method/property that is null.
https://github.com/dotnet/dotnet-monitor/blob/efbb575789e1137fb4416a33a3279ab64a26d38a/src/Tests/Microsoft.Diagnostics.Monitoring.HostingStartup.UnitTests/ParameterCapturing/ObjectFormatting/Formatters/DebuggerDisplay/ExpressionBinderTests.cs#L55

schmittjoseph: Nevermind, looks like I misread what got changed here!

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ObjectFormatting/Formatters/DebuggerDisplay/ExpressionBinder.cs
```diff
@@ -26,7 +27,7 @@ internal record struct ExpressionEvaluator(BoundEvaluator Evaluate, Type ReturnT
 
             if (debuggerDisplay.Expressions.Count == 0)
             {
-                return (_, _) => debuggerDisplay.FormatString;
+                return (_, _) => new() { FormattedValue = debuggerDisplay.FormatString, Flags = ParameterEvaluationFlags.None };
```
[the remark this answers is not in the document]
jander-msft: There's only one caller, so it should be able to take the formatted string and turn it into a `ObjectFormatterFunc`.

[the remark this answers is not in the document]
clguiman: I can't return a `string?` because the `ObjectFormatterFunc` takes in a `formatSpecifier` and the `object` as input and that's used to construct the final string (see lines 52 and 62).
I could return a new delegate type that takes an `object` and `FormatSpecifier` as input and returns a `string` instead of `ObjectFormatterResult`, but that seems like a very specific type that's only used in a single place.
What do you think?

documentation/api/definitions.md
```diff
@@ -116,6 +116,8 @@ Object describing a captured parameter.
 | `value` | string | The parameter value. |
 | `typeName` | string | The parameter type name. |
 | `moduleName` | string | The parameter type module name. |
+| `evalFailReason` | string | The reason why evaluation failed. If missing the evaluation was successful. |
+| `isNull` | bool | Whether the evaluated parameter is null (value is omitted if false). |
```
schmittjoseph, line 120: It seems like it might make more sense to remove this field and simply make the `value` field a nullable-string instead. If the value is null, then we use null literal to represent it.

clguiman: That makes sense. I've updated the code to make the parameter value and activityId nullable strings. Nullable types are not enabled on the project level, so I had to enable them in the files I touched and I also had to include a workaround for the `required` modifier in .NET 6

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ObjectFormatting/ObjectFormatter.cs
```diff
@@ -16,6 +17,19 @@ internal enum FormatSpecifier
         NoSideEffects = 2
     }
 
+    internal readonly struct ObjectFormatterResult
+    {
+        public string FormattedValue { get; init; }
+
+        public ParameterEvaluationFlags Flags { get; init; }
+
+        public static readonly ObjectFormatterResult Empty = new() { FormattedValue = string.Empty, Flags = ParameterEvaluationFlags.None };
```
schmittjoseph, line 26: Unused?

src/Tools/dotnet-monitor/ParameterCapturing/ParameterCapturingEvents.cs
```diff
@@ -88,8 +90,19 @@ public static class CapturedParameterPayloads
...
             public const int ParameterValue = 5;
-            public const int ParameterAttributes = 6;
-            public const int ParameterTypeIsByRef = 7;
+            public const int ParameterValueEvaluationFlags = 6;
+            public const int ParameterAttributes = 7;
+            public const int ParameterTypeIsByRef = 8;
+        }
+
+        [Flags]
+        public enum ParameterEvaluationFlags
```
schmittjoseph, line 99: It looks like we don't use more than 1 flag at a given time (i.e. they are exclusive), let's make this a regular enum instead?

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ObjectFormatting/ObjectFormatter.cs
```diff
@@ -16,6 +17,19 @@ internal enum FormatSpecifier
         NoSideEffects = 2
     }
 
+    internal readonly struct ObjectFormatterResult
+    {
+        public string FormattedValue { get; init; }
+
+        public ParameterEvaluationFlags Flags { get; init; }
```
schmittjoseph, line 24: Let's have a default value of `ParameterEvaluationFlags.None` instead of requiring everyone to set it? We have to explicitly set this flag to `None` in several places when IMO it should be the default value

src/Tools/dotnet-monitor/ParameterCapturing/CapturedParameters.cs
```diff
@@ -28,7 +30,7 @@ public void AddParameter(ParameterInfo parameter)
             _parameters.Add(parameter);
         }
 
-        public string ActivityId { get; }
+        public string? ActivityId { get; }
 
         public ActivityIdFormat ActivityIdFormat { get; }
```
schmittjoseph, line 35: Shouldn't this also be nullable now?

clguiman: The `ActivityIdFormat` enum has an `Unknown` value. From practice I noticed that's set when `ActivityId` is `null`

src/Microsoft.Diagnostics.Monitoring.HostingStartup/ParameterCapturing/ObjectFormatting/Formatters/DebuggerDisplay/ExpressionBinder.cs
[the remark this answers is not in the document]
clguiman: I updated the code on Joe's suggestion to default the eval result to `Success`, so there's no need to always set it.

schmittjoseph (approved):

Issue #6569, opened 2024-05-02
Upgrade to version 8 causes hanging of processes which dotnet monitor is supposed to be monitoring


jwreford99:
### Description
We recently upgraded our image from `mcr.microsoft.com/dotnet/monitor:7` (which I think was specifically on `mcr.microsoft.com/dotnet/monitor@sha256:0cda57851419ac60878166c29dd18f09915ca2047de60444847f91338db8a326`) to `mcr.microsoft.com/dotnet/monitor:8` (specifically `mcr.microsoft.com/dotnet/monitor@sha256:9324ee5e07e7ce8428772aa6269438b3a6688649d1746f41190acb0a045ff857`) in our production environment and some of the dotnet processes being run in the container which is supposed to be being monitored started emitting logs of the form:
```
The runtime has been configured to pause during startup and is awaiting a Diagnostics IPC ResumeStartup command from a Diagnostic Port.
DOTNET_DiagnosticPorts="/dotnet-diagnostics/port.sock"
DOTNET_DefaultDiagnosticPortSuspend=0
```
When the process emitted that log it then just hung in the `SNl` state, and we terminated it with an automatic cutoff for duration of time (see notes in configuration but these C# processes are supposed to be very short lived, which I suspect is where the issue stems from.)

At the same time there also seems to be a log message from the `dotnet-monitor` container saying:
```
{
  "Timestamp": "2024-05-02T08:39:56.8474094Z",
  "EventId": 99,
  "LogLevel": "Warning",
  "Category": "Microsoft.Diagnostics.Tools.Monitor.ServerEndpointInfoSource",
  "Message": "Unable to fully initialize endpoint for process 792382. Some monitoring features may be unavailable.",
[...]

PR #6570, opened 2024-05-02
Add semaphore around service scopes dictionary

jander-msft:
###### Summary

The `_activeEndpointServiceScopes` dictionary is largely protected from simultaneous mutation using the `_activeEndpointsSemaphore` semaphore. However, there is one place where this lock is not taken which is likely causing the issue reported in #6569. Add the semaphore usage so that the dictionary is fully protected.

###### Release Notes Entry
Fix an issue that could cause monitored processes to fail to resume when in listen mode: [#6569](https://github.com/dotnet/dotnet-monitor/issues/6569)


schmittjoseph (approved):

schmittjoseph:
Let's mention this in the release notes as a bug fix

PR #6655, opened 2024-05-16
Introduce BackgroundService utility class in StartupHook

clguiman:
###### Summary
As part of the move from HostingStartup to StartupHook we need a mechanism similar to `Microsoft.Extensions.Hosting.BackgroundService` to manage the lifetime of `ParameterCapturingService`.

This PR adds the `Microsoft.Diagnostics.Monitoring.StartupHook.BackgroundService` abstract class.  `ParameterCapturingService` can derive from it and `DiagnosticsBootstrapper` can instantiate and start the service. The integration will be part of a separate PR when I'll move all the files to the StartupHook project.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/BackgroundService.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/BackgroundService.cs
@@ -1,0 +1,55 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Diagnostics.Monitoring.StartupHook
+{
+    internal abstract class BackgroundService : IDisposable
+    {
+        private readonly CancellationTokenSource _cts = new();
+        private Task? _executeTask;
+        private long _disposedState;
+
+        public void Start()
+        {
+            _executeTask = Task.Run(async () =>
+            {
+                try
+                {
+                    await ExecuteAsync(_cts.Token);
+                }
+                catch (OperationCanceledException)
+                {
+                    // Ignore
+                }
+                catch (Exception ex)
+                {
+                    BackgroundTaskException = ex;
+                }
+            });
+        }
+
+        public void Stop()
+        {
+            _cts.Cancel();
+        }
+
+        public Exception? BackgroundTaskException { get; private set; }
+
+        public virtual void Dispose()
+        {
+            if (!DisposableHelper.CanDispose(ref _disposedState))
+                return;
+
+            _executeTask?.Wait();
+            _executeTask = null;
+
+            _cts.Dispose();
+        }
+
+        protected abstract Task ExecuteAsync(CancellationToken stoppingToken);
+    }
+}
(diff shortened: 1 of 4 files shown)


src/Microsoft.Diagnostics.Monitoring.StartupHook/BackgroundService.cs
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        public Exception? BackgroundTaskException { get; private set; }
+
+        public virtual void Dispose()
+        {
+            if (!DisposableHelper.CanDispose(ref _disposedState))
+                return;
+
+            _executeTask?.Wait();
```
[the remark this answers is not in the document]
clguiman: Is 1 second too much?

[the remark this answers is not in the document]
jander-msft: That's probably fine. We can adjust it later if need be.

src/Microsoft.Diagnostics.Monitoring.StartupHook/BackgroundService.cs
```diff
@@ -0,0 +1,50 @@
...
+                // ignore
+            }
+        }
+
+        public virtual void Dispose()
+        {
+            if (!DisposableHelper.CanDispose(ref _disposedState))
+                return;
+
+            _cts.Cancel();
```
schmittjoseph, line 44: Don't we also need to wrap `_cs.Cancel()` in a try/catch since it could throw? (same applies to `Stop`)

jander-msft: It should only throw if `_cts` is disposed, which it should be since there is the disposal guard on the previous lines.

schmittjoseph: `_cts.Cancel()` would throw if anything registers a delegate to be called when the corresponding token gets cancelled and it happens to throw.

e.g.

```csharp
  _cts.Token.Register(() =>
  {
      throw new InvalidOperationException("bug");
  });

  _cts.Cancel(); // <- throws 
```

Since we hand off `_cts`'s cancellation token to `abstract ExecuteAsync` it's possible that something like this could happen

clguiman: That's a good point. I think I should catch `AggregateException` (it wrapps any exception thrown from registered callbacks) and let `ObjectDisposedException` pass through. Any thoughts?

src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/BackgroundService/BackgroundServiceTests.cs
```diff
@@ -0,0 +1,129 @@
...
+    public sealed class BackgroundServiceTests
+    {
+        [Fact]
+        public void ConstructionWorks()
+        {
+            using BackgroundService _ = new MockBackgroundService();
+        }
+
+        [Fact]
+        public async Task StartBackgroundTask()
```
schmittjoseph, line 22: nit:

For test names, let's separate the name of the method from what aspect of it is being tested.

e.g.

- `Start_StartsBackgroundTask`
- `Stop_TriggersCancellation`
- ...

jander-msft (approved):

schmittjoseph (approved):

PR #6665, opened 2024-05-16
Fix ErrorMessage_CouldNotMapToAction usage

schmittjoseph:
###### Summary

While working on another feature I came across a bug where we don't use `ErrorMessage_CouldNotMapToAction` correctly - we weren't supplying the necessary argument to it.

###### Release Notes Entry
Fix the error log message shown when an action could not be registered.

### DIFF
--- a/src/Tools/dotnet-monitor/CollectionRules/ActionListExecutor.cs
+++ b/src/Tools/dotnet-monitor/CollectionRules/ActionListExecutor.cs
@@ -7,6 +7,7 @@ using Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options;
 using Microsoft.Extensions.Logging;
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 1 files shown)


src/Tools/dotnet-monitor/CollectionRules/ActionListExecutor.cs
```diff
@@ -83,7 +84,11 @@ public ActionListExecutor(ILogger<CollectionRuleService> logger, ICollectionRule
 
                         if (!_actionOperations.TryCreateFactory(actionOption.Type, out factory))
                         {
-                            throw new InvalidOperationException(Strings.ErrorMessage_CouldNotMapToAction);
+                            throw new InvalidOperationException(string.Format(
+                                CultureInfo.InvariantCulture,
```
schmittjoseph, line 88: NOTE: We're pretty inconsistent with what culture we use when formatting error messages (some places use `CultureInfo.CurrentCulture`, while others use the invariant culture). This should be a separate effort to standardize everything but for now I've opted to just use the invariant culture here.

jander-msft (approved):

jander-msft:
Do you think this needs to go into release notes? Also, please backport where applicable (presumably all in-support versions).

PR #6686, opened 2024-05-20
Prevent operations from getting stuck in the `Starting` state

### LINKED ISSUE #6683: Operations can indefinitely be stuck in `Starting` state
###### Current design
When an operation is first created it enters a `Starting` state and typically signals a task completion source once it has started and the operation transitions to a `Running` state. If an error occurs while running the operation transitions to the `Faulted` state.

###### Edge cases

1) If the operation encounters an error (i.e. throws an exception) before it has a chance to transition to the `Running` state, the operation will be stuck in the `Starting` state and never transition to `Faulted`.
2) If the operation has a bug and completes without ever signaling the started task completion source (e.g. it should transition from `Starting` -> `Completed`), the operation will be stuck in the `Starting` state.

schmittjoseph:
###### Summary

When an operation is first created it enters a `Starting` state and typically signals a task completion source once it has started and the operation transitions to a `Running` state. If an error occurs while running the operation transitions to the `Failed` state.

There are 2 edge cases that could cause an operation to get indefinitely stuck in the `Starting` state:
1) If the operation encounters an error (i.e. throws an exception) before it has a chance to transition to the `Running` state, the operation will be stuck in the `Starting` state and never transition to `Failed`.
2) If the operation has a bug and completes without ever signaling the started task completion source (e.g. it should transition from `Starting` -> `Completed`), the operation will be stuck in the `Starting` state.

This PR fixes these issues and adds test coverage for `EgressOperationService`. Since there previously was no test coverage, I've had to do some minor refactoring to enable easier testing.

Closes #6683 

###### Release Notes Entry
Fix an issue where operations could finish, or fail, but their state would be still be reported as `Starting` ([issue #6683](https://github.com/dotnet/dotnet-monitor/issues/6683)).

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.WebApi.UnitTests/Operation/EgressOperationServiceTests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.WebApi.UnitTests/Operation/EgressOperationServiceTests.cs
@@ -138,7 +138,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.UnitTests.Operation
                 throw testException;
             });
 
-            TaskCompletionSource<ExecutionResult<EgressResult>> egressResultCompletionSource = new();
+            TaskCompletionSource<ExecutionResult<EgressResult>> egressResultCompletionSource = new(TaskCreationOptions.RunContinuationsAsynchronously);
 
             _ = mockStore
                 .Setup(m => m.CompleteOperation(request.OperationId, It.IsAny<ExecutionResult<EgressResult>>()))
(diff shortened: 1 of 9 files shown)


schmittjoseph:
NOTE: `7.X.X` does not have the concept of the `Starting` operation state so this doesn't need to be backported there.

src/Tests/Microsoft.Diagnostics.Monitoring.WebApi.UnitTests/Operation/EgressOperationServiceTests.cs
```diff
@@ -0,0 +1,195 @@
...
+            EgressRequest request = CreateEgressRequest((startCompletionSource, token) =>
+            {
+                if (isStarted)
+                {
+                    _ = startCompletionSource.TrySetResult();
+                }
+                throw testException;
+            });
+
+            TaskCompletionSource<ExecutionResult<EgressResult>> egressResultCompletionSource = new();
```
jander-msft, line 141:
```suggestion
            TaskCompletionSource<ExecutionResult<EgressResult>> egressResultCompletionSource = new(TaskCreationOptions.RunContinuationsAsynchronously);
```

jander-msft (approved):

jander-msft (approved):

PR #6699, opened 2024-05-21
Validate IPC message commandset before acknowledging

schmittjoseph:
###### Summary

Currently when we send an IPC command via the profiler, we respond back with a status code that signals the message was received. However this doesn't account for the scenario where the profiler doesn't know how to handle the message's command set, meaning we would have acknowledged the message but then effectively discarded it without informing dotnet-monitor.

This PR updates the notify-only profiler to check if it knows how to handle a message before acknowledging it. If the message is unsupported, we reply with `E_NOT_SUPPORTED`.  In dotnet-monitor this will be thrown as:

> System.Runtime.InteropServices.COMException (0x80070032): The request is not supported. (0x80070032)

### DIFF
--- a/src/Profilers/MonitorProfiler/Communication/CommandServer.cpp
+++ b/src/Profilers/MonitorProfiler/Communication/CommandServer.cpp
@@ -13,7 +13,7 @@ CommandServer::CommandServer(const std::shared_ptr<ILogger>& logger, ICorProfile
 {
 }
 
-HRESULT CommandServer::Start(const std::string& path, std::function<HRESULT(const IpcMessage& message)> callback)
+HRESULT CommandServer::Start(const std::string& path, std::function<HRESULT(const IpcMessage& message)> callback, std::function<HRESULT(const IpcMessage& message)> validateMessage)
 {
     if (_shutdown.load())
     {
@@ -31,6 +31,7 @@ HRESULT CommandServer::Start(const std::string& path, std::function<HRESULT(cons
 #endif
 
     _callback = callback;
+    _validateMessage = validateMessage;
 
     IfFailLogRet_(_logger, _server.Bind(path));
     _listeningThread = std::thread(&CommandServer::ListeningThread, this);
@@ -75,7 +76,17 @@ void CommandServer::ListeningThread()
         if (FAILED(hr))
         {
             _logger->Log(LogLevel::Error, _LS("Unexpected error when receiving data: 0x%08x"), hr);
+            // Best-effort shutdown, ignore the result.
+            client->Shutdown();
             continue;
+        }
+
+        bool doEnqueueMessage = true;
+        hr = _validateMessage(message);
+        if (FAILED(hr))
+        {
+            _logger->Log(LogLevel::Error, _LS("Failed to validate message: 0x%08x"), hr);
+            doEnqueueMessage = false;
         }
 
         *reinterpret_cast<HRESULT*>(response.Payload.data()) = hr;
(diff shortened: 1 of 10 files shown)


src/Profilers/MonitorProfiler/Communication/MessageCallbackManager.cpp
```diff
@@ -27,7 +36,7 @@ bool MessageCallbackManager::TryRegister(unsigned short commandSet, std::functio
 
 HRESULT MessageCallbackManager::DispatchMessage(const IpcMessage& message)
 {
-    std::lock_guard<std::mutex> lock(m_mutex);
+    std::lock_guard<std::mutex> dispatchLock(m_dispatchMutex);
```
jander-msft, line 40: Shouldn't this also acquire a lock on `m_lookupMutex` since it is accessing the callback list? I wonder if the guard should be moved into `TryGetCallback` so then the callers don't have to worry about getting the lock.

jander-msft (approved):

src/Profilers/MonitorProfiler/Communication/MessageCallbackManager.cpp
schmittjoseph: So this is where a `shared_mutex` would make things more obvious as to the intent. 

I'm (poorly) replicating support for multiple readers by having 1 mutex for each distinct reader:
- 1 thread will be repeatedly calling `DispatchMessage`
- 1 thread will be repeatedly calling `IsRegistered`

These 2 operations need to be able to occur at the same time, since `DispatchMessage` can take a long time and we don't want to hold up the other thread that calls `IsRegistered`.

Meanwhile N threads can be calling into either `TryRegister` or `Unregister`. These are the "writers" and so they acquire both mutexes to guarantee exclusivity. 

----

> I wonder if the guard should be moved into TryGetCallback so then the callers don't have to worry about getting the lock.

`TryGetCallback` is a helper method used by both "readers" and "writers" which require holding the locks for different scopes so that's why we don't grab a lock inside of there.

schmittjoseph: So this is where a shared_mutex would make things more obvious as to the intent.

I'm (kinda) replicating support for multiple readers by having 1 mutex for each distinct reader:

1 thread will be repeatedly calling DispatchMessage
1 thread will be repeatedly calling IsRegistered
These 2 operations need to be able to occur at the same time, since DispatchMessage can take a long time and we don't want to hold up the other thread that calls IsRegistered.

Meanwhile N threads can be calling into either TryRegister or Unregister. These are the "writers" and so they acquire both mutexes to guarantee exclusivity.

I wonder if the guard should be moved into TryGetCallback so then the callers don't have to worry about getting the lock.

TryGetCallback is a helper method used by both "readers" and "writers" which require holding the locks for different scopes so that's why we don't grab a lock inside of there.

src/Profilers/MonitorProfiler/Communication/CommandServer.cpp
```diff
@@ -75,25 +79,39 @@ void CommandServer::ListeningThread()
...
 
+        bool doEnqueueMessage = true;
+        hr = _validateMessageCallback(message);
+        if (FAILED(hr))
+        {
+            _logger->Log(LogLevel::Error, _LS("Failed to validate message: 0x%08x"), hr);
+            doEnqueueMessage = false;
+        }
+
         *reinterpret_cast<HRESULT*>(response.Payload.data()) = hr;
```
wiktork, line 95: Do we show the HR response on the client logs for this? If not we probably should to make it easier to figure out failures.

wiktork (approved):

PR #6707, opened 2024-05-22
Move the ParameterCapturing code in StartupHook

clguiman:
###### Summary
* Moved the ParameterCapturing source files in the StartupHook project and updated the namespace to match the new location.
* The HostingStartup project now builds a stub assembly. I'll remove it and its test project in a separate PR.
* `StartupHook.DiagnosticsBootstrapper` now manages the lifetime of the `ParameterCapturingService`
* Replaced `ParameterCapturingTests.NonAspNetAppFailsOperation` with `ParameterCapturingTests.CapturesParametersInNonAspNetApps` since now we support parameter capturing in non-aspnet applications. In a separate PR I'll add a test to verify the failure when targeting an app that doesn't have startup hooks enabled.
* Updated `CaptureParametersOperation.EnsureEndpointCanProcessRequestsAsync` to replace the check for the hosting startup feature with a check for startup hook.


src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Microsoft.Diagnostics.Monitoring.UnitTestApp.csproj
```diff
@@ -13,11 +13,12 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\Microsoft.Diagnostics.Monitoring.HostingStartup\Microsoft.Diagnostics.Monitoring.HostingStartup.csproj" />
+    <ProjectReference Include="..\..\Microsoft.Diagnostics.Monitoring.StartupHook\Microsoft.Diagnostics.Monitoring.StartupHook.csproj" />
```
jander-msft, line 16: Why is this being included in the unit test app? This means that the assembly is in the application's default probing path, which we want to avoid since we want to mimic real applications as much as possible.

src/Tools/dotnet-monitor/Strings.resx
```diff
@@ -930,8 +930,8 @@
   <data name="ParameterCapturingNotAvailable_Reason_PreventedHostingStartup" xml:space="preserve">
     <value>The process has prevented hosting startup assemblies from loading using the ASPNETCORE_PREVENTHOSTINGSTARTUP environment variable.</value>
   </data>
-  <data name="ParameterCapturingNotAvailable_Reason_HostingStartupDidNotLoad" xml:space="preserve">
-    <value>This feature is only available on processes using ASP.NET Core. If the process is using ASP.NET Core and has successfully started, ensure that it has not been configured to prevent hosting startup assemblies from loading.</value>
+  <data name="ParameterCapturingNotAvailable_Reason_StartupHookDidNotLoad" xml:space="preserve">
+    <value>This feature requires that the DOTNET_STARTUP_HOOKS environment variable is set on the target process.</value>
```
jander-msft, line 934: This explanation isn't really correct. In .NET 8+, we don't use that environment variable to load the startup hook but use a diagnostic IPC command to dynamically load it. Maybe just say "This feature requires the startup hook to be loaded in the target process." That should be sufficient since I believe the `LogFormatString_StartupHookApplyFailed` string will be logged in .NET Monitor's output shortly before `ErrorMessage_ParameterCapturingNotAvailable` string is logged.

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Microsoft.Diagnostics.Monitoring.UnitTestApp.csproj
clguiman: This project also tests Function Probes which are now in the StartupHook assembly. Should I reference the source files in the unit test project so I can remove the assembly dependency, or should I create a separate unit tests project for function probes scenario?

jander-msft: Chatted with @clguiman about this. This change isn't introducing a new dependency on the functionality that is moving. Let's keep it as you have it and then we can separately consider if a new app should be created for testing these.

clguiman:
I fixed the `FunctionProbesTests`, but the `StartupHook` project reference is still there. It will be addressed in a separate PR

schmittjoseph (approved):

PR #6751, opened 2024-05-30
Remove HostingStartup

clguiman:
###### Summary

* Remove `Microsoft.Diagnostics.Monitoring.HostingStartup.csproj` and all references in other projects.
* Remove `Microsoft.Diagnostics.Monitoring.HostingStartup.UnitTests.csproj`.
* Update `DiagnosticsBootstrapper` to stop loading hosting startup assemblies.
* Remove the `HOSTINGSTARTUP` compile-time constant checks.
* Remove all HostingStartup documentation references.
* Replaced the`Microsoft.Diagnostics.Tools.Monitor.HostingStartup` namespace with `Microsoft.Diagnostics.Tools.Monitor.StartupHook`.


documentation/api/parameters.md
```diff
@@ -131,8 +131,6 @@ Content-Type: application/x-ndjson
 - The target application must use ASP.NET Core.
 - The target application cannot have [Hot Reload](https://learn.microsoft.com/visualstudio/debugger/hot-reload) enabled.
 - `dotnet-monitor` must be set to `Listen` mode, and the target application must start suspended. See [diagnostic port configuration](../configuration/diagnostic-port-configuration.md) for information on how to do this.
```
clguiman, line 133: @schmittjoseph @jander-msft  do you know if this requirement is still valid? Does the StartupHook require to be present at app start?

schmittjoseph: Nope :) We can now remove this requirement!

src/Microsoft.Diagnostics.Monitoring.StartupHook/Microsoft.Diagnostics.Monitoring.StartupHook.csproj
```diff
@@ -14,7 +14,6 @@
     <Compile Include="..\Microsoft.Diagnostics.Monitoring.WebApi\Stacks\NameIdentificationEvents.cs" Link="Exceptions\Eventing\NameIdentificationEvents.cs" />
     <Compile Include="..\Tools\dotnet-monitor\DisposableHelper.cs" Link="DisposableHelper.cs" />
     <Compile Include="..\Tools\dotnet-monitor\Exceptions\ExceptionEvents.cs" Link="Exceptions\Eventing\ExceptionEvents.cs" />
-    <Compile Include="..\Tools\dotnet-monitor\StartupHook\StartupHookIdentifiers.cs" Link="StartupHookIdentifiers.cs" />
```
clguiman, line 17: FYI - I removed this file because the only identifier inside (`StartupHookIdentifiers.EnvironmentVariables.HostingStartupPath`) was removed as part of this change, so I didn't really see the need for an empty type declaration.

documentation/api/parameters.md
jander-msft: IIRC .NET Monitor currently only injects the startup hook in listen mode. I don't think that the "must start suspended" portion is valid anymore given the new implementation of the probes callback, but we should double-check that and add tests to verify and ensure that remains correct.

src/Tools/dotnet-monitor/InProcessFeatures/InProcessFeatures.cs
```diff
@@ -32,10 +32,8 @@ internal sealed class InProcessFeatures : IInProcessFeatures
 
         public bool IsMutatingProfilerRequired => IsParameterCapturingEnabled;
```
schmittjoseph, line 33: Now that parameter capturing lives entirely in the startup hook, it **should** work on attach. In theory all of the code we have in the mutating profiler should support this but we've never validated it. Have you had been able to validate parameter capturing works when connecting to an already running process?  (We might want to add a test case for this seperately -- we don't need to exercise the whole parameter capturing path (unless its easy), just that function probes work

documentation/api/parameters.md
clguiman: Does that UnitTestApp have the capability to test this new scenario? We'd need to start up dotnet-monitor in `Connect` mode, then issue a parameter capture request and verify data flows back.

jander-msft (approved):

documentation/api/parameters.md
jander-msft: Yes, scenarios can be run with connect mode. I don't recall if we allow no-suspend listen mode, but that should be easy to add.

src/Tools/dotnet-monitor/InProcessFeatures/InProcessFeatures.cs
jander-msft: If it does work, I think we can adapt the existing `ParameterCapturingTests` to test both connect and listen mode.

schmittjoseph (approved):

src/Tools/dotnet-monitor/InProcessFeatures/InProcessFeatures.cs
schmittjoseph: I mean validating that this working in `Listen` + no suspend mode, we don't load in-proc features in Connect

PR #6778, opened 2024-06-05
Enable parameter capturing in no suspend mode.

clguiman:
###### Summary
*  Moved the `InProcessFeaturesService` registration before the `ProfilerService` registration because the `IEndpointInfoSourceCallbacks` need to first reach the in process service so it can set the `DotnetMonitor_InProcessFeatures_ParameterCapturing_Enable` environment variable which is later checked by the profiler initialization.
* Reworked `ParameterCapturingTests` to include the new `shouldSuspendTargetApp` arg and added tests to verify the new scenarios: `CapturesParametersNoSuspend` & `AppWithStartupHookFailsInNoSuspend`.
* Updated `ScenarioRunner.SingleTarget()` to support starting the target process befor the monitor process.

* Updated docs.

### DIFF
--- a/documentation/api/parameters.md
+++ b/documentation/api/parameters.md
@@ -128,12 +128,10 @@ Content-Type: application/x-ndjson
 
 ## Additional Requirements
 
-- The target application must use ASP.NET Core.
 - The target application cannot have [Hot Reload](https://learn.microsoft.com/visualstudio/debugger/hot-reload) enabled.
 - `dotnet-monitor` must be set to `Listen` mode. See [diagnostic port configuration](../configuration/diagnostic-port-configuration.md) for information on how to do this.
-- If the target application has the `dotnet-monitor` startup hook manually configured then the feature will only work if the target application is started suspended.
+- If the target application is using .NET 7 then the dotnet-monitor startup hook must be manually configured and the target application must start suspended. In .NET 8+ this is not a requirement.
 - This feature relies on a [ICorProfilerCallback](https://docs.microsoft.com/dotnet/framework/unmanaged-api/profiling/icorprofilercallback-interface) implementation. If the target application is already using an `ICorProfiler` that isn't notify-only, this feature will not be available.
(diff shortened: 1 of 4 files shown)


jander-msft (approved):

schmittjoseph (approved):

jander-msft (approved):

clguiman:
I noticed flakiness around the nosuspend tests caused by the `DOTNET_STARTUP_HOOKS` env var being set on the target app (The `ParameterCapturingService` might start up before the monitor is ready), so I reworked the `ScenarioRunner` to support starting the target app before the monitor.
I also added the `AppWithStartupHookFailsInNoSuspend` test so it's clearly state we don't support this scenario.

documentation/api/parameters.md
```diff
@@ -130,7 +130,8 @@ Content-Type: application/x-ndjson
 
 - The target application must use ASP.NET Core.
 - The target application cannot have [Hot Reload](https://learn.microsoft.com/visualstudio/debugger/hot-reload) enabled.
-- `dotnet-monitor` must be set to `Listen` mode, and the target application must start suspended. See [diagnostic port configuration](../configuration/diagnostic-port-configuration.md) for information on how to do this.
+- `dotnet-monitor` must be set to `Listen` mode. See [diagnostic port configuration](../configuration/diagnostic-port-configuration.md) for information on how to do this.
+- If the target application has the `dotnet-monitor` startup hook manually configured then the feature will only work if the target application is started suspended.
```
schmittjoseph, line 134: suggestion:

Let's combine these with the below and talk in terms of what .NET version the target app is using
> - If a target application is using .NET 7 then the `dotnet-monitor` startup hook must be configured. This is automatically done in .NET 8+.

Perhaps something like

> If the target application is using .NET 7 then the `dotnet-monitor` startup hook must be manually configured and the target application must start suspended.

jander-msft (approved):

PR #6831, opened 2024-06-14
[Egress provider - S3] Support KMS encryption for artifacts

jwreford99:
###### Summary
AWS offer the ability to use KMS (Key Management Service) encryption keys to encrypt new objects into an S3 bucket. There is more documentation [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingKMSEncryption.html), but the key benefits of this are:

> - Centrally create, view, edit, monitor, enable or disable, rotate, and schedule deletion of KMS keys.
> - Define the policies that control how and by whom KMS keys can be used.
> - Audit their usage to prove that they are being used correctly.
>
> The security controls in AWS KMS can help you meet encryption-related compliance requirements. You can use these KMS keys to protect your data in Amazon S3 buckets.

As such it would be a useful option given the sensitive data which dotnet monitor exports (e.g. memory dumps) to be able to enable this for dotnet monitor deployments.

For KMS encryption it provides two options: 
 - No specific key provided ("AWS managed key") - AWS will create a KMS key for the bucket and uses that one for future PUT operations. This gives limited control as to what can be done with that key going forwards
 - A specific key provided ("customer managed key") - the customer creates and manages their own KMS key and has full control over actions relating to it

This PR allows for both options while also allowing users to continue to use the default encryption which AES256 encryption.

It introduces two new configuration options:
 - useKmsEncryption: A boolean flag which controls whether the Egress should use KMS server side encryption. 
 - kmsEncryptionKey: If UseKmsEncryption is true, this specifies the arn of the "customer managed" KMS encryption key to be used for server side encryption. If no value is set for this field then S3 will use an AWS managed key for KMS encryption. 

Documentation for how to specify server side encryption with KMS can be found at https://docs.aws.amazon.com/AmazonS3/latest/userguide/specifying-kms-encryption.html

I could not see a sensible way to unit test these changes as it would require mocking out the S3 client which is not currently possible in the existing test harness and would require a significant refactor to make it possible. However, if you think this is needed I would be happy to follow your guidance

###### Release Notes Entry
The S3 egress provider now provides the ability to use KMS encryption keys to encrypt artifacts placed in S3 buckets


documentation/configuration/egress-configuration.md
```diff
@@ -184,6 +186,25 @@ The Queue Message's payload will be the blob name (`<BlobPrefix>/<ArtifactName>`
...
+
+  ```json
+  {
+      "Egress": {
+          "S3Storage": {
+              "monitorS3Blob": {
+                  "endpoint": "http://localhost:9000",
+                  "bucketName": "myS3Bucket",
+                  "useKmsEncryption": true,
+                  "kmsEncryptionKey": "arn:aws:kms:eu-west-1:012345678900:key/bb16676b-5cef-4eab-aa4c-ac6cd7961047"
```
kkeirstead, line 200: Nit: Not a suggestion, just gauging people's thoughts regarding configuration complexity - if a user species this field (`kmsEncryptionKey`), should they also need to remember to explicitly the `useKmsEncryption` field, or should that be implied? My understanding is we still need that field since `KmsEncryption` can be on without the user specifying their own key, but if they _do_ specify a key, presumably they want to use it.

kkeirstead: Also, unrelated, but should the value for the Key ID be made generic here?

jwreford99: @kkeirstead you are right that we need the field because you can want to use KMS encryption without specifying a specific KMS key. I have no strong opinions on if you need to specify both if you are specifying a key, but it felt like it would make sense to need `useKmsEncryption` if you wanted KMS encryption, and I guess I could add a validation rule to the effect so that people "fail early" if they ask for an encryption key but don't say "useKmsEncryption"


An alternative API would be to still just have one field but to have a "magic value" which means "use an AWS controlled key". However, that could obviously be confusing. I don't pretend to be the expert on this, so happy to follow the shape of other similar things that exist 😄 

With respect to making the key ID generic do you just mean to have a something like `arn:aws:kms:{region}:{account-id}:key/{resource-id}` rather than a dummy value?

kkeirstead: > With respect to making the key ID generic do you just mean to have a something like `arn:aws:kms:{region}:{account-id}:key/{resource-id}` rather than a dummy value?

That would probably be fine - looking at [this page](https://github.com/dotnet/dotnet-monitor/blob/main/documentation/api-key-setup.md) in our docs, it looks like we do `fffff...` to make it obvious that something's not a real value.

jwreford99: I've update it to the generic key ID because it feels like that makes more sense in this context 😄

schmittjoseph (approved):
LGTM, thanks for the contribution!

jwreford99:
@schmittjoseph thanks for the speedy review. Just checking, do I need to do anything to make sure this ends up in the `release/8.x` branch / the dotnet monitor nightly build? 😄

schmittjoseph:
> @schmittjoseph thanks for the speedy review. Just checking, do I need to do anything to make sure this ends up in the `release/8.x` branch / the dotnet monitor nightly build? 😄

Nope! This will automatically be available in the next `8.1-preview` nightly build.  As for official builds, it'll be available in both the `8.1` and `9.0` releases once they are available.

jwreford99:
Great thanks very much!

PR #6839, opened 2024-06-15
Add mvid and mdtoken to callstackframe

WilliamXieMSFT:
###### Summary

This PR adds moduleVersionId guid and methodDef token to callstack frames.

Example output from invoking /exceptions route:

```json
{
    "id": 40,
    "timestamp": "2024-06-14T23:39:07.9509456Z",
    "typeName": "AppDiagnosticsNet6.Controllers.MyException",
    "moduleName": "AppDiagnosticsNet.dll",
    "message": "I have an inner exception.",
    "activity": {
      "id": "00-5e1784b5de0c77f7fb60abed6e43f3b9-ecd287b342b2dcec-00",
      "idFormat": "W3C"
    },
    "innerExceptions": [
      {
        "id": 39
      }
    ],
    "stack": {
      "threadId": 49328,
      "threadName": null,
      "frames": [
        {
          "methodName": "ThrowInnerException",
*         "methodToken": 100663635,
          "parameterTypes": [],
          "typeName": "AppDiagnosticsNet6.Controllers.ExceptionsController",
          "moduleName": "AppDiagnosticsNet.dll",
*         "moduleVersionId": "bf769014-c2e2-496a-93b7-76fbbcd04be5"
        },
```

###### Release Notes Entry

CallStackFrame JSON results now include methodtoken and moduleversionid.


schmittjoseph:
Changes largely look good to me so far, the outstanding things I'm aware of are:
- Doc changes
- Update the profiler's /stacks feature to emit these fields as well (this will probably fix the failing CI)
- Tests

src/Tools/dotnet-monitor/Exceptions/ExceptionsOperation.cs
```diff
@@ -203,6 +204,7 @@ await using (Utf8JsonWriter writer = new(stream, new JsonWriterOptions() { Inden
                         writer.WriteEndArray(); // end parameterTypes
                         writer.WriteString("typeName", frame.TypeName);
                         writer.WriteString("moduleName", frame.ModuleName);
+                        writer.WriteString("moduleVersionId", frame.ModuleVersionId);
```
schmittjoseph, line 207: Let's be explicit about how we serialize the mvid? e.g. `writer.WriteString("moduleVersionId", frame.ModuleVersionId.ToString("D"))`

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/CallStackResults.cs
```diff
@@ -11,12 +12,18 @@ public class CallStackFrame
         [JsonPropertyName("methodName")]
         public string MethodName { get; set; }
 
+        [JsonPropertyName("methodToken")]
```
schmittjoseph, line 15: I'm a bit surprised this isn't causing a test failure. This should be changing the api surface and require a regeneration of our `openapi.json` file

src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/Exceptions/Eventing/ExceptionsEventListener.cs
```diff
@@ -68,6 +68,7 @@ protected override void OnEventWritten(EventWrittenEventArgs eventData)
                             ToUInt64(eventData.Payload[NameIdentificationEvents.FunctionDescPayloads.FunctionId]),
```
schmittjoseph, line 68: Can we add a couple tests (or update existing ones) to validate that the new data is being reported correctly for both exception and stacks?

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/CallStackResults.cs
WilliamXieMSFT: Spoke offline - there are missing schemas in the json (like callstackframe) that will need to be addressed outside this PR.

src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/Exceptions/Eventing/ExceptionsEventListener.cs
WilliamXieMSFT: I've updated both the startuphook unit tests and the exceptions functional tests. Let me know if there are other areas where I can add more tests.

schmittjoseph: I'm assuming this PR also updates `/stacks` to report this information so ideally we'd update those tests by updating the following lines:
https://github.com/dotnet/dotnet-monitor/blob/32fa38302fa3293889835062e611ea753b564618/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/StacksTests.cs#L536
https://github.com/dotnet/dotnet-monitor/blob/32fa38302fa3293889835062e611ea753b564618/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/StacksTests.cs#L64


If we take a project reference on `Microsoft.Diagnostics.Monitoring.UnitTestApp.dll` we can determine the expected mvid and method tokens for the frames.

WilliamXieMSFT: The mvid and mdtokens show up in the JSON response so I've updated `JsonValidation` instead of `PlainTextValidation`

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/StacksTests.cs
```diff
@@ -33,10 +33,13 @@ public class StacksTests
         private readonly TemporaryDirectory _tempDirectory;
 
         private const string ExpectedModule = @"Microsoft.Diagnostics.Monitoring.UnitTestApp.dll";
+        private static readonly Guid ExpectedModuleVersionId = new Guid("b271d014-2be3-4972-9835-7523a19ecff8");
```
schmittjoseph, line 36: We can't hard code the expected guid / method tokens, since they may change with each build.

We should instead use reflection to get them now that we have a project reference.

See this test for an example of getting a `MethodInfo`:
- https://github.com/dotnet/dotnet-monitor/blob/32fa38302fa3293889835062e611ea753b564618/src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/Scenarios/FunctionProbes/FunctionProbesScenario.cs#L96

src/Tests/Microsoft.Diagnostics.Monitoring.UnitTestApp/AssemblyInfo.cs
```diff
@@ -0,0 +1,20 @@
...
+// components.  If you need to access a type in this assembly from COM, set the ComVisible
+// attribute to true on that type.
+
+[assembly: ComVisible(false)]
+
+// The following GUID is for the ID of the typelib if this project is exposed to COM.
+
+[assembly: Guid("67036149-b174-4033-a15f-ed5079896dfb")]
+
+[assembly:InternalsVisibleTo("Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
schmittjoseph, line 20: We define `InternalsVisibleTo` via our csproj files, e.g.

https://github.com/dotnet/dotnet-monitor/blob/32fa38302fa3293889835062e611ea753b564618/src/Microsoft.Diagnostics.Monitoring.WebApi/Microsoft.Diagnostics.Monitoring.WebApi.csproj#L30

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/StacksTests.cs
WilliamXieMSFT: Looks like I need to update the exceptions tests too

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Identification/ExceptionGroupIdentifierCache.cs
```diff
@@ -83,13 +83,19 @@ public ulong GetOrAdd(MethodBase method)
                 return methodId;
 
             // Dynamic methods do not have metadata tokens
-            uint metadataToken = 0;
+            uint methodToken = 0;
             try
             {
-                metadataToken = Convert.ToUInt32(method.MetadataToken);
+                methodToken = Convert.ToUInt32(method.MetadataToken);
```
schmittjoseph, line 89: Double checking -- this appears to be a bug fix, correct? We were incorrectly reporting the method token as the parent class token previously?

documentation/api/definitions.md
```diff
@@ -35,8 +35,11 @@ First Available: 8.0 Preview 7
 | Name | Type | Description |
 |---|---|---|
 | `methodName` | string | Name of the method for this frame. This includes generic parameters. |
+| `methodToken` | int | TypeDef token for the method. |
+| `parameterTypes` | string[] | Array of parameter types. Empty array if none. |
 | `typeName` | string | Name of the class for this frame. This includes generic parameters. |
 | `moduleName` | string | Name of the module for this frame. |
+| `moduleVersionId` | guid | Unique identifier used to distinguish between two versions of the same module. Format: `00000000-0000-0000-0000-000000000000`. |
```
schmittjoseph, line 42:
```suggestion
| `moduleVersionId` | guid | Unique identifier used to distinguish between two versions of the same module. An empty value: `00000000-0000-0000-0000-000000000000`. |
```

To be inline with the rest of the docs.

src/Tests/Microsoft.Diagnostics.Monitoring.StartupHook.UnitTests/Exceptions/Eventing/ExceptionsEventSourceTests.cs
```diff
@@ -142,6 +142,58 @@ public void ExceptionsEventSource_WriteStackFrame_Event(ulong id, ulong methodId
...
+            Assert.Equal(name, function.Name);
+            // We would normally expect the following to return an array of the stack frame IDs
+            // but in-process listener doesn't decode non-byte arrays correctly.
+            Assert.Equal(Array.Empty<ulong>(), function.TypeArgs);
+            Assert.Equal(Array.Empty<ulong>(), function.ParameterTypes);
+        }
+
+        [Theory]
+        [InlineData(0, "00000000-0000-0000-0000-000000000000", "")]
+        [InlineData(1, "792ce009-1dbd-46d7-9e47-0a55bff0f5a4", "Module")]
```
schmittjoseph, line 179:
```suggestion
        [InlineData(1, NonEmptyGuidString, "Module")]
```

Let's reuse the existing const in this test class

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/StacksTests.cs
```diff
@@ -40,6 +40,17 @@ public class StacksTests
...
         private const string ExpectedThreadName = "TestThread";
 
+        private static MethodInfo GetMethodInfo(string methodName)
+        {
+            // Strip off any generic type information. Pseudo frames (eg. [NativeFrame]) will return null anyways.
+            if (methodName.Contains('['))
+            {
+                methodName = methodName[..methodName.IndexOf('[')];
+            }
+
```
schmittjoseph, line 50: nit: Rather than letting psuedo frames enter the below check and rely on it returning null, let's add an explicit check to make the intention clear.

e.g.

```csharp
// Return null on psuedo frames (e.g. [NativeFrame])
if (methodName.Length == 0)
{
  return null;
}

```

wiktork (approved):

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Identification/ExceptionGroupIdentifierCache.cs
WilliamXieMSFT: Correct

schmittjoseph (approved):

PR #6899, opened 2024-06-27
Update build asset staging to use managed identity

jander-msft:
###### Summary

Use managed identity for:
- Downloading drops via `darc gather-drop`. The key part is to run this within an AZ CLI task so an access token can be acquired from the service connection to authenticate with maestro.
- Publishing the asset layout to Azure blob storage. The key part is run this within an AZ CLI task so that managed identity credentials can be used to authenticate with blob storage.

Note: Much of these changes were adopted from dotnet/diagnostics, especially for the DiagnosticsReleaseTool. You can compare these changes to the ones in that repository: https://github.com/dotnet/diagnostics/tree/main/eng/release/DiagnosticsReleaseTool

Validation build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2484218&view=results


global.json
```diff
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "8.0.104",
+    "dotnet": "8.0.106",
```
schmittjoseph, line 3: Is this change needed?

schmittjoseph (approved):
LGTM.

Request for next time: Let's perform the analyzer/style changes as a separate PR first since it took a bit of time to sort through the managed identity specific code vs styling changes.

jander-msft:
> LGTM.
> 
> Request for next time: Let's perform the analyzer/style changes as a separate PR first since it took a bit of time to sort through the managed identity specific code vs styling changes.

I've undone the stylistic/analyzer changes. That'll probably help minimize collisions when backporting too.

schmittjoseph (approved):

PR #6924, opened 2024-07-01
[Nullable Context] Enable on `Microsoft.Diagnostics.Monitoring.WebApi`

schmittjoseph:
###### Summary

This is the first PR in a series to enable nullable aware context throughout our code. To help with ease of review and minimize the chance of introducing regressions, I'll be organizing the PRs as follows:
- Each project will have nullable aware context enabled in its own PR.
  - These PRs will be **large** but **only** contain trivial, low/no-risk, changes.  Namely they'll primarily contain the addition of the new nullable syntaxes (`?` and `!`) and static analysis attributes (e.g. `[NotNullWhen(..)]`).
  - Nullable context will be disabled around pieces of code that contain some level of risk to update. These sections of code will be addressed in follow up PRs so they can be properly reviewed. (e.g. `#nullable disable ... #nullable restore`)
  - If you see anything that you feel like could introduce a regression or is non-trivial in these PRs, leave a comment and I'll pull out those changes.
 
- All PRs in this effort will **not** contain any breaking changes.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/DiagnosticServices.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/DiagnosticServices.cs
@@ -26,14 +26,14 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
             _logger = logger;
         }
 
-        public async Task<IEnumerable<IProcessInfo>> GetProcessesAsync(DiagProcessFilter processFilterConfig, CancellationToken token)
+        public async Task<IEnumerable<IProcessInfo>> GetProcessesAsync(DiagProcessFilter? processFilterConfig, CancellationToken token)
         {
-            IEnumerable<IProcessInfo> processes = null;
+            IEnumerable<IProcessInfo>? processes = null;
 
             try
             {
                 using CancellationTokenSource extendedInfoCancellation = CancellationTokenSource.CreateLinkedTokenSource(token);
-                IList<Task<IProcessInfo>> processInfoTasks = new List<Task<IProcessInfo>>();
+                IList<Task<IProcessInfo?>> processInfoTasks = new List<Task<IProcessInfo?>>();
                 foreach (IEndpointInfo endpointInfo in await _endpointInfoSource.GetEndpointInfoAsync(token))
                 {
                     // CONSIDER: Can this processing be pushed into the IEndpointInfoSource implementation and cached
@@ -71,7 +71,8 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
 
                 processes = processInfoTasks
                     .Where(t => t.Result != null)
-                    .Select(t => t.Result);
+                    .Select(t => t.Result)
+                    .Cast<IProcessInfo>();
             }
             catch (UnauthorizedAccessException)
             {
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/DiagnosticServices.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/DiagnosticServices.cs
@@ -28,7 +28,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
 
         public async Task<IEnumerable<IProcessInfo>> GetProcessesAsync(DiagProcessFilter? processFilterConfig, CancellationToken token)
         {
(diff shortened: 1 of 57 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/Validation/IntegerOrHexStringAttribute.cs
```diff
@@ -3,26 +3,27 @@
...
 using System.ComponentModel.DataAnnotations;
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 
 namespace Microsoft.Diagnostics.Monitoring.WebApi.Validation
 {
     public class IntegerOrHexStringAttribute : ValidationAttribute
     {
-        protected override ValidationResult IsValid(object value, ValidationContext validationContext)
+        protected override ValidationResult? IsValid(object? value, ValidationContext validationContext)
```
schmittjoseph, line 13: note: `ValidationResult.Success` is null, and a null return value indicates there are no issues.

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/CallStackResults.cs
```diff
@@ -10,16 +10,16 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Models
     public class CallStackFrame
```
schmittjoseph, line 10: NOTE:

The `MethodName`, `TypeName`, `ModuleName`, and `ThreadName` should be marked with the `[Required]` attributes. However that is beyond the scope of this PR effort and also wouldn't make any difference to our `openapi.json` since these types aren't included.

src/Microsoft.Diagnostics.Monitoring.WebApi/DiagnosticServices.cs
```diff
@@ -26,14 +26,14 @@ internal sealed class DiagnosticServices : IDiagnosticServices
             _logger = logger;
         }
 
-        public async Task<IEnumerable<IProcessInfo>> GetProcessesAsync(DiagProcessFilter processFilterConfig, CancellationToken token)
+        public async Task<IEnumerable<IProcessInfo>> GetProcessesAsync(DiagProcessFilter? processFilterConfig, CancellationToken token)
         {
-            IEnumerable<IProcessInfo> processes = null;
+            IEnumerable<IProcessInfo>? processes = null;
```
jander-msft, line 31: Could you assign this `Enumerable.Empty<IProcessInfo>()` so it doesn't have to be nullable?

schmittjoseph: Sure, in general I've been trying to keep the code as close to original as possible given the amount of changes, and deferring cleanup to later PRs.

In this case though there's no risk of this change so I'll update it

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/EventPipeProvider.cs
```diff
@@ -15,19 +18,19 @@ public class EventPipeProvider
...
         [Required]
-        public string Name { get; set; }
+        public string Name { get; set; } = string.Empty;
 
         [JsonPropertyName("keywords")]
 #if !SCHEMAGEN
         [IntegerOrHexString]
 #endif
-        public string Keywords { get; set; } = "0x" + EventKeywords.All.ToString("X");
+        public string? Keywords { get; set; } = "0x" + EventKeywords.All.ToString("X");
```
jander-msft, line 27: Is this really nullable given that it has a default non-null value?

schmittjoseph: From an `openapi.json` perspective it is since there's no `[Required]` attribute (and we have null handling for it).

Practically though, it should be marked as non-nullable and `[Required]` since we throw an exception if it is set to null.


With all of that said, I'm trying to avoid any changes to `openapi.json`, even if it doesn't seem right. This is something we should clean up seperately.

schmittjoseph: I'll create an issue to track these sorts of problems

jander-msft (approved):

PR #6932, opened 2024-07-01
[Nullable Context] Enable on `Microsoft.Diagnostics.Monitoring.ConfigurationSchema`

schmittjoseph:
###### Summary

Enable nullable aware context for `Microsoft.Diagnostics.Monitoring.ConfigurationSchema`.  Just like https://github.com/dotnet/dotnet-monitor/pull/6924 this PR only contains no/low risk changes.

#### Why not use `required`  members?

Throughout this PR (and others in this series) you'll see I initialize option class members with a default value when they are `[Required]` instead of marking them as `required`. This is because in several places we constrain our options with the `new constraint`, meaning they must have a public parameterless constructor. Changing this would be a much larger effort.

### DIFF
--- a/src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleOptions.cs
+++ b/src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleOptions.cs
@@ -19,11 +19,13 @@ namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options
             Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_CollectionRuleOptions_Filters))]
         public List<ProcessFilterDescriptor>? Filters { get; } = new List<ProcessFilterDescriptor>(0);
 
+#nullable disable
         [Display(
             ResourceType = typeof(OptionsDisplayStrings),
             Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_CollectionRuleOptions_Trigger))]
         [Required]
-        public CollectionRuleTriggerOptions Trigger { get; set; } = new();
+        public CollectionRuleTriggerOptions Trigger { get; set; }
+#nullable enable
 
         [Display(
             ResourceType = typeof(OptionsDisplayStrings),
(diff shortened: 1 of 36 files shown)


src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleOptions.cs
```diff
@@ -15,23 +17,25 @@ internal sealed partial class CollectionRuleOptions
         [Display(
             ResourceType = typeof(OptionsDisplayStrings),
             Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_CollectionRuleOptions_Filters))]
-        public List<ProcessFilterDescriptor> Filters { get; } = new List<ProcessFilterDescriptor>(0);
+        public List<ProcessFilterDescriptor>? Filters { get; } = new List<ProcessFilterDescriptor>(0);
```
jander-msft, line 20: The properties that are get-only and already have a default value shouldn't need to be nullable, correct?

jander-msft (dismissed):

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleOptions.cs
schmittjoseph: Just double checked, you're correct. Even if you explicitly specify `null` in the schema, it'll default to this value. I'll push a new iteration reverting  these

jander-msft (approved):

src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleOptions.cs
schmittjoseph: Now that the PR build failed I remember why I did this -- the schema itself states that these fields are `nullable`, so they must be marked as such. This is an issue with our schema and not this PR, so I'll restore my original change and add this to the ongoign github issue.

jander-msft (approved):

Issue #6929, opened 2024-07-01
[Nullable Context] Outstanding issues


schmittjoseph:
While enabling nullable context throughout the code, we've found several places that need more attention:
- [ ] [`EventPipeProvider.Keywords`](https://github.com/dotnet/dotnet-monitor/pull/6924#discussion_r1661371090)
- [ ] [`StorageOptions.DumpTempFolder`](https://github.com/dotnet/dotnet-monitor/pull/6933#discussion_r1661618736)
- [ ] [`ProcessFilterDescriptor.Value`](https://github.com/dotnet/dotnet-monitor/pull/6933#discussion_r1661618659)
- [x] [`CollectionRuleOptions.{Filters,Actions}`](https://github.com/dotnet/dotnet-monitor/pull/6932/files#r1663237955)
- [ ] [`ICollectionRuleTriggerOperations.TryValidateOptions`](https://github.com/dotnet/dotnet-monitor/pull/6949#discussion_r1687043047)
- [ ] [`ActivityExtensions.GetSpandId`](https://github.com/dotnet/dotnet-monitor/pull/6949#discussion_r1664777202)

PR #6949, opened 2024-07-03
[Nullable Context] Enable on `dotnet-monitor`

schmittjoseph:
###### Summary

Enable nullable aware context for `dotnet-monitor`.  Just like https://github.com/dotnet/dotnet-monitor/pull/6924 this PR only contains no/low risk changes. If you see anything that you feel like could introduce a regression or is non-trivial, leave a comment and I'll pull out those changes.

### DIFF
--- a/src/Tools/dotnet-monitor/ActivityExtensions.cs
+++ b/src/Tools/dotnet-monitor/ActivityExtensions.cs
@@ -7,7 +7,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor
 {
     internal static class ActivityExtensions
     {
-        public static string GetSpanId(this Activity activity)
+        public static string? GetSpanId(this Activity activity)
         {
             switch (activity.IdFormat)
             {
@@ -19,7 +19,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor
             return string.Empty;
         }
 
-        public static string GetParentId(this Activity activity)
+        public static string? GetParentId(this Activity activity)
         {
             switch (activity.IdFormat)
             {
@@ -31,7 +31,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor
             return string.Empty;
         }
 
-        public static string GetTraceId(this Activity activity)
+        public static string? GetTraceId(this Activity activity)
         {
             switch (activity.IdFormat)
             {
--- a/src/Tools/dotnet-monitor/Auth/ApiKey/MonitorApiKeyConfiguration.cs
+++ b/src/Tools/dotnet-monitor/Auth/ApiKey/MonitorApiKeyConfiguration.cs
@@ -16,9 +16,9 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Auth.ApiKey
     internal class MonitorApiKeyConfiguration : AuthenticationSchemeOptions
     {
         public bool Configured { get; set; }
-        public string Subject { get; set; }
-        public SecurityKey PublicKey { get; set; }
-        public IEnumerable<ValidationResult> ValidationErrors { get; set; }
-        public string Issuer { get; set; }
+        public string? Subject { get; set; }
+        public SecurityKey? PublicKey { get; set; }
+        public IEnumerable<ValidationResult>? ValidationErrors { get; set; }
+        public string? Issuer { get; set; }
     }
 }
--- a/src/Tools/dotnet-monitor/CollectionRules/Options/Actions/CollectTraceOptions.Validate.cs
+++ b/src/Tools/dotnet-monitor/CollectionRules/Options/Actions/CollectTraceOptions.Validate.cs
@@ -8,6 +8,7 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Linq;
 
@@ -20,17 +21,32 @@ namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Actions
     partial record class CollectTraceOptions :
         IValidatableObject
     {
+
+        [MemberNotNullWhen(true, nameof(StoppingEvent))]
+        private bool HasStoppingEvent()
+        {
+            return StoppingEvent != null;
+        }
+
+        [MemberNotNullWhen(true, nameof(Providers))]
+        private bool HasProviders()
+        {
+            return Providers != null && Providers.Count > 0;
+        }
+
+        [MemberNotNullWhen(true, nameof(Profile))]
+        private bool HasProfile()
+        {
+            return Profile.HasValue;
+        }
+
         IEnumerable<ValidationResult> IValidatableObject.Validate(ValidationContext validationContext)
         {
             List<ValidationResult> results = new();
 
-            bool hasProfile = Profile.HasValue;
-            bool hasProviders = null != Providers && Providers.Any();
-            bool hasStoppingEvent = null != StoppingEvent;
-
-            if (hasProfile)
+            if (HasProfile())
             {
-                if (hasStoppingEvent)
+                if (HasStoppingEvent())
                 {
                     results.Add(new ValidationResult(
                         string.Format(
@@ -40,7 +56,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Actions
                             nameof(StoppingEvent))));
                 }
 
-                if (hasProviders)
+                if (HasProviders())
                 {
                     // Both Profile and Providers cannot be specified at the same time, otherwise
                     // cannot determine whether to use providers from the profile or the custom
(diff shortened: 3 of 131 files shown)


src/Tools/dotnet-monitor/ConvertUtils.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using System;
+
+namespace Microsoft.Diagnostics.Tools.Monitor;
+
+internal static class ConvertUtils
+{
+    public static string ToString(object value, IFormatProvider? provider)
+    {
+        return Convert.ToString(value, provider)!; // Since value is not null this will never return null.
```
schmittjoseph, line 14: Ref: https://learn.microsoft.com/en-us/dotnet/api/system.convert.tostring?view=net-8.0#system-convert-tostring(system-object-system-iformatprovider)
> The string representation of value, or [Empty](https://learn.microsoft.com/en-us/dotnet/api/system.string.empty?view=net-8.0#system-string-empty) if value is an object whose value is `null`. If value is `null`, the method returns `null`.

src/Tools/dotnet-monitor/ActivityExtensions.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.Diagnostics.Tools.Monitor
 {
     internal static class ActivityExtensions
     {
-        public static string GetSpanId(this Activity activity)
+        public static string? GetSpanId(this Activity activity)
         {
             switch (activity.IdFormat)
```
wiktork, line 12: Why is this nullable? It seems like every code path returns a non null string.

src/Tools/dotnet-monitor/Auth/ApiKey/MonitorApiKeyConfiguration.cs
```diff
@@ -16,9 +16,9 @@ namespace Microsoft.Diagnostics.Tools.Monitor.Auth.ApiKey
     internal class MonitorApiKeyConfiguration : AuthenticationSchemeOptions
     {
         public bool Configured { get; set; }
-        public string Subject { get; set; }
-        public SecurityKey PublicKey { get; set; }
-        public IEnumerable<ValidationResult> ValidationErrors { get; set; }
-        public string Issuer { get; set; }
+        public string? Subject { get; set; }
+        public SecurityKey? PublicKey { get; set; }
+        public IEnumerable<ValidationResult>? ValidationErrors { get; set; }
```
wiktork, line 21: nit: It's usually easier to work with empty collections than null

schmittjoseph: It is, but I'm aiming to minimize code changes in this PR due to the size. That kind of cleanup should be done as separate PRs. See this PR description (https://github.com/dotnet/dotnet-monitor/pull/6924) for the approach I'm taking.

src/Tools/dotnet-monitor/ActivityExtensions.cs
schmittjoseph: `Activity.Id` can be null.

https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity.id?view=net-8.0

src/Tools/dotnet-monitor/CollectionRules/Options/Actions/CollectTraceOptions.Validate.cs
```diff
@@ -20,17 +21,32 @@ namespace Microsoft.Diagnostics.Tools.Monitor.CollectionRules.Options.Actions
     partial record class CollectTraceOptions :
         IValidatableObject
     {
+
+        [MemberNotNullWhen(true, nameof(StoppingEvent))]
```
wiktork, line 25: So this attribute allows usage of StoppingEvent without being certain that it's not null correct?

schmittjoseph: This attribute tells the compiler that when `HasStoppingEvent` returns true, the caller can guarantee that `StoppingEvent` is not null.

src/Tools/dotnet-monitor/CollectionRules/Triggers/ICollectionRuleTriggerOperations.cs
```diff
@@ -33,7 +33,7 @@ internal interface ICollectionRuleTriggerOperations
         /// </summary>
         bool TryValidateOptions(
             string triggerName,
-            object options,
+            object? options,
```
wiktork, line 36: We migrated most options here to be record types, which could guarantee non-nullability. If we push that approach across the board, we could potentially change this to non-nullable. I don't think there's a base type for records though.

schmittjoseph: Updated https://github.com/dotnet/dotnet-monitor/issues/6929 to track this since it's out of scope of this PR.

src/Tools/dotnet-monitor/EndpointInfo/ClientEndpointInfoSource.cs
```diff
@@ -68,7 +68,7 @@ public async Task<IEnumerable<IEndpointInfo>> GetEndpointInfoAsync(CancellationT
 
             await Task.WhenAll(endpointInfoTasks);
 
-            return endpointInfoTasks.Where(t => t.Result != null).Select(t => t.Result);
+            return endpointInfoTasks.Where(t => t.Result != null).Select(t => t.Result).Cast<IEndpointInfo>();
         }
     }
```
wiktork, line 73: Does cast throw for nulls? If so, then can the original List be non-nullable?

src/Tools/dotnet-monitor/HostBuilder/HostBuilderHelper.cs
```diff
@@ -21,7 +21,7 @@ internal static class HostBuilderHelper
 
         public static IHostBuilder CreateHostBuilder(HostBuilderSettings settings)
         {
-            string aspnetUrls = string.Empty;
+            string? aspnetUrls = string.Empty;
```
wiktork, line 24: This seems like it should be non-nullable.

schmittjoseph: This can be null - we set it using `webBuilder.GetSetting(WebHostDefaults.ServerUrlsKey);` which may be null. If it is, we apply our own defaults.

src/Tools/dotnet-monitor/StartupHook/StartupHookApplicator.cs
```diff
@@ -48,14 +48,15 @@ public async Task<bool> ApplyAsync(string tfm, string fileName, CancellationToke
                 return true;
             }
 
-
-            if (_endpointInfo.RuntimeVersion.Major < 8)
+#nullable disable
+            if (_endpointInfo.RuntimeVersion?.Major < 8)
```
wiktork, line 52: Is this disablement still needed? You are guarding runtime version and altered logger  as well.

src/Tools/dotnet-monitor/ValidationResultExtensions.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.ComponentModel.DataAnnotations;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    internal static class ValidationResultExtensions
+    {
```
wiktork, line 10: Is this the correct condition? Doesn't null imply success?

wiktork: Also consider adding some comments here

wiktork (dismissed):

src/Tools/dotnet-monitor/ValidationResultExtensions.cs
schmittjoseph: > Is this the correct condition? Doesn't null imply success?

It does, null is explicitly success. I'll add some comments to clarify this

src/Tools/dotnet-monitor/StartupHook/StartupHookApplicator.cs
schmittjoseph: It is required, but it's at the wrong line... good catch.

The `_logger.` call below uses `fileInfo.PhysicalPath` which may return null and should be handled in a separate PR.

src/Tools/dotnet-monitor/ActivityExtensions.cs
schmittjoseph: I've updates https://github.com/dotnet/dotnet-monitor/issues/6929 to keep track of this.

src/Tools/dotnet-monitor/EndpointInfo/ClientEndpointInfoSource.cs
schmittjoseph: Discussed offline, refactoring this code is out of scope of this PR

wiktork (approved):

PR #6950, opened 2024-07-03
[Nullable Context] Proposal for better null-state static analysis with `IPostConfigureOptions`

schmittjoseph:
###### Summary

This PR is a proposal for how we can have better null-state static analysis when dealing with our Options classes the rely on `IPostConfigureOptions` to set default values.

(Also partially addresses #6929)

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.Options/StorageOptions.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.Options/StorageOptions.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
 
         internal bool Configured
         {
-            [MemberNotNullWhen(true, nameof(DefaultSharedPath), nameof(DumpTempFolder), nameof(SharedLibraryPath))]
+            [MemberNotNullWhen(true, nameof(DumpTempFolder), nameof(SharedLibraryPath))]
             get;
             set;
         }
--- a/src/Microsoft.Diagnostics.Monitoring.Options/StorageOptions.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.Options/StorageOptions.cs
@@ -1,8 +1,8 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using Microsoft.Diagnostics.Monitoring.Options;
 using System.ComponentModel.DataAnnotations;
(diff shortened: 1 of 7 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/DumpService.cs
```diff
@@ -36,8 +37,10 @@ public async Task<Stream> DumpAsync(IEndpointInfo endpointInfo, Models.DumpType
                 throw new ArgumentNullException(nameof(endpointInfo));
             }
 
-            // Guaranteed to not be null by StoragePostConfigureOptions.PostConfigure.
-            string dumpTempFolder = _storageOptions.CurrentValue.DumpTempFolder!;
+            StorageOptions options = _storageOptions.CurrentValue;
+            OptionsUtils.ThrowIfNotConfigured<StorageOptions>(options.Configured);
```
schmittjoseph, line 41: This properly informs the compiler's null context because:
- `Options.ThrowIfNotConfigured` statically tell the compiler that if `options.Configured` is false it will never return.
- `options.Configured`'s getter method statically tells the compiler that if it is true then several members (e.g. `DumpTempFolder`) will not be null.

Ref: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis

schmittjoseph: It's also important to note that I haven't found a way of generalizing this new `Configured` field behind an interface, since the interface's definition can't have the concrete-implementation's attributes for null-state static analysis.

src/Microsoft.Diagnostics.Monitoring.Options/StorageOptions.cs
```diff
@@ -18,10 +18,17 @@ public class StorageOptions
...
-        [Experimental]
+        [Options.Experimental]
         [Display(
             ResourceType = typeof(OptionsDisplayStrings),
             Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_StorageOptions_SharedLibraryPath))]
         public string? SharedLibraryPath { get; set; }
+
+        internal bool Configured
+        {
+            [MemberNotNullWhen(true, nameof(DumpTempFolder))]
```
schmittjoseph, line 29: note: This support specifying multiple members, e.g.
`[MemberNotNullWhen(true, nameof(DumpTempFolder), nameof(SharedLibraryPath)]`

PR #6951, opened 2024-07-03
[Nullable Context] Enable on S3Storage

ashyuuu:
###### Summary
Enable nullable aware context for S3Storage.

### DIFF
--- a/src/Extensions/S3Storage/MultiPartUploadStream.cs
+++ b/src/Extensions/S3Storage/MultiPartUploadStream.cs
@@ -16,7 +16,7 @@ internal sealed class MultiPartUploadStream : Stream
 {
     private readonly byte[] _buffer;
     private int _offset;
-    private readonly string? _bucketName;
+    private readonly string _bucketName;
     private readonly string _objectKey;
     private readonly string _uploadId;
     private readonly IS3Storage _client;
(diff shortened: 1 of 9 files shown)


src/Extensions/S3Storage/S3StorageEgressProviderOptions.cs
```diff
@@ -14,38 +14,38 @@ internal sealed partial class S3StorageEgressProviderOptions
...
             Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_S3StorageEgressProviderOptions_Endpoint))]
-        public string Endpoint { get; set; }
+        public string? Endpoint { get; set; }
 
         [Display(
             ResourceType = typeof(OptionsDisplayStrings),
             Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_S3StorageEgressProviderOptions_BucketName))]
         [Required(AllowEmptyStrings = false)]
-        public string BucketName { get; set; }
+        public string? BucketName { get; set; }
```
schmittjoseph, line 23: Since this configuration option is `[Required]` we know that it must have a value and not be null. 

The pattern I've been using elsewhere for required config options is to make them non-nullable with a default value of an empty string.

```suggestion
        public string BucketName { get; set; } = string.Empty;
```

schmittjoseph: With changing this I suspect we'll also want to update a few other files to make `BucketName` non-nullable.

src/Extensions/S3Storage/MultiPartUploadStream.cs
```diff
@@ -27,7 +27,7 @@ internal sealed class MultiPartUploadStream : Stream
     public const int MinimumSize = 5 * 1024 * 1024; // the minimum size of an upload part (except for the last part)
     private readonly int _bufferSize;
 
-    public MultiPartUploadStream(IS3Storage client, string bucketName, string objectKey, string uploadId, int? bufferSize)
+    public MultiPartUploadStream(IS3Storage client, string? bucketName, string objectKey, string uploadId, int? bufferSize)
```
schmittjoseph, line 30: Follow up to: https://github.com/dotnet/dotnet-monitor/pull/6951#discussion_r1664891327

We should be able to make these bucket name variables non-nullable now that the configuration option is non-nullable.

schmittjoseph: Same applies to other files in this PR as well (such as the `S3Storage.cs` file which has a `_bucketName` variable).

PR #7030, opened 2024-07-24
[feature/9.x] Sync branch with origin/main

kkeirstead:
###### Summary

Automation failed via https://github.com/dotnet/dotnet-monitor/actions/runs/10032368298, so manually merging latest from main to feature/9.x.


schmittjoseph (changes requested):
Let's fix the automation as it breaks instead of manually merging. Based on the failure it looks like the merge driver isn't handling modify/delete conflicts correctly.

We'll likely need update this config, or update the workflow. I can take a closer look unless you'd like to.

https://github.com/dotnet/dotnet-monitor/blob/784e3c7654b7da15b926ef588df162d0bcaae893/.gitattributes#L51-L59

kkeirstead:
> Let's fix the automation as it breaks instead of manually merging. Based on the failure it looks like the merge driver isn't handling modify/delete conflicts correctly.
> 
> We'll likely need update this config, or update the workflow. I can take a closer look unless you'd like to.
> 
> https://github.com/dotnet/dotnet-monitor/blob/784e3c7654b7da15b926ef588df162d0bcaae893/.gitattributes#L51-L59

Sounds good if you can take a look - Justin asked for some adjustments so I went ahead and made those manually here. If we can get the automation working we can close this out, otherwise we'll just use this manual merge instead.

kkeirstead:
Replaced by https://github.com/dotnet/dotnet-monitor/pull/7038

PR #7052, opened 2024-07-29
Update view-merged-configuration : update command

MKasaei00:
Fix command to dotnet-monitor

###### Summary
Fix dotnet-monitor command


schmittjoseph (approved):
Thanks for the fix!

PR #7113, opened 2024-08-06
Fix spelling in release notes

jander-msft:
###### Summary

Fix spelling so that spell checker doesn't flag these as errors.


kkeirstead (approved):

schmittjoseph (approved):
Also be sure to update the affected PR's release notes entry to match this change

Issue #2761, opened 2022-10-24
Can dotnet monitor expose mulitple dotnet processes' metrics?


LaurenceChau:
## Background and Motivation
Currently, one dotnet monitor can only expose one dotnet process metrics. Although in container world, it is usual to have only one process in a container, our process is running on a machine right now and we have no plan to migrate to container shortly. If we want to get multiple processes metrics, we have to deploy multiple dotnet monitors in a machine.

Can dotnet monitor expose multiple dotnet processes' metrics so that one dotnet monitor can monitor all dotnet processes in a machine? 

## Proposed Feature

current dotnet monitor metrics format is below
```
systemruntime_cpu_usage_ratio 0 1666605036059
systemruntime_cpu_usage_ratio 0 1666605041049
systemruntime_cpu_usage_ratio 0 1666605046048
```

sugget adding the process_id and process label to the metrics, e.g.
```
systemruntime_cpu_usage_ratio{process_id="1234", process="TestProcess1"} 0 1666605036059
systemruntime_cpu_usage_ratio{process_id="1234", process="TestProcess1"} 0 1666605041049
systemruntime_cpu_usage_ratio{process_id="1234", process="TestProcess1"} 0 1666605046048

systemruntime_cpu_usage_ratio{process_id="4321", process="TestProcess2"} 0 1666605036059
systemruntime_cpu_usage_ratio{process_id="4321", process="TestProcess2"} 0 1666605041049
systemruntime_cpu_usage_ratio{process_id="4321", process="TestProcess2"} 0 1666605046048
```

In this way, dotnet-monitor can expose multiple metrics that Prometheus can understand.

## Usage Examples
One dotnet-monitor can monitor multiple dotnet processes metrics when processes are running on a machine.

jander-msft:
As implied, we'd likely do this via labels, which has been request in #430

PR #7200, opened 2024-08-26
Multiple processes metrics

piotrkryger:
###### Summary

This PR covers functionality requested in #2761 :

 - [AllowMultipleProcessesMetrics](https://github.com/piotrkryger/dotnet-monitor/blob/faa3572000b1df1db2c3e7e8043692771a46ed5d/src/Microsoft.Diagnostics.Monitoring.Options/MetricsOptions.cs#L57) configuration flag was added; it allows to return metrics from multiple processes via `/metrics` endpoint. By default its value is `false` and `/metrics` endpoint behaves as currently: when multiple processes match default process filters, no data is returned. When set to `true`, `/metrics` endpoint can return metrics of multiple processes, and:
   - all metrics returned by `/metrics` endpoint will contain additional labels: `process_name` and `process_id`. Those labels identify process which is the source of given metric
   - `process_name` label returns name of app pool for IIS processes (that is, when system process name is `w3wp`) and "regular" system process name for other processes

###### Release Notes Entry

Added option to track metrics of multiple processes via /metrics endpoint

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsService.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsService.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi
                 return;
             }
 
-            if (!_optionsMonitor.CurrentValue.AllowMultipleProcesses)
+            if (!_optionsMonitor.CurrentValue.AllowMultipleProcessesMetrics)
             {
                 await GetMetricsFromSingleProcessAsync(stoppingToken);
             }
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsService.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsService.cs
@@ -7,6 +7,8 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Options;
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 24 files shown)


piotrkryger:
@dotnet-policy-service agree

src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsService.cs
```diff
@@ -41,37 +46,104 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+        {
+            var trackedProcesses = new Dictionary<int, Task>();
+            while (!stoppingToken.IsCancellationRequested)
+            {
+                stoppingToken.ThrowIfCancellationRequested();
+
+                try
+                {
+                    DiagProcessFilter filter = DiagProcessFilter.FromConfiguration(_processFilterMonitor.CurrentValue);
+                    IEnumerable<IProcessInfo> processes = await _services.GetProcessesAsync(filter, stoppingToken);
```
wiktork, line 113: I think we'll need a separate filter for processes we monitor vs. processes we collect metrics for.

src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsService.cs
```diff
@@ -41,37 +46,104 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+                    {
+                        SingleProcessMetricsService[] allCompleted = trackedProcesses.Values
+                            .OfType<Task<SingleProcessMetricsService>>()
+                            .Where(task => task.IsCompleted)
+                            .Select(task => task.Result)
+                            .ToArray();
+
+                        foreach (var completed in allCompleted)
+                        {
+                            _store.RemoveMetricsForPid(completed.ProcessId);
```
wiktork, line 139: Rather than keeping a list of tasks and checking their results, you could create continuations to StartMetricsPipelineForProcess that remove the pid from a set and remove the metrics from the store.

src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/MetricsStoreService.cs
```diff
@@ -4,20 +4,77 @@
...
+                _metricStores.Values.CopyTo(all, 0);
+            }
+
+            return all;
+        }
+        private static IDictionary<string, string> GetProcessLabels(IProcessInfo process)
+        {
+            return new Dictionary<string, string>()
+            {
+                { "process_id", process.EndpointInfo.ProcessId.ToString() },
```
wiktork, line 71: These will likely need to be configurable

wiktork:
Thank you for your contribution. I put some initial feedback in the PR, but please hold off on additional changes until we figure out some of these design points:
- Should we support multiple metrics on connect mode
- Configuration options: allow the same level of configurability but across N processes or keep all options the same.
- How do we extract a meaningful process name? Ideally we don't need specific understanding of w3wp or dotnet.

piotrkryger:
Thank you for your feedback. I'll wait for your decisions.

> (...)
> * How do we extract a meaningful process name? Ideally we don't need specific understanding of w3wp or dotnet.

I suppose some extensibility for "process label factory" could be introduced, similarly to egress extensibility model.

PR #7342, opened 2024-09-19
Ensure exception event source listener is running before applying startup hook

schmittjoseph:
###### Summary

Ensure that the exception event source listener pipeline is running before applying the startup hook / resuming the process. This is to prevent an issue where some early exception data events could be missed if there are sent before the event source listener pipeline is ready. This can result in the name-cache being out-of-sync and thus the reported call stacks missing information such as mvids. 

To fix this, we need to update `DiagnosticLifetimeBackgroundService` to support waiting on services to fully start before returning from `StartAsync`. 

> [!NOTE]
> This doesn't address the issue when attaching to a non-suspended .NET process that has been explicitly configured to run our startup hook via the `DOTNET_STARTUP_HOOKS` environment variable. Solving for this specific case will require additional work and is outside of the scope of this PR.

###### Release Notes Entry
Fixed an issue that could sometimes result in exceptions having empty incomplete stack information reported in exception history.


src/Tools/dotnet-monitor/Exceptions/ExceptionsService.cs
```diff
@@ -33,23 +32,23 @@ public ExceptionsService(
...
                 new EventExceptionsPipelineSettings(),
                 store);
         }
 
-        protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+
+        protected override async Task<Task> ExecuteAsync(CancellationToken stoppingToken)
         {
-            if (!_options.Value.GetEnabled() || !await _startupHookService.Applied.WaitAsync(stoppingToken))
+            if (!_options.Value.GetEnabled())
```
schmittjoseph, line 45: NOTE: This results in a different error handling behavior.  Previously if the startup hook failed to apply we wouldn't start the exception service. Now we always start the pipeline and don't react if the startup hook fails to apply. IMO this is acceptable since nothing will fail/crash in the pipeline but it's worth calling attention to.

wiktork: By not react, you mean we simply wait and do not produce any data?

wiktork (dismissed):

src/Tools/dotnet-monitor/Exceptions/ExceptionsService.cs
```diff
@@ -33,23 +32,22 @@ public ExceptionsService(
...
-            _startupHookService = startupHookService ?? throw new ArgumentNullException(nameof(startupHookService));
 
             _pipeline = new EventExceptionsPipeline(
                 new DiagnosticsClient(endpointInfo.Endpoint),
                 new EventExceptionsPipelineSettings(),
                 store);
         }
 
-        protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+        protected override async Task<Task> ExecuteAsync(CancellationToken stoppingToken)
```
jander-msft, line 42: If you need ExceptionsService to wait for the exceptions pipeline to start before returning from `IDiagnosticLifetimeService.StartAsync`, then it should implement `IDiagnosticLifetimeService` directly.

jander-msft: If that works here, then please undo the changes to DiagnosticLifetimeBackgroundService.

jander-msft (changes requested):

src/Tools/dotnet-monitor/Exceptions/EventExceptionsPipeline.cs
```diff
@@ -134,11 +134,6 @@ private void Callback(TraceEvent traceEvent)
 #endif
             }
         }
-
-        public new Task StartAsync(CancellationToken token)
```
schmittjoseph, line 138: Restore

jander-msft (approved):

src/Tools/dotnet-monitor/Exceptions/ExceptionsService.cs
```diff
@@ -33,29 +32,38 @@ public ExceptionsService(
...
             // Collect exceptions and place them into exceptions store
-            await _pipeline.RunAsync(stoppingToken);
+            await _pipeline.StartAsync(cts.Token);
         }
 
-        public override async ValueTask DisposeAsync()
+        public async ValueTask StopAsync(CancellationToken cancellationToken)
         {
-            await base.DisposeAsync();
+            await _pipeline.StopAsync(cancellationToken);
```
schmittjoseph, line 62: This will throw when exception history aren't enabled

jander-msft (approved):

PR #7383, opened 2024-10-01
Update Learning Path Staleness Check

kkeirstead:
###### Summary

I did some looking around, and I think I figured out why the learning path automation is broken. This hasn't been tested, so I'd like to check it in, run the automation, and then iterate if needed - the changes are pretty minimal and pertain only to the automation.

There are two slight bugs that I'm trying to address: the first is that if a line is no longer present in the file, the `newLineNumberFirst` and `newLineNumberLast` in theory are the same (0), in which case we try to apply a suggestion when we shouldn't (in `AssembleOutput`, we check if there's a value for `newLineNumber`, which in JS if the value is 0 this would be false). The second is that we aren't trimming the `prevContentLines`, which could cause issues for changes in tabbing.

The error being thrown was occurring when attempting to apply suggestions - it was looking for the `->` separator, but for a few files that didn't have a suggestion, there was no separator - causing it to fail. It might also make sense to add some extra handling around that so it doesn't crash and burn, but I first wanted to validate my hypothesis before doing that.

### DIFF
--- a/.github/actions/learning-path-staleness-check/index.js
+++ b/.github/actions/learning-path-staleness-check/index.js
@@ -24,7 +24,7 @@ function UpdateModifiedFiles(fileName, path, learningPathFile)
 {
   modifiedFilesUrlToFileName[path] = fileName;
 
-  modifiedFilesPathToLearningPathFile[path] = modifiedFilesPathToLearningPathFile[path] ? modifiedFilesPathToLearningPathFile[path] : new Set();;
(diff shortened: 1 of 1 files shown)


schmittjoseph (approved):
LGTM. For what it's worth, you can test these types of changes on your fork by removing this line in your repo:
https://github.com/dotnet/dotnet-monitor/blob/0476970c1bfa23c529e979fd9648900ac7183e39/.github/workflows/check-learning-path-links.yml#L11

PR #7390, opened 2024-10-02
Add System.Text.Json to Dependabot

jander-msft:


.github/dependabot.template.yml
```diff
@@ -48,6 +48,9 @@ updates:
         - "version-update:semver-major"
         - "version-update:semver-minor"
 #@ end
+      - dependency-name: System.Text.Json
```
jander-msft, line 51: This needs to be an ignore

wiktork (approved):

PR #7434, opened 2024-10-06
Replace COMPlus with DOTNET prefix

am11:
The configuration system in coreclr and mono started supporting `DOTNET_` prefix in net6.0. Using `DOTNET_` is preferred because [it has precedence](https://github.com/dotnet/runtime/blob/a3d1e1cdb61a65b64246dee45998f2701c8987b8/src/coreclr/inc/clrconfignocache.h#L81) over `COMPlus_`.

This matches the behavior by prioritizing DOTNET_ prefix and moving COMPlus_ to fallback in product code. Tests are using the new prefix.


jander-msft (approved):

jander-msft:
Thanks for updating the prefix usage!

PR #7493, opened 2024-10-10
Record `StackTraceHiddenAttribute` when collecting stacks and exceptions

schmittjoseph:
###### Summary

When collecting `/stacks` and `/exceptions` data, also record where the [`StackTraceHiddenAttribute`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.stacktracehiddenattribute?view=net-8.0) is present. To keep PRs small and easier to review,  **this PR does not change the API surface**, it is only focused on collecting this information and sending it to dotnet-monitor over event source.  A follow up PR will expose this new information and add functional tests. 

Note that we use a `UINT32` to transmit this information over event source instead of a boolean. This is due to a bug in event source regarding the size of bools. A separate PR will be made to fix an existing instance of this elsewhere in dotnet-monitor.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
@@ -90,7 +90,7 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Eventing
             SetValue(ref data[NameIdentificationEvents.ClassDescPayloads.ModuleId], ModuleId);
             SetValue(ref data[NameIdentificationEvents.ClassDescPayloads.Token], Token);
             SetValue(ref data[NameIdentificationEvents.ClassDescPayloads.Flags], Flags);
-            SetValue(ref data[NameIdentificationEvents.ClassDescPayloads.StackTraceHidden], StackTraceHidden);
+            SetBool(ref data[NameIdentificationEvents.ClassDescPayloads.StackTraceHidden], StackTraceHidden, out int _);
             SetValue(ref data[NameIdentificationEvents.ClassDescPayloads.TypeArgs], typeArgsSpan);
 
             WriteEventWithFlushing(ExceptionEvents.EventIds.ClassDescription, data);
@@ -120,7 +120,7 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.Exceptions.Eventing
             SetValue(ref data[NameIdentificationEvents.FunctionDescPayloads.ClassId], ClassId);
             SetValue(ref data[NameIdentificationEvents.FunctionDescPayloads.ClassToken], ClassToken);
             SetValue(ref data[NameIdentificationEvents.FunctionDescPayloads.ModuleId], ModuleId);
-            SetValue(ref data[NameIdentificationEvents.FunctionDescPayloads.StackTraceHidden], StackTraceHidden);
+            SetBool(ref data[NameIdentificationEvents.FunctionDescPayloads.StackTraceHidden], StackTraceHidden, out int _);
             SetValue(ref data[NameIdentificationEvents.FunctionDescPayloads.Name], namePinned);
             SetValue(ref data[NameIdentificationEvents.FunctionDescPayloads.TypeArgs], typeArgsSpan);
             SetValue(ref data[NameIdentificationEvents.FunctionDescPayloads.ParameterTypes], parameterTypesSpan);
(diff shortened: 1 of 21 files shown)


src/Profilers/MonitorProfiler/Stacks/StacksEventProvider.cpp
```diff
@@ -46,6 +46,7 @@ HRESULT StacksEventProvider::WriteClassData(ClassID classId, const ClassData& cl
         static_cast<UINT64>(classData.GetModuleId()),
         classData.GetToken(),
         static_cast<UINT32>(classData.GetFlags()),
+        classData.GetStackTraceHidden(),
```
schmittjoseph, line 49: note: this relies on an implicit cast from `bool` to `UINT32`. If folks would prefer that this become an explicit `static_cast<UINT32>` let me know.

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/TraceEventExtensions.cs
```diff
@@ -2,11 +2,17 @@
...
     internal static class TraceEventExtensions
     {
+        public static bool GetBoolPayload(this TraceEvent traceEvent, int index)
+        {
+            return Convert.ToBoolean(GetPayload<uint>(traceEvent, index));
+        }
+
         public static T GetPayload<T>(this TraceEvent traceEvent, int index)
         {
             return (T)traceEvent.PayloadValue(index);
```
schmittjoseph, line 18: note: I plan on updating this with a `DEBUG` check to prevent us accidentally calling it with `<bool>`. This will happen in a separate PR since we have an instance of using `<bool>` here elsewhere in dotnet-monitor that needs to get fixed.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Eventing/ExceptionsEventSource.cs
```diff
@@ -79,16 +79,18 @@ public void ClassDescription(
...
-            Span<EventData> data = stackalloc EventData[5];
+            Span<EventData> data = stackalloc EventData[6];
             Span<byte> typeArgsSpan = stackalloc byte[GetArrayDataSize(TypeArgs)];
             FillArrayData(typeArgsSpan, TypeArgs);
 
             SetValue(ref data[NameIdentificationEvents.ClassDescPayloads.ClassId], ClassId);
             SetValue(ref data[NameIdentificationEvents.ClassDescPayloads.ModuleId], ModuleId);
             SetValue(ref data[NameIdentificationEvents.ClassDescPayloads.Token], Token);
             SetValue(ref data[NameIdentificationEvents.ClassDescPayloads.Flags], Flags);
+            SetValue(ref data[NameIdentificationEvents.ClassDescPayloads.StackTraceHidden], StackTraceHidden);
```
jander-msft, line 93: While we are not trying to support any kind of version compatibility for the event data between .NET Monitor and the startup hook event sources, I would prefer that we would append new event data to the existing definition rather than inserting it in the middle.

schmittjoseph: This was an intentional decision -- Since like you said we're not trying to support any kind of version compatibility I'm of the opinion that we shouldn't worry about it yet and should be free to re-order events if it makes sense.

Fixed-sized payloads should generally be specified before variable-length payloads (such as strings / arrays) for improved & faster deserialization. The only exception being if we cared about backwards compatibility and utilized the `Version` attribute in which case it makes sense to either append or create a new event name / event source name altogether.

src/Profilers/MonitorProfiler/Stacks/StacksEventProvider.cpp
jander-msft: I think that's okay. But I want us to make sure that C++ implicit conversion of bool to UINT32 (`unsigned __int32`) (and that's something the native compiler is doing) is the same as .NET's conversion of bool to UInt32 (which is provided by a library, not compiled).

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/TraceEventExtensions.cs
jander-msft: I don't suppose there is a way to special case this such that if it is called with `bool` that we just do the specialization instead of the generic fallback?

schmittjoseph: I initially tried to do that, but wasn't able to figure a way of convincing msbuild to do the necessary type conversions short of something like this which I'm not a massive fan of:
```csharp
        public static T GetPayload<T>(this TraceEvent traceEvent, int index)
        {
            if (typeof(T) == typeof(bool))
            {
                return (T)(object)Convert.ToBoolean(GetPayload<uint>(traceEvent, index));
            }
            ...
```

With regards to this PR though, I can't change the default behavior of `GetPayload<bool>` since we have other code incorrectly using it today that needs to get fixed first. If there's a way of nicely special casing `GetPayload<bool>` we can do it in follow-up PR that fixes the existing code that uses it.

src/Profilers/MonitorProfiler/Stacks/StacksEventProvider.cpp
schmittjoseph: The `bool` to `UINT32` implicit conversion would be considered an integral promotion by C++ and so would be mapped to `true -> 1, false -> 0` which is identical to `Convert.ToUInt32(boolean)` which is equivalent to `return boolean ? 1u : 0u`

PR #7504, opened 2024-10-14
Update documentation around `/stacks` and `parameterTypes`

schmittjoseph:
###### Summary

Our documentation stated that parameter type information is collected as part of the `/stacks` feature -- this isn't the case. Update our documentation.


wiktork:
I have a list of /stack improvements, I'll make an item out of it for future reference.

wiktork (approved):

PR #7511, opened 2024-10-15
Use `JsonSerializer` to serialize exception data

schmittjoseph:
###### Summary

Switch over to using `JsonSerializer` to serialize our exception data instead of manually serializing. This will make it less likely for us to accidentally introduce bugs here, standardize with our other features, and also make writing more complex json tests for `/exceptions` easier.

**This PR does not change the API surface**.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Models/CallStackResults.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Models/CallStackResults.cs
@@ -15,6 +15,13 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Models
         [JsonPropertyName("methodToken")]
         public uint MethodToken { get; set; }
 
+        [JsonPropertyName("parameterTypes")]
+        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+        public IList<string>? FullParameterTypes { get; set; }
+
+        [JsonIgnore]
+        internal IList<string>? SimpleParameterTypes { get; set; }
+
         [JsonPropertyName("typeName")]
         public string TypeName { get; set; } = string.Empty;
 
@@ -30,11 +37,6 @@ namespace Microsoft.Diagnostics.Monitoring.WebApi.Models
         [JsonIgnore]
         internal IList<string> FullGenericArgTypes { get; set; } = new List<string>();
 
-        [JsonIgnore]
-        internal IList<string> SimpleParameterTypes { get; set; } = new List<string>();
-
-        [JsonIgnore]
-        internal IList<string> FullParameterTypes { get; set; } = new List<string>();
         //TODO Bring this back once we have a relative il offset value.
         //[JsonPropertyName("offset")]
         //public ulong Offset { get; set; }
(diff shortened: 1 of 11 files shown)


src/Tools/dotnet-monitor/Exceptions/ExceptionsOperation.cs
```diff
@@ -141,81 +139,28 @@ private async Task WriteJsonInstance(Stream stream, IExceptionInstance instance,
...
-                    writer.WriteStartArray("frames");
-
-                    StringBuilder builder = new StringBuilder();
-
-                    foreach (var frame in instance.CallStack.Frames)
-                    {
-                        writer.WriteStartObject();
-
-                        string assembledMethodName = frame.MethodName;
-                        if (frame.FullGenericArgTypes.Count > 0)
```
schmittjoseph, line 189: It looks like `/stacks` and `/exceptions` diverge here. Exception frames include generic arg types while stacks don't... I'll work on restoring this behavior for `/exceptions`

src/Microsoft.Diagnostics.Monitoring.WebApi/Models/CallStackResults.cs
```diff
@@ -1,20 +1,46 @@
...
+        public string MethodNameWithGenericArgTypes
+        {
+            get
+            {
+                StringBuilder builder = new(MethodName);
+                NameFormatter.BuildGenericArgTypes(builder, FullGenericArgTypes);
+                return builder.ToString();
+            }
+            // Only intended for use test code.
+            set
```
wiktork, line 24: internal set

schmittjoseph: When it's `internal`, `JsonSerializer` is unable to set this field in our tests when we deserialize our api output

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/StackUtilities.cs
```diff
@@ -17,23 +17,22 @@ internal enum StackFormat
 
     internal static class StackUtilities
     {
-        public static Models.CallStack TranslateCallStackToModel(CallStack stack, NameCache cache, bool methodNameIncludesGenericParameters = true)
+        public static Models.CallStack TranslateCallStackToModel(CallStack stack, NameCache cache, bool parameterTypesSupported = true)
```
jander-msft, line 20: Can we change the parameter to indicate what it's actually doing? Something like `ensureParameterTypeFieldsNotNull`. I don't know what effect `parameterTypesSupported` is supposed to have without reading the method body.

src/Microsoft.Diagnostics.Monitoring.WebApi/Utilities/StackUtilities.cs
```diff
@@ -17,23 +17,22 @@ internal enum StackFormat
...
             stackModel.ThreadId = stack.ThreadId;
             stackModel.ThreadName = stack.ThreadName;
 
-            StringBuilder builder = new();
             foreach (CallStackFrame frame in stack.Frames)
             {
                 var frameModel = CreateFrameModel(frame, cache);
-                if (methodNameIncludesGenericParameters)
+
+                if (parameterTypesSupported)
```
jander-msft, line 30: Consider adding this as a parameter on `CreateFrameModel` and let it handle all the semantics of what should be included in the model.

jander-msft (approved):

PR #7525, opened 2024-10-17
Refactor server endpoint management

schmittjoseph:
###### Summary

This PR refactors how server endpoint management is done. **There are no functional changes**.  Currently there is a single service responsible for accepting new endpoints, enforcing a pruning policy, and determining if an endpoint is active. Now it is broken into 3 distinct services:
- `ServerEndpointInfoSource: IEndpointInfoSourceInternal`
  - Responsible for setting up the endpoint server, accepting connections, and tearing down endpoints when they are no longer used (not determined by this service)
- `IServerEndpointTracker`
  - Responsible for tracking endpoints accepted by `ServerEndpointInfoSource` and enforcing a pruning policy. When this class determines a tracked endpoint should be removed, it signals an event that `ServerEndpointInfoSource` is subscribed to.
- `IServerEndpointStateChecker`
  - Responsible for determining the current state of a given endpoint (`Active`, `Unresponsive`, or `Error`). It does not enforce pruning policy. 

This new separation will allow us to:
- Easily change any 1 part of this system independent of the rest. For example we could introduce a new pruning policy without impacting the server or endpoint state checking, or we could swap out how endpoint state is determined.
- Add more unit tests in the future since we can mock different layers of this system.


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTestCommon/EndpointUtilities.cs
```diff
@@ -72,7 +73,16 @@ public async Task<ServerSourceHolder> StartServerAsync(EndpointInfoSourceCallbac
...
-            ServerEndpointInfoSource source = new(scopeFactoryMock.Object, portOptions, callbacks, operationTrackerService);
+            IServerEndpointStateChecker endpointChecker = new ServerEndpointStateChecker(operationTrackerService);
+
+            IServerEndpointTracker endpointTracker = new ServerEndpointTracker(endpointChecker);
+
+            ServerEndpointInfoSource source = new(
+                scopeFactoryMock.Object,
+                endpointTracker,
+                portOptions,
+                Mock.Of<ILogger<ServerEndpointInfoSource>>(),
```
jander-msft, line 84: nit: `NullLogger<T>.Instance` can be used instead of a mock

src/Tools/dotnet-monitor/EndpointInfo/ServerEndpointInfoSource.cs
```diff
@@ -123,6 +117,9 @@ public async ValueTask DisposeAsync()
                 {
                 }
             }
+
+            // Clear all endpoints
+            await _endpointTracker.ClearAsync(CancellationToken.None);
```
jander-msft, line 122: I think this should be the responsibility of the IEndpointTracker implementation, presumably in an IAsyncDisposable.DisposeAsync method implementation rather than this implementation telling it how to manage its endpoint list.

src/Tools/dotnet-monitor/EndpointInfo/IServerEndpointTracker.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Threading.Tasks;
+using System.Threading;
+using System;
+
+namespace Microsoft.Diagnostics.Tools.Monitor
+{
+    internal record class EndpointRemovedEventArgs(IEndpointInfo Endpoint, ServerEndpointState State);
+
+    internal interface IServerEndpointTracker
+        : IHostedService
```
jander-msft, line 16: This (`IHostedService`) should not be the base of the `IServerEndpointTracker` interface. It is an implementation detail of the `ServerEndpointTracker` implementation. Please remove and make sure that the implementation is registered as both an `IHostedService` and an `IServerEndpointTracker`.

schmittjoseph: Agreed, this PR is currently draft since I'm mostly looking for PR builds to pass before opening it up

jander-msft: Sorry, missed that it was a draft, or maybe it changed while I was reviewing it.

jander-msft (approved):

wiktork (approved):

PR #7571, opened 2024-10-28
[Experimental] Add new server endpoint pruning algorithm

schmittjoseph:
###### Summary

This PR adds a new, experimental, pruning algorithm for server endpoints. It is explicitly opt-in via an undocumented environment variable (`DotnetMonitor_Experimental_ServerEndpointPruningAlgorithmV2`). As we try this new mode out and build confidence in it, we can replace the current algorithm with it.

New Algorithm:
- When retrieving a list of active endpoints (`IServerEndpointTracker.GetEndpointInfoAsync`) return the last set of validated endpoints. It does **not** re-check the endpoints.
- Periodically check the state of endpoints. If one becomes unresponsive, it now has a 1 minute grace period to become active again before being pruned. While the endpoint is unresponsive but in this grace period, we still report the endpoint as active.

NOTE: I haven't updated any integration tests to use this new pruning algorithm as it's opt-in, undocumented.

### DIFF
--- a/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfo/ServerEndpointTrackerV2Tests.cs
+++ b/src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfo/ServerEndpointTrackerV2Tests.cs
@@ -1,0 +1,168 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Diagnostics.Monitoring.TestCommon;
+using Microsoft.Diagnostics.Monitoring.WebApi;
+using Microsoft.Diagnostics.Tools.Monitor;
+using Microsoft.Extensions.Options;
+using Moq;
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.Diagnostics.Monitoring.Tool.UnitTests.EndpointInfo
+{
+    [TargetFrameworkMonikerTrait(TargetFrameworkMonikerExtensions.CurrentTargetFrameworkMoniker)]
+    public sealed class ServerEndpointTrackerV2Tests
+    {
+        private readonly MockTimeProvider _timeProvider = new();
+        private readonly IOptions<DiagnosticPortOptions> _diagPortOptions = Extensions.Options.Options.Create(new DiagnosticPortOptions()
+        {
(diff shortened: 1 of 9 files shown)


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfo/ServerEndpointTrackerV2Tests.cs
```diff
@@ -0,0 +1,168 @@
...
+            // Act
+            await tracker.PruneEndpointsAsync(cts.Token);
+
+            // Assert
+            IEnumerable<IEndpointInfo> activeEndpoints = await tracker.GetEndpointInfoAsync(cts.Token);
+            Assert.Empty(activeEndpoints);
+
+            Assert.NotNull(endpointRemovedArgs);
+            Assert.Equal(ServerEndpointState.Error, endpointRemovedArgs.State);
+            Assert.Equal(endpointRemovedArgs.Endpoint, endpoint);
```
jander-msft, line 88: I think these parameters should be switched: `endpoint` is the expected value whereas `endpointRemovedArgs.Endpoint` is the tested value.

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfo/ServerEndpointTrackerV2Tests.cs
```diff
@@ -0,0 +1,168 @@
...
+            // Act
+            await tracker.PruneEndpointsAsync(cts.Token);
+
+            // Assert
+            IEnumerable<IEndpointInfo> activeEndpoints = await tracker.GetEndpointInfoAsync(cts.Token);
+            Assert.Empty(activeEndpoints);
+
+            Assert.NotNull(endpointRemovedArgs);
+            Assert.Equal(ServerEndpointState.Unresponsive, endpointRemovedArgs.State);
+            Assert.Equal(endpointRemovedArgs.Endpoint, endpoint);
```
jander-msft, line 165: I think these parameters should be switched: `endpoint` is the expected value whereas `endpointRemovedArgs.Endpoint` is the tested value.

jander-msft (dismissed):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTests/EndpointInfo/ServerEndpointTrackerV2Tests.cs
schmittjoseph: Whoops, good catch

jander-msft (approved):

PR #7658, opened 2024-11-12
Remove `release/6.x` from workflows

schmittjoseph:
###### Summary

.NET Monitor 6 is no longer supported, remove it from dependabot and branch syncing


.github/dependabot.template.yml
```diff
@@ -10,7 +10,6 @@
 #@   return {
 #@     "9": [ "net9.0", "net8.0", "net7.0", "net6.0" ],
```
schmittjoseph, line 11: note: I wasn't sure if we still wanted to references net6.0 here or not. Given that net7.0 is still hanging around I opted to keep it, at least for this PR

jander-msft (approved):

PR #7914, opened 2025-02-03
Force callstacks to only run on untainted threads

wiktork:
###### Summary
Use another thread to process non-managed commands in the profiler. This fixes an issue where /parameters (which causes a managed callback in the inproc code) breaks /stacks (which requires a thread that has never ran managed code).

###### Release Notes Entry
Fix an issue where `/parameters` causes `/stacks` to return CORPROF_E_UNSUPPORTED_CALL_SEQUENCE


jander-msft (approved):

jander-msft:
Let's add release notes for this change and backport to all in-support versions.

jander-msft (approved):

PR #7926, opened 2025-02-06
Change deployment samples to be ready-to-deploy files

jander-msft:
###### Summary

This change allows users to easily run the sample deployments without having to copy/paste them first; just checkout the repository and run the deployment. For Kubernetes, I've only created a single sample based on .NET Monitor 8 since the prior versions are out of support and the sample applies to .NET Monitor 8 and forward. Additionally, I changed the app image to use the chiseled variant so that the users of the app and the .NET Monitor container match.


samples/Kubernetes/deployment.yaml
```diff
@@ -0,0 +1,61 @@
...
+        volumeMounts:
+        - mountPath: /diag
+          name: diagvol
+        resources:
+          limits:
+            cpu: 250m
+            memory: 512Mi
+      - name: monitor
+        image: mcr.microsoft.com/dotnet/monitor:8
+        # DO NOT use the --no-auth argument for deployments in production; this argument is used for demonstration
```
kkeirstead, line 33: Just a bit of a quirk with some of our comments in a couple of places - if someone references this yaml in isolation, these don't mean much and don't tell you where to get clarification.

samples/Docker/compose.yaml
```diff
@@ -0,0 +1,51 @@
...
+    volumes:
+      - diagvol:/diag
+    deploy:
+      resources:
+        limits:
+          cpus: '0.250'
+          memory: 512M
+
+  monitor:
+    image: mcr.microsoft.com/dotnet/monitor:8
```
kkeirstead, line 18: Any reason we're using 8 for these examples instead of 9?

wiktork (approved):

samples/Docker/compose.yaml
jander-msft: Just copy/paste, we should fix up docs to reference the latest major tag wholesale.

samples/Kubernetes/deployment.yaml
jander-msft: I don't think many folks will find this file in isolation. They are likely to get to it from the kubernetes and docker markdown files. It might be good to have readme markdowns for each sample that explain the nuances right next to these files. I'll think about it and propose a separate change for this.

PR #7974, opened 2025-02-28
Add `MVID` and `MethodToken` To Parameters

kkeirstead:
###### Summary

Added `MVID` and `MethodToken` to Parameter Capture output.

###### Release Notes Entry

Added `ModuleVersionId` and `MethodToken` Fields to Parameter Capture Output

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/ParameterCapturing/Eventing/ParameterCapturingEventSource.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/ParameterCapturing/Eventing/ParameterCapturingEventSource.cs
@@ -125,7 +125,7 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.ParameterCapturing.Eventi
             string methodModuleName,
             string methodDeclaringTypeName,
             uint methodToken,
-            Guid mvid
+            Guid moduleVersionId
             )
         {
             Span<EventData> data = stackalloc EventData[10];
@@ -144,7 +144,7 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.ParameterCapturing.Eventi
             SetValue(ref data[ParameterCapturingEvents.CapturedParametersStartPayloads.MethodModuleName], pinnedMethodModuleName);
             SetValue(ref data[ParameterCapturingEvents.CapturedParametersStartPayloads.MethodDeclaringTypeName], pinnedMethodDeclaringTypeName);
             SetValue(ref data[ParameterCapturingEvents.CapturedParametersStartPayloads.MethodToken], methodToken);
-            SetValue(ref data[ParameterCapturingEvents.CapturedParametersStartPayloads.Mvid], mvid);
+            SetValue(ref data[ParameterCapturingEvents.CapturedParametersStartPayloads.ModuleVersionId], moduleVersionId);
 
             WriteEventWithFlushing(ParameterCapturingEvents.EventIds.ParametersCapturedStart, data);
         }
--- a/src/Microsoft.Diagnostics.Monitoring.StartupHook/ParameterCapturing/Eventing/ParameterCapturingEventSource.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.StartupHook/ParameterCapturing/Eventing/ParameterCapturingEventSource.cs
@@ -123,10 +123,12 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.ParameterCapturing.Eventi
             int managedThreadId,
             string methodName,
             string methodModuleName,
-            string methodDeclaringTypeName
+            string methodDeclaringTypeName,
+            uint methodToken,
+            Guid mvid
             )
         {
-            Span<EventData> data = stackalloc EventData[8];
+            Span<EventData> data = stackalloc EventData[10];
 
             using PinnedData pinnedActivityId = PinnedData.Create(activityId);
             using PinnedData pinnedMethodName = PinnedData.Create(methodName);
@@ -141,6 +143,8 @@ namespace Microsoft.Diagnostics.Monitoring.StartupHook.ParameterCapturing.Eventi
             SetValue(ref data[ParameterCapturingEvents.CapturedParametersStartPayloads.MethodName], pinnedMethodName);
             SetValue(ref data[ParameterCapturingEvents.CapturedParametersStartPayloads.MethodModuleName], pinnedMethodModuleName);
             SetValue(ref data[ParameterCapturingEvents.CapturedParametersStartPayloads.MethodDeclaringTypeName], pinnedMethodDeclaringTypeName);
+            SetValue(ref data[ParameterCapturingEvents.CapturedParametersStartPayloads.MethodToken], methodToken);
+            SetValue(ref data[ParameterCapturingEvents.CapturedParametersStartPayloads.Mvid], mvid);
 
             WriteEventWithFlushing(ParameterCapturingEvents.EventIds.ParametersCapturedStart, data);
         }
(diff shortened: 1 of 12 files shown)


src/Microsoft.Diagnostics.Monitoring.StartupHook/ParameterCapturing/Eventing/ParameterCapturingEventSource.cs
```diff
@@ -123,10 +123,12 @@ public void CapturedParameterStart(
             int managedThreadId,
             string methodName,
             string methodModuleName,
-            string methodDeclaringTypeName
+            string methodDeclaringTypeName,
+            uint methodToken,
```
schmittjoseph, line 127: These are fixed-length values, so they should be placed before the variable-length ones to make event source more efficient (e.g. place them before the first `string`). While this would technically a breaking change in the event source contract (appending isn't), we're the only consumer of this so we can safely do it.

src/Microsoft.Diagnostics.Monitoring.StartupHook/ParameterCapturing/MethodSignature.cs
```diff
@@ -46,6 +46,10 @@ public static class Names
 
         public string MethodName { get; } = GetMethodName(method);
 
+        public int MethodToken { get; } = method.MetadataToken;
+
+        public Guid ModuleVersionId { get; } = method.Module.ModuleVersionId;
+
```
schmittjoseph, line 52: I'm not 100% certain yet that this is actually safe-to-do here. These properties may throw on access for certain methods. I don't believe the parameter capturing infrastructure ever tries to access either of these properties except for when it goes to calculate ancillary boxing instructions which is guarded against this problem.

The rest of the infrastructure operates on function ids. It does exclude dynamic assemblies from being instrumented.

We either need to guard when accessing these, or validate that it is **always** safe to do given the parameter capturing infrastructure design. I'll think on it more, but it might be safest just to guard.

kkeirstead: Assuming we catch an exception, do we just want these 0'd out (in the case of MVID just an empty GUID), or should we specifically indicate that there was no value?

schmittjoseph: Let's use 0'd out values as backups, this will be consistent with what we do for stack frames that don't have this data

schmittjoseph:
Let's also add/update a function test for this --`src\Tests\Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests\ParameterCapturingTests.cs` already has several, and the `Exceptions` functional tests should already have an example of how to get the expected mvid & mdToken IIRC.

src/Tools/dotnet-monitor/ParameterCapturing/ParameterCapturingEvents.cs
```diff
@@ -67,12 +67,14 @@ public static class CapturedParametersStartPayloads
         {
             public const int RequestId = 0;
             public const int CaptureId = 1;
-            public const int ActivityId = 2;
-            public const int ActivityIdFormat = 3;
-            public const int ThreadId = 4;
-            public const int MethodName = 5;
-            public const int MethodModuleName = 6;
-            public const int MethodDeclaringTypeName = 7;
+            public const int MethodToken = 2;
```
wiktork, line 70: This is an interesting scenario to consider for situations where the inproc component stays the same while dotnet-monitor changes. I don't think we have to worry about that yet, but it seems like it would be possible to make this non-breaking change if we wanted to.

schmittjoseph: See https://github.com/dotnet/dotnet-monitor/pull/7974#discussion_r1975912301 -- this is an intention decision and inline with past decisions around this (https://github.com/dotnet/dotnet-monitor/pull/7493#discussion_r1799777520)

wiktork (dismissed):

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/ParameterCapturingTests.cs
```diff
@@ -157,6 +157,9 @@ await RunTestCaseCore(
 
                 Assert.Equal(expectedCapturedMethod.TypeName, actualMethod.TypeName);
                 Assert.Equal(expectedCapturedMethod.MethodName, actualMethod.MethodName);
+                Assert.Equal(expectedCapturedMethod.ModuleName, actualMethod.ModuleName);
+                Assert.NotEqual(Guid.Empty, actualMethod.ModuleVersionId);
```
schmittjoseph, line 161: We can calculate the actual expected values here. I believe the Exceptions functional tests already do this

src/Microsoft.Diagnostics.Monitoring.WebApi/ParameterCapturing/CapturedParametersTextFormatter.cs
```diff
@@ -20,7 +20,7 @@ protected override async Task WriteCapturedMethodAsync(CapturedMethod capture, C
         {
             StringBuilder builder = new();
 
-            builder.AppendLine(FormattableString.Invariant($"[{capture.CapturedDateTime}][thread {capture.ThreadId}] {GetValueOrUnknown(capture.ActivityId)}[format: {capture.ActivityIdFormat}]"));
+            builder.AppendLine(FormattableString.Invariant($"[{capture.CapturedDateTime}][thread {capture.ThreadId}] {GetValueOrUnknown(capture.ActivityId)}[format: {capture.ActivityIdFormat}][methodToken: {capture.MethodToken}][moduleVersionId: {capture.ModuleVersionId}]"));
```
schmittjoseph, line 23: We probably don't care about the mvid/methodtoken when doing text formatting, but the feature is experimental so we can always change this in the future

schmittjoseph (approved):

src/Microsoft.Diagnostics.Monitoring.WebApi/ParameterCapturing/CapturedParametersTextFormatter.cs
jander-msft: Agreed, text output shouldn't contain tokens in it.

schmittjoseph (approved):

src/Microsoft.Diagnostics.Monitoring.StartupHook/ParameterCapturing/MethodSignature.cs
```diff
@@ -58,6 +62,30 @@ private static string GetMethodName(MethodInfo method)
             return builder.ToString();
         }
 
+        private static int GetMethodToken(MethodInfo method)
+        {
+            try
+            {
+                return method.MetadataToken;
+            }
+            catch
```
jander-msft, line 71: Do we actually get exceptions under any circumstances? If not, we should probably not wrap these with generic exception handling.

schmittjoseph: See https://github.com/dotnet/dotnet-monitor/pull/7974#discussion_r1975922750

schmittjoseph (dismissed):

src/Microsoft.Diagnostics.Monitoring.StartupHook/ParameterCapturing/MethodSignature.cs
jander-msft: I think comments in source would help alleviate someone from wondering why these are guarded.

PR #7977, opened 2025-03-01
Add `Capabilities` to `/info` Route

kkeirstead:
###### Summary

Adds a list of capabilities to the `/info` route - these are always present, and have an `enabled` property for consumers to determine if a feature is present. Note that this does not include information about _how_ a feature is configured (i.e. it may be enabled, but not configured correctly for a given use case). If we'd like to adjust which properties are included just let me know.

###### Release Notes Entry

Add `Capabilities` to `/info` route.


kkeirstead:
Note to self - update docs for `DotnetMonitorInfo` and the `/info` route example

schmittjoseph:
Let's be sure to also add tests for this

src/Microsoft.Diagnostics.Monitoring.Options/MonitorCapability.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.Diagnostics.Monitoring.Options
+{
+    public class MonitorCapability : IMonitorCapability
```
schmittjoseph, line 6: I'm not sure if this class adds value here. We already have `IMonitorCapability` and we can have simple concrete implementations for each capability. Each concrete implementation would then be able to have isolated logic to determine if the feature is on or not (some abstraction may be wanted for in-proc features).

kkeirstead: @jander-msft can weigh in, but if I'm understanding this correctly that's more aligned with the initial approach I took, and then we transitioned to this model that made everything standardized

wiktork: Maybe a compromise here would be that the class structure stays the same but the default implementing class doesn't have upfront knowledge about all of its capabilities. This could be done at the registration level.

wiktork: We can also make the interface public but the class internal

src/Tools/dotnet-monitor/CapabilitiesPostConfigureOptions.cs
```diff
@@ -0,0 +1,51 @@
...
+using Microsoft.Diagnostics.Monitoring.Options;
+using Microsoft.Diagnostics.Monitoring.WebApi;
+using Microsoft.Extensions.Options;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Diagnostics.Monitoring
+{
+    public class CapabilityPostConfigureOptions :
```
wiktork, line 13: Consider pushing this to base class functionality without any vector based storage. As it is right now, this class is instantiated 4 times, each instance supporting all 4 choices, but only actually being called against one.

kkeirstead: I refactored this out, let me know if you had something else in mind.

src/Microsoft.Diagnostics.Monitoring.Options/MonitorCapability.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.ComponentModel.DataAnnotations;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Diagnostics.Monitoring.Options
+{
+    public class MonitorCapability
```
schmittjoseph, line 9: This seems like it could just be an interface

src/Microsoft.Diagnostics.Monitoring.Options/MonitorCapability.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        public const string Exceptions = "exceptions";
```
schmittjoseph, line 11: IMO it's a bit odd for this class to have these constants. Let's move them to be owned by the classes the define each capability (e.g. `CallStacksCapabilityPostConfigureOptions` would define the `CallStacks` const)

kkeirstead: What did you have in mind for capabilities that don't have a `PostConfigureOptions` (i.e. `HttpEgress`)?

schmittjoseph: We could have a separate static class holding these constants if that's easier, my larger point here is that it's odd for this class in particular to own it, especially since it's public

src/Tools/dotnet-monitor/CapabilitiesPostConfigureOptions.cs
```diff
@@ -0,0 +1,28 @@
...
+    public abstract class CapabilityPostConfigureOptions<T> : IPostConfigureOptions<T> where T : class
+    {
+        private MonitorCapability? _capability;
+
+        public CapabilityPostConfigureOptions(MonitorCapability? capability)
+        {
+            _capability = capability;
+        }
+
+        public abstract void PostConfigure(string? name, T options);
```
wiktork, line 18: I think this is ok, but just to offer some additional options for consideration:
- Override seal PostConfigure and have a abstract method similiar to `abstract bool GetIsEnabled(T options). The base class PostConfigure implementation can call GetIsEnabled and PostConfigure(bool) in the implementation.
- Alternatively, Make CapabilityPostConfigureOptions<T> non-abstract and have it take an additonal parameter of Func<T, bool>. During registration this might look something like `RegisterCapability<CallStackOptions>( (CallStackOptions options) => options.Enabled`

src/Tools/dotnet-monitor/ServiceCollectionExtensions.cs
```diff
@@ -240,6 +241,25 @@ public static IServiceCollection ConfigureStorage(this IServiceCollection servic
             return services;
         }
 
+        public static IServiceCollection ConfigureCapabilities(this IServiceCollection services, bool noHttpEgress)
+        {
+            services.AddSingleton(new MonitorCapability() { Name = MonitorCapabilityConstants.Metrics });
```
wiktork, line 246: nit: Consider creating a ConfigureCapability function that handles the necessary singleton registration for a single capability that requires post configuration.

src/Microsoft.Diagnostics.Monitoring.Options/MonitorCapability.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Diagnostics.Monitoring.Options
+{
+    public class MonitorCapability
+    {
+        [Required]
+        [JsonPropertyName("name")]
+        public string Name { get; set; } = string.Empty;
```
jander-msft, line 13: If this is required, then let's remove the default value and put a `required` keyword on it.

src/Microsoft.Diagnostics.Monitoring.Options/MonitorCapabilityConstants.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.Diagnostics.Monitoring.Options
+{
+    internal static class MonitorCapabilityConstants
+    {
+        public const string Exceptions = "exceptions";
+        public const string ParameterCapturing = "parameters";
+        public const string CallStacks = "callstacks";
```
jander-msft, line 10: nit: Call stacks is two words; let's put an underscore between the two words:
```suggestion
        public const string CallStacks = "call_stacks";
```

src/Tests/Microsoft.Diagnostics.Monitoring.Tool.FunctionalTests/InfoTests.cs
```diff
@@ -67,7 +70,27 @@ public Task InfoEndpointValidationTest(DiagnosticPortConnectionMode mode)
                         Assert.Equal(runner.DiagnosticPortPath, info.DiagnosticPortName);
                     }
 
+                    Assert.Equal(5, info.Capabilities.Length); // Update if capabilities change
+                    Assert.Contains(info.Capabilities, capability => capability.Name == MonitorCapabilityConstants.Exceptions);
```
jander-msft, line 74: These checks can be refactored into something like:
```cs
AssertCapability(true, MonitorCapabilityConstants.HttpEgress);
AssertCapability(enableInProcessFeatures, MonitorCapabilityConstants.Exceptions);
...
void static AssertCapability(bool expectEnabled, string capabilityName)
{
    MonitorCapability? capability = info.Capabilities.First(c => c.Name == capabilityName);
    Assert.NotNull(capability, $"Expected {capabilityName} capability to exist.");
    Assert.Equal(expectedEnabled, capability.Enabled);
}
```

src/Tools/dotnet-monitor/ServiceCollectionExtensions.cs
```diff
@@ -240,6 +241,28 @@ public static IServiceCollection ConfigureStorage(this IServiceCollection servic
...
+            services.AddSingleton(new MonitorCapability() { Name = MonitorCapabilityConstants.HttpEgress, Enabled = !noHttpEgress });
+
+            return services;
+        }
+
+        private static IServiceCollection ConfigureCapability<TOptions, TPostOptions>(this IServiceCollection services, string capabilityName)
+            where TOptions : class
+            where TPostOptions : class, IPostConfigureOptions<TOptions>
+        {
+            services.AddSingleton(new MonitorCapability() { Name = capabilityName });
```
jander-msft, line 260: I'm going to proposal an alternative design that (1) allows the capability properties to be read-only, (2) doesn't require relying on the options construction infrastructure, and (3) avoids the possibility of forgetting to register a `MonitorCapability` and/or a `IPostConfigureOptions<T>` pairing:
- Add a `IMonitorCapability` interface with `string Name` and `bool Enabled` getter-only properties.
- Instead of `IPostConfigureOptions<T>` implementations, just implement `IMonitorCapability` for each capability. This implementation can import `IOptions<T>` and then just set the `Enabled` property in its constructor.
- Add each `IMonitorCapability` implementation as a singleton.
- Still use `MonitorCapability` for API model purposes; this can be a projection of the `IMonitorCapability` implementations.

kkeirstead: I was a little unsure about the last point on here, but I think the latest changes should more-so be following what you had in mind.

src/Microsoft.Diagnostics.Monitoring.Options/MonitorCapability.cs
```diff
@@ -1,18 +1,22 @@
...
 
 namespace Microsoft.Diagnostics.Monitoring.Options
 {
-    public class MonitorCapability
+    public class MonitorCapability : IMonitorCapability
     {
-        [Required]
         [JsonPropertyName("name")]
-        public string Name { get; set; } = string.Empty;
+        public string Name { get; init; }
```
jander-msft, line 11: I would add the `[Required]` attribute back to make the OpenAPI description such that the property is not nullable.

jander-msft, line 11: Can the init be removed from both of the properties since they are set through the constructor? I think STJ should still include them according to https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializeroptions.ignorereadonlyproperties?view=net-9.0

jander-msft (approved):

jander-msft (approved):

jander-msft (approved):

PR #7984, opened 2025-03-03
Add managed entry point to process info and collection rule filters

jander-msft:


src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/AppRunner.cs
```diff
@@ -22,6 +22,8 @@ namespace Microsoft.Diagnostics.Monitoring.TestCommon.Runners
     [DebuggerDisplay(@"\{AppRunner:{_runner.StateForDebuggerDisplay,nq}\}")]
     public sealed class AppRunner : IAsyncDisposable
     {
+        public const string AssemblyNameWithoutExtension = "Microsoft.Diagnostics.Monitoring.UnitTestApp";
```
wiktork, line 25: nit: can we use nameof or typeof for this?

wiktork (approved):

PR #8008, opened 2025-03-14
Use openapi generator

sbomer:
Switches from Swashbuckle.AspNetCore.SwaggerGen to Microsoft.AspnetCore.OpenApi.


src/Tools/dotnet-monitor/OpenApi/OpenApiOptionsExtensions.cs
```diff
@@ -0,0 +1,239 @@
...
+                            var schema = response.Value.Content[ContentTypes.ApplicationJson].Schema;
+                            schema.AdditionalProperties.AdditionalPropertiesAllowed = false;
+                        }
+                    }
+                }
+                return Task.CompletedTask;
+            });
+
+            // Add missing descriptions on types
+            options.AddSchemaTransformer((schema, context, cancellationToken) => {
```
jander-msft, line 167: Could these descriptions be added directly to the types with the DescriptionAttribute? If those are not detected, then could you change this transformer to check if associated schema type has a DescriptionAttribute and apply its value to the schema? This would make it a more generic solution.

src/Tools/dotnet-monitor/OpenApi/OpenApiOptionsExtensions.cs
```diff
@@ -0,0 +1,239 @@
...
+                            Type = ReferenceType.Schema,
+                            Id = nameof(LogLevel)
+                        }
+                    };
+                }
+                return Task.CompletedTask;
+            });
+
+            // Fix up "additionalProperties" for the response type of GetCollectionRulesDescription
+            options.AddOperationTransformer((operation, context, cancellationToken) => {
```
jander-msft, line 151: Any idea why this particular operation needs this type of fix?

documentation/openapi.json
```diff
@@ -2053,7 +2061,8 @@
           }
         },
         "additionalProperties": false,
-        "description": "Represents the details of a given process used in an operation."
+        "description": "Represents the details of a given process used in an operation.",
+        "nullable": true
```
jander-msft, line 2065: Do you know why some of these object schemas have `"nullable": true` on them? I can rationalize the one on OperationError (because all of its properties are nullable), but I can't for this schema. It feels strange to me that an object schema itself would declare that it is nullable rather than the usage of the schema would also allow null.

src/Microsoft.Diagnostics.Monitoring.Options/MonitorCapability.cs
```diff
@@ -10,15 +10,15 @@ public class MonitorCapability : IMonitorCapability
...
-        public string Name { get; }
+        [MinLength(1)]
+        public string Name { get; set; }
 
         [JsonPropertyName("enabled")]
-        public bool Enabled { get; }
+        public bool Enabled { get; init; }
 
-        public MonitorCapability(string name, bool enabled)
+        public MonitorCapability(string name)
```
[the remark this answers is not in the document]
jander-msft: I think making both required is fine. This type is only reported via the HTTP API and is not provided by the client (e.g. as a request body).

documentation/openapi.json
```diff
@@ -351,7 +357,8 @@
             "in": "query",
             "description": "The profiles enabled for the trace session.",
             "schema": {
-              "$ref": "#/components/schemas/TraceProfile"
```
jander-msft, line 354: This doesn't look right to me. The parameter type is an enum. The parameter schema in OpenAPI shouldn't be just a `string` with a default value.

documentation/openapi.json
```diff
@@ -1725,15 +1734,13 @@
           "ActionExecuting",
           "Throttled",
           "Finished"
-        ],
-        "type": "string"
```
jander-msft, line 1729: Every example of an enum schema that I've run across in OpenAPI has the `type` property set to `string`. Could you add a schema transformer that fixes up enum schemas?

jander-msft (approved):

PR #8013, opened 2025-03-17
Use minimal APIs

sbomer:
Moves from controllers to minimal APIs.


src/Tools/dotnet-monitor/Startup.cs
```diff
@@ -29,18 +30,8 @@ public Startup(IConfiguration configuration)
...
-            .AddApplicationPart(typeof(DiagController).Assembly);
-
-            services.AddControllers(options =>
-            {
-                options.Filters.Add(typeof(EgressValidationUnhandledExceptionFilter));
+            services.ConfigureHttpJsonOptions(options => {
+                options.SerializerOptions.Converters.Add(new JsonStringEnumConverter());
             });
 
             services.Configure<ApiBehaviorOptions>(options =>
```
jander-msft, line 37: `ApiBehaviorOptions` seems to be an MVC concept. Could you check to see if this is required?

src/Tools/dotnet-monitor/Startup.cs
```diff
@@ -96,9 +92,17 @@ public static void Configure(IApplicationBuilder app, IWebHostEnvironment env, I
 
             app.UseEndpoints(builder =>
             {
-                builder.MapControllers();
+                var serviceProvider = builder.ServiceProvider;
```
jander-msft, line 95: This seems to be unused.

jander-msft (approved):
I think this change looks good. I'm going to ask one of the other folks from @dotnet/dotnet-monitor to take a look as well.

PR #8067, opened 2025-04-08
[main] Remove .NET 7 and lower references except for startup hook

jander-msft:
###### Summary

- Remove all dependencies for .NET 7 and lower (except for .NET 6 for the startup hook).
- Remove all tests variants for .NET 7 and lower
- Remove installation of .NET 7 runtimes and lower.
- Remove preprocessor directives for .NET 8 and lower and make minor code adjustments as necessary.
- Remove dependabot updates for .NET 7 and lower.


src/Tests/Microsoft.Diagnostics.Monitoring.TestCommon/Runners/AppRunner.cs
```diff
@@ -110,10 +110,12 @@ public AppRunner(ITestOutputHelper outputHelper, Assembly testAssembly, int appI
                 AssemblyNameWithoutExtension,
                 tfm);
 
+#pragma warning disable CS0618 // Type or member is obsolete
             _startupHookPath = AssemblyHelper.GetAssemblyArtifactBinPath(
                 testAssembly,
                 "Microsoft.Diagnostics.Monitoring.StartupHook",
                 TargetFrameworkMoniker.Net60);
```
wiktork, line 117: I'm assuming we're keeping the startup hook since to ensure we're not depending on apis that don't exist in older frameworks correct?

eng/release/DiagnosticsReleaseTool/DiagnosticsReleaseTool.csproj
```diff
@@ -22,7 +22,7 @@
 
     <PackageReference Include="Azure.Identity" Version="[1.12.0]" />
     <PackageReference Include="Azure.Storage.Blobs" Version="[12.21.2]" />
-    <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.20468.1" />
+    <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.20478.1" />
```
wiktork, line 25: Why the version change?

wiktork (approved):

eng/release/DiagnosticsReleaseTool/DiagnosticsReleaseTool.csproj
jander-msft: I think the older version was in one of the older dotnetX feeds that I removed so it could find it anymore. I bumped it to the next available version. We should probably update this project to use CPM, but want to leave that for another change.

PR #8126, opened 2025-04-18
[feature/10.0] Move binding logic to descriptors

sbomer:
This is in preparation for using the configuration binding source generator, which needs the concrete type of the settings object to be available at the `Bind` callsite. We have that type available in the descriptors, so that seems like a natural place to do it. This change replaces the generic descriptor types with concrete ones that use the specific settings types.


src/Tests/Microsoft.Diagnostics.Monitoring.Tool.UnitTestCommon/CollectionRules/Actions/CallbackAction.cs
```diff
@@ -73,6 +75,20 @@ public ICollectionRuleAction Create(IProcessInfo processInfo, BaseRecordOptions
         }
     }
 
+    internal sealed class CallbackActionDescriptor : ICollectionRuleActionDescriptor
```
jander-msft, line 78: I assume that these cannot be handled by a single generic class because the configuration source generator wouldn't be able to intercept the Bind call? Because if it could, I think all of these could be collapsed into one class.

src/Tools/dotnet-monitor/ServiceCollectionExtensions.cs
```diff
@@ -196,37 +196,38 @@ public static IServiceCollection ConfigureCollectionRules(this IServiceCollectio
...
-        public static IServiceCollection RegisterCollectionRuleAction<TFactory, TOptions>(this IServiceCollection services, string actionName)
+        public static IServiceCollection RegisterCollectionRuleAction<TFactory, TOptions, TDescriptor>(this IServiceCollection services)
             where TFactory : class, ICollectionRuleActionFactory<TOptions>
             where TOptions : BaseRecordOptions, new()
+            where TDescriptor : class, ICollectionRuleActionDescriptor, new()
         {
             services.AddSingleton<TFactory>();
             services.AddSingleton<CollectionRuleActionFactoryProxy<TFactory, TOptions>>();
-            services.AddSingleton<ICollectionRuleActionDescriptor, CollectionRuleActionDescriptor<TFactory, TOptions>>(sp => new CollectionRuleActionDescriptor<TFactory, TOptions>(actionName));
+            services.AddSingleton<ICollectionRuleActionDescriptor, TDescriptor>(sp => new TDescriptor());
```
jander-msft, line 206: Do the descriptors actually need to be explicitly constructed? Couldn't service registration just do this for us e.g. `services.AddSingleton<ICollectionRuleActionDescriptor, TDescriptor>()`. Same question for the other descriptor registrations. If not, you can also remove the `new()` constraint on `TDescriptor`.

jander-msft (approved):

PR #8154, opened 2025-04-25
Support resetting state when connecting to a previously monitored app

wiktork:
###### Summary

**Scenario**
Dotnet-monitor profiling/exceptions/parameter capture is installed in an application. Then dotnet-monitor is terminated but the application is not. Dotnet-monitor is then restarted and is once again observing the same application.

**Startuphook/profiler changes**
- Startup hooks only install once. This behavior has been maintained.
- Profiler previously attempted to install again. We now detect that a profiler is already installed and reuse that channel.

**Command handling**
There are 3 threads that handle commands:
- The connection thread, listening for a connection with the client. The message is processed and dispatched as appropriate:
- For a simple message such as callstack collection, the message is sent to the native queue. 
- For startup hook messages (such as exceptions/parameter capture), the message is sent to the managed queue to be picked up by the startup hook for further processing
- The new __reset__ message. This message sends stop/start messages to the native and managed queues above. These effectively block until a response is received from the stop calls to ensure that all cached data has been cleared.

**Exception reset**
- Exceptions rely on an event handler and occur synchronously on any thread.
- For stopping, we first unhook from the exception handler. Then we synchronously wait for all outstanding exception processing to finish. Note the client is awaiting these operations to finish and will not finish the profiler initialization until then.
- After all the data drained, we reset the pipelines and attach the event handler.

**Parameters capture reset**
- We request that all outstanding requests are turned off. No further synchronization is necessary because parameter capturing has a concept of RequestId that matches up requests with returned data. So if a new client received old data, it simply discards it.

**Callstacks**
- No action is needed, but we could further optimize by cancelling outstanding stack walks.

TODO:
- Figure out a way to call stop when the client is disconnected and no new client is started.
- Ideally this can be used for independent stop/start calls in the future. However, this will require additional changes to the client for exceptions.
- Add tests for reset scenario.

###### Release Notes Entry

Added support for resetting in process features when dotnet-monitor is restarted.


src/Microsoft.Diagnostics.Monitoring.StartupHook/ParameterCapturing/Pipeline/ParameterCapturingPipeline.cs
```diff
@@ -206,6 +206,17 @@ public void SubmitRequest(StartCapturingParametersPayload payload, IFunctionProb
             }
         }
 
+        public void RequestStopAll()
```
wiktork, line 209: Per Joe:
- Only one instrumentation session can occur so if we have overlapping requests (old client, new client), we may get an error.

src/Profilers/MonitorProfiler/MainProfiler/MainProfiler.cpp
```diff
@@ -283,6 +283,10 @@ HRESULT MainProfiler::ProfilerCommandSetCallback(const IpcMessage& message)
     {
     case ProfilerCommand::Callstack:
         return ProcessCallstackMessage();
+    case ProfilerCommand::Start:
```
wiktork, line 286: Per Joe/Justin:
- Do more testing around app "death" instead of dotnet-monitor death.

src/Tools/dotnet-monitor/Profiler/ProfilerService.cs
```diff
@@ -68,7 +72,21 @@ public async Task ApplyProfilersAsync(IEndpointInfo endpointInfo, CancellationTo
...
+
+                // TODO Ideally we would detect that dotnet-monitor was disconnected and cleanup inproc collection
+                // TODO Configuration changes across dotnet-monitor may not be respected if profiler feature are already installed
+                // For example, parameter capture is on, but new config says it's off.
+
+                if (env.TryGetValue(ProfilerIdentifiers.EnvironmentVariables.RuntimeInstanceId, out string? runtimeInstanceId))
+                {
+                    // Profiler already applied. We will reset state to discard all previous data collection.
+
+                    await _profilerChannel.SendMessage(endpointInfo, new JsonProfilerMessage(StartupHookCommand.ResetState, new EmptyPayload()), cancellationToken);
```
wiktork, line 85: Per Joe:
- Add timeout to prevent permanent wait.

src/Profilers/MonitorProfiler/Communication/CommandServer.cpp
```diff
@@ -94,70 +93,162 @@ void CommandServer::ListeningThread()
...
+
+    HRESULT hr = S_OK;
+
+    if ((message.CommandSet == static_cast<unsigned short>(CommandSet::StartupHook)) &&
+        (message.Command == static_cast<unsigned short>(StartupHookCommand::Stop) || message.Command == static_cast<unsigned short>(StartupHookCommand::ResetState)))
+    {
+        CallbackInfo managedCallbackInfo;
+        CallbackInfo nativeCallbackInfo;
+
+        CreateControlMessage(CommandSet::StartupHook, StartupHookCommand::Stop, managedCallbackInfo);
```
wiktork, line 152: Per Joe:
- Can we add a generic stop so that other future features can easily pick up that a reset happened

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/ExceptionPipeline.cs
```diff
@@ -13,6 +14,8 @@ internal sealed class ExceptionPipeline :
 
         private long _disposedState;
 
+        private int _processingState;
```
wiktork, line 17: Add config for inproc exceptions as started or stopped initial state.

src/Microsoft.Diagnostics.Monitoring.Options/ExceptionsOptions.cs
```diff
@@ -27,5 +27,11 @@ public sealed class ExceptionsOptions :
             ResourceType = typeof(OptionsDisplayStrings),
             Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_ExceptionsOptions_CollectionFilters))]
         public ExceptionsConfiguration? CollectionFilters { get; set; }
+
+        [Display(
+            ResourceType = typeof(OptionsDisplayStrings),
+            Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_ExceptionsOptions_CollectOnStartup))]
+        [DefaultValue(ExceptionsOptionsDefaults.CollectOnStartup)]
+        public bool? CollectOnStartup { get; set; }
```
wiktork, line 35: This option is respected during resets, but not during initialization.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/CurrentAppDomainFirstChanceExceptionSource.cs
```diff
@@ -16,10 +16,19 @@ internal sealed class CurrentAppDomainFirstChanceExceptionSource :
         private long _disposedState;
 
         public CurrentAppDomainFirstChanceExceptionSource()
```
jander-msft, line 18: nit: Could we remove the ctor?

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/CurrentAppDomainUnhandledExceptionSource.cs
```diff
@@ -15,10 +15,19 @@ internal sealed class CurrentAppDomainUnhandledExceptionSource :
         private long _disposedState;
 
         public CurrentAppDomainUnhandledExceptionSource()
```
jander-msft, line 17: nit: Can we remove the ctor?

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/ExceptionPipeline.cs
jander-msft, line 17: Could Semaphore/SemaphoreSlim be used instead?

src/Microsoft.Diagnostics.Monitoring.StartupHook/ParameterCapturing/Pipeline/ParameterCapturingPipeline.cs
```diff
@@ -206,6 +206,17 @@ public void SubmitRequest(StartCapturingParametersPayload payload, IFunctionProb
...
+        {
+            foreach (var key in _allRequests.Keys)
```
jander-msft, line 211: I think this throws an exception because the dictionary is modified during enumeration. Try iterating `_allRequests.Values`, call `StopRequest?.TrySetResult()` on each, and then clear the entire dictionary after.

You may also want to put the dictionary into a local and immediately recreate the dictionary to avoid concurrent access and then issue the stops.

src/Profilers/MonitorProfiler/Communication/CommandServer.cpp
```diff
@@ -94,70 +93,138 @@ void CommandServer::ListeningThread()
...
+            return false;
     }
 }
 
-void CommandServer::ClientProcessingThread()
+HRESULT CommandServer::ReceiveMessage(std::shared_ptr<IpcCommClient> client, IpcMessage& message)
 {
-    HRESULT hr = _profilerInfo->InitializeCurrentThread();
-
+    HRESULT hr = client->Receive(message);
```
jander-msft, line 199: You can use `IfFailLogRet_(_logger, ...)` instead.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/CurrentAppDomainUnhandledExceptionSource.cs
```diff
@@ -15,10 +15,19 @@ internal sealed class CurrentAppDomainUnhandledExceptionSource :
...
+        {
+        }
+
+        public override void Start()
         {
             AppDomain.CurrentDomain.UnhandledException += CurrentDomain_UnhandledException;
         }
 
+        public override void Stop()
```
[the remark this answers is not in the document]
wiktork: I validated that this is a noop

src/Microsoft.Diagnostics.Monitoring.StartupHook/ParameterCapturing/Pipeline/ParameterCapturingPipeline.cs
wiktork: I was under the impression that ConcurrentDictionary has no problems with this because it's already snapshotting but let me double check

wiktork: I verified that this works correctly and ConcurrentDictionary effectively snapshots when iteration starts.

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/ExceptionPipeline.cs
wiktork: I don't think I can use these easily since I don't really have a fixed count here. I ultimately landed on an Interlocked reader/writer lock solution.

schmittjoseph (approved):
Signing off on the parameter capturing + profiler changes

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/ExceptionPipeline.cs
```diff
@@ -31,15 +32,19 @@ public ExceptionPipeline(ExceptionSourceBase exceptionSource, Action<ExceptionPi
...
+            // We must do this first to prevent any new exception callbacks.
             _exceptionSource.ExceptionAvailable -= ExceptionSource_ExceptionAvailable;
 
-            while (0 != Interlocked.CompareExchange(ref _processingState, 0, 0) && !DisposableHelper.IsDisposed(ref _disposedState))
+            // Wait until all outstanding exception callbacks are finished. Once _processingState reaches 0, we will set the WriteFlag.
+            // This will prevent any further exception handlers from processing.
+            while (0 != Interlocked.CompareExchange(ref _processingState, WriteFlag, 0) && !DisposableHelper.IsDisposed(ref _disposedState))
             {
                 // Wait for all exceptions to be processed
                 Thread.Sleep(100);
```
jander-msft, line 50: Should SpinWait be used here too?

jander-msft (approved):

src/Microsoft.Diagnostics.Monitoring.StartupHook/Exceptions/Pipeline/ExceptionPipeline.cs
wiktork: I purposely did not. My thinking is that there are N threads that could be throwing exceptions and I want those to have as little contention as possible and I don't want to put the thread to sleep if possible.  The Stop thread on the other hand should not be eating cpu if there are many outstanding exceptions being processed. It should yield its timeslice liberally to the rest of the app. Starvation is not a concern because I first unhook the exception handler.

src/Microsoft.Diagnostics.Monitoring.Options/OptionsDisplayStrings.resx
```diff
@@ -780,4 +780,7 @@
     <value>Performs a match based on the name of the managed entry point assembly of the process.</value>
     <comment>The description provided for the ManagedEntryPointAssemblyName parameter on ProcessFilterDescriptor.</comment>
   </data>
+  <data name="DisplayAttributeDescription_ExceptionsOptions_CollectOnStartup" xml:space="preserve">
+    <value>True if exception collection should begin immediately</value>
```
jander-msft, line 784:
```suggestion
    <value>[Experimental] Determines if exception collection should begin immediately</value>
```

jander-msft (dismissed):

jander-msft (approved):

src/Microsoft.Diagnostics.Monitoring.StartupHook/ParameterCapturing/Pipeline/ParameterCapturingPipeline.cs
wiktork: Pending testing, basic parameter capture works as expected

jander-msft (approved):

PR #8165, opened 2025-04-30
[feature/10.0] Use validation source generator

sbomer:
This uses the new model for source generated validation from https://github.com/dotnet/aspnetcore/issues/46349.

For now this checks in the generated code, with some tweaks to work around the following issues:
- https://github.com/dotnet/aspnetcore/issues/61379
- https://github.com/dotnet/aspnetcore/issues/61388
- https://github.com/dotnet/aspnetcore/issues/61525
- https://github.com/dotnet/aspnetcore/issues/61589
- https://github.com/dotnet/aspnetcore/issues/61737

and the generator is disabled.

Once those issues are fixed, we can:
- remove checked-in ValidatableInfoResolver.cs
- turn the generator back on
- remove a couple of the types from ValidatableTypes.cs (see comments)


sbomer:
The new source-generated validation model doesn't expose sync overloads, making it unsuitable for options validation (even though the implementation doesn't do anything async under the hood, as far as I could tell).

It's also mainly designed to be integrated with parameter validation in aspnet, not options validation. For that reason we're pursuing an approach using the existing options validations APIs instead, in https://github.com/dotnet/dotnet-monitor/pull/8254.

PR #8180, opened 2025-05-05
[feature/10.0] Use configuration binding source generator

sbomer:
This opts into the configuration-binding source generator.

The ConfigureOptions<T> helper must be inlined, because the generator expects to know the concrete type argument at each services.Configure callsite.

The source generator doesn't support the `ValidationResult` type, so this is replaced with another type.

Follow-up to https://github.com/dotnet/dotnet-monitor/pull/8126.


src/Tools/dotnet-monitor/CollectionRules/Options/CollectionRuleOptions.cs
```diff
@@ -37,6 +37,18 @@ internal sealed partial class CollectionRuleOptions
             Description = nameof(OptionsDisplayStrings.DisplayAttributeDescription_CollectionRuleOptions_Limits))]
         public CollectionRuleLimitsOptions? Limits { get; set; }
 
-        internal List<ValidationResult> ErrorList { get; } = new List<ValidationResult>();
+        internal List<ErrorValidationResult> ErrorList { get; } = new List<ErrorValidationResult>();
```
jander-msft, line 40: Could you add a source comment explaining why `ValidationResult` cannot be used here?

jander-msft (approved):

PR #8233, opened 2025-05-12
[feature/10.0] Replace reflection-based options mapping logic

sbomer:
This replaces the reflection-based logic from CommonOptionsExtensions.cs with a statically-typed version.
The statically typed version could be source generated, but for now it is written by hand (with plenty help from copilot).

CommonOptionsExtensions has been moved into the test tree, and a new testcase has been added that:
- uses reflection to fully populate a RootOptions instance
- maps it to a dictionary using both the old and the new logic
- checks that the results are equal

Includes a fix to the mapping logic to treat `Uri` as a built-in type (calling ConvertUtils.ToString instead of recursing into its properties).


src/Tools/dotnet-monitor/ConfigurationTokenParser.cs
```diff
@@ -119,11 +119,12 @@ public bool TryCloneSettings(object? originalSettings, ref object? settings)
...
         }
 
-        public static IEnumerable<PropertyInfo> GetPropertiesFromSettings(object? settings, Predicate<PropertyInfo>? predicate = null) =>
-            settings?.GetType()
-                .GetProperties(BindingFlags.Public | BindingFlags.Instance)
-                .Where(p => p.PropertyType == typeof(string) && (predicate?.Invoke(p) ?? true)) ??
-            Enumerable.Empty<PropertyInfo>();
+        public static IEnumerable<PropertyInfo> GetPropertiesFromSettings(Type type, Predicate<PropertyInfo>? predicate = null)
+        {
+            return type.GetProperties(BindingFlags.Public | BindingFlags.Instance)
```
wiktork, line 124: Can you clarify why this works correctly for Type.GetProperties but not object?.GetType().GetProperties?

PR #8555, opened 2025-08-25
[TESTING] Only 10.0 builds

wiktork:


wiktork:
TODO Fix known race condition in build

PR #9484, opened 2026-06-03
[release/10.0] Merge main to release/10

wiktork:

### DIFF
--- a/eng/dependabot/net8.0/Versions.props
+++ b/eng/dependabot/net8.0/Versions.props
@@ -10,7 +10,7 @@
     <!-- Microsoft.Extensions.Logging.Console -->
     <MicrosoftExtensionsLoggingConsole80Version>8.0.1</MicrosoftExtensionsLoggingConsole80Version>
     <!-- Microsoft.NETCore.App -->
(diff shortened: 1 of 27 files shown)


eng/dependabot/net8.0/Versions.props
```diff
@@ -12,6 +12,6 @@
     <!-- Microsoft.NETCore.App -->
     <MicrosoftNETCoreApp80Version>8.0.27</MicrosoftNETCoreApp80Version>
     <!-- System.Text.Json -->
-    <SystemTextJson80Version>8.0.6</SystemTextJson80Version>
+    <SystemTextJson80Version>10.0.3</SystemTextJson80Version>
```
kkeirstead, line 15: Any reason this didn't get the bump to 10.0.8 that 9.0 is using? I see this is different than the previous merge

wiktork: Good catch. Let me look into it.

wiktork: So it looks like 10.0.3 was the matching dependency on the diagnostics side; we can go higher, but not lower. Then dependabot updated some of the entries but not across all the branches.

kkeirstead (approved):

PR #9637, opened 2026-07-15
Temporarily turn off sbom validation

wiktork:


hoyosjs (approved):
Does this also disable the one that comes from 1es.Publish... tasks? Or is that one working fine?

PR #9723, opened 2026-08-13
Decode escaped counter tags in metrics output

hoyosjs:
MetricsEventSource escapes meter tag keys and values in newer event versions so delimiters inside tags remain distinguishable from separators. Decode that transport representation before exposing tags through Prometheus or `/livemetrics` JSON.

- Recover original meter tag key/value pairs for Prometheus labels.
- Decode value, meter, and instrument tags in JSON output.
- Preserve legacy EventCounters `:`-separated metadata and null tags.
- Cover commas, equals signs, backslashes, combinations, empty values, histograms, and EventCounters metadata.

This depends on a future `Microsoft.Diagnostics.Monitoring.EventPipe` package containing `CounterTagFormatter` from dotnet/diagnostics commit `19fd00d3e`. The current pinned package does not contain that helper, so the dependency must flow before this PR can leave draft.

### DIFF
--- a/src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/CounterTags.cs
+++ b/src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/CounterTags.cs
@@ -1,0 +1,77 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Diagnostics.Monitoring.EventPipe;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    /// <summary>
+    /// Recovers the original key/value pairs from the tag strings carried on a counter payload.
+    /// </summary>
+    internal static class CounterTags
+    {
+        // System.Diagnostics.Metrics tags reach dotnet-monitor in a canonical escaped form
+        // ('\' -> '\\', ',' -> '\,', '=' -> '\=') so a ',' or '=' inside a key or value stays
+        // distinguishable from the ',' separating pairs and the '=' separating a key from its value.
+        // That form is a transport detail and must be decoded before tags are exposed to a caller.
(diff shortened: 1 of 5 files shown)


src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/CounterTags.cs
```diff
@@ -0,0 +1,77 @@
...
+
+namespace Microsoft.Diagnostics.Monitoring.WebApi
+{
+    /// <summary>
+    /// Recovers the original key/value pairs from the tag strings carried on a counter payload.
+    /// </summary>
+    internal static class CounterTags
+    {
+        // System.Diagnostics.Metrics tags reach dotnet-monitor in a canonical escaped form
+        // ('\' -> '\\', ',' -> '\,', '=' -> '\=') so a ',' or '=' inside a key or value stays
```
noahfalk, line 16: I suggested an alternative encoding in the runtime PR. If we use it we'd need update this.

src/Microsoft.Diagnostics.Monitoring.WebApi/Metrics/JsonCounterLogger.cs
```diff
@@ -75,10 +75,10 @@ protected override async Task SerializeAsync(ICounterPayload counter)
                         counter.DisplayName,
                         counter.Unit,
                         counter.CounterType.ToString(),
-                        CounterUtilities.AppendPercentile(counter.ValueTags, quantile.Percentage),
+                        FormatValueTags(counter, CounterUtilities.AppendPercentile(counter.ValueTags, quantile.Percentage)),
                         quantile.Value,
-                        counter.CounterMetadata.MeterTags,
-                        counter.CounterMetadata.InstrumentTags);
+                        CounterTags.Format(counter.CounterMetadata.MeterTags),
```
wiktork, line 80: This might affect our Prometheus scenario as well, but we can address that ourselves as well.

PR #9769, opened 2026-08-26
[S3 egress] Support S3-compatible endpoints: SessionToken, endpoint on every credential path, DisablePayloadSigning

### LINKED ISSUE #9768: S3 egress cannot upload to S3-compatible services without chunked/trailered payload support (e.g. Cloudflare R2)
### Description

The S3 egress provider cannot upload any artifact to an S3-compatible service that implements neither chunked-signed payloads nor flexible-checksum trailers. Cloudflare R2 is the common case. With valid static credentials and a reachable bucket, every egress attempt fails:

```
S3 storage egress failed: STREAMING-AWS4-HMAC-SHA256-PAYLOAD-TRAILER not implemented
```

Setting `AWS_REQUEST_CHECKSUM_CALCULATION=when_required` on the container removes the checksum trailer but only moves the failure one step, because the payload is still chunk-signed:

```
S3 storage egress failed: STREAMING-AWS4-HMAC-SHA256-PAYLOAD not implemented
```

The same upload succeeds when the request is signed with `UNSIGNED-PAYLOAD` — e.g. the AWS CLI uploads multi-part objects to the same bucket with the same credentials without issue. This is also Cloudflare's own documented integration path for the AWS SDK for .NET: their [R2 example for aws-sdk-net](https://developers.cloudflare.com/r2/examples/aws/aws-sdk-net/#upload-and-retrieve-objects) sets `DisablePayloadSigning = true` on upload requests for exactly this reason.

**No configuration can work around this.** The AWS SDK's opt-out is `DisablePayloadSigning`, a **per-request** property (`PutObjectRequest`/`UploadPartRequest`) with no client-config, environment-variable, or profile binding — and `S3StorageEgressProvider.EgressAsync` unconditionally takes the multi-part path, so every artifact type (dumps, gcdumps, traces, logs, live metrics) is affected.

Two details worth noting for a fix:

1. **The checksum trailer and payload signing are independent wire features, and both must be addressed.** Disabling only payload signing still emits `STREAMING-UNSIGNED-PAYLOAD-TRAILER` (the SDK forces the flexible checksum into a trailer whenever `DisablePayloadSigning` is set), which these endpoints also reject. A working option has to set `UNSIGNED-PAYLOAD` on the requests *and* move the client to `RequestChecksumCalculation`/`ResponseChecksumValidation` `WHEN_REQUIRED`.
2. An unsigned payload is integrity-protected only by the transport, so the option should be validated to require an HTTPS endpoint (the SDK's signer throws `AmazonClientException` otherwise — which, being a sibling of `AmazonS3Exception` in AWS SDK v4, today escapes `EgressAsync`'s catch and leaves the initiated multi-part upload un-aborted).

### Adjacent defects in the same area

While fixing this we also hit two related gaps in the S3 provider:

- **Temporary (STS-style) credentials are unusable**: there is no `sessionToken` option, and a request signed with only an access key id + secret access key is rejected by the service when the credentials are temporary.
- **`endpoint`, `regionName` and `forcePathStyle` are silently ignored** when authenticating via `awsProfileName` or the default credential chain — they are only applied when `accessKeyId`/`secretAccessKey` are both set, which makes the profile/default-chain paths unusable against a custom S3-compatible endpoint (`No RegionEndpoint or ServiceURL configured`).

### Configuration

`mcr.microsoft.com/dotnet/monitor:10` (also reproduced on a 10.0.3 source build), Linux x64 containers, S3 egress extension, target service Cloudflare R2. Not specific to the container environment.

### Regression?

No — the S3 provider has behaved this way since chunked signing / flexible checksums became the AWS SDK default.

### Other information

We run a patched build of the S3 extension in production against Cloudflare R2; the fix (an opt-in `disablePayloadSigning` option plus the `sessionToken` and endpoint-handling corrections) is proposed in the linked PR, with wire-level tests asserting the `x-amz-content-sha256` form via a recording HttpMessageHandler.

GureevLeonid:
###### Summary

Fixes #9768.

Three changes that make the S3 egress provider work against S3-compatible services (Cloudflare R2, MinIO, and similar), in two commits:

**1. `sessionToken` option** — temporary (STS-style) credentials consist of an access key id, a secret access key **and** a session token; all three must be presented on every request. The provider had no way to configure the third, so temporary credentials were unusable. Validated to require both companions; switches the client to `SessionAWSCredentials`.

**2. `endpoint`/`regionName`/`forcePathStyle` honored on every credential path** — previously they were only applied when `accessKeyId` and `secretAccessKey` were both set, and silently ignored with `awsProfileName` or the default credential chain (`No RegionEndpoint or ServiceURL configured`). The connection configuration is now built identically on every path: the endpoint of the storage service is orthogonal to how the caller authenticates. A "Behavior change" note in the docs covers the one observable difference (a configured `regionName` now wins over a profile-sourced region on those paths).

**3. `disablePayloadSigning` option** — the SDK uploads with a signed streaming payload plus a flexible-checksum `aws-chunked` trailer; services that implement neither reject every artifact (`STREAMING-AWS4-HMAC-SHA256-PAYLOAD-TRAILER not implemented`). `DisablePayloadSigning` is a per-request SDK property with no config/env binding, and `EgressAsync` always takes the multi-part path, so no configuration could work around it.

##### Why the option also moves checksums to `WHEN_REQUIRED`

The trailer and the payload signature are independent wire features. Disabling only the signature still emits `STREAMING-UNSIGNED-PAYLOAD-TRAILER` (`ChecksumUtils` forces the checksum into a trailer whenever `DisablePayloadSigning` is set), which the same endpoints reject. A knob that requires a second undocumented knob to function is a trap, so the one option sets `UNSIGNED-PAYLOAD` on the requests *and* `RequestChecksumCalculation`/`ResponseChecksumValidation = WHEN_REQUIRED` on the client. The option description and schema say so.

Setting `DisablePayloadSigning` is also Cloudflare's own documented integration path for this SDK — their [R2 example for the AWS SDK for .NET](https://developers.cloudflare.com/r2/examples/aws/aws-sdk-net/#upload-and-retrieve-objects) sets it on upload requests — so the option mirrors the vendor-recommended usage rather than inventing a workaround.

##### Wire-level evidence (recorded via a stub HttpClientFactory against the real AmazonS3Client)

With `disablePayloadSigning: true`, the full provider sequence signs as:

| request | `x-amz-content-sha256` | body form |
|---|---|---|
| `GetBucketAcl` (bucket probe) | hex SHA-256 (empty) | none |
| `PutBucket` | hex SHA-256 (empty) | 0-byte |
| `InitiateMultipartUpload` | hex SHA-256 (empty) | 0-byte |
| `UploadPart` | **`UNSIGNED-PAYLOAD`** | raw bytes, no `aws-chunked` |
| `CompleteMultipartUpload` | hex SHA-256 of the XML | XML |
| `PutObject` (empty-artifact path) | **`UNSIGNED-PAYLOAD`** | 0-byte |
| `AbortMultipartUpload` | hex SHA-256 (empty) | none |

— i.e. exactly the shape the AWS CLI produces, and the shape R2 accepts.

##### Safety

- An unsigned payload is integrity-protected by the transport, so validation rejects `disablePayloadSigning` with a non-HTTPS endpoint (`Uri`-parsed, whitespace/case tolerant). An unset endpoint passes — the AWS-hosted endpoints are HTTPS; an endpoint resolved from `AWS_ENDPOINT_URL_S3` escapes options-level validation and is rejected by the SDK's signer at upload time instead.
- That signer refusal is an `AmazonClientException` — a **sibling** of `AmazonS3Exception` in AWS SDK v4, not a subclass — thrown after the multi-part upload is initiated. `EgressAsync` now catches it too, so the upload is aborted instead of leaving orphaned (billed) parts, and the message is wrapped like other egress failures.
- Default `false`; Amazon S3 behavior is unchanged. `UseChunkEncoding` deliberately not touched: the SDK's post-marshall handler skips that assignment entirely when `DisablePayloadSigning` is set.

##### Tests

- Wire-level assertions through the real `AmazonS3Client` + recording `HttpMessageHandler`: `x-amz-content-sha256` is `UNSIGNED-PAYLOAD` with the option on (PutObject and UploadPart — the path every artifact takes) and not with it off.
- Checksum-default assertions compare against an `AmazonS3Config` built in-process, so the suite stays green under `AWS_REQUEST_CHECKSUM_CALCULATION=when_required` in the environment (the very variable a user might set while debugging this).
- Credentials tests cover `SessionAWSCredentials`/`BasicAWSCredentials` selection and option validation; endpoint-validation tests cover https/http/scheme-less/whitespace.
- `documentation/schema.json` is generator-produced (`SchemaGenerationTests` green).

This change runs in production against Cloudflare R2 (process dumps egressing from Kubernetes sidecars); the first artifact uploads failed with the exact error strings above on the stock image and succeed with this change.

Docs claim `First Available: 10.1` / `(10.1+)` following the 8.1/9.1 minor convention — maintainers, please adjust if the next feature release is numbered differently.

###### Release Notes Entry
Added `sessionToken` and `disablePayloadSigning` options to the S3 egress provider and honored `endpoint`/`regionName`/`forcePathStyle` on every credential path, enabling egress to S3-compatible services such as Cloudflare R2 and MinIO.

### DIFF
--- a/documentation/configuration/egress-configuration.md
+++ b/documentation/configuration/egress-configuration.md
@@ -155,6 +155,7 @@ First Available: 8.0
 | bucketName | string | true | The name of the S3 Bucket to which the blob will be egressed. |
 | accessKeyId | string | false | The AWS AccessKeyId for IAM user to login.  |
 | secretAccessKey | string | false | The AWS SecretAccessKey associated AccessKeyId for IAM user to login. To login by access key id the 'secretAccessKey' must be set. |
+| sessionToken | string | false | (10.1+) The AWS SessionToken that accompanies temporary (STS-issued) credentials, e.g. those returned by `sts:AssumeRole` or by an S3-compatible service that issues short-lived credentials. When set, both 'accessKeyId' and 'secretAccessKey' must also be set. |
 | awsProfileName | string | false | The AWS profile name to be used for login. |
 | awsProfilePath | string | false | The AWS profile path, if profile details not stored in default path. |
 | regionName | string | false | A Region is a named set of AWS resources in the same geographical area. This option specifies the region to connect to. If the Endpoint is specified, this is the AuthenticationRegion; otherwise, it is the RegionEndpoint. |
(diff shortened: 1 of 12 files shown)


documentation/configuration/egress-configuration.md
```diff
@@ -221,10 +223,117 @@ First Available: 8.0
...
+</details>
+
+> **Note:** Leave this off for Amazon S3, which implements both features. Turning it off where it is
+> not needed loses the end-to-end integrity check that the payload signature and checksum provide.
+
+### S3 endpoint handling on every authentication path
+
+First Available: 10.1
+
+> **Behavior change:** `endpoint`, `regionName` and `forcePathStyle` are applied on **every**
```
wiktork, line 320: Are there any breaking changes here? I would like to get this into a patch release rather than a minor release. Also, this would only be in dotnet monitor 10+. I am not planning on a backport to 8/9

documentation/configuration/egress-configuration.md
```diff
@@ -221,10 +223,117 @@ First Available: 8.0
...
+                  "regionName": "auto",
+                  "disablePayloadSigning": true
+              }
+          }
+      }
+  }
+  ```
+</details>
+
+> **Note:** Leave this off for Amazon S3, which implements both features. Turning it off where it is
```
wiktork, line 313: nit: Turning it on?

